PR #87, opened 2022-07-24
Denis/encoding

Trinidadec:


text/encoding/main.f.js
```diff
@@ -0,0 +1,53 @@
...
+const operator = require('../../types/function/operator/main.f.js')
+const result = require('../../types/result/main.f.js')
+const list = require('../../types/list/main.f.js')
+
+/** @typedef {result.Result<number,number>} Utf8Result */
+
+/** @type {(input:number) => list.List<Utf8Result>} */
+const codePointToUtf8 = input =>
+{
+    if (input >= 0x0000 && input <= 0x007f)
```
sergey-shandar, line 11:
```js
if (...) { return ... }
if (... ) {
   return ...;
}
```

text/encoding/main.f.js
```diff
@@ -0,0 +1,41 @@
+const { todo } = require('../../dev/main.f.js')
+const operator = require('../../types/function/operator/main.f.js')
+const result = require('../../types/result/main.f.js')
+const list = require('../../types/list/main.f.js')
+
```
sergey-shandar, line 5: const { ok, error } = result;

sergey-shandar (approved):

PR #263, opened 2023-12-07
`switch (true)` instead of `if () { return }`

sergey-shandar:
This PR is created to track the progress of TypeScript issue.


sergey-shandar:
I believe it's related to this bug https://github.com/microsoft/TypeScript/issues/43458

PR #278, opened 2024-10-21
Node 23

sergey-shandar:


sergey-shandar:
There's a [bug](https://github.com/nodejs/node/issues/55410) in Node.js 23. It looks like it's fixed but the fix is not released yet. We will wait.

PR #314, opened 2024-12-09
@ts-self

sergey-shandar:
A workaround for "slow-types", missing `.d.ts`, and TSDoc 
https://github.com/jsr-io/jsr/issues/370
https://github.com/jsr-io/jsr/issues/444
https://github.com/jsr-io/jsr/issues/494


sergey-shandar:
It works now https://jsr.io/@functionalscript/functionalscript/doc/types/list

PR #332, opened 2024-12-22
Introduce unary_plus at nanvm_lib::interface - and test it (Rust only)

sasha-gil:


nanvm-lib/src/interface.rs
```diff
@@ -69,6 +70,56 @@ pub trait Any: PartialEq + Sized + Clone + fmt::Debug {
...
+                    Nullish::Null => return Self::new_simple(Simple::Number(0.0)),
+                    Nullish::Undefined => return Self::new_simple(Simple::Number(f64::NAN)),
+                },
+                Simple::Boolean(v) => {
+                    return Self::new_simple(Simple::Number(if v { 1.0 } else { 0.0 }))
+                }
+                Simple::Number(v) => return self.clone(),
+            }
+        }
+        if let Ok(v) = self.clone().try_to::<Self::String16>() {
```
sergey-shandar, line 87: I would recommend using the `unpack` function instead. In this case, the match will be simpler.

nanvm-lib/src/interface.rs
```diff
@@ -69,6 +70,56 @@ pub trait Any: PartialEq + Sized + Clone + fmt::Debug {
         // bigint and function
         todo!()
     }
+
+    fn unary_plus(&self) -> Self {
```
sergey-shandar, line 74: If the function returns `Self`, let's pass `self` as an input parameter instead of `&self`. In this case, we can move the object without ever calling the function `clone`.

nanvm-lib/tests/test.f.mjs
```diff
@@ -23,6 +23,7 @@ export default {
             },
             nullish: () => {
                 n(false)(undefined)
+                n(false)(null)
```
sergey-shandar, line 26: Add the same operations into the JS file.

nanvm-lib/src/interface.rs
sasha-gil: Will do!

nanvm-lib/tests/test.f.mjs
sasha-gil: I am ready to do so but I need instructions on how to run this JS in a local enlistment.

sergey-shandar: `npm test` will run all tests. The log is huge so I run `npm test > _.txt` sometimes.

sergey-shandar (approved):

nanvm-lib/tests/test.rs
```diff
@@ -102,7 +111,82 @@ fn eq<A: Any>() {
...
+    let object0: A = [].to_object_unknown();
+    assert_is_nan(Any::unary_plus(object0), "object");
+    // TODO: test objects with valueOf, toString functions.
+    // TODO: test Function.
+}
+
 #[test]
 fn naive_eq() {
     eq::<naive::Any>();
+    unary_plus::<naive::Any>();
```
sergey-shandar, line 191: next PR: combine `eq` and other operations into a generic functions.

sergey-shandar (commented):
Switch to data-driven tests (both TS and Rust).

```ts
type Test = [any, any, (any) => void]
```

PR #340, opened 2024-12-23
Main module

Trinidadec:


module.f.ts
```diff
@@ -0,0 +1,42 @@
+import * as tokenizer from './djs/tokenizer/module.f.ts'
```
sergey-shandar, line 1: Rename the file to `module.ts`. We use `.f.ts` for pure FunctionalScript without I/O.

sergey-shandar (approved):

PR #353, opened 2025-01-01
Implement unary minus, refactor tests

sasha-gil:
TypeScript tests for unary_minus will be added shortly.

### DIFF
--- a/nanvm-lib/src/interface.rs
+++ b/nanvm-lib/src/interface.rs
@@ -135,7 +135,12 @@ pub trait Any: PartialEq + Sized + Clone + fmt::Debug {
         // Symbol, so we don't return error results. We use unary_plus as a helper function here,
         // handling BigInt case when the error result of unary_plus indicates that case.
         match Self::unary_plus(v) {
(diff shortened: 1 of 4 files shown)


nanvm-lib/src/interface.rs
```diff
@@ -38,8 +38,14 @@ pub trait Function<U: Any<Function = Self>>:
...
 #[derive(Debug, PartialEq)]
-pub enum RuntimeError {
-    TypeError,
+pub enum TypeError<U: Any> {
+    BigIntToNumber(U), // Cannot convert a given BigInt value to a number
+                       // More error variants will be added in the future as needed.
+}
+
+#[derive(Debug, PartialEq)]
+pub enum RuntimeError<U: Any> {
```
sergey-shandar, line 47: IMHO, it would be better if we use `Any` for as a run-time error type because FunctionalScript can `throw` anything. For example: https://github.com/functionalscript/functionalscript/blob/main/dev/module.f.ts

sergey-shandar (approved):

PR #355, opened 2025-01-03
BNF types and an interface for a `RangeMap`

sergey-shandar:


bnf/module.f.ts
```diff
@@ -0,0 +1,214 @@
...
+ * - An "or" operation (`Or`)
+ * - A terminal range (`TerminalRange`)
+ * - A string (representing a literal sequence of characters)
+ */
+export type DataRule = Sequence | Or | TerminalRange | string
+
+/**
+ * Represents a lazy grammar rule.
+ * A function that resolves to a `DataRule`
+ */
```
sasha-gil, line 63: Could you please provide brief hints here on why / when this is useful? I vaguely recall your explanation from a live meeting, but having some hints in written would not hurt!

types/range_map/module.f.ts
```diff
@@ -1,3 +1,34 @@
...
+ *
+ * // Create range maps
+ * const range1 = rmOps.fromRange([0, 10])(true)
+ * const range2 = rmOps.fromRange([5, 15])(false)
+ *
+ * // Merge range maps
+ * const merged = toArray(rmOps.merge(range1)(range2))
+ *
+ * // Retrieve values from the merged range map
+ * if (rmOps.get(7)(merged) !== true) { throw 'error' }
```
sasha-gil, line 26: It looks like this statement says: the merged result has 'true' at position 7 while range2 was supposed to knock down that value to 'false', right?

types/range_map/test.f.ts
```diff
@@ -1,132 +1,155 @@
...
+        const range1 = rmOps.fromRange([0, 10])(true)
+        const range2 = rmOps.fromRange([5, 15])(false)
+
+        // Merge range maps
+        const merged = toArray(rmOps.merge(range1)(range2))
+
+        // Retrieve values from the merged range map
+        if (rmOps.get(7)(merged) !== true) { throw 'error' }
+        if (rmOps.get(12)(merged) !== false) { throw 'error' }
+        if (rmOps.get(-1)(merged) !== false) { throw 'error' }
```
sasha-gil, line 38: I don't get the outcome of merging range2 into the initial range1 state:
- values at positions 0..4 should remain true thanks to unaffected range1 state, no doubts here;
- values at positions ..-1 should remain false because that's the default value specified at rmOps initialization; same story for values at positions 16.., right?
- values at positions 11..15 are re-affirmed as false not only by default of rmOps, but also because range2 has them set to false explicitly;
- but, what's about values at positions 5..10 (7 included) - this test code states that there is true at position 7 - if so, range2 merging in essentially did not affect the result of rmOps.merge(range1)?

bnf/module.f.ts
sergey-shandar: Done

types/range_map/module.f.ts
sergey-shandar: There's a user's defined `union` function, which is defined as
```js
union: a => b => a || b
```
so when `a` is `true` and `b` is false, the result of the merge is `true`.

sergey-shandar: Ignore the comment, I've changed the example to use numbers and XOR.

types/range_map/test.f.ts
sergey-shandar: See the updated example.

Trinidadec (approved):

PR #377, opened 2025-01-27
Reuse elements of NaNVM's big integer implementation

sasha-gil:
The current implementation of BigInt in functionalscript\nanvm-lib is incomplete. This update copied over elements of big_uint.rs, big_int.rs from nanvm to back BigInt with proper implementation. For now, just a couple of operation is implemented; more to follow.


nanvm-lib/src/naive.rs
```diff
@@ -1,38 +1,49 @@
...
+        b: &rc::Rc<RefCell<Vec<Self::Item>>>,
+    ) -> bool {
         a.as_ptr() == b.as_ptr()
     }
 }
 
 pub struct Complex<P: Policy> {
     header: P::Header,
-    items: rc::Rc<[P::Item]>,
+    items: rc::Rc<RefCell<Vec<P::Item>>>,
```
sergey-shandar, line 46: IMHO, we shouldn't use `Vec` here. Let's create a new `Rc<[...]>` instead when needed.

sasha-gil: All right - I'll probably restart this update from scratch. For now, I am archiving this PR.

PR #429, opened 2025-04-16
Initial writeup of bytecode notes under issues/lang

sasha-gil:


issues/lang/9100-call-like-instructions.md
```diff
@@ -0,0 +1,35 @@
...
+The interpreter accesses metadata upon decoding the starting part of a call instruction. Knowing the
+number of parameters and result presence flag, it continues decoding argument instructions filling
+out an internal data structure best fit for the given host function. For example,
+
+```js
+const c = a + b
+```
+
+produces bytecode for calling host's implementation of operator plus, followed by slot
+specifications for a, b and c at the caller stack frame. This allows to execute that built-in
```
sergey-shandar, line 25:
```suggestion
specifications for `a`, `b` and `c` at the caller stack frame. This allows to execute that built-in
```

sergey-shandar (approved):

PR #447, opened 2025-05-17
Fix join utility function bug

sergey-shandar:
## Summary
- ensure `join` doesn't prefix a separator for the first element

## Testing
- `npx tsc --noEmit -p tsconfig.json` *(fails: Cannot find module 'node:fs' or its corresponding type declarations)*


sergey-shandar:
It doesn't make any sense.

PR #465, opened 2025-05-20
Fix parser: allow identifier object keys

sergey-shandar:
## Summary
- support identifiers as object literal keys and shorthand properties
- test parsing identifier object keys and shorthand export

## Testing
- `npx tsc`
- `npm run test22`
- `cargo test`
- `cargo clippy`
- `cargo fmt -- --check`


sergey-shandar:
This PR doesn't implement the shorten properties feature correctly.

PR #510, opened 2025-07-09
Digital signatures

sergey-shandar:


crypto/secp/README.md
```diff
@@ -2,11 +2,31 @@
...
+We have a [monoid](https://en.wikipedia.org/wiki/Monoid) on the finite set `S` with an operation `*` and an identity `O`. We can apply the operation `*` `N` times on any object `s` from `S` using only `O(log(N))` operations by using the [exponentiation by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring) algorithm. At the same time, if some one knows `s` and `s^N` but doesn't know `N`, it's very hard to find `N` for some monoids, for example for [elliptic curves](https://en.wikipedia.org/wiki/Elliptic-curve_cryptography). For `N ~ 2^256` we need about `256` operations to compute `s^N`, but we need about `2^256` operations to find `N` from `s^N` using a brute force algorithm. This's is called a discrete logarithm problem. See [discrete logarithm](https://en.wikipedia.org/wiki/Discrete_logarithm).
```
sasha-gil, line 5:
```suggestion
We have a [monoid](https://en.wikipedia.org/wiki/Monoid) on the finite set `S` with an operation `*` and an identity `O`. We can apply the operation `*` `N` times on any object `s` from `S` using only `O(log(N))` operations by using the [exponentiation by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring) algorithm. At the same time, if some one knows `s` and `s^N` but doesn't know `N`, it's very hard to find `N` for some monoids, for example for [elliptic curves](https://en.wikipedia.org/wiki/Elliptic-curve_cryptography). For `N ~ 2^256` we need about `256` operations to compute `s^N`, but we need about `2^256` operations to find `N` from `s^N` using a brute force algorithm. This's is called a discrete logarithm problem (DLP for short). See [discrete logarithm](https://en.wikipedia.org/wiki/Discrete_logarithm).
```suggestion
We have a [monoid](https://en.wikipedia.org/wiki/Monoid) on the finite set `S` with an operation `*` and an identity `O`. We can apply the operation `*` `N` times on any object `s` from `S` using only `O(log(N))` operations by using the [exponentiation by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring) algorithm. At the same time, if some one knows `s` and `s^N` but doesn't know `N`, it's very hard to find `N` for some monoids, for example for [elliptic curves](https://en.wikipedia.org/wiki/Elliptic-curve_cryptography). For `N ~ 2^256` we need about `256` operations to compute `s^N`, but we need about `2^256` operations to find `N` from `s^N` using a brute force algorithm. This's is called a discrete logarithm problem. See [discrete logarithm](https://en.wikipedia.org/wiki/Discrete_logarithm).
```
```

File: crypto/secp/README.md [hunk 2]
```diff
@@ -2,11 +2,31 @@
...
+
+|elliptic curves                                          |DLP and monoid                                    |
+|---------------------------------------------------------|--------------------------------------------------|
+|`a + b`, add two points, `a+b = b+a`, `(a+b)+c = a+(b+c)`|`a * b`, multiplication                           |
+|`O`, a point on infinity. `a + O = O + a = a`            |`O`, an identity                                  |
+|`a * n`, scalar multiplication                           |`a ^ n`, exponent (using the fast exponent method)|
+|find `n` for specific `a` and `b`, such that `a * n = b` |`n = log(a, a ^ n)`                               |
+|for any `n` there is `k = 1/n` such that `a * n * k = a` |`a ^ n ^ k = a`                                   |
+
+The scalar multiplier `n` is defined on a prime `N` field. We can `k` such that `(k * n)%N = 1`. The function that returns `/n` is called `reciprocal`, see [../prime_field/module.f.ts](../prime_field/module.f.ts#L35) and it uses [Euclidean division](https://en.wikipedia.org/wiki/Euclidean_division).
```
sasha-gil, line 21:
```suggestion
The scalar multiplier `n` is defined on a prime `N` field. We can `k` such that `(k * n)%N = 1`. The function that returns `1/n` is called `reciprocal`, see [../prime_field/module.f.ts](../prime_field/module.f.ts#L35) and it uses [Euclidean division](https://en.wikipedia.org/wiki/Euclidean_division).
```

crypto/secp/README.md
```diff
@@ -2,11 +2,31 @@
...
+
+`G` is a base point on the elliptic curve. It has `x` and `y`, or a compressed representation has `x` and a boolean.
```
sasha-gil, line 23: Please explain that G = (x,y) is a well known point that is a part of a specific encryption protocol.

PR #770, opened 2026-05-09
Bnf tokenizer

Trinidadec:


sergey-shandar (commented):
## Review: BNF Tokenizer

### Overview

This PR renames `djs/parser-new/` → `djs/tokenizer-new/` to better reflect its purpose, and introduces a significant new test file that builds a tokenizer pipeline on top of the existing `descentParser`/BNF infrastructure. It also exposes `AstTag` as a public type and cleans up the `deno.json` exports.

---

### Changes Summary

| File | Change |
|---|---|
| `bnf/data/module.f.ts` | Export `AstTag` (was internal) |
| `deno.json` | Remove old `parser-new` export |
| `djs/parser-new/test.f.ts` | Deleted |
| `djs/parser-new/module.f.ts` | Renamed → `djs/tokenizer-new/module.f.ts` |
| `djs/tokenizer-new/test.f.ts` | New — 893 lines of tokenizer logic + tests |

---

### Issues

**1. `toJsToken` silently drops `number` and `string` tokens**

`filterFunc` passes `'number'` and `'string'` through, but `toJsToken` hits the `default` branch and returns `null` for both. These tokens are filtered in, then immediately discarded. This is either an incomplete implementation or a logic bug — the commented-out tests suggest the former.

```ts
// filterFunc passes 'number' and 'string'...
case 'number':
case 'string':
    return true

// ...but toJsToken has no case for them
const toJsToken: (tk: Token) => JsToken | null = tk => {
    switch(tk[0]) {
        case '{': case '}': case '[': case ']': case ':': case ',':
            return {kind: tk[0]}
        default:
            return null  // number/string silently dropped here
    }
}
```

**2. Debug comment left in production code**

In `tokenizeString`:
```ts
//return JSON.stringify(toArray(filterTokens))
```
This should be removed before merging.

**3. `djs/tokenizer-new/module.f.ts` not exported in `deno.json`**

The old `./djs/parser-new/module.f.ts` entry was removed from `deno.json`, but no entry for `./djs/tokenizer-new/module.f.ts` was added. If external consumers need access to `jsGrammar`, this is a breaking omission. If this module is intentionally private, that should be confirmed.

**4. Massive volume of commented-out tests**

The `djs` test section has ~7 active test cases and over 100 commented-out ones covering strings, numbers, identifiers, keywords, operators, whitespace, and comments. This signals the tokenizer pipeline is incomplete. Shipping commented-out dead code makes it hard to distinguish "not implemented" from "intentionally excluded." Consider either:
- Moving WIP tests to a tracking issue and deleting the comments, or
- Using a placeholder structure that makes it obvious these are pending

**5. Missing newline at end of file**

`djs/tokenizer-new/test.f.ts` is missing a trailing newline (`\ No newline at end of file`). Minor but inconsistent with project conventions.

**6. `tagToToken` fallback for `undefined` is fragile**

```ts
case 'undefined': return 'undefined'
```
This converts an `undefined` tag to the string `'undefined'`, which then fails `filterFunc` (since `'undefined'` is not in the allowlist). This may be intentional (to discard untagged nodes), but it's fragile — a future addition of `'undefined'` to `filterFunc` would silently include what were intended to be untagged structural nodes.

---

### Code Quality

- **Functional style is consistent** with the rest of the codebase (curried functions, explicit type annotations on separate lines).
- **The `scanFunc`/`filterFunc`/`toJsToken` pipeline** is a reasonable approach for flattening the AST into a token stream, but the incomplete `toJsToken` makes it unclear what the final shape of a token is supposed to be for `number`/`string`.
- **`descentParserCpOnly` is duplicated** from the deleted `djs/parser-new/test.f.ts` into the new file verbatim — fine for test helper code.

---

### Suggestions

1. **Resolve `toJsToken`** — either implement `number`/`string` cases or remove them from `filterFunc` with a comment explaining why they're excluded.
2. **Remove the debug comment** (`//return JSON.stringify(...)` line).
3. **Add `djs/tokenizer-new/module.f.ts` to `deno.json`** if it should be publicly accessible, or confirm it's intentionally private.
4. **Replace commented-out tests** with a tracking issue and delete the comments from the file.
5. **Add trailing newline** to `test.f.ts`.

---

### Summary

The rename from `parser-new` to `tokenizer-new` is a clear improvement, and exposing `AstTag` is a minimal, reasonable change. The main concern is that `djs/tokenizer-new/test.f.ts` merges work-in-progress code: `toJsToken` is incomplete (silently drops tokens), and ~90% of the `djs` tests are commented out. This PR would benefit from either completing the tokenizer or scoping it down to what is actually finished.

---

PR #779, opened 2026-05-13
Normalize `or` in RTTI: flatten, unknown collapse, subset drop, dedup

sergey-shandar:
Addresses [issue 130](./issues/130-or-optimization.md) — goal 2 and a useful slice of goal 1. The `or` constructor in `fs/types/rtti/module.f.ts` now normalizes its variants at construction time, so this analysis is performed once and shared by `orValidate`, `orParse`, and any other consumer.

## Summary

- **Flatten nested `or`** (goal 2): variants whose thunk resolves to `['or', ...]` — both `or(...)` calls and manually-constructed `() => ['or', ...]` thunks — are inlined into the outer union. A visited set keeps self-referential schemas terminating (a thunk is resolved at most once per `or(...)` call).
- **Unknown collapse** (goal 1, "any type ⊆ `unknown`"): any `unknown` variant collapses the whole union to `unknown`.
- **Primitive subset drop** (goal 1, "primitive const ⊆ primitive type"): `or(42, number)` → `or(number)`, similarly for `boolean`/`string`/`bigint`.
- **Dedup** (goal 1, "duplicate variants collapse"): reference equality for thunks and object/tuple consts, `Object.is` for primitive consts — so `NaN` collapses with itself while `+0` and `-0` stay distinct, matching `constPrimitiveValidate` semantics.

Deferred to a follow-up:
- Deep structural subset/equal on struct/tuple consts (the "narrower const tuple/struct is a subset of a wider one" example in goal 1).
- The canonical ordering required by goal 3 — needs a total order on `Type` that includes thunks, which depends on the structural equal/subset above.

## Test plan

- [x] `npm test` — 1492/1492 pass (18 new subtests under `/or/...`).
- [x] `npx tsc` — clean.
- [x] New tests cover: shallow + deep flatten, manually-constructed `or` thunks, self-referential `or` termination, unknown collapse across const/thunk/nested positions, primitive subset drop for each primitive type and the keep-without-thunk case, dedup by reference and by `Object.is` (incl. `NaN` collapse and `±0` distinction), and that `or()` / `never` still produce an empty variant list.
- [x] Existing `or` / `option` / `never` tests in `validate/test.f.ts`, `parse/test.f.ts`, `ts/test.f.ts` unchanged and passing.

https://claude.ai/code/session_01Ekq7xkSfanpfAyLj2U9yQ9

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ekq7xkSfanpfAyLj2U9yQ9)_


fs/types/rtti/test.f.ts
```diff
@@ -12,11 +14,118 @@ const tests: Tests = {
...
+    const info = t()
+    if (info[0] !== 'or') { throw `expected an or thunk, got ${info[0]}` }
+    return info.slice(1) as readonly Type[]
+}
+
+const assertLen = (expected: number) => (vs: readonly Type[]) => {
+    if (vs.length !== expected) { throw `expected ${expected} variant(s), got ${vs.length}` }
+}
+
+const assertEq = (a: Type, b: Type) => { if (a !== b) { throw 'variants differ' } }
```
sergey-shandar: We have a similar function in `fs/dev/module.f.ts`

fs/types/rtti/module.f.ts
```diff
@@ -173,9 +173,110 @@ export const record: MakeType1<'record'> = type1('record')
...
+        const tag = variantTag(t)
+        if (tag === 'unknown') { return [unknown] }
+        if (tag === 'boolean' || tag === 'number' || tag === 'string' || tag === 'bigint') {
+            primThunks.add(tag)
+        }
+    }
+    const result: Type[] = []
+    for (const t of flat) {
+        const tt = typeof t
+        if ((tt === 'boolean' || tt === 'number' || tt === 'string' || tt === 'bigint')
```
sergey-shandar: Code duplications:
```
tag === 'boolean' || tag === 'number' || tag === 'string' || tag === 'bigint'
```
and
```
tt === 'boolean' || tt === 'number' || tt === 'string' || tt === 'bigint'
```

fs/types/rtti/module.f.ts
```diff
@@ -173,9 +173,110 @@ export const record: MakeType1<'record'> = type1('record')
...
+            && primThunks.has(tt)) {
```
sergey-shandar: Why do we need `(tt === 'boolean' || tt === 'number' || tt === 'string' || tt === 'bigint')` if we then check `primThunks.has(tt)`?

fs/types/rtti/module.f.ts
```diff
@@ -173,9 +173,110 @@ export const record: MakeType1<'record'> = type1('record')
...
+ * - `Object.is` covers primitive consts (so `NaN` and `±0` behave like
+ *   `constPrimitiveValidate`).
+ *
+ * Deep structural equality on struct/tuple consts is intentionally out of
+ * scope — see goals 1 and 3 of issue 130.
+ */
+const sameVariant = (a: Type, b: Type): boolean => {
+    const ta = typeof a
+    if (ta !== typeof b) { return false }
+    // For primitive consts, use `Object.is` so `+0` and `-0` are distinct (and
```
sergey-shandar: Why not use `Object.is` always?

Should it be?
```ts
const sameVariant = Object.is
```

fs/types/rtti/module.f.ts
```diff
@@ -173,9 +173,110 @@ export const record: MakeType1<'record'> = type1('record')
...
+        for (const t of ts) {
+            if (typeof t === 'function' && !visited.has(t)) {
+                visited.add(t)
+                const info = t()
+                if (info[0] === 'or') {
+                    walk(info.slice(1) as readonly Type[])
+                    continue
+                }
+            }
+            out.push(t)
```
sergey-shandar: No `push` operations.

fs/types/rtti/module.f.ts
```diff
@@ -173,9 +173,110 @@ export const record: MakeType1<'record'> = type1('record')
...
+ * - Each thunk is resolved at most once; thunks reached a second time are
+ *   kept as-is, so self-referential `or` schemas terminate.
+ */
+const flattenOr = (types: readonly Type[]): readonly Type[] => {
+    const out: Type[] = []
+    const visited = new Set<Type>()
+    const walk = (ts: readonly Type[]): void => {
+        for (const t of ts) {
+            if (typeof t === 'function' && !visited.has(t)) {
+                visited.add(t)
```
sergey-shandar: Construct a new set instead of adding an item.

fs/types/rtti/module.f.ts
```diff
@@ -173,9 +173,110 @@ export const record: MakeType1<'record'> = type1('record')
...
+ * a future change — see goals 1 and 3 of issue 130.
+ */
+const reduceOr = (types: readonly Type[]): readonly Type[] => {
+    const flat = flattenOr(types)
+    const primThunks = new Set<string>()
+    for (const t of flat) {
+        const tag = variantTag(t)
+        if (tag === 'unknown') { return [unknown] }
+        if (tag === 'boolean' || tag === 'number' || tag === 'string' || tag === 'bigint') {
+            primThunks.add(tag)
```
sergey-shandar: Construct a new set.

fs/types/rtti/test.f.ts
sergey-shandar: Done — dropped the local `assertLen`/`assertEq`; tests now import `assertEq` from `fs/dev/module.f.ts` and call `assertEq(v.length, N)` directly.

---

fs/types/rtti/module.f.ts
sergey-shandar: Fixed — both call sites now use `primitive0List` (via a small `isPrim0` helper), so the four-tag string list isn't duplicated.

---

fs/types/rtti/module.f.ts
sergey-shandar: Good point — you're right, `primThunks` only ever contains the four `Primitive0` tag strings, so `primThunks.has(typeof t)` already implies `typeof t` is one of them. The redundant `tt === 'boolean' || ...` check is gone; the filter is now just `primThunks.has(typeof t)`.

---

fs/types/rtti/module.f.ts
sergey-shandar: Agreed — `Object.is` already does the right thing for both: it distinguishes `+0`/`-0`, treats `NaN` as equal to itself, and behaves like `===` for objects/functions (reference equality). Dropped `sameVariant`; the dedup step now calls `Object.is` directly.

---

fs/types/rtti/module.f.ts
sergey-shandar: Rewritten without `push`: both `flattenOr` and `reduceOr` now use `reduce` with spreads to build their result arrays immutably.

---

fs/types/rtti/module.f.ts
sergey-shandar: Done — `visited` is now threaded through `reduce` as part of the accumulator and rebuilt with `new Set([...visited, t])` whenever we descend into a new thunk. No more `Set.add`.

---

fs/types/rtti/module.f.ts
sergey-shandar: Done — `primThunks` is now built immutably inside a `reduce`: when a primitive tag is encountered the accumulator becomes `new Set([...p, tag])` instead of mutating in place.

---

fs/types/rtti/module.f.ts
```diff
@@ -173,9 +173,91 @@ export const record: MakeType1<'record'> = type1('record')
...
+ * `Object.is` is used for deduplication, so `NaN` collapses with itself and
+ * `+0` and `-0` stay distinct — matching `constPrimitiveValidate`.
+ *
+ * Full structural subset (tuples/structs/`or`/recursive schemas) is left to
+ * a future change — see goals 1 and 3 of issue 130.
+ */
+const reduceOr = (types: readonly Type[]): readonly Type[] => {
+    const flat = flattenOr(types)
+    type Collect = readonly[boolean, ReadonlySet<string>]
+    const isPrim0 = (s: string): boolean =>
```
sergey-shandar: Don't create nested `const` objects (including functions) if they can be declared outside of the function!

AGENTS.md
```diff
@@ -44,8 +44,9 @@ where `<...Module documentation...>` should be documentation for the module.
...
   - Avoid side effects and mutability.
-- When a sibling module already has the type or helper you need, import it — add `export` to the existing declaration if it's not yet exported, rather than duplicating it (e.g. `parse` reuses `Path`, `ValidationError`, `verror`, `prependPath`, `primitive0Validate`, `constPrimitiveValidate` from `validate`).
-- Don't mutate arrays or objects in place. Avoid `.push`, `.pop`, `.shift`, `.unshift`, `.splice`, `.sort`, `.reverse`, and index/property assignment on accumulators. Build new values with `.map`, `.filter`, `.flatMap`, spread, and `Object.fromEntries(entries.map(...))`.
+- When a sibling module already has the type or helper you need, import it — add `export` to the existing declaration if it's not yet exported, rather than duplicating it (e.g. `parse` reuses `Path`, `ValidationError`, `verror`, `prependPath`, `primitive0Validate`, `constPrimitiveValidate` from `validate`). The same applies to canonical constants and standard built-ins: use the existing `primitive0List` instead of inlining `['boolean', 'number', 'string', 'bigint']`, and reach for built-ins like `Object.is` directly instead of wrapping them in a hand-rolled helper.
```
sergey-shandar:
```suggestion
- When a sibling module already has the type or helper you need, import it — add `export` to the existing declaration if it's not yet exported, rather than duplicating it (e.g. `parse` reuses `Path`, `ValidationError`, `verror`, `prependPath`, `primitive0Validate`, `constPrimitiveValidate` from `validate`).
```

PR #792, opened 2026-05-16
issues: close 91 (Utf8 nominal type) as won't-do

sergey-shandar:
## Summary

Closes [issue 91](./issues/README.md) — _Create a separate nominal type for UTF-8_ — as **won't do**.

After implementing the type (see reverted commit) and looking at all current callers, every concrete consumer of `utf8(...)` (crypto, hashing, `writeFile`) treats the result as opaque bytes. A nominal `Utf8` would just force `utf8ToVec(utf8(...))` wrapping at every call site without catching any real bug. The `utf8` / `utf8ToString` function names already carry the encoding contract.

(The intersection-with-`Vec` shortcut from the first commit was also inconsistent with how `Vec` and `Id` are defined — standalone `Nominal<..., bigint>` with explicit `asBase`. Going consistent would have meant the verbose option, which isn't worth it for this type.)

This PR only marks issue 91 closed in `issues/README.md` with that rationale.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `npm test` — 1479 tests pass (back to baseline)


fs/text/module.f.ts
```diff
@@ -31,7 +43,14 @@ export type Utf8 = Vec
  * @returns The resulting UTF-8 bit vector, MSB first.
  */
 export const utf8 = (s: string): Utf8 =>
-    u8ListToVecMsb(fromCodePointList(stringToCodePointList(s)))
+    u8ListToVecMsb(fromCodePointList(stringToCodePointList(s))) as Utf8
```
sergey-shandar: We should use `asUtf8`.

PR #797, opened 2026-05-18
RTTI: extract shared kernel from validate/parse into rtti/common (i133)

sergey-shandar:
## Summary

Addresses [i133](./issues/README.md) — "Investigate common parts in `rtti/validate` and `rtti/parse`."

Both `validate` and `parse` traverse the same `Type` shape and return the same `Result<T, ValidationError>` outcome — only the per-variant handling differs (`validate` returns the original value, `parse` constructs a fresh one). The shared parts were living in `validate` and re-imported by `parse`, so `parse` depended on `validate` for incidental reasons.

This PR extracts the shared parts into a new `fs/types/rtti/common/module.f.ts`:

- Error shape (`Path`, `ValidationError`) and path bookkeeping (`verror`, `prependPath`).
- Primitive checks (`primitive0Validate`, `constPrimitiveValidate`).
- The `Result<T>` / `Validate<T>` signatures.
- `match`: a flat-`Kind` recognizer that collapses the two-level `Const | Thunk` ADT (and its inner tag) into a single discriminated union, so the top-level dispatch in each consumer is one `switch` rather than nested case analysis.

`validate` and `parse` now both build on `match` and share no code through each other. `validate` re-exports the moved helpers so its public API is unchanged.

This also gives future schema-driven consumers (the data form sketched in [i143](./issues/README.md)) a stable shared base to build on.

## Test plan

- [x] `npx tsc` clean.
- [x] `npm test` — all 1479 tests pass; `rtti/common`, `rtti/validate`, and `rtti/parse` all report 100% line/branch/function coverage.
- [x] Public API of `rtti/validate` preserved (helpers re-exported from `common`).
- [x] `parse` no longer imports from `validate`.
- [x] `deno.json` `exports` regenerated via `npm run index` (no further changes).

---
_Generated by [Claude Code](https://claude.ai/code/session_011yDdt7Z4WSyVFv6YNW5xgd)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- RTTI: extract shared kernel (error shape, primitive checks, `match` recognizer) from `validate`/`parse` into a new `rtti/common` module ([i133](./issues/README.md)) [797](https://github.com/functionalscript/functionalscript/pull/797)
 - NodeProgram: move `Env` to `fs/types/effects/node` and add as second parameter [795](https://github.com/functionalscript/functionalscript/pull/795)
-- RTTI: extract shared kernel (error shape, primitive checks, `match` recognizer) from `validate`/`parse` into a new `rtti/common` module ([i133](./issues/README.md)) [796](https://github.com/functionalscript/functionalscript/pull/796)
 
 ## 0.15.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 - NodeProgram: move `Env` to `fs/types/effects/node` and add as second parameter [795](https://github.com/functionalscript/functionalscript/pull/795)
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 - NodeProgram: move `Env` to `fs/types/effects/node` and add as second parameter [795](https://github.com/functionalscript/functionalscript/pull/795)
+- RTTI: extract shared kernel (error shape, primitive checks, `match` recognizer) from `validate`/`parse` into a new `rtti/common` module ([i133](./issues/README.md)) [796](https://github.com/functionalscript/functionalscript/pull/796)
```
sergey-shandar: It should be on top.

fs/types/rtti/common/module.f.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Recognizes a schema `Type` and returns its variant as a flat `Kind`.
+ *
+ * - `Thunk` schemas are evaluated once to read the `Info` descriptor, then
+ *   tagged by their leading tag (`'const'`, `'array'`, `'record'`,
+ *   `'unknown'`, `'or'`, or a `Tag0` primitive).
+ * - `Const` schemas (primitives, tuples, structs) are routed directly to
+ *   `tuple`, `struct`, or `constPrimitive`.
+ */
+export const match = (rtti: Type): Kind => {
```
sergey-shandar: This is a translation from one representation into another. I think a visitor would be a better fit.

PR #860, opened 2026-05-26
text: extract shared UTF-8/UTF-16 code-point decoder (i168)

sergey-shandar:
## Summary

Resolves [i168](./issues/168-utf-codepoint-decoder.md).

`fs/text/utf8` and `fs/text/utf16` shared a byte-for-byte streaming code-point
decoder skeleton — append a `null` EOF sentinel, run a `stateScan` that
dispatches unit vs. EOF, then `flat` the result — plus a duplicated `errorMask`
constant. Only the per-unit bit-twiddling and the leftover-state error code
genuinely differ.

This PR extracts the shared layer into a new `fs/text/code_point` module:

- `errorMask` — the invalid-sequence tag, previously duplicated verbatim in both modules.
- `decoder(byteOp, eofOp, init)` — a factory that wraps the two direction-specific
  steps into a single `List`-to-`List` conversion, owning the EOF sentinel, the
  unit-vs-EOF dispatch, and the `flat(stateScan(...))` plumbing.

`utf8` and `utf16` now build `toCodePointList` from the shared `decoder` and
import `errorMask`. The per-unit steps (`utf8ByteToCodePointOp`,
`utf16ByteToCodePointOp`) and EOF steps stay local, since they are the real
per-encoding delta.

Net: 48 insertions, 97 deletions.

## Test plan

- [x] `npx tsc` passes.
- [x] `npm test` passes (full suite, exit 0); `fs/text/code_point/module.f.ts` reports 100% line/branch/function coverage.
- [x] `cd fs/text && npm run fst` — all 57 text tests pass.

https://claude.ai/code/session_011MaSnTYeARm4t38QnuwRon

---
_Generated by [Claude Code](https://claude.ai/code/session_011MaSnTYeARm4t38QnuwRon)_


fs/text/code_point/module.f.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @param byteOp - The per-unit decoding step.
+ * @param eofOp - The end-of-input step that flushes the remaining state.
+ * @param init - The initial decoding state.
+ * @returns A function converting a list of code units into a list of code points.
+ */
+export const decoder = <Unit, S, Cp>(
+    byteOp: StateScan<Unit, S, List<Cp>>,
+    eofOp: (state: S) => readonly [List<Cp>, S],
+    init: S,
```
sergey-shandar: is it always `null`?

sergey-shandar: Yes — both `utf8` and `utf16` start from the empty (`null`) state, so it was always `null`. Dropped the `init` parameter and initialize internally to `null`; the state is now typed `S | null` so a null start stays enforced by the type system. Done in 031a6f9.

---

issues/README.md
```diff
@@ -318,8 +318,8 @@
...
-- [x] [167-bit-vec-msb-concat](./167-bit-vec-msb-concat.md). DRY: four modules (`crypto/sign`, `asn.1`, `sul/id`, `sul/level/literal`) each re-bind the identical `listToVec(msb)` under a different local name. Export the bound `msbConcat` from `bit_vec` so consumers import it instead of re-deriving.
-- [ ] [168-utf-codepoint-decoder](./168-utf-codepoint-decoder.md). DRY: `utf8` and `utf16` share a byte-for-byte streaming decoder skeleton (`eofList = [null]`, the unit-vs-EOF scan op, and `toCodePointList = flat(stateScan(op)(null)(flat([input, eofList])))`) plus a duplicated `errorMask`. Extract a `decoder(byteOp, eofOp, init)` factory and move `errorMask` to a shared code-point module.
+- [X] [167-bit-vec-msb-concat](./167-bit-vec-msb-concat.md). DRY: four modules (`crypto/sign`, `asn.1`, `sul/id`, `sul/level/literal`) each re-bind the identical `listToVec(msb)` under a different local name. Export the bound `msbConcat` from `bit_vec` so consumers import it instead of re-deriving.
```
sergey-shandar:
```suggestion
- [x] [167-bit-vec-msb-concat](./167-bit-vec-msb-concat.md). DRY: four modules (`crypto/sign`, `asn.1`, `sul/id`, `sul/level/literal`) each re-bind the identical `listToVec(msb)` under a different local name. Export the bound `msbConcat` from `bit_vec` so consumers import it instead of re-deriving.
```

sergey-shandar:
```suggestion
- [x] [167-bit-vec-msb-concat](./167-bit-vec-msb-concat.md). DRY: four modules (`crypto/sign`, `asn.1`, `sul/id`, `sul/level/literal`) each re-bind the identical `listToVec(msb)` under a different local name. Export the bound `msbConcat` from `bit_vec` so consumers import it instead of re-deriving.
```

PR #866, opened 2026-05-27
Add design issues 184–193: refactoring and abstraction proposals

sergey-shandar:
## Summary

This PR adds nine new design issues (184–193) to the `issues/` directory, documenting refactoring and abstraction opportunities across the codebase. These are investigative proposals for improving code reuse, separation of concerns, and type honesty.

## Issues Added

- **184** (`min`/`max`): Extract a generic comparable `min`/`max` from `operator` and `bigint` into `function/compare`, eliminating duplication across two consumers.

- **185** (`byte_set`): Reuse `bigint.mask` in `byte_set.range` instead of re-deriving the bit-mask arithmetic.

- **186** (`sul/id`): Reuse `sha2.fromV8` to pack SHA-2 word vectors into a bigint, removing local re-derivation.

- **187** (`asn.1`): Reuse `bigint.divUp`/`roundUp` for byte-rounding arithmetic and retype them from `Reduce` to `Unary` to reflect their config/operand split.

- **188** (`nullable`): Extract `fromUndefined` helper to centralize the `undefined → null` normalization used by `array.at` and `object.at`.

- **189** (`asn.1`): Introduce a `decodeAll` unfold combinator to eliminate duplicate "consume a `Vec` until empty" loops in `decodeObjectIdentifier` and `decodeSequence`.

- **190** (`fs/text`): Own the single code-unit/code-point ↔ string boundary by exposing `charFromCode`, `charFromCodePoint`, and reader helpers, eliminating scattered `String.*` re-bindings across `html`, `fsc`, `js/tokenizer`, and `bnf`.

- **191** (`bigfloat`): Extract a `withSign` helper to factor the "operate on magnitude, restore sign" envelope used by `round53` and `decToBin`.

- **192** (`effects/node`): Introduce an `errorExit` helper for the "print error, exit 1" pattern used by CLI entry points `cas` and `fjs`.

- **193** (`btree`): Investigate a shared `Path` fold engine for `set` and `remove` to unify their identical "rebuild parent branch at slot `{0,2,4}`" dispatch logic.

## Key Themes

- **DRY**: All issues identify code duplicated across two or more real consumers.
- **Separation of concerns**: Moving logic to its natural module home (e.g., bit arithmetic to `bigint`, text boundaries to `fs/text`).
- **Type honesty**: Correcting misleading type aliases (e.g., `divUp`/`roundUp` as `Reduce` when they are really `Unary` operations).
- **Naming patterns**: Extracting recurring idioms (e.g., `decodeAll`, `withSign`, `errorExit`) so future code reuses them.

## Notes

- Issue 193 is marked "investigate" rather than a mechanical refactor, pending confirmation that the two `Path` fold handlers can be unified without `as` casts.
- All other issues are straightforward extractions or reuses with clear consumers and minimal risk.
- These proposals align with the project's `AGENTS.md` guidance on avoiding scattered `String` references, re-bound helpers, and reaching into built-ins from `.f.ts` modules.

https://claude.ai/code/session_013zo9vHRPZj3voUo15a7M1A


issues/184-min-max-comparable.md
```diff
@@ -0,0 +1,74 @@
...
+
+| function | consumer |
+|---|---|
+| `operator.min`/`max` | `fs/types/number/module.f.ts:8` (`reduce(minOp)`/`reduce(maxOp)`) |
+| `bigint.min`/`max` | `fs/types/bit_vec/module.f.ts:24`, `fs/asn.1/module.f.ts:71` (`max`) |
+
+The only difference is the literal element type. `max` is even spelled two ways
+(`a > b ? a : b` vs `a < b ? b : a`) that are equivalent for a total order.
+
+## Proposed abstraction
```
sergey-shandar, line 35: Better proposal: use the same technique as `cmp` does:

```ts
export type Cmp1 = boolean | string | number | bigint

export type Cmp2<A, B> =
    [A, B] extends [boolean, boolean] ? boolean :
    [A, B] extends [string, string] ? string :
    [A, B] extends [number, number] ? number :
    [A, B] extends [bigint, bigint] ? bigint :
    never

export const cmp = <A extends Cmp1>(a: A) => <B extends Cmp2<A, B>>(b: B): Sign =>
    a as any < b ? -1 : a as any > b ? 1 : 0
```

Also, call these functions `min` and `max` and retire the old functions.

sergey-shandar: Done in 7a7da4f. Updated 184 to define `min`/`max` in `function/compare` with the same `Cmp1`/`Cmp2` + `a as any < b` technique `cmp` uses, and to retire the per-type copies in `operator` and `bigint` (consumers point at the single `compare.min`/`max`). Dropped the old `minBy`/`maxBy`-via-`cmp` framing.

---

sergey-shandar: Got it — fixed the emphasis in 14554bd. The proposal now makes the **signature** the point: `min`/`max` carry the same `Cmp1`/`Cmp2` generic guard as `cmp`, so `min(1)("a")` is a compile error. The body is called out as an implementation detail — likely just `cmp(a)(b) < 0 ? a : b` (no `as any` at all), with the inline `as any` form noted only as an equally-valid internal choice that doesn't affect the protective signature.

---

issues/187-byte-rounding-divup.md
```diff
@@ -0,0 +1,80 @@
...
+- Clarity/type-honesty: naming the curried first parameter as configuration
+  (`(divisor) => Unary`) rather than a fold accumulator documents the shape and
+  removes a latent misuse.
+
+## Caveats
+
+- The runtime behavior and curried shape are unchanged by the retype; only the
+  declared type alias changes. The `Reduce` import in `bigint` is still used by
+  `addition`/`xor`/etc.
+- Confirm no caller currently relies on `divUp`/`roundUp` being assignable to
```
sergey-shandar: If we replace `>>` with `divUp`, may it affect performance? Would be better to have something like `divUpE2` ?

sergey-shandar: Good catch — yes, it would. `asn.1` rounds with shifts (`>> 3n` / `<< 3n`); reusing `divUp(8n)` swaps that for a general bigint `/`, a needless slowdown on a power-of-two divisor.

Revised 187 in 80af8cf to add exactly your `divUpE2`: a shift-based, power-of-two variant parameterized by the exponent (reusing the existing `mask`):

```ts
export const divUpE2 = (e: bigint): Unary => { const m = mask(e); return v => (v + m) >> e }
export const roundUpE2 = (e: bigint): Unary => { const d = divUpE2(e); return v => d(v) << e }
```

`asn.1` uses `divUpE2(3n)`, and `crypto/sign`'s `divUp(8n)`/`roundUp(8n)` (also 8 = 2³) migrate to `divUpE2(3n)`/`roundUpE2(3n)` — second consumer, and a shift speedup for `sign` too. The general `divUp`/`roundUp` stay for `bit_vec`, which calls `divUp(n << 1n)` — a non-power-of-two divisor — and both still get the `Reduce` → `(b) => Unary` retype.

---

PR #885, opened 2026-05-30
effects: add foldStep/forEachStep combinators (i209)

sergey-shandar:
Resolves [i209](./issues/209-effect-fold-step.md).

## Summary

- Adds `foldStep` and `forEachStep` to `fs/types/effects/module.f.ts` — sequential state-threading and void-accumulator siblings of the existing parallel `all` combinator.
- Replaces three open-coded variants of the same pattern:
  - `fs/dev/tf/module.f.ts` — `modules.reduce((acc, [k,v]) => acc.step(runModule(...)(k,v)), pure(zero))`
  - `fs/djs/transpiler/module.f.ts` — `pathsArray.reduce<Effect<ReadFile, ParseContext>>(...)`
  - `fs/cas/module.f.ts` — mutable `let i = begin; for (const j of v) { i = ... }` with a `TODO: make it lazy` comment (now gone)
- Adds proof tests at `fs/types/effects/proof.f.ts`.

## Design refinement vs the issue

The issue proposed an uncurried `f: (item, state) => Effect<O, S>` shape. Both existing reduce sites already had **curried** inner functions (`runModule(reporter)(k, v)` returns `(state) => Effect`, `foldNextModuleOp(p)` returns `(context) => Effect`), so the implementation uses the curried form `f: item => state => Effect<O, S>`. This lets the call sites pass the existing functions through directly:

```ts
foldStep<ReadFile, string, ParseContext>(foldNextModuleOp)(contextWithStack)(pathsArray)
```

instead of needing an eta-wrapper `(p, ctx) => foldNextModuleOp(p)(ctx)`.

## Test plan

- [x] `npm run fst` — 1480 pass / 0 fail (1475 baseline + 5 new proof tests)
- [x] `npm test` — 1472 pass / 8 fail (same 8 pre-existing throw-test failures on the Node `--test` path; not related to this change)

---
_Generated by [Claude Code](https://claude.ai/code/session_01F44MMzHin2AJNPixRM8H7x)_


fs/types/effects/module.f.ts
```diff
@@ -56,6 +56,34 @@ export const do_ =
...
+ * builds `f(x₀)(init).step(f(x₁)).step(f(x₂)).…` and yields a single
+ * `Effect<O, S>` that produces the final state.
+ *
+ * Sequential — each step depends on the previous state. Compare to `all`,
+ * which fans out independent effects.
+ */
+export const foldStep =
+    <O extends Operation, T, S>(f: (item: T) => (state: S) => Effect<O, S>) =>
+    (init: S) =>
+    (items: readonly T[]): Effect<O, S> =>
```
sergey-shandar: We can use `reduce` from `./fs/types/list/module.f.ts` and then accept `List<T>` type, which is superset of `readonly T[]`.

fs/dev/tf/module.f.ts
```diff
@@ -211,10 +212,8 @@ const { entries } = Object
 export const runModuleMap = <O extends Operation>(reporter: Reporter<O>) => (moduleMap: ModuleMap): Effect<O | All, number> => {
     const { summary } = reporter
     const modules = entries(moduleMap).filter(([k]) => isTest(k))
-    return modules.reduce(
-        (acc: Effect<O | All, TestState>, [k, v]) => acc.step(runModule(reporter)(k, v)),
-        pure({ time: 0, pass: 0, fail: 0 })
-    )
+    return all(...modules.map(([k, v]) => runModule(reporter)(k, v)(zero)))
+    .step(m => pure(m.reduce(mergeState, zero)))
```
sergey-shandar, line 215: In some cases, when we don't care about order, we need `allReduce` instead of `foldStep`. Create an issue to introduce `allReduce` (or if you have a better name).

PR #901, opened 2026-05-31
docs: add three P4 structural refactoring proposals

sergey-shandar:
Add three issue proposals documenting opportunities to reduce code duplication at the structural level across the codebase.

## Summary

These three issues identify parallel implementations of the same control-flow patterns in different parts of the system, where extracting a shared abstraction would improve maintainability and lock down implicit contracts.

## Issues Added

- **65Z-effect-runner-stepper**: Extract the effect-system trampoline loop (the `while (true)` driver that peels operations off effects and continues) into a shared `drive` combinator. Both `asyncRun` and the mock `run` independently implement identical loop logic that differs only in how a single operation is dispatched and how state is threaded. Centralizing this pins down the effect-runner contract and prevents future runners (e.g., worker-based sandbox) from reimplementing the same loop.

- **65Z-tf-test-tree-walker**: Factor the dynamic test-tree traversal (collect leaves, recurse into function-return sub-trees, fan out with `all`, merge results) into a shared `walkTests` combinator. Both `runModule` and `registerModule` independently walk the same tree structure with identical recursion logic but different per-leaf actions. Lifting the traversal separates the recursion structure from the per-leaf reducer and documents the contract for future test reporters or instrumenters.

- **65Z-asn1-tag-codec-table**: Collapse the parallel encode/decode tag switches in `recordToRaw` and `rawToRecord` into a single codec table keyed by tag. Currently every supported tag must be maintained in two separate `switch` statements; a table-based approach makes it impossible to drop a tag from only one side and ensures new tags are added to both encode and decode in one place.

## Rationale

All three follow the same pattern: two or more implementations of the same structural logic, held in sync by convention rather than by the type system. Each proposal:
- Identifies the shared shape (loop, traversal, dispatch table)
- Proposes extracting it as a named abstraction
- Explains why centralizing the contract matters for future maintainability

Per `AGENTS.md`, these are speculative refactorings (P4 priority) that should only ship when the abstraction makes existing code shorter and clearer, not on the promise of future consumers.

https://claude.ai/code/session_019qmhyav99jaTfc39J8fozT


issues/65Z-effect-runner-stepper.md
```diff
@@ -0,0 +1,151 @@
...
+  with `S = void`) is one option; a single combinator with two callbacks
+  (`onPure`, `runOp`) is another.
+- TypeScript's variance on the existential `cmd` makes the type plumbing
+  here non-trivial. A small spike is warranted before committing to a
+  signature.
+- The `await` keyword sits inside the loop in `asyncRun`. If the shared
+  driver is a synchronous function, the async runner needs to wrap each
+  iteration in a continuation-returning callback (or the driver becomes
+  async-aware itself). Either is fine; pick whichever keeps `asyncRun`
+  short.
```
sergey-shandar: I'm not sure how you can use synchronous `drive` to `await` operation. The syntax looks the same, but we can't access the results of async operations from within a synchronous function.

sergey-shandar: You're right — a synchronous `drive` cannot `await`; the sketch was unimplementable. Updated the proposal in 6725756 to split into two honest options: (A) extract only the pure-tuple classifier and let each runner keep its own loop, or (B) share an async-only `driveAsync` and leave the sync mock alone. Added the sync-cannot-host-async point at the top of the caveats list.

---

sergey-shandar: Agreed — `asyncDrive` can't help the mock either, so Option B doesn't dedupe anything that matters, and Option A is too small to justify a shared module. Marked the issue **done** with a "will not fix" decision section explaining the rationale (42f8759). It can be deleted in the next cleanup pass.

---

sergey-shandar: Dropped the file entirely in 8a9db62. PR now ships only the two remaining proposals (ASN.1 codec table, tf test-tree walker).

---

PR #906, opened 2026-05-31
tf: update README to reflect proof export and current file conventions

sergey-shandar:
- Fix typo all.tets.ts → all.test.ts; correct import to module.js
- Replace *.test.f.ts / default export with *.proof.f.ts / proof export
- Document vanilla TS/JS opt-in naming (proof.ts, *.proof.ts, etc.)
- Add fjs t and playwright to the supported-runner list; fix deno flag order
- Fix invalid throw-by-name example (throw is a reserved word)

https://claude.ai/code/session_01SsSDUP8rdqFtzQiPbawpyb

### DIFF
--- a/fs/dev/tf/README.md
+++ b/fs/dev/tf/README.md
@@ -1,5 +1,19 @@
 # Test Framework
 
+## Installation
+
+Install from npm:
+
+```sh
+npm install functionalscript
+```
+
+or from JSR:
+
+```sh
+npx jsr add @functionalscript/functionalscript
+```
+
 ## Running tests
 
 ### `fjs t` (built-in runner)
@@ -12,7 +26,7 @@ fjs t
 
 ### External runners (Node, Bun, Deno, Playwright)
 
-For external runners, create a file the runner will pick up (e.g. `*.test.ts`, which Node/Bun/Deno/Playwright all load by default):
+For external runners, create an entry file that matches the runner's discovery pattern (many runners pick up files matching `*.test.ts` or `*.test.js` by default):
 
 ```ts
 import { run } from 'functionalscript/fs/dev/tf/module.js'
@@ -26,6 +40,8 @@ Then invoke the runner:
 - `deno test --allow-read --allow-env`
 - `npx playwright test`
 
+Developers can also implement their own runners as long as they follow the proof-tree conventions described below.
+
 ## Design: Dependency-Free Tests
 
 Unlike most test frameworks (Jest, Mocha, Vitest, …), test files do **not** import anything from the test framework. A test is just a plain function or object — no `describe`, `it`, `expect`, or assertion library required. This means:
@@ -37,22 +53,22 @@ Unlike most test frameworks (Jest, Mocha, Vitest, …), test files do **not** im
 
 ### Discovery: the `proof` export
 
-A module is treated as a test module if and only if it exports a property named **`proof`**. Discovery is by value — "does the module export `proof`?" — not by filename alone.
+A module is treated as a test module if and only if it exports a property named **`proof`**. Discovery is by value — "does the module have a proof?" — not by filename alone.
 
 The framework loads modules in two tiers:
 
 | Language | Load gate |
 |----------|-----------|
 | FunctionalScript (`.f.ts` / `.f.js`) | **all** files are loaded — FS modules have no import side effects by construction |
-| Vanilla TypeScript / JavaScript | opt-in by filename: files whose name ends in `proof.ts`, `proof.js`, `proof.mts`, or `proof.mjs` (e.g. `math.proof.ts`, `proof.ts`) |
+| Vanilla TypeScript / JavaScript | opt-in by filename: files whose name ends in `proof.ts`, `proof.js`, `proof.mts`, or `proof.mjs` (e.g. `math.proof.ts`, `proof.js`) |
 
 A loaded file that does not export `proof` is silently skipped; filename alone never causes a file to be executed as a test suite.
 
-This design is intentional: keeping "is this a proof?" as a property of the *module's value* rather than its path means proofs remain self-describing even when files are stored by content hash (no stable filename) in a Merkle DAG.
+This design is intentional: keeping "does the module have a proof?" as a property of the *module's value* rather than its path means proofs remain self-describing even when files are stored by content hash (no stable filename) in a Merkle DAG, or when source code is published and copied as text.
 
 ### The `proof` export
 
-The named export `proof` is the test tree. A **leaf test** is a zero-argument function (`f.length === 0`); functions with parameters are ignored and not called.
+The named export `proof` is the test tree. A zero-argument function (`f.length === 0`) is a test node; functions with parameters are ignored and not called. A test node passes if it returns normally, and fails if it throws. Its return value may itself contain further test nodes (see [Return value as sub-tree](#return-value-as-sub-tree)).
 
 ```ts
 export const proof = {
@@ -63,7 +79,7 @@ export const proof = {
 
 ### Nested objects
 
-Objects (and arrays) are traversed recursively. Each key becomes a path segment in the output:
+Objects (and arrays) are traversed recursively. Each key becomes a path segment in the output. Only **own enumerable keys** are visited (as returned by `Object.entries`); prototype properties are excluded.
 
 ```ts
 export const proof = {
@@ -101,6 +117,21 @@ export const proof = {
 }
 ```
 
+This is useful for generating tests dynamically. The following example produces one named sub-test per input case:
+
+```ts
+const cases: readonly [number, number, number][] = [[1, 1, 2], [0, 0, 0], [2, 3, 5]]
+
+export const proof = {
+    add: () => Object.fromEntries(
+        cases.map(([a, b, expected]) => [
+            `${a}+${b}`,
+            () => { if (a + b !== expected) throw `${a}+${b} !== ${expected}` },
+        ])
+    ),
+}
+```
+
 Only **return values** of non-throw tests are walked as sub-trees, and the `throw` flag always resets to false for the sub-tree. Thrown values are discarded and never traversed, even if they are objects containing zero-parameter functions.
 
 ## Convention: only real Promises are awaited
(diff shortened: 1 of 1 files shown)


fs/dev/tf/README.md
```diff
@@ -1,18 +1,20 @@
 # Test Framework
 
-To enable testing in your project, you can add a simple test runner file `all.tets.ts` like this:
+To enable testing in your project, add a test runner file `all.test.ts` like this:
```
sergey-shandar: Actually, any file that will be loaded by the desired test runner. For example, most test runners load `*.test.ts`, `*.test.js`.

sergey-shandar: BTW, `fjs` doesn't require the file.

fs/dev/tf/README.md
```diff
@@ -23,14 +25,21 @@ Unlike most test frameworks (Jest, Mocha, Vitest, …), test files do **not** im
 
 ## Writing Tests
```
sergey-shandar, line 52: [This document](./issues/65Y-proof-by-export.md) can be used as a good source to fill the document with details.

fs/dev/tf/README.md
```diff
@@ -54,27 +63,27 @@ export default {
...
+export const proof = {
     throw: {
         divByZero: () => { throw new Error('division by zero') },  // passes
         noThrow: () => 42,                                          // fails
     },
 }
 ```
 
-A function named `throw` is also recognised as a throw-test regardless of its position in the tree:
+A function whose `.name` property is `'throw'` is also recognised as a throw-test regardless of its position in the tree. Because `throw` is a reserved word it cannot be used as a variable name directly; the typical way to obtain such a function is via object property access:
```
sergey-shandar: This is out of date. We no longer use the function's property name as a sign that the function is expected to throw. However, if a path to the function has a `throw` property (for example `import('proof.ts').proof[5].throw.my()`, we consider such a function as expected to throw.

fs/dev/tf/README.md
```diff
@@ -54,27 +63,27 @@ export default {
...
 ```ts
-const throw = () => { throw 'expected' }
-export default { b: throw }   // passes — function name triggers throw semantics
+const t = { throw: () => { throw 'expected' } }.throw
+export const proof = { b: t }   // passes — function name triggers throw semantics
 ```
 
 ### Return value as sub-tree
 
 When a non-throw test function returns an object or another function, the return value is walked as a fresh sub-tree. This allows lazy or computed test trees:
```
sergey-shandar, line 114: Make sure that the document mentions that only functions with zero arguments (`f.length === 0`) are called. All other functions are skipped.

fs/dev/tf/README.md
```diff
@@ -1,18 +1,30 @@
 # Test Framework
```
sergey-shandar, line 2: Should we add that a user should install the FunctionalScript first, for example, using either NPM or JSR package? But also that developers can use third-party or their own runners if the runners follow the conventions described below.

fs/dev/tf/README.md
```diff
@@ -1,18 +1,30 @@
...
+
+FunctionalScript's own runner discovers all proof modules automatically — no entry file required:
+
+```sh
+fjs t
+```
+
+### External runners (Node, Bun, Deno, Playwright)
+
+For external runners, create a file the runner will pick up (e.g. `*.test.ts`, which Node/Bun/Deno/Playwright all load by default):
```
sergey-shandar: `*.test.ts` is not a file name example; it's one of the possible patterns. I would rephrase it.

fs/dev/tf/README.md
```diff
@@ -23,25 +35,38 @@ Unlike most test frameworks (Jest, Mocha, Vitest, …), test files do **not** im
...
+### Discovery: the `proof` export
+
+A module is treated as a test module if and only if it exports a property named **`proof`**. Discovery is by value — "does the module export `proof`?" — not by filename alone.
+
+The framework loads modules in two tiers:
+
+| Language | Load gate |
+|----------|-----------|
+| FunctionalScript (`.f.ts` / `.f.js`) | **all** files are loaded — FS modules have no import side effects by construction |
+| Vanilla TypeScript / JavaScript | opt-in by filename: files whose name ends in `proof.ts`, `proof.js`, `proof.mts`, or `proof.mjs` (e.g. `math.proof.ts`, `proof.ts`) |
```
sergey-shandar:
```suggestion
| Vanilla TypeScript / JavaScript | opt-in by filename: files whose name ends in `proof.ts`, `proof.js`, `proof.mts`, or `proof.mjs` (e.g. `math.proof.ts`, `proof.js`) |
```
To show some variances.

fs/dev/tf/README.md
```diff
@@ -23,25 +35,38 @@ Unlike most test frameworks (Jest, Mocha, Vitest, …), test files do **not** im
...
+
+A loaded file that does not export `proof` is silently skipped; filename alone never causes a file to be executed as a test suite.
 
-### Functions
+This design is intentional: keeping "is this a proof?" as a property of the *module's value* rather than its path means proofs remain self-describing even when files are stored by content hash (no stable filename) in a Merkle DAG.
```
sergey-shandar: or source code published, copied as text.

fs/dev/tf/README.md
```diff
@@ -23,25 +35,38 @@ Unlike most test frameworks (Jest, Mocha, Vitest, …), test files do **not** im
...
+export const proof = {
+    myTest: () => 42,                      // passes — returns normally
+    failing: () => { throw 'oops' },       // fails — throws
 }
 ```
 
 ### Nested objects
 
-Objects are traversed recursively. Each key becomes a path segment in the output:
+Objects (and arrays) are traversed recursively. Each key becomes a path segment in the output:
```
sergey-shandar: Add that we consider only own keys, prototype properties are excluded.

fs/dev/tf/README.md
```diff
@@ -23,25 +35,38 @@ Unlike most test frameworks (Jest, Mocha, Vitest, …), test files do **not** im
...
+
+A loaded file that does not export `proof` is silently skipped; filename alone never causes a file to be executed as a test suite.
 
-### Functions
+This design is intentional: keeping "is this a proof?" as a property of the *module's value* rather than its path means proofs remain self-describing even when files are stored by content hash (no stable filename) in a Merkle DAG.
 
-A zero-parameter function is a test case. It passes if it returns without throwing:
+### The `proof` export
+
+The named export `proof` is the test tree. A **leaf test** is a zero-argument function (`f.length === 0`); functions with parameters are ignored and not called.
```
sergey-shandar: It's not really a leaf because it may generate new tests.

fs/dev/tf/README.md
sergey-shandar: "is this a proof?" Would it be better to ask "does the module have a proof?"

fs/dev/tf/README.md
```diff
@@ -23,25 +35,38 @@ Unlike most test frameworks (Jest, Mocha, Vitest, …), test files do **not** im
...
 
 ```ts
-export default {
-    myTest: () => 42,         // passes — returns normally
-    failing: () => { throw 'oops' },  // fails — throws unexpectedly
+export const proof = {
+    myTest: () => 42,                      // passes — returns normally
+    failing: () => { throw 'oops' },       // fails — throws
 }
```
sergey-shandar, line 77: Make another example to demonstrate practical test generation, for example, generate multiple tests dynamically.

fs/dev/tf/README.md
```diff
@@ -90,3 +110,19 @@ When a test function returns a value, the framework checks `value instanceof Pro
...
 As a consequence, **exporting a function named `then` from a test module is forbidden**: the module namespace object would become a thenable, corrupting dynamic `import()` resolution. See [issues/lang/3240-export.md](../../issues/lang/3240-export.md).
+
+## Proof location and scope
+
+A proof's *location* determines what it can see. Three tiers exist:
+
+| Mechanism | Scope | Runs when | Use for |
+|-----------|-------|-----------|---------|
+| Module-level `assertEq(...)` | public + private | **every module load** | light, cheap, deterministic checks only |
+| `export const proof` co-located in module | public + private | under the runner | white-box unit proofs |
```
sergey-shandar: It would be good to have links to white-box black-box description. For example, on Wikipedia.

fs/dev/tf/README.md
sergey-shandar: ...don't need to import anything... ?

fs/dev/tf/README.md
```diff
@@ -54,33 +95,43 @@ export default {
...
+    add: () => Object.fromEntries(
+        cases.map(([a, b, expected]) => [
+            `${a}+${b}`,
+            () => { if (a + b !== expected) throw `${a}+${b} !== ${expected}` },
+        ])
+    ),
+}
+```
+
 Only **return values** of non-throw tests are walked as sub-trees, and the `throw` flag always resets to false for the sub-tree. Thrown values are discarded and never traversed, even if they are objects containing zero-parameter functions.
```
sergey-shandar: "and the `throw` flag always resets to false for the sub-tree" that doesn't make much sense considering that only non-throw functions are walked. I would suggest removing it.

fs/dev/tf/README.md
```diff
@@ -1,18 +1,46 @@
...
+Install from npm:
+
+```sh
+npm install functionalscript
+```
+
+or from JSR:
+
+```sh
+npx jsr add @functionalscript/functionalscript
```
sergey-shandar: A better example would be Deno instead of `npx jsr`.

fs/dev/tf/README.md
```diff
@@ -23,25 +51,38 @@ Unlike most test frameworks (Jest, Mocha, Vitest, …), test files do **not** im
...
 
-Objects are traversed recursively. Each key becomes a path segment in the output:
+Objects (and arrays) are traversed recursively. Each key becomes a path segment in the output. Only **own enumerable keys** are visited (as returned by `Object.entries`); prototype properties are excluded.
 
 ```ts
-export default {
+export const proof = {
     math: {
         add: () => 1 + 1,
         mul: () => 2 * 2,
```
sergey-shandar: Add an array to the example.

fs/dev/tf/README.md
sergey-shandar: I mean, the test/proof files may import something but they don't have to.

fs/dev/tf/README.md
```diff
@@ -23,29 +51,46 @@ Unlike most test frameworks (Jest, Mocha, Vitest, …), test files do **not** im
...
 ### Nested objects
 
-Objects are traversed recursively. Each key becomes a path segment in the output:
+Objects (and arrays) are traversed recursively. Each key becomes a path segment in the output. Only **own enumerable keys** are visited (as returned by `Object.entries`); prototype properties are excluded.
 
 ```ts
-export default {
+export const proof = {
     math: {
         add: () => 1 + 1,
```
sergey-shandar: Can we add some asserts to the tests, for example `if (1 + 1 !== 2) throw 'an impossible error!'

PR #918, opened 2026-06-01
docs: add issues 661 for effect decoder and path normalization

sergey-shandar:
## Summary

Add two issue documents proposing refactorings to improve code organization and reduce duplication in the effects system and path handling.

## Changes

- **661-effect-value-decoder.md**: Proposes extracting a single `node` decoder function to centralize the tuple-based representation of `Effect` values. Currently, the discriminant logic (`value.length === 1` for pure vs. do nodes) and tuple destructuring are duplicated across three interpreter loops and five assertion sites. The proposal adds a `Node<O, T>` discriminated union type and `node()` accessor to the core effects module, allowing all consumers to use named fields instead of positional indices.

- **661-path-posix-separators.md**: Proposes extracting the Windows-to-POSIX separator conversion (`replaceAll('\\', '/')`) into a dedicated `toPosix` export from `fs/path/module.f.ts`. Currently, this logic is duplicated inline in both `fs/path/parse` and `fs/dev/loadModuleMap`, violating the principle that path manipulation belongs in `fs/path`.

Both issues follow the DRY and separation-of-concerns principles outlined in `AGENTS.md`, with clear scope and implementation guidance.

https://claude.ai/code/session_01V6cg2QoBUETUGhWTn7zgT8

### DIFF
--- a/issues/661-effect-value-decoder.md
+++ b/issues/661-effect-value-decoder.md
@@ -109,11 +109,11 @@ places**, none of which is the module that owns the type.
 ## Proposed decoder
 
 Add one accessor to the core module that turns the positional tuple into a
-named, discriminated node. It is the only place that knows the layout:
+named, discriminated form. It is the only place that knows the layout:
 
 ```ts
 // fs/types/effects/module.f.ts
-export type Node<O extends Operation, T> =
+export type Decoded<O extends Operation, T> =
     | { readonly done: true, readonly result: T }
     | {
         readonly done: false,
@@ -122,8 +122,8 @@ export type Node<O extends Operation, T> =
         readonly continuation: Do<O, T>[2],
     }
 
-/** Decodes an effect's next node: a pure result, or a command to perform. */
-export const node = <O extends Operation, T>({ value }: Effect<O, T>): Node<O, T> =>
+/** Decodes an effect's next step: a pure result, or a command to perform. */
+export const decode = <O extends Operation, T>({ value }: Effect<O, T>): Decoded<O, T> =>
     value.length === 1
         ? { done: true, result: value[0] }
         : { done: false, command: value[0], payload: value[1], continuation: value[2] }
@@ -135,20 +135,20 @@ vs. state-thread) differs, which is exactly the part that *should* differ:
 ```ts
 // async runner
 while (true) {
-    const n = node(effect)
-    if (n.done) { return n.result }
-    effect = n.continuation(await map[n.command](...n.payload))
+    const d = decode(effect)
+    if (d.done) { return d.result }
+    effect = d.continuation(await map[d.command](...d.payload))
 }
 ```
 
 ```ts
 // mock runner
 while (true) {
-    const n = node(e)
-    if (n.done) { return [s, n.result] }
-    const [ns, m] = o[n.command](s, ...n.payload)
+    const d = decode(e)
+    if (d.done) { return [s, d.result] }
+    const [ns, m] = o[d.command](s, ...d.payload)
     s = ns
-    e = n.continuation(m)
+    e = d.continuation(m)
 }
 ```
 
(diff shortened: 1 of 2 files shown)


issues/661-effect-value-decoder.md
```diff
@@ -0,0 +1,198 @@
...
+  pure value vs. a command — not *how to tell them apart*.
+
+## Proposed decoder
+
+Add one accessor to the core module that turns the positional tuple into a
+named, discriminated node. It is the only place that knows the layout:
+
+```ts
+// fs/types/effects/module.f.ts
+export type Node<O extends Operation, T> =
```
sergey-shandar: Node can be confused with Node.js. We need a better name.

sergey-shandar: Renamed `Node`/`node` → `Decoded`/`decode` to avoid confusion with Node.js (3783635).

---

issues/661-effect-value-decoder.md
```diff
@@ -0,0 +1,198 @@
...
+Both runners *already* write `const [command, payload, continuation] = value`
+and then call `operation(...payload)`; that compiles today against the
+decorrelated union types of `Do<O, T>`. `decode` returns exactly those same
+three bindings under names, so the type situation is unchanged — it only names the
+discriminant. No `as` casts are introduced; the length check narrows the tuple
+union structurally (no type predicate needed), in line with `AGENTS.md`.
+
+## Why not a `match`/`fold` combinator
+
+A `match(onPure, onDo)` higher-order form reads well for pure expression code
```
sergey-shandar: What about this idea:

```ts
const match = map => effect => {
    const d = decode(effect) // we can inline the function code, because we don't need it anymore.
    return d.done ? ['done', d.result] : ['cont', map[d.command](...d.payload)]
}
```

Then sync cycle:

```ts
const ma = match(o)
while (true) {
    const [tag, result] = ma(e)
    if (tag === 'done') { return [s, result] }
    const [ns, m] = result(s) // note we change the operation function signature from (s, ...payload) to (...payload)(s) to achieve the match
    s = ns
    e = d.continuation(m)
}
```

Then async cycle:

```ts
const ma = match(map)
while (true) {
    const [tag, result] = ma(effect)
    if (tag === 'done') { return result }
    effect = d.continuation(await result)
}
```

In both cases, we apply an additional operation after the match, either update a state `const [ns,m] = result(s)` or await the result.

sergey-shandar: Adopted this — much better, since it shares the `map[command](...payload)` dispatch too, not just the discriminant. Updated the proposal (93b45f4):

- `match(map)(effect)` returns `['done', result]` or `['cont', operationOutput, continuation]`, with `decode` as its internal helper.
- The async/sync split collapses to one eliminator line per loop — `r[2](await r[1])` vs `r[2](r[1](s))` — exactly as you sketched.
- Curried the mock op map to `(...payload) => (state) => [state, result]`, and noted it as a breaking change whose one consumer is the virtual FS (`node/virtual/module.f.ts`), so that migration is called out in the scope section (or `decode` can land first, currying as a follow-up).

I dropped my earlier "why not match/fold" section — it was wrong; deferring the eliminator past the dispatch is what makes it work.

---

issues/661-effect-value-decoder.md
```diff
@@ -0,0 +1,232 @@
...
+re-do the `map[command](...payload)` dispatch. Per @sergey-shandar's review, we
+can share that too with a `match` combinator that decodes *and* dispatches,
+returning the operation's output while deferring the one world-specific step —
+`await` (async) or state-threading (sync) — to the caller:
+
+```ts
+// fs/types/effects/module.f.ts
+export const match =
+    <O extends Operation, R>(map: { readonly [K in O[0]]: (...payload: Pr<O, K>[0]) => R }) =>
+    <T>(effect: Effect<O, T>): readonly['done', T] | readonly['cont', R, Do<O, T>[2]] => {
```
sergey-shandar:
```ts
type Map<...> = { readonly [K in O[0]]: (...payload: Pr<O, K>[0]) => R })
type MatchResult<...> = readonly['done', T] | readonly['cont', R, Do<O, T>[2]]
```

sergey-shandar: Done (4278f73) — lifted both into named aliases:

```ts
export type OperationMap<O extends Operation, R> = {
    readonly [K in O[0]]: (...payload: Pr<O, K>[0]) => R
}
export type MatchResult<O extends Operation, T, R> =
    | readonly['done', T]
    | readonly['cont', R, Do<O, T>[2]]
```

I named the map `OperationMap` rather than `Map` to avoid shadowing the global `Map` (same spirit as the `Node` rename). Bonus: `OperationMap<O, R>` generalizes both existing maps — `ToAsyncOperationMap` is `OperationMap<O, Promise<…>>` and the curried `MemOperationMap` is `OperationMap<O, (state) => [state, …]>` — so they could be redefined in terms of it. Let me know if you'd prefer the plain `Map` name anyway.

---

issues/661-effect-value-decoder.md
```diff
@@ -0,0 +1,244 @@
...
+    }
+```
+
+What unifies async and sync is the *shape* `R` of the operation's output: if both
+operation maps return a value the loop finishes with a single world-specific
+eliminator, the two loops become the same skeleton plus one line.
+
+```ts
+// async runner — map: (...payload) => Promise<result>, so R = Promise<...>
+const ma = match(map)
```
sergey-shandar: Can we have a better name instead of `ma`?

sergey-shandar: Renamed `ma` → `next` in both runner examples (6337078) — it's `match(map)` partially applied, so `next(effect)` reads as "advance one step." Avoided `step` since that's already the method name on `Effect`.

---

PR #940, opened 2026-06-04
function/compare: add generic min/max, retire per-type copies (i184)

sergey-shandar:
Fixes [i184](./issues/184-min-max-comparable.md).

## Summary

- Add `min`/`max` to `fs/types/function/compare/module.f.ts` next to `cmp`, reusing the same `Cmp1`/`Cmp2<A, B>` guard so call sites mixing incompatible primitive types (e.g. `min(1)("a")`) fail to compile.
- Retire the duplicated `Reduce<number>`-typed `min`/`max` from `function/operator` and the bigint-typed `min`/`max` from `types/bigint`.
- Point the four consumers (`types/number`, `types/bit_vec`, `asn.1`, `types/bigint/proof`) at the single generic pair. `number.min`/`max` instantiate as `min<number>`/`max<number>` so `reduce` can infer the fold type.
- Tests for `min`/`max` move from `operator/proof.f.ts` to `compare/proof.f.ts`, extended with `bigint` and `string` cases plus commented mixed-type calls that document the compile-time guard.

## Test plan

- [x] `npx tsc` — passes
- [x] `npm test` — same 10 pre-existing failures as `main` (all in `nanvm-lib` / `fs/io`, unrelated)
- [x] New `min`/`max` cases in `compare/proof.f.ts` pass

---
_Generated by [Claude Code](https://claude.ai/code/session_013KbYwgau11sYbmtcVV3ZFN)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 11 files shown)


sergey-shandar:
Update CHANGELOG.md

PR #948, opened 2026-06-05
Add issue 665: formalize command line parsing with command maps

sergey-shandar:
https://claude.ai/code/session_01C4seTrLV49wpAnY1sjmiKb


issues/665-command-line-parsing-refactor.md
```diff
@@ -0,0 +1,68 @@
...
+### Step 1 — Define a `CommandMap` type
+
+```ts
+type CommandEntry = {
+  readonly aliases?: readonly string[]
+  readonly description: string
+  readonly handler: (args: readonly string[]) => Effect<NodeOp, number>
+}
+
+type CommandMap = { readonly [command: string]: CommandEntry }
```
sergey-shandar: I would prefer that `CommandEntry` be self-descriptive. It means it should contain all aliases, and we store it as `type Commands = readonly CommandEntry[]`. Then we transform it into a map where each `alias` is present.

issues/665-command-line-parsing-refactor.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Proposal
+
+### Step 1 — Define a `CommandMap` type
+
+```ts
+type CommandEntry = {
+  readonly aliases?: readonly string[]
+  readonly description: string
+  readonly handler: (args: readonly string[]) => Effect<NodeOp, number>
```
sergey-shandar, line 28: handler is either an effect function or a next set of commands. However, we may do only an effect function on the first iteration.

PR #954, opened 2026-06-05
Refactor SHA2 sigma functions to use shared factory pattern

sergey-shandar:
## Summary
Refactored the SHA2 sigma function implementations to eliminate code duplication by introducing a shared `sigma` factory function that both `bigSigma` and `smallSigma` use.

## Key Changes
- Introduced a higher-order `sigma` factory function that accepts a transformation function as a parameter
- The factory abstracts the common pattern of applying three rotation/shift operations and XORing the results
- `bigSigma` now uses `sigma(rotr)` to apply rotation to all three parameters
- `smallSigma` now uses `sigma(c => x => x >> c)` to apply rotation to the first two parameters and right-shift to the third
- Removed duplicate function bodies that were previously defined separately for `bigSigma` and `smallSigma`

## Implementation Details
The refactoring maintains identical behavior while reducing code duplication. The `sigma` factory takes a higher-order function that determines how the third parameter `c` is transformed, allowing both sigma variants to share the same core logic. This improves maintainability and makes the relationship between the two functions explicit.

https://claude.ai/code/session_01USXwPNbBkemA3DsEZTszB2


fs/crypto/sha2/module.f.ts
```diff
@@ -74,23 +74,23 @@ const base = ({ logBitLen, k, bs0, bs1, ss0, ss1 }: BaseInit): Base => {
...
     }
 
-    const bigSigma = ([a, b, c]: V3) => {
-        const ra = rotr(a)
-        const rb = rotr(b)
-        const rc = rotr(c)
-        return (x: bigint) => ra(x) ^ rb(x) ^ rc(x)
-    }
+    const sigma =
+        (third: (c: bigint) => (x: bigint) => bigint) =>
```
sergey-shandar: This should be `(third: Binary<bigint>) =>`

PR #965, opened 2026-06-06
Add issue templates for refactoring and cleanup tasks

sergey-shandar:
## Summary

This PR adds five new issue templates documenting refactoring and cleanup opportunities across the codebase. These are design-level improvements that consolidate duplicated logic, improve separation of concerns, and align inconsistent implementations with shared contracts.

## Issues Added

- **666-json-schema-visit**: Route `toJsonSchema` through the shared `rtti/common.visit` kernel, eliminating a fourth hand-rolled walker over the `Type` ADT and gaining exhaustiveness safety (P3)

- **666-crypto-sign-fromcurve**: Eliminate duplicate RFC6979 helper derivation in `sign` by using the existing `fromCurve` abstraction, and extract the "bits2int then modular reduction" step into a named `bits2intModQ` helper (P4)

- **666-js-tokenizer-position-layer**: Separate position/metadata tracking from the JS tokenizer core by exposing a raw `tokenizeRaw` entry point, allowing JSON tokenizer to avoid the dummy-path workaround (P4)

- **666-utf16-encode-errormask**: Align UTF-16 encoder's invalid code point handling with UTF-8's by preserving the `errorMask` tag instead of silently truncating (P5)

- **666-utf8-continuation-helpers**: Name the repeated UTF-8 continuation-byte bit patterns (`contByte`, `contPayload`) and lead-byte tag/mask pairs to eliminate ~20 instances of hand-written masks across four functions (P5)

## Notes

All issues are marked `open` and represent opportunities for incremental cleanup. The highest-priority item (666-json-schema-visit) is proposed to land together with issue i662 as part of a broader effort to route all remaining `Type`-ADT printers through the shared `visit` kernel.

https://claude.ai/code/session_012wgn4FFtV3DjZ8uzXNZx8y


issues/666-crypto-sign-fromcurve.md
```diff
@@ -0,0 +1,76 @@
...
+   ```
+
+   RFC6979's "extra modular reduction" (documented in the comment at `:148-153`)
+   thus appears twice, with the rationale comment split across the two sites.
+
+## Proposal
+
+1. In `sign`, replace `const a = all(q)` with `const a = fromCurve(c)` and read
+   `q` (and `bits2int`) from `a`; keep destructuring only `div` and `g` from `c`,
+   which `fromCurve` does not provide:
```
sergey-shandar: "which `fromCurve` does not provide". Should we change "fromCurve" to provide `div` (or `nf`) and `g`?

sergey-shandar: Yes — and a bit more than `div`/`g`. `sign` reaches into the curve in three places: `div` (`:186`), and both `g` and `mul` at `:172` (`c.mul(k)(g)`), so adding only `div`/`g` would still leave it touching `c` for `mul`.

I've updated the issue (7a76556) to recommend making `fromCurve` the complete "curve → signing context": it returns `Signer = All & { div, mul, g }` (`{ ...all(c.nf.p), div: c.nf.div, mul: c.mul, g: c.g }`), so `sign` destructures everything from `fromCurve(c)` and never re-reaches into `c`. `all(q)` stays the pure `q`-only `All` factory that `computeK` consumes (it needs none of `div`/`mul`/`g`), and since `Signer extends All`, `computeK(a)` still type-checks.

---

sergey-shandar: I mean, can we add `nf` and `g` to `All`?

sergey-shandar: Adding them to `All` directly runs into two snags, so I kept them on the `fromCurve` return type instead (updated in 14a76e7):

1. **`all` is called with a bare subgroup order, no curve.** `all(q: bigint)` builds the RFC6979 helpers from `q` alone and is used that way — including in `fs/crypto/sign/proof.f.ts` (`all(7n)`, `all(17n)`, `all(5n)`, `all(11n)`, `all(q)`). None of those callers has an `nf`/`g` to supply, so putting `nf`/`g` on `All` would force every bare `all(q)` site to invent curve data (or make the fields nullable, which just pushes `undefined` into `sign`).
2. **`mul` isn't on `nf`.** It's a field of `Curve` (`fs/crypto/secp/module.f.ts:44`, `mul: Fold<bigint, Point>`), so `nf` + `g` alone still wouldn't cover `sign`'s `c.mul(k)(g)`.

So I went with `Signer = All & { nf, mul, g }` returned by `fromCurve` (exposing the whole `nf` per your preference — it carries `div` and `q = nf.p`), with `All` staying the pure `q`-only factory `computeK` consumes. If you'd rather have one type and are fine changing `all`'s signature / updating the bare `all(q)` proofs, I can switch the proposal to fold everything into a single curve-derived `All` instead — let me know which you prefer.

---

sergey-shandar: What I don't like about this is that `&`. I prefer composition over mixing and deriving, for example 
```ts
type Signer = { 
    all: All /* may be a better name than `All` */
    ...
}
```

sergey-shandar: Good call — switched to composition (no `&`) in 7d26dea:

```ts
export type Signer = {
    readonly rfc6979: Rfc6979   // the q-only conversions — was `All`
    readonly nf: Curve['nf']    // gives div, and q via nf.p
    readonly mul: Curve['mul']
    readonly g: Curve['g']
}
export const fromCurve = (c: Curve): Signer =>
    ({ rfc6979: all(c.nf.p), nf: c.nf, mul: c.mul, g: c.g })
```

This is actually cleaner than the intersection version: the RFC6979 record stays **completely unchanged**, `computeK` keeps taking just that record (`computeK(rfc6979)`), and the bare `all(q)` call sites in the proofs are untouched. `sign` reads `const { rfc6979, nf: { div }, mul, g } = fromCurve(c)`.

I also folded in your `All` naming note — the issue now suggests renaming `All` → `Rfc6979` (used as the field type), flagged as optional polish that can be split out if it churns proofs/imports. Let me know if you'd prefer a different name.

---

sergey-shandar: Should we add "prefer composition over intersection" to `AGENTS.md`?

sergey-shandar: Yes — done in e5d9abc. Added a bullet to `AGENTS.md` alongside the other type-discipline rules (`as`, type predicates, literal pinning):

> Prefer composition over intersection types. When a type needs an existing record plus extra fields, embed the record as a named field rather than mixing it in with `&`. Write `type Signer = { rfc6979: Rfc6979, nf: PrimeField, g: Point }`, not `type Signer = Rfc6979 & { nf: PrimeField, g: Point }`. Intersection blurs where each field came from, couples the composite to the exact shape of the part, and tempts you to widen the part to fit the whole … A named field keeps the part **unchanged** — independently constructed and consumed — and reads as plain data you destructure …

I tied the rationale to the data-first reasoning behind the existing `as`/type-predicate rules. (Rebased onto your `main` merge before pushing.)

---

PR #978, opened 2026-06-07
Add parse function import and basic error handling test

sergey-shandar:
## Summary
Added import of the `parse` function from the module and introduced a basic test case to verify that the parse function throws an error when given an empty string input.

## Key Changes
- Imported `parse` function alongside existing `jsGrammar` import from `./module.f.ts`
- Added `parseTodo` test case that validates the parse function throws an error on empty string input
  - Test attempts to call `parse('')` and expects it to throw
  - Fails the test if parse does not throw as expected

## Implementation Details
The new test follows the existing proof pattern in the file, using a try-catch block to verify error handling behavior. This appears to be a placeholder test (indicated by the "Todo" naming convention) for parse function validation.

https://claude.ai/code/session_01L34HcEMWn9uMFfRi9QfJaf


fs/djs/tokenizer-new/proof.f.ts
```diff
@@ -881,6 +881,11 @@ export const proof = {
     //         if (result !== '[{"kind":"yield"},{"kind":"eof"}]') { throw result }
     //     },
     // ],
+    parseTodo: () => {
```
sergey-shandar: Use `throw` as the name of the property instead of `parseTodo`. Then you don't need `try catch`.

Trinidadec (approved):

PR #982, opened 2026-06-07
issues: abandon JSR publishing, keep deno/bun testing

sergey-shandar:
## Summary

- Adds `issues/667-abandon-jsr.md` proposing to stop publishing to JSR
- Deno can consume NPM directly, so JSR is a redundant second source of truth
- Prerequisite: publish `deno doc` output to the website before removing JSR
- Deno and Bun are still used for testing — only the publish step is dropped


sergey-shandar:
@codex review

PR #1001, opened 2026-06-09
Add immutable `capture` effect and refactor emergent-testing reporter

sergey-shandar:
- Summarizes the Write-based reporter refactor.
- Notes that events are serialized to virtual stdout and parsed from finalState.stdout.
- Calls out that NodeOp and the virtual runner remain unchanged.
- References the issue as done.


sergey-shandar:
@codex review

PR #1003, opened 2026-06-09
fs/path: add non-normalizing `join` and route inline segment joins through it

sergey-shandar:
## Summary

Implements [i668-path-join-segments](./issues/668-path-join-segments.md).

- Adds `join: Reduce<string> = a => b => `${a}/${b}`` to `fs/path/module.f.ts`. Unlike `concat`, `join` does **not** parse/normalize the result, so absolute roots and `.`/`..` segments survive verbatim — which is what the CAS store root and recursive `readdir` walks need.
- Replaces the five hand-rolled `` `${a}/${b}` `` joins in `fs/cas`, `fs/dev`, and `fs/effects/node/virtual` with calls to it. This addresses the *Separation of concerns* example in AGENTS.md ("path manipulation belongs in `fs/path`, not inline in a loader").
- Renames the existing `join` import from `fs/types/string` to `listJoin` inside `fs/path/module.f.ts` so the `join` name is free for the path-specific export.
- Adds `joinTest` proof coverage in `fs/path/proof.f.ts`.
- Marks `issues/668-path-join-segments.md` as **done**.

This is a behaviour-preserving refactor — each `join(a)(b)` produces the exact same string the inline template literal produced before.

## Test plan

- [x] `npx tsc` passes
- [x] `fjs t` in `fs/path` — all 18 proofs pass (including new `joinTest`)
- [x] `fjs t` in `fs/cas` — all 11 proofs pass
- [x] `fjs t` in `fs/dev` — all 12 proofs pass
- [x] `fjs t` in `fs/effects` — all 29 proofs pass
- [x] full `fjs t` — 1635/1635 pass

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mbyd51uSMyFKKYR8kUtWzP)_

### DIFF
--- a/fs/cas/module.f.ts
+++ b/fs/cas/module.f.ts
@@ -4,7 +4,7 @@
  * @module
  */
 import { computeSync, sha256, type Sha2 } from '../crypto/sha2/module.f.ts'
-import { parse } from '../path/module.f.ts'
+import { join, parse } from '../path/module.f.ts'
 import type { Vec } from '../types/bit_vec/module.f.ts'
 import { cBase32ToVec, vecToCBase32 } from '../cbase32/module.f.ts'
 import { forEachStep, pure, type Effect, type Operation } from '../effects/module.f.ts'
@@ -36,7 +36,7 @@ const toPath = (key: Vec): string => {
     const s = vecToCBase32(key)
     const [a, bc] = split(s)
     const [b, c] = split(bc)
-    return `${prefix}/${a}/${b}/${c}`
+    return join(prefix)(join(a)(join(b)(c)))
 }
 
 /**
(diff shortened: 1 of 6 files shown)


fs/cas/module.f.ts
```diff
@@ -36,7 +36,7 @@ const toPath = (key: Vec): string => {
     const s = vecToCBase32(key)
     const [a, bc] = split(s)
     const [b, c] = split(bc)
-    return `${prefix}/${a}/${b}/${c}`
+    return join(prefix)(join(a)(join(b)(c)))
```
sergey-shandar: This is terrible! I propose to have two functions, with these signatures:
- the main one `(...list: readonly string[]) string`, most of the time we should use this one;
- and concatenation `Reduce<string>`.

sergey-shandar: Made `join` variadic — `(...list: readonly string[]) => string`. The CAS 3-segment case is now `join(prefix, a, b, c)`, and the store-root + parent-dir join collapses to a single `join(path, ...parts.slice(0, -1))` since `parse` already returns segments.

On the second function — for "concatenation `Reduce<string>`", I read the existing `concat: Reduce<string>` (the normalizing one) as filling that slot, and the variadic now covers every non-normalizing call site in this PR, so I didn't add a separate non-normalizing binary form. Happy to add one if you want it exported too — what should it be called? `concat` is taken by the normalizing version; the alternatives I considered are renaming the existing `concat` (e.g. to `normalizedConcat`) and freeing the name, or using something like `concatSegment`.

---

PR #1049, opened 2026-06-12
Route source-literal rendering through owning modules

sergey-shandar:
## Summary

This PR consolidates duplicate source-literal rendering logic by routing calls through the modules that own these concerns, following the pattern established in i190-text-char-code-boundary.

## Changes

- **bigint literals**: `fs/types/ts` now imports and uses `serialize` from `fs/types/bigint` instead of re-implementing the `${a}n` template inline, mirroring the approach already used by `fs/djs/serializer`.

- **string literals**: `fs/json/serializer` exports a new `stringLiteral` function that wraps `JSON.stringify` for bare-string quoting, and redefines `stringSerialize` in terms of it (no behavior change).

- **string literal consumers**: `fs/emergent_testing` now imports and uses `stringLiteral` from `fs/json/serializer` for its three quoting sites (property access, dynamic imports, and terminal output) instead of calling `JSON.stringify` directly.

## Implementation Details

- The bigint change is a clean peer import within `fs/types/` with no layering concerns.
- The string literal extraction in `fs/json/serializer` maintains backward compatibility while exposing a reusable bare-string form.
- The `fs/types/ts` string-quoting case (two sites) is left unchanged to avoid inverting layering (`fs/types/` depending on `fs/json/`); this is documented as a judgment call pending a potential future low-level JS-syntax module.
- All changes preserve existing behavior and test coverage.

https://claude.ai/code/session_012p4BgzsRPywnyk3aBtkdfq

### DIFF
--- a/issues/66C-emit-literals-via-owner-modules.md
+++ b/issues/66C-emit-literals-via-owner-modules.md
@@ -1,0 +1,151 @@
+# 66C-emit-literals-via-owner-modules. Route source-literal rendering through the owning modules
+
+**Priority:** P4
+**Status:** open
+
+## Problem
+
+Several modules render a value as a piece of JS/TS/JSON **source text** — a
+bigint literal (`123n`) or a double-quoted string literal (`"abc"`). That
+rendering is a small, well-defined concern, and the codebase already has a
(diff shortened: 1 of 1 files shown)


sergey-shandar:
@codex review

PR #1050, opened 2026-06-12
Add edge case test for cycle with empty array

sergey-shandar:
## Summary
Enhanced the `cycleTest` to cover an additional edge case where `cycle()` is called with an empty array.

## Key Changes
- Converted `cycleTest` from a single test function to an array of test functions
- Added a new test case that verifies `cycle([])` correctly returns an empty array `[]`
- Retained the existing test case for `cycle([1, 2, 3])` with `take(10)`

## Implementation Details
The test structure was refactored to allow multiple test scenarios for the `cycle` function. The new test validates that the cycle function handles the edge case of an empty input array gracefully, ensuring it doesn't cause unexpected behavior or infinite loops.

https://claude.ai/code/session_0132yrL2LEkWJZuXqvkZDjYo

### DIFF
--- a/fs/types/list/proof.f.ts
+++ b/fs/types/list/proof.f.ts
@@ -40,10 +40,16 @@ const stringifyTest = () => {
     if (s !== '[1,2,3]') { throw s }
 }
 
-const cycleTest = () => {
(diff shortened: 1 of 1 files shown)


sergey-shandar:
@codex review

PR #1051, opened 2026-06-12
Drop the private Io interface from the node effect runner

sergey-shandar:
Implements [i664-drop-io-interface](https://github.com/functionalscript/functionalscript/blob/claude/gracious-fermat-vuslxv/issues/664-drop-io-interface.md).

The private `Io` type in `fs/effects/node/module.ts` was pure indirection: a single instance bound to Node globals, never a mocking seam (the mem-runners in `fs/effects/mock` and `fs/effects/node/virtual` interpret `Effect<NodeOp, T>` directly), and with no external consumer.

## Changes

- Inline the single `io` handler table into a module-level `asyncRun({ … })` wired straight to the Node globals; `runProgram`/`fromIo` collapse into a `NodeProgramOptions` constant plus `runEffect`.
- Delete the private `Io`, `App`, and `Run` types, and the structural types that existed only to shape `Io` fields (`Fs`, `Dirent`, `ReadDir`, `RmOptions`, `MakeDirectoryOptions`, `Console`, `Performance`, `Process`, `Writable`, `TryCatch`, `Sandbox`, `Http`, `ChildProcess`).
- Prune `io` members the interpreter never read: `console`, `tryCatch`, `asyncTryCatch` — and `performance`, which the issue's design missed (the `sandbox` handler closes over the global `performance` directly). Noted in the issue file.
- Keep the narrowed structural view of `node:http`'s `createServer` (`method`/`url`/`headers` required), since the official types declare them optional/wide while the effect-level `RequestListener` requires them present; documented at the alias.
- Add the missing `@module` header documenting why there is deliberately no IO seam, per the "capture design decisions in the codebase" rule.
- Mark the issue **Status: done**.

## No public API change

`run` / `runEffect` and all effect types are untouched; this is a behaviour-preserving internal simplification (−282/+140 lines).

## Testing

- `npx tsc` — clean
- `fjs t` (`node ./fs/fjs/module.ts t`) — 1729 tests pass, 0 fail; this run itself exercises the refactored runner end to end (`exec`, `readdir`, `readFile`, `sandbox`, `write`, …)

https://claude.ai/code/session_0192Kk1pHPcEgVhLHhVxfDvn

---
_Generated by [Claude Code](https://claude.ai/code/session_0192Kk1pHPcEgVhLHhVxfDvn)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- `effects/node`: drop the private `Io` indirection from the node effect runner — inline the single handler table into a module-level `asyncRun({ … })` wired straight to the Node globals, delete the dead `Io`/`App`/`Run` types and the `io` members the interpreter never read (`console`, `tryCatch`, `asyncTryCatch`, `performance`); pure refactor, `run`/`runEffect` unchanged (i664-drop-io-interface) [#1051](https://github.com/functionalscript/functionalscript/pull/1051)
 - `fjs`: add a `proof.f.ts` covering the CLI command handlers (help output, the `compile` missing-argument error path, and the `run` handler's import-and-invoke and import-failure paths) via the virtual Node-effect interpreter [#1047](https://github.com/functionalscript/functionalscript/pull/1047)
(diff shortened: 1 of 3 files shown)


sergey-shandar:
@codex review

PR #1053, opened 2026-06-12
Add decode/match accessors for the effect value representation (i661)

sergey-shandar:
Implements [i661](issues/661-effect-value-decoder.md): the `Pure | Do` tuple layout (`value.length === 1` discriminant, `[command, payload, continuation]` indexing) was re-derived by hand in the async runner, the mock runner, an inline proof runner, and five proof assertions — 8+ sites, none of them the module that owns the type.

## What changed

- **`fs/effects/module.f.ts`** now owns the representation:
  - `decode` turns an effect's next step into a named, discriminated form (`{ done: true, result } | { done: false, command, payload, continuation }`). It is the only place that knows the tuple layout.
  - `match(map)(effect)` decodes *and* dispatches the command to an `OperationMap<O, R>`, returning `['done', T] | ['cont', R, continuation]` — only the world-specific eliminator (`await` for async, state threading for sync) stays in each runner.
- **Runners** (`fs/effects/module.ts`, `fs/effects/mock/module.f.ts`) become the shared `match` skeleton plus one eliminator line.
- **`MemOperationMap` is curried**: `(state, ...payload) => [state, r]` → `(...payload) => (state) => [state, r]`, separating a command's arguments (fixed at issue time) from the state the runner threads. Migrated mechanically: `fs/effects/node/virtual`, `fs/effects/memory/proof.f.ts`, `fs/mcp/proof.f.ts`, `fs/emergent_testing/proof.f.ts`.
- **Proofs** (`fs/effects/proof.f.ts`, `fs/effects/node/proof.f.ts`) use `decode` instead of poking at the tuple; new `decode` / `match.done` / `match.cont` proofs cover both branches of each new export against a minimal `AddOp`.

## Divergence from the issue sketch

`match`'s effect parameter is `<O1 extends O, T>(effect: Effect<O1, T>)` rather than `Effect<O, T>`: `Effect` is not covariant in `O` (the continuation consumes the operation result, an input position), so `RunInstance`'s `<O1 extends O>` effects would not be assignable otherwise. Recorded in the issue's resolution section, which is now marked **done**.

## Testing

- `npx tsc` — clean, no casts introduced.
- `node ./fs/fjs/module.ts t` — 1764 pass, 0 fail (includes the mock/virtual runner proofs that exercise both runners end to end).

https://claude.ai/code/session_01CRQ5Etje6SzgEYFhYKEcPL

---
_Generated by [Claude Code](https://claude.ai/code/session_01CRQ5Etje6SzgEYFhYKEcPL)_

### DIFF
--- a/fs/effects/memory/proof.f.ts
+++ b/fs/effects/memory/proof.f.ts
@@ -14,7 +14,7 @@ type MemoryState = {
 const initial: MemoryState = { next: 0, values: {} }
 
 const mock: MemOperationMap<MemOp, MemoryState> = {
-    memCreate: (state, value) => {
+    memCreate: value => state => {
         const id = `k${state.next}`
         const key: Key<unknown> = asNominal(id)
         return [{
@@ -22,9 +22,9 @@ const mock: MemOperationMap<MemOp, MemoryState> = {
             values: { ...state.values, [id]: value },
         }, key]
     },
-    memRead: (state, key) =>
+    memRead: key => state =>
         [state, state.values[asBase(key)]],
-    memWrite: (state, key, value) => {
+    memWrite: (key, value) => state => {
         const id = asBase(key)
         if (!(id in state.values)) { throw id }
         return [{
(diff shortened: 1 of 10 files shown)


sergey-shandar:
@codex review

PR #1055, opened 2026-06-13
fix(types): use ? for infinite-key mapped types (i66C)

sergey-shandar:
## Summary

- Introduces `StringMap<K, T>` in `fs/types/object/module.f.ts`: a single conditional type that resolves to `{ readonly[k in string]?: T }` for infinite key sets (`string`) and `{ readonly[k in K]: T }` for finite literal-union key sets. Also adds `definedEntries` alongside the existing `definedValues` helper.
- Migrates `Map<T>` to `Map<T> = StringMap<string, T>` so all infinite-key maps are correctly optional.
- Applies the fix across 17 sites in 11 files. Mutually-recursive types (`djs.Object`, `AstObject`, `Obj<T>`, `bnf.Variant`) use the inline `{ readonly[k in string]?: T }` form because TypeScript's circular-reference detection cannot resolve through conditional types.
- Updates downstream call sites: `definedEntries`/`definedValues` for iterations, `!` assertions for guarded lookups, `?? []` for the FSM state lookup.
- Adds compile-time proof assertions to `fs/types/object/proof.f.ts`.

## Test plan

- [x] `deno check fs/**/*.ts` — zero type errors
- [x] All existing proof functions unchanged in behaviour


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1057, opened 2026-06-13
djs/serializer: hoist sharedRef predicate (i66B)

sergey-shandar:
## Summary

Implements [i66B-djs-serializer-shared-ref-predicate](./issues/66B-djs-serializer-shared-ref-predicate.md).

`fs/djs/serializer/module.f.ts` previously open-coded the "is this value shared / worth hoisting to a const?" test twice:

```ts
const refCounter = refs.get(value)
if (refCounter !== undefined && refCounter[1] > 1) { ... }
```

in `getConstants` (decide which values become consts) and `serializeWithConst` (emit a `c<N>` reference to one), with raw `[1] > 1` / `[0]` tuple indexing at each site.

Lift the lookup-and-test into one module-scope helper that returns the `RefCounter` only when the value is shared:

```ts
const sharedRef = (refs: Refs) => (v: Unknown): RefCounter | undefined => {
    const rc = refs.get(v)
    return rc !== undefined && rc[1] > 1 ? rc : undefined
}
```

`getConstants` uses it for the `!== undefined` check; `serializeWithConst` keeps the returned counter so `[0]` is read from the same result without a second lookup. `stringify`'s `constSerialize` is intentionally left alone — it has the different "lookup-or-invariant-violation" contract (the entry came from `getConstants`, so `undefined` is a bug), per the issue's caveat.

Pure refactor — no behavior change. Existing `fs/djs/serializer/proof.f.ts` continues to exercise the same call paths.

## Test plan

- [x] `npx tsc` — clean
- [x] `node ./fs/fjs/module.ts t` — 1764/1764 pass, including all 14 `fs/djs/serializer/proof.f.ts` cases

---
_Generated by [Claude Code](https://claude.ai/code/session_013KJtJKLt5Q8UxgDjJPM1po)_


sergey-shandar:
@codex review

Trinidadec (approved):

PR #1080, opened 2026-06-15
fix(json): stringify skips undefined-valued object properties (66F)

sergey-shandar:
## Summary

- `stringify` in `fs/json/module.f.ts` now uses `definedEntries` (already in `fs/types/object`) instead of `Object.entries`, so properties with `undefined` values are silently dropped — matching `JSON.stringify` behaviour
- Removes the `defined` filter workaround in `fs/mcp/stdio/module.f.ts`; `stringifyJson` simplifies to `stringify(sort)`
- Adds a proof test: `stringify(sort)({ x: undefined }) === '{}'`

## Test plan

- [ ] `proof.undefined` test in `fs/json/proof.f.ts` passes
- [ ] `fs/mcp/stdio` proof suite still passes (no regression from removing the filter)


sergey-shandar:
@codex review

PR #1086, opened 2026-06-15
Add union conflict proof test for range handler merging

sergey-shandar:
This change adds a proof-of-concept test that exercises an edge case in the `union` function that is unreachable through the public API.

## Summary
Added a `proof` export containing a `unionConflict` test case that validates the error handling behavior when attempting to merge two distinct non-default handlers for the same range.

## Key Changes
- Added `proof` object with `unionConflict` function that:
  - Creates two separate `CreateToResult` handlers that both return `unexpectedSymbol`
  - Attempts to merge them using the private `union` function
  - Verifies that the union operation throws an error containing references to both handlers
  - Confirms the error is properly structured as an array with the two conflicting handlers

## Implementation Details
- This test path is unreachable through the public `init` API since it has no overlapping ranges
- The test is placed in the module where the private `union` function is in scope, allowing direct testing of this internal behavior
- The error validation ensures the thrown error is an array containing exactly the two conflicting handler references

https://claude.ai/code/session_017gyVcmmZXyDMrXtc82khf6


fs/fsc/module.f.ts
```diff
@@ -89,3 +89,21 @@ export const init: ToResult = create([
     range('09')(() => a => [[fromCharCode(a)], unexpectedSymbol]),
     ...[...punctuation].map(single),
 ])(undefined)
+
+export const proof = {
+    // union throws when two distinct non-def handlers are merged for the same range;
+    // this path is unreachable through the public API (init has no overlapping ranges),
+    // so we exercise it here where the private union function is in scope.
+    unionConflict: () => {
```
sergey-shandar: Read fs/emergent_testing/README.md on how to handle exceptions in proofs.

PR #1097, opened 2026-06-16
base_n: extract shared Vec ↔ string bit-codec factory for base64/cbase32

sergey-shandar:
## Summary

Implements i66F-base-n-codec-factory. `fs/base64` and `fs/cbase32` had the same two inner loops, differing only in chunk width (`6n` vs `5n`), alphabet, and an optional decode-side character normalization. Extract that raw codec into a single `baseN(bits, alphabet, normalize?)` factory under a new `fs/base_n` module and rewrite both consumers through it.

- Each format keeps its own padding strategy (`base64`'s octet-alignment + `=` padding, `cbase32`'s 1-then-0s stop-bit sentinel) — only the alphabet ↔ bit-width codec is shared.
- Clears the YAGNI gate in i039-radix-encoding: a second real `Vec → string` consumer (`base64`) now exists alongside `cbase32`.
- New module ships with a co-located `proof.f.ts` covering empty input, aligned/unaligned encode, decode round-trip, invalid-character rejection, and both `normalize`-present/absent branches.

## Test plan

- [x] `npx tsc` passes
- [x] `node ./fs/fjs/module.ts t` — 1828/1828 tests pass (up from 1820; +8 in `fs/base_n`)
- [x] All downstream consumers (`fs/cas`, `fs/cas/mcp`) still pass

---
_Generated by [Claude Code](https://claude.ai/code/session_017GTRzMbktfuuQmgVsPkCRf)_


sergey-shandar:
@codex review

PR #1099, opened 2026-06-16
Make CAS store path configurable via home directory

sergey-shandar:
## Summary
This PR makes the Content Addressable Store (CAS) configurable to use a home directory path instead of always defaulting to the current directory. This allows the CAS to be used with different storage locations, improving flexibility and testability.

## Key Changes
- **Added `home` field to `NodeProgramOptions`**: A new required `home: string` field is added to track the home directory path, initialized from `process.env.HOME` or `process.cwd()` in the Node.js runtime.

- **Refactored `fileKvStore` function**: 
  - Changed from an arrow function expression to a block function to compute `storePrefix` once
  - Updated all file operations to use `join(path, ...)` to properly resolve paths relative to the provided root
  - Fixed the `list()` operation to use the computed `storePrefix` instead of hardcoded `.cas`

- **Updated CAS command handlers**: All command handlers (`add`, `get`, `list`, `mcp`) now:
  - Accept `home` from the handler context
  - Create a fresh CAS instance using `fileKvStore(home)` instead of using a global instance
  - This ensures each command operates on the correct home directory

- **Updated all test/proof files**: Added `home: '.'` to `NodeProgramOptions` initialization in:
  - `fs/cas/proof.f.ts`
  - `fs/cli/proof.f.ts`
  - `fs/emergent_testing/proof.f.ts`
  - `fs/fjs/proof.f.ts`
  - `fs/text/sgr/proof.f.ts`

- The global `const c = cas(sha256)(fileKvStore('.'))` instance was removed and replaced with per-command instantiation, allowing each command to use the appropriate home directory
- The `storePrefix` is now computed once at function initialization rather than being reconstructed in the `list()` operation
- Path handling is now consistent across all operations using `join()` to properly resolve relative paths

https://claude.ai/code/session_01RiNnhogXRWc21wiwHutqVy

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -407,7 +407,7 @@
           "run": "deno install --frozen"
         },
         {
-          "run": "deno test --allow-read --allow-env --coverage && deno coverage --include='.*module\\.f\\.ts'"
+          "run": "deno test --allow-read --allow-env --allow-sys --coverage && deno coverage --include='.*module\\.f\\.ts'"
         },
         {
           "run": "git reset --hard HEAD && git clean -fdx"
--- a/deno.json
+++ b/deno.json
@@ -1,8 +1,8 @@
 {
   "tasks": {
     "fjs": "deno run --allow-read --allow-write --allow-env --allow-net ./fs/fjs/module.ts",
-    "test": "deno test --allow-read --allow-env",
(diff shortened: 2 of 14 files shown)


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1112, opened 2026-06-17
Add proof of concept for union conflict handling

sergey-shandar:
## Summary
Added a proof-of-concept test case demonstrating the union conflict error path in the tokenizer module. This validates that the union function correctly throws when attempting to merge two distinct non-default handlers for the same range.

## Changes
- Added `proof` export object containing a `throw.unionConflict` test case
- The test case demonstrates the unreachable code path that occurs when conflicting handlers are merged
- Includes documentation noting that this scenario cannot occur through the public API due to the absence of overlapping ranges in practice

## Implementation Details
- The proof case defines three handler functions (default, a, and b) with identical signatures
- Calls `union(def)(a)(b)` to trigger the conflict detection logic
- Serves as a reference implementation for the union conflict behavior

https://claude.ai/code/session_01CEUPvP8okD2sSkTr2cwbjX


sergey-shandar:
@codex review

pageofswrds (approved):
If I understand properly:

1) The tokenizer breaks source text (code) down into tokens, so that it can be normalized then hashed
2) Union is used to assemble a lookup table (i.e. rulebook) for how the tokenizer processes character→handler
3) This PR is basically just about throwing an error if your initial rulebook is invalid (which is why it's unreachable; it only fires at build time)

LGTM 😄 – but my model (who helped me understand this code) asked me to call this out:

"One question: AGENTS.md says to restructure unreachable branches rather than cover them, but this proves one instead. I read the throw as a defensive guard against a future contributor adding an overlapping rule, so keep-and-prove seems reasonable — but was removing it considered?"

PR #1134, opened 2026-06-20
Add issue templates for virtual FS and SUL literal refactorings

sergey-shandar:
This PR adds two issue templates documenting refactoring opportunities in the codebase:

## Summary
Two new issue documents have been added to track code quality improvements: one for consolidating repeated entity-kind discrimination logic in the virtual filesystem module, and another for eliminating duplicate level construction in the SUL literal encoder.

## Issues Added

**66M-virtual-fs-entity-kind** (Priority P3)
- Documents the problem of entity-kind discrimination being re-derived inline at twelve call sites in `fs/effects/node/virtual/module.f.ts` with four different spellings
- Proposes introducing three helper functions (`isJsModule`, `isFile`, `isDir`) to centralize the structural contract of the `Entity` union type
- Includes concrete examples of the current scattered checks and how they would be simplified
- Recommends a minimal approach that improves readability without requiring broader restructuring

**66M-sul-literal-level-reuse** (Priority P4)
- Documents the duplication of level construction in `fs/sul/level/literal/module.f.ts` where `level(0n)`, `level(2n)`, and `level(7n)` are built twice
- Proposes passing already-constructed `Level` objects to `literalToVec` instead of exponents, eliminating redundant construction and the single-source-of-truth violation
- Shows how this change makes the relationship between the encoder pipeline and decoders explicit

Both issues follow the existing issue template format and include clear problem statements, proposals with code examples, and concrete task lists for implementation.

https://claude.ai/code/session_012zzeBcAvTnfGpbEnd1LvJc


issues/66M-virtual-fs-entity-kind.md
```diff
@@ -0,0 +1,166 @@
...
+
+Define the three kind-tests **once**, right after the `Entity` / `Dir` /
+`JsModule` declarations, and have every site call them. No type predicates
+(`x is U`) — per `AGENTS.md` those are error-prone — just plain booleans that
+name the structural intent:
+
+```ts
+const isJsModule = (e: Entity): boolean => typeof e === 'function'
+const isFile     = (e: Entity): boolean => e instanceof Array
+const isDir      = (e: Entity): boolean => typeof e === 'object' && !(e instanceof Array)
```
sergey-shandar: apply type narrowing instead of `boolean`.

sergey-shandar: Done in 4edd6da. Replaced the `boolean` helpers with a value-carrying tagged classifier:

```ts
type Classified =
    | readonly ['file', readonly Vec[]]
    | readonly ['module', JsModule]
    | readonly ['dir', Dir]

const classify = (e: Entity): Classified =>
    e instanceof Array      ? ['file', e]   :
    typeof e === 'function' ? ['module', e] :
                              ['dir', e]
```

Inside the ternary `e instanceof Array` / `typeof e === 'function'` narrow each arm automatically (no `x is U` predicate), and consumers `switch` on the tag so the payload comes out as a real `Dir`/`Vec[]` — which also lets the proposal drop the existing `as Dir` / `as readonly Vec[]` casts at the call sites.

---

issues/66M-virtual-fs-entity-kind.md
```diff
@@ -0,0 +1,172 @@
...
+`file as readonly Vec[]` (`:110`), `content as Dir` (`:155`), `sub as Dir`
+(`:191`, `:224`), `existing as Dir` (`:211`), `file as readonly Vec[]` (`:259`).
+`AGENTS.md` treats `as` like Rust's `unsafe`, so those casts are the thing to
+remove.
+
+Define three predicate helpers next to the `Entity` declarations, where each
+predicate's body **is** exactly the narrowing condition it asserts:
+
+```ts
+const isFile     = (e: Entity): e is readonly Vec[] => e instanceof Array
```
sergey-shandar: Call it `isBinFile` because JsModule is also a file.

issues/66M-virtual-fs-entity-kind.md
```diff
@@ -0,0 +1,172 @@
...
+const isJsModule = (e: Entity): e is JsModule       => typeof e === 'function'
+const isDir      = (e: Entity): e is Dir            => typeof e === 'object' && !(e instanceof Array)
```
sergey-shandar, line 111:
```suggestion
const isDir      = (e: Entity): e is Dir            => !(isBinFile(e) || isJsModule(e))
```

issues/66M-virtual-fs-entity-kind.md
sergey-shandar: Done in c545939. Renamed `isFile` → `isBinFile` throughout, and relabeled the kind table row to "binary file". Noted in the prose that a `JsModule` counts as a file too (`readdir` already treats it as one), which is why the array case gets the `Bin` qualifier.

---

issues/66M-virtual-fs-entity-kind.md
sergey-shandar: Applied the intent — `isDir` now composes from the other two predicates rather than re-spelling the structural test — but with the negation, since the literal suggestion looks inverted:

```ts
const isDir = (e: Entity): e is Dir => !isBinFile(e) && !isJsModule(e)
```

`isBinFile(e) || isJsModule(e)` is the condition for *not* a directory (a value that's an array or a function can't be a `Dir`), so using it as the body of `e is Dir` would assert `Dir` while testing the opposite — exactly the divergent predicate the guidance warns about. The exhaustive negation over the closed three-member union keeps the body matching the asserted type. Shout if you actually meant a separate `isFile = isBinFile(e) || isJsModule(e)` helper (with `isDir = !isFile(e)`) and I'll restructure that way.

---

sergey-shandar: Ok, keep the old implementation.

sergey-shandar: Done in 7977f13. Reverted `isDir` to its direct structural test:

```ts
const isDir = (e: Entity): e is Dir => typeof e === 'object' && !(e instanceof Array)
```

and updated the surrounding prose/tasks to drop the composition. `isBinFile` / `isJsModule` stay as the matched predicates.

---

PR #1158, opened 2026-06-23
Refactor CAS file operations to unify streaming logic

sergey-shandar:
## Summary
This PR refactors the Content Addressable Storage (CAS) file operations to eliminate code duplication and improve separation of concerns. A new `casAddFile` function is introduced as the canonical way to stream files through the CAS, with both CLI and MCP layers delegating to it.

## Key Changes

- **New `casAddFile` function**: Extracted the core logic of streaming a file through `cas.write()` into a reusable function that both CLI and MCP operations can delegate to
- **Simplified CLI `add` command**: Removed manual file reading and stream construction; now uses `casAddFile` directly instead of `readFile` + `c.write()`
- **Updated MCP `cas_add` handler**: Refactored to use `casAddFile` for the `type:'url'` case, then chains `rm()` to delete the source file on success
- **Improved type signatures**: Updated return types throughout to include `Rm` operation where file deletion is now performed
- **Removed unused imports**: Cleaned up `listEffectCons`, `listEffectEnd`, and `ReadFile` from CLI module as they're no longer needed

## Implementation Details

- The `casAddFile` function encapsulates the file streaming pattern, making it the single source of truth for how files are added to the CAS
- The MCP layer now explicitly handles cleanup (file deletion) as a separate step after successful upload, rather than relying on a separate `casUpload` function
- The CLI layer is simplified by delegating to `casAddFile` instead of manually constructing effect chains
- Error handling remains unchanged: on failure, source files are preserved to allow retries

https://claude.ai/code/session_01V1VTNYsR7jCRD5VVxipBrH


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1171, opened 2026-06-27
Add todo issues for DRY improvements across codebase

sergey-shandar:
## Summary

This PR adds seven new todo issues documenting opportunities to reduce code duplication and improve consistency across the codebase. These are design documents for future refactoring work, not implementation changes.

## Issues Added

- **RTTI: share the container entry-loop between `parse` and `validate`** (P3)
  - Documents duplication of container iteration logic in two different shapes with behavioral inconsistency
  - Proposes hoisting a shared `eachEntry` helper to `common/` to unify the loops and give `parse` early-exit semantics

- **Share Unicode code-point predicates in `code_point`** (P3)
  - Identifies duplicate code-point classification predicates (BMP, surrogate, supplementary ranges) defined independently in utf8 and utf16
  - Proposes moving predicates to `code_point/module.f.ts` as single source of truth
  - Includes secondary fix: replace stray `0xffffffff` error sentinel in utf16 with shared `errorMask`

- **Lift chunk-boundary padding into a `base_n` strategy layer** (P4)
  - Documents identical padding boundary arithmetic in base64 and cbase32
  - Proposes optional `paddedBaseN` factory to centralize the `len % bits` math and pad/unpad strategies

- **66O-read-streamfile-dedup: `fileCas.read` should delegate to `streamFile`** (P3)
  - Identifies near-identical recursive chunk-reader loops in two methods
  - Proposes `read` delegate to `streamFile` with effect-type widening, removing duplicate EOF/error invariant

- **Export `okResult` from `mcp` next to `errorResult`** (P3)
  - Documents `ToolsCallResult` success constructor duplicated in `cas/mcp` while error constructor is in `mcp`
  - Proposes moving `okResult` to `mcp/module.f.ts` and expressing `errorResult` in terms of it

- **RTTI: one source of truth for `Unknown` / `Primitive`** (P3)
  - Identifies inconsistent imports: `parse` uses rtti-local definitions while `common` and `validate` import from `djs`
  - Proposes unified import from rtti-local source to honor the stated keep-rtti-free-of-djs goal

- **Define `TerminalRange` once, shared between `bnf` and `bnf/data`** (P4)
  - Documents identical type declaration in two modules with explicit comment "The same as the functional TerminalRange"
  - Proposes defining once in `bnf` and importing in `bnf/data`

## Notes

All issues follow the existing todo format with Problem, Proposal, and Tasks sections. They represent independent refactoring opportunities; none block each other.

https://claude.ai/code/session_01QNragX2oKNAAU5CKHTnJht


fs/types/rtti/todo/unknown-primitive-source-of-truth.md
```diff
@@ -0,0 +1,61 @@
...
+  `djs`, which is the dependency direction `rtti/ts` says it wants to avoid, and
+- makes all sibling files agree on one import.
+
+Verify the two definitions are structurally identical before switching (the
+comment asserts they are "currently equivalent"); if they have drifted, that
+divergence is itself a bug this change surfaces.
+
+### Tasks
+
+- [ ] Confirm `djs.Unknown`/`Primitive` and `rtti/ts.Unknown`/`Primitive` are
```
sergey-shandar: In the future, `rtti.Unknown` may contain types that are not supported by DJS, for example, functions. So, `rtti` should have its own definition and it may not match `djs.Unknwon`. At the same time, djs.Unknown is a subset of rtti.Unknown.

sergey-shandar: Good point — updated the proposal in 2194fcd. Reframed it around `djs.Unknown ⊆ rtti.Unknown` (subset, not equivalence): `rtti` keeps its own `Unknown` as the source of truth and may later widen it to types DJS can't represent (e.g. functions). The change is now just re-pointing `common`/`validate` at the rtti-local definition (the import inconsistency this proposal targets), plus a task to fix the `rtti/ts` doc comment's "currently equivalent" wording to state the subset relationship.

---

sergey-shandar:
@codex review

fs/base_n/todo/baseN-padding-strategy.md
```diff
@@ -0,0 +1,82 @@
...
+
+The boundary arithmetic — `extra = bits - len % bits`, then
+`concat(v)(vec(extra)(padValue))` — is structurally identical; only the pad
+*value* (zeros + an external `=` rule vs. a sentinel bit) and the matching unpad
+rule differ.
+
+There is also a subtle asymmetry worth normalizing during the extraction: base64
+special-cases `rem === 0n` to add **0** pad bits, while cbase32's
+`5n - len % 5n` yields a full extra symbol (`5n`) when `len % 5n === 0n`. Centralizing
+the boundary math fixes the "already aligned" case once for both.
```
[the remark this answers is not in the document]
sergey-shandar: Correct — fixed in dad9bfa. I removed the "asymmetry to normalize" framing. cbase32's full sentinel block on 5-bit-aligned input is required by its in-band sentinel scheme (`cBase32ToVec` scans backward for the sentinel `1` and rejects inputs without one; `proof.f.ts` proves the extra terminator), unlike base64 which adds 0 pad bits and relies on the `=` count. The proposal now states the aligned-case behavior is strategy-owned and must not be unified across the two codecs.

---

fs/text/code_point/todo/share-code-point-predicates.md
```diff
@@ -0,0 +1,95 @@
...
+    return [[0xffffffff], state]
+}
+```
+
+Every other error branch in the same function uses `errorMask`
+(`word | errorMask`, `state | errorMask`). `errorMask` is
+`0b1000_0000_…_0000` (`0x80000000`), which is **not** `0xffffffff`, so this one
+branch tags malformed units with a different value than the rest of the decoder.
+Downstream consumers that test `result & errorMask` to detect errors will not
+recognize this path. This is a small consistency/correctness fix that should ride
```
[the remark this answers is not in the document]
sergey-shandar: Correct — fixed in dad9bfa. You're right that `0xffffffff` already has the error bit set and that `utf16/proof.f.ts:19-20` asserts the `4294967295` value, so replacing it with `errorMask` would change observable output and break the proof. I dropped that change entirely. The proposal now records it as out-of-scope (with the reasoning: the all-ones sentinel is deliberately distinct from `word | errorMask`, which preserves the offending in-range word — a non-u16 unit has no meaningful word to preserve) so it isn't re-proposed. The issue is now only about sharing the classification predicates.

---

sergey-shandar:
@codex review

fs/base_n/todo/basen-padding-strategy.md
```diff
@@ -0,0 +1,98 @@
+## Lift chunk-boundary padding into a `base_n` strategy layer
```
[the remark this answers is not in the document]
sergey-shandar: Done in f4bc419 — renamed to `fs/base_n/todo/basen-padding-strategy.md` (lowercase kebab-case), matching the sibling `various-basen-encodings.md` and the documented `todo/{slug-kebab}.md` convention.

---

sergey-shandar:
@codex review

fs/text/code_point/todo/share-code-point-predicates.md
```diff
@@ -0,0 +1,92 @@
...
+
+The `0xffffffff` out-of-range sentinel stays as is (see "Not in scope" above).
+
+### Tasks
+
+- [ ] Add the BMP / surrogate / supplementary / validity predicates to
+      `fs/text/code_point/module.f.ts`, built from one set of range constants;
+      add `proof` coverage for each.
+- [ ] Replace the utf8 `bigRange`/`smallRange`/`isValidCodePoint` definitions
+      with imports from `code_point`.
```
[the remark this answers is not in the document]
sergey-shandar: Good catch — addressed in a79360f. `isValidCodePoint` is `export const` in utf8 but its only consumer is `fromVec` in the same module (no external importer). The proposal now requires a deliberate decision rather than a silent drop: either re-export it from `utf8` to keep the public API stable, or drop the export (defensible since nothing external uses it and AGENTS.md says only export with a real external consumer) — but record that as an intentional API change, with the canonical export living on `code_point`. The task line now says "don't drop it silently."

---

fs/types/rtti/todo/parse-validate-shared-entry-loop.md
```diff
@@ -0,0 +1,101 @@
...
+### Proposal
+
+Hoist a single early-exit entry-loop helper into
+`fs/types/rtti/common/module.f.ts`, e.g.:
+
+```ts
+// runs `item` over each entry, bailing on the first error with the path prefixed
+export const eachEntry = <R>(
+    entries: ReadonlyArray<readonly [string, Unknown]>,
+    item: (k: string, v: Unknown) => CommonResult<R, ValidationError>,
```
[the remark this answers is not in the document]
sergey-shandar: Correct — fixed in a79360f. Made `eachEntry` generic over the entry-value type (`<V, R>`) instead of hard-coding `Unknown`. The proposal now spells out the two entry sources: `containerValidate`/`containerParse` iterate the value's own entries (`V = Unknown`), while `constContainerValidate`/`constContainerParse` iterate schema entries `Object.entries(rtti)` (`V = Type`, possibly thunk functions), with their `item` closing over `value` and applying `validate/parse(schemaEntry)(getItem(value, k))`. Hard-coding `Unknown` would have forced unsafe casts and defeated the one-shared-loop goal.

---

sergey-shandar:
@codex review

fs/cas/todo/66o-read-streamfile-dedup.md
```diff
@@ -0,0 +1,71 @@
...
+        const chunk = result[1]
+        return length(chunk) === 0n ? end() : cons(ok(chunk), loop(offset + chunkBytes))
+    })
+```
+
+The only real difference is the declared effect type: `FileCasOperation` in
+`read` vs. `ReadBytes` in `streamFile`. `ReadBytes ⊆ FileCasOperation`, and the
+codebase already relies on exactly this widening — `casAddFile`
+(`fs/cas/module.f.ts:271`) widens `streamFile`'s `List<ReadBytes, …>` to a
+generic `O` with a documented sound cast. So the EOF/error streaming invariant is
```
pageofswrds: **The cited precedent for this cast doesn't exist as described.** `casAddFile` (lines 267–271) performs no `as` cast — it returns the *union* `Effect<O | ReadBytes, …>` and lets `cas.write`'s generic absorb `ReadBytes`; the word "cast" appears only in its comment. The proposal at line 53 (`read: (hash) => streamFile(...)`) can't reuse that mechanism: `read`'s return type is pinned to `List<FileCasOperation, …>` by the `FileCas` interface, so it can't widen to a union the way `casAddFile` does. Expect to need an **explicit** cast here — the "sound cast pattern `casAddFile` already uses" isn't actually there to copy. Worth a `tsc` check before committing to the one-liner.

*— tycho*

pageofswrds (commented):
## Review — factual accuracy of the todo docs

These seven todos add no code, so I reviewed them the way that matters for design docs: checking each factual claim — line citations, quoted code, counts, cross-references — against the actual source on `main`.

**Four of seven verified fully accurate:** `code_point` predicates; the rtti `parse`/`validate` entry-loop (including the behavioral-divergence claim — confirmed `parse`'s eager `Array.map` really does evaluate every entry while `validate` short-circuits); `bnf` `TerminalRange`; and the rtti `Unknown`/`Primitive` import disagreement + subset claim. The negative-knowledge sections (the cbase32 sentinel, the `0xffffffff` "not in scope") all hold up — nicely careful work.

The four inline comments below flag the slips: three quote/cross-reference drifts and one proposal that over-claims its own feasibility. None blocks merging (these are todos — merging changes no behavior), but the two on the `cas` todo are worth fixing **before** anyone implements it, since they'd send the implementer down a wrong path.

---
*Review authored by **tycho**, an AI agent, working with the reviewer. Each finding was verified against the source before posting.*

fs/cas/todo/66o-read-streamfile-dedup.md
```diff
@@ -0,0 +1,71 @@
...
+
+`fileCas.read`'s inner `loop` (`fs/cas/module.f.ts:140-152`):
+
+```ts
+const loop = (offset: number): List<FileCasOperation, IoResult<Vec>> =>
+    readBytes(p, offset, chunkBytes)
+    .step((result): List<FileCasOperation, IoResult<Vec>> => {
+        const [t, v] = result
+        if (t === 'error') { return cons<FileCasOperation, IoResult<Vec>>(result, end()) }
+        return length(v) === 0n ? end() : cons(ok(v), loop(offset + chunkBytes))
```
pageofswrds: **These "verbatim" quotes use constructors that aren't in the module.** The real code (lines 145/149/150) uses `nonEmpty(...)` and `elEmpty()`, not `cons(...)` / `end()` — same semantics, but `cons`/`end` appear nowhere in `fs/cas/module.f.ts`. The `streamFile` quote just below (lines 30–32) has the same substitution vs. the real lines 253/256/257. The logic you're documenting is right; only the identifiers drifted — but anyone copy-pasting the block while refactoring would hit undefined names.

*— tycho*

fs/base_n/todo/basen-padding-strategy.md
```diff
@@ -0,0 +1,98 @@
...
+      a sentinel block, so its `cBase32ToVec` backward sentinel scan still finds a
+      terminator).
+- [ ] Remove the now-unused `msb`/`vec`/`length`/`empty` imports from `base64`
+      and `cbase32`.
+- [ ] Run `npx tsc` and `fjs t`; confirm round-trip proofs still pass.
+
+### Related
+
+- `fs/base_n/todo/various-basen-encodings.md` — note its "Base64" entry is now
+  stale: `fs/base64/module.f.ts` is implemented and already uses `baseN`.
```
pageofswrds: **This "Related" note would close a still-open todo.** The `various-basen-encodings.md` "Base64" entry asks specifically for an *identifier-safe* base64 (alphabet ending in `_`, `$`) — that variant was never built. `fs/base64/module.f.ts` is the standard RFC 4648 `+/` codec, a different alphabet. So base64 shipping on `baseN` doesn't make that entry stale; the requested identifier variant is still open.

*— tycho*

fs/mcp/todo/share-ok-result.md
```diff
@@ -0,0 +1,62 @@
...
+but the success constructor lives privately in the CAS adapter
+(`fs/cas/mcp/module.f.ts:263-264`):
+
+```ts
+const okResult = (text: string): ToolsCallResult =>
+    ({ content: [{ type: 'text', text }] })
+```
+
+`cas/mcp` already imports `errorResult` from `mcp` and uses `okResult` heavily
+(eight call sites). Both constructors build the same
```
pageofswrds: **Count is off by one** — `okResult` has nine call sites in `fs/cas/mcp/module.f.ts` (lines 163, 182, 215, 218, 230, 231, 244, 247, 256), not eight. Minor, but the migration task below leans on the number.

*— tycho*

pageofswrds (approved):
I've given this a look from my product perspective and it gets a big ol' LGTM. I'm learning a lot about the codebase! Here are the things I reviewed:
 
- **code_point:** We looked into the architecture – very interesting, to think about how deep normalization goes. The solution to encoding between utf8 and utf16 seems very elegant, LGTM
- **rtti.Unknown:** This is a new pattern for me, but if I understand right, djs.Unknown is a subset (⊆) of rtti.Unknown, so this PR is about making sure rtti is allowed to evolve without getting blocked by djs.
- **paddedBaseN**: As I understand, the code for base64 and cbase32 are very similar: it first computes how many pad bits before reaching the next chunk boundary, and then appends that many bits before running the raw codec. So the proposal is to bring those two steps into a shared 'factory'. The difficulty is that, because they encode things differently, if paddedBaseN gets it wrong then you get incorrect encodings. Personally, I think this is a great pattern, I tend to lean in favor of abstraction (especially now that we have LLMs to make sure the abstractions actually hold weight).

I've submitted approval, but my personal agent (@tycho) has written some comments/feedback that maybe your agents can digest.

fs/cas/todo/66o-read-streamfile-dedup.md
sergey-shandar: You're right — fixed in 77123bc. `casAddFile` absorbs `ReadBytes` via its union return type `Effect<O | ReadBytes, …>` (the generic on `cas.write`), not an `as` cast. `read` can't reuse that because its return type is pinned to `List<FileCasOperation, …>` by the `FileCas` interface. I rewrote the caveat to say an explicit cast (or restructuring) is likely needed, that `tsc` should confirm it before committing to the one-liner, and that an unavoidable `as` should be weighed against the dedup value per AGENTS.md. Also corrected the "Related" note so it no longer claims `casAddFile` performs a reusable cast.

---

fs/cas/todo/66o-read-streamfile-dedup.md
sergey-shandar: Good catch — fixed in 77123bc. The quotes were verbatim when written, but the `main` merges (#1174 "List improvement") renamed the list constructors `cons`/`end` → `nonEmpty`/`elEmpty` (imported as `nonEmpty, empty as elEmpty` from `fs/effects/list/module.f.ts`). Updated both code blocks and the prose to the current identifiers and refreshed the line ranges.

---

fs/base_n/todo/basen-padding-strategy.md
sergey-shandar: Correct — fixed in 77123bc. The `various-basen-encodings.md` "Base64" entry asks for an identifier-safe base64 (`0-9 A-Z a-z _ $`), which is a different alphabet from `fs/base64`'s standard RFC 4648 `+/` codec, so it's not superseded. Reworded the "Related" note to say the identifier variant is still unbuilt and that this issue only refactors padding shared by the existing standard base64 and cbase32.

---

fs/mcp/todo/share-ok-result.md
sergey-shandar: Fixed in 77123bc — corrected to nine call sites and listed the exact lines (163, 182, 215, 218, 230, 231, 244, 247, 256).

---

sergey-shandar:
@codex review

PR #1173, opened 2026-06-27
Add proof object with test coverage for casMcpServer and collectRead

sergey-shandar:
## Summary
Added a `proof` export object containing test cases to improve code coverage for critical paths in the MCP server module that are not exercised by integration tests.

## Key Changes
- Imported `vec` function from bit_vec module to support test data generation
- Added `proof` object with two test cases:
  - `casMcpServer`: Calls the server initialization function to cover its Effect-building body, which is not exercised in integration tests since it drives a real stdio server
  - `collectReadOverflow`: Tests the overflow guard in `collectRead` by feeding a pure stream with two chunks that together exceed `maxLength`, ensuring the error branch executes without requiring real I/O

## Implementation Details
The `collectReadOverflow` test constructs a synthetic stream using `cons` and `end` primitives with two `ok` results: the first chunk is exactly half of `maxLength`, and the second chunk is `maxLength/2 + 1`, which causes their sum to exceed the limit and trigger the overflow error handling path.

https://claude.ai/code/session_01DSSkXHF2WfS1eoJNtMfuGN


sergey-shandar:
@codex review

sergey-shandar:
@codex review

sergey-shandar:
@codex review

fs/cas/mcp/module.f.ts
sergey-shandar:
```suggestion
import { empty, length as bitVecLength, maxLength, maxLengthBytes, msb, vec, type Vec } from '../../types/bit_vec/module.f.ts'
```

PR #1176, opened 2026-06-27
Make Pure<T> lazy by representing it as a thunk

sergey-shandar:
## Summary

This PR implements the proposal to make `Pure<T>` lazy by changing its representation from a strict tuple `readonly [T]` to a lazy thunk `() => T`. This eliminates redundant function calls in effect list processing and simplifies the representation of lazy streams.

## Key Changes

- **Changed `Pure<T>` representation**: From `readonly [T]` to `() => T`, making pure values lazy thunks
- **Simplified `List<O, T>` type**: Removed the intermediate `Next<O, T>` wrapper type. Lists now directly return `Effect<O, NonEmpty<O, T> | undefined>` instead of `Effect<O, () => NonEmpty<O, T> | undefined>`
- **Updated `pure()` function**: Now wraps values in a thunk (`() => v`) instead of a tuple (`[v]`)
- **Updated `decode()` function**: Changed detection logic from checking tuple length to checking if value is a function
- **Removed thunk invocations**: Updated all stream processing code to work directly with list cells instead of invoking intermediate thunks:
  - `fs/cas/proof.f.ts`: Removed `nodeThunk()` calls in stream drains
  - `fs/cas/mcp/module.f.ts`: Removed thunk invocation in `collectRead`
  - `fs/cas/module.f.ts`: Removed thunk invocation in file CAS stream processing
  - `fs/effects/node/module.f.ts`: Removed thunk invocation in write loop
- **Removed TODO document**: Deleted `fs/effects/todo/lazy.md` as the proposal is now implemented

## Implementation Details

The change makes `Effect` itself the single suspension boundary for lazy streams. Previously, effectful list operations required two suspension points: the effect and an additional thunk. Now the effect directly returns the next list cell or `undefined`, reducing overhead in hot paths like file chunk reading and stream processing.

All existing code that consumes pure values or list cells has been updated to work with the new lazy representation.

https://claude.ai/code/session_01EojqYPBAvGYMrSYygiHDL6


fs/effects/list/module.f.ts
```diff
@@ -5,35 +5,27 @@ export type NonEmpty<O extends Operation, T> = {
     readonly tail: List<O, T>
 }
 
-export type Next<O extends Operation, T> =
```
sergey-shandar, line 8: Keep the `Next` type because we use `Effect<O, Next<O, T>>` as a replacement for `List<O, T>` In some cases.

sergey-shandar: Restored `Next` in 32ee2d3. It's now `NonEmpty<O, T> | undefined` (the cons cell or end-of-stream marker), and `empty`/`nonEmpty` are typed `Effect<O, Next<O, T>>` so it can stand in for `List<O, T>` where the recursive payload can't be written as a return type.

---

fs/effects/list/module.f.ts
```diff
@@ -5,35 +5,37 @@ export type NonEmpty<O extends Operation, T> = {
...
  *
- * Built as an `Effect` object literal rather than through `pure`. `pure` returns
- * `Effect<never, …>`; widening that to an arbitrary operation set `O` is sound (a pure
- * value performs no operations) and the compiler accepts it for a non-recursive
- * payload, but not when the payload recursively mentions `O` — as `List`'s cons
- * cell does. Writing the literal directly lets the contextual return type drive the
- * check, so the recursive payload type-checks without a cast. Construct streams through
- * these two combinators.
- *
- * Note: we use `Effect<O, Next<O, T>>` because TypeScript can't convert `pure(...)` to
```
sergey-shandar, line 25: keep this note

sergey-shandar: Restored the note verbatim in 0891830.

---

fs/effects/module.f.ts
```diff
@@ -33,7 +33,7 @@ export type Do<O extends Operation, T> =
     DoK<O, T, O[0]>
 
 export const pure = <T>(v: T): Effect<never, T> => ({
```
sergey-shandar, line 35: We also need a function that accepts a thunk instead of a value. For example, if we need to create the next item for a list, we don't want to instantiate all items.

sergey-shandar: Added `lazy` in 4da3ad2:

```ts
export const lazy = <T>(t: () => T): Effect<never, T> => ({
    value: t,
    step: f => f(t())
})
```

It mirrors `pure` but takes a thunk and evaluates it on demand, so the next list item can be produced without instantiating the rest — e.g. `lazy(() => ({ first, tail: next() }))` defers building `tail` until the cell is decoded. Added proofs covering the value and that the thunk isn't forced until decode.

Left `pure` and the existing `nonEmpty`/`empty` signatures as-is (they take already-built values); let me know if you'd also like `nonEmpty` to accept a thunk tail.

---

sergey-shandar:
@codex review

PR #1178, opened 2026-06-28
Document cas_get metadata-only streaming proposal

sergey-shandar:
## Summary

This PR adds a design document outlining a proposal to fix `cas_get` failures on large blobs when only metadata is requested.

## Changes

- **New file:** `fs/cas/mcp/todo/cas-get-large-files.md`
  - Documents the problem: `cas_get` with `content: false` fails on blobs larger than 128 KiB because it collects the entire blob into a single `Vec` even when only metadata is needed
  - Proposes a streaming MIME/metadata detector (`detectStream`) that derives metadata without materializing the full blob content
  - Outlines implementation tasks including adding incremental UTF-8 validation, rewiring `cas_get` to use the streaming path for metadata-only requests, and adding proof tests
  - Clarifies the separation between inspection (cheap, streaming, size-independent) and transfer (bounded by `maxLength`)

## Key Points

- The issue prevents agents from inspecting large blobs to decide whether to fetch via URL or inline content
- The proposal keeps magic-byte detection logic in `fs/mime` while adding a streaming variant
- Metadata derivation (length, MIME type, text vs binary) can be done incrementally without holding the full blob in memory
- Only the `content: true` path would continue to use `collectRead`, where the `maxLength` ceiling remains a genuine limitation

https://claude.ai/code/session_015fqpYTTLUbeTH6AxNBcPpa

### DIFF
--- a/fs/bnf/todo/layered-parser.md
+++ b/fs/bnf/todo/layered-parser.md
@@ -26,8 +26,9 @@ Every layer reuses the same BNF engine.
 
 ### Transducers, recognizers, and streaming
 
-A transducer's step is Mealy-style — `(state, inSymbol) => (state, outSymbol*)`,
-producing zero or more output symbols per input. A **recognizer** (see
+A transducer's step is Mealy-style, which is exactly the existing
+`StateScan<I, S, O>` (`fs/types/function/operator`): `(inSymbol, state) =>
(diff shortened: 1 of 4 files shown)


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1184, opened 2026-06-28
Add `todo` test marker for captured-but-unfixed bugs

sergey-shandar:
## Summary

This PR introduces a `todo` test marker to the emergent testing framework, allowing developers to capture proofs of known bugs without breaking CI. Tests marked with `todo` are expected to fail today but will signal when the bug is fixed by turning red.

## Problem

Currently, there's no way to commit executable proofs of bugs without either:
- Breaking CI (if written as a normal test)
- Hiding the test in comments/markdown where it silently rots

## Solution

Add a `todo` marker that inverts test expectations: a `todo` leaf passes when it throws (documenting the current buggy behavior) and fails when it returns (signaling the bug is fixed).

## Key Design

- **Two flags with XOR/OR logic:**
  - `throws`: inherited from ancestor `throw` keys (OR-accumulated)
  - `todo`: local to the leaf's own last key (not inherited)
  - `expectFailure = throws XOR todo` (determines pass/fail inversion)
  - `leafOnly = throws OR todo` (skips return-value walk and subtests)

- **Truth table:**
  - `{ add: fn }` → expects return normally
  - `{ throw: { x: fn } }` → expects throw
  - `{ todo: fn }` → expects throw (green today when buggy)
  - `{ throw: { todo: fn } }` → expects return normally (green today when not yet throwing)

- **Pure leaf behavior:** `todo` never walks return values or spawns subtests, regardless of XOR direction

- **Static guard:** Enforces that `todo` keys hold zero-argument functions at collection time

## Implementation Changes

All changes are localized to `fs/emergent_testing/module.f.ts`:
- Add `todo: boolean` to `TestEntry`
- Derive `expectFailure` (XOR) and `leafOnly` (OR) from flags
- Keep `throws` inherited; set `todo` from leaf's last key only
- Switch inversion logic to use `expectFailure`
- Skip walks/registration and drop star suffix based on `leafOnly`
- Update `Reporter.result` to annotate passing `todo` leaves with `# TODO`
- Add static validation for `todo` key structure

## Migration

Existing code using `throw: { todo }` will need to be migrated, as the new semantics change its behavior from passing to failing.

https://claude.ai/code/session_01EkkLa6DPxLZRsNEAaRhACy

### DIFF
--- a/fs/emergent_testing/todo/todo-property.md
+++ b/fs/emergent_testing/todo/todo-property.md
@@ -18,110 +18,171 @@ either breaking CI or hiding the test:
 
 A proof author needs a way to say "this test documents a known bug; it is
 *expected to fail today*, so a failure must not break CI — but the moment the
-behaviour is fixed, flip the test green so we are reminded to remove the
-marker."
+behaviour is fixed, flip the test red so we are reminded to remove the marker."
 
 The existing `throw` marker (`fs/emergent_testing/module.f.ts`) already
-demonstrates the mechanism we want: a structural property key in the proof tree
-that *inverts* the pass/fail expectation for every leaf reachable through it.
+demonstrates the mechanism: a structural property in the proof tree that
+*inverts* a leaf's pass/fail expectation.
 
 ### Proposal
 
-Add a second structural marker, `todo`, alongside `throw`. Like `throw`, it is a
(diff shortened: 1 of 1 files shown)


sergey-shandar:
@codex review

PR #1185, opened 2026-06-29
Document cas_get content oversizing issue and error handling

sergey-shandar:
## Summary

This PR adds documentation for an existing issue in the `cas_get` handler where oversized blobs (exceeding 128 KiB) are incorrectly reported as "not found" instead of returning a proper "too large" error message.

## Changes

- **Added issue tracker document** (`fs/cas/todo/66p-cas-get-content-too-big-message.md`):
  - Documents the problem: `cas_get({ content: true })` collapses both "hash not found" and "blob exceeds maxLength" errors into the same "no such hash" message
  - Explains the impact: clients cannot distinguish between genuinely absent hashes and oversized blobs that exist but cannot be fetched inline
  - Notes that the metadata-only path already works correctly for any size
  - Proposes distinguishing the two error cases and returning a descriptive message for oversized blobs
  - Suggests pre-checking blob size before attempting `collectRead` to avoid the size limit error
  - Includes concrete tasks for implementation and testing
  - References related issues (66j, 66k, 66n)

## Context

This is a documentation-only change that captures an existing bug in the CAS (Content Addressable Store) module. The issue affects the inline content fetch path when blobs exceed the 128 KiB vector length limit, causing misleading error messages to MCP clients. The fix will require distinguishing error types in `collectRead` or pre-checking blob size before attempting inline materialization.

https://claude.ai/code/session_01Sr6AifXuN9S88mChqGKMvc

### DIFF
--- a/fs/cas/todo/66p-cas-get-content-too-big-message.md
+++ b/fs/cas/todo/66p-cas-get-content-too-big-message.md
@@ -1,0 +1,88 @@
+## 66P-cas-get-content-too-big-message. `cas_get({ content: true })` reports oversized blobs as "not found"
+
+**Priority:** P3
+**Status:** open
+**Blocked by:** —
+
+### Problem
+
+In `fs/cas/mcp/module.f.ts`, `cas_get` with `content: true` materializes the blob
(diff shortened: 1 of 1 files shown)


sergey-shandar:
@codex review

PR #1186, opened 2026-06-29
cas_get: distinguish oversized blobs from missing hashes

sergey-shandar:
## Summary

When `cas_get` is called with `content: true` on a blob larger than the inline limit (128 KiB), it now returns a distinct "too large" error instead of misreporting the blob as absent. This allows callers to distinguish between a genuinely missing hash and an existing blob that exceeds the buffering limit.

## Key Changes

- **Module logic** (`fs/cas/mcp/module.f.ts`):
  - Refactored `content: true` path to call `detectStream` first (size-independent metadata detection) before attempting to buffer the blob
  - Added explicit size check against `maxLengthBytes` (128 KiB) before calling `collectRead`
  - Oversized blobs now return a descriptive error message including the actual byte size, limit, and `url` alternative
  - Genuinely absent hashes still return "no such hash" error
  - Removed unused `detectVec` import

- **Test coverage** (`fs/cas/mcp/proof.f.ts`):
  - Added `getContentLargeBlobTooLargeError`: verifies that `content: true` on a present blob larger than `maxLength` returns the "too large" error, not "no such hash"
  - Added `getContentMissingHashIsError`: verifies that `content: true` on a genuinely absent hash still returns "no such hash"

- **Documentation** (`fs/cas/mcp/README.md`):
  - Added new "Inline-content size limit" section explaining the 128 KiB limit
  - Documented the distinction between "too large" (blob exists but oversized) and "no such hash" (genuinely absent)
  - Included example error message and guidance on alternatives (use `url` or omit `content`)

## Implementation Details

The fix separates two failure modes that were previously conflated:
1. **Stream error** (absent hash) → "no such hash"
2. **Size exceeded** (blob exists but too large) → "blob too large to fetch inline"

By running the size-independent `detectStream` machine first, we can determine the blob's size and type before attempting to materialize it, allowing us to short-circuit with the appropriate error message before `collectRead` is called.

https://claude.ai/code/session_01Mmogv8xHrkfB2L2xbStgzB


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1187, opened 2026-06-29
Add inline content size limit (128 KiB) to cas_add and cas_get

sergey-shandar:
## Summary

This PR adds an explicit size limit for inline content in `cas_add` and `cas_get` operations. Inline content (`text` and `base64` types) is now capped at 128 KiB (131,072 bytes), with larger payloads rejected with a descriptive error message that directs users to the `type: 'url'` streaming alternative.

## Key Changes

- **cas_add validation**: Added a check to reject inline content (`text`/`base64`) larger than `maxLength` (128 KiB). When exceeded, returns an `isError` result with a message that includes the actual byte size and points users to `type: 'url'` for streaming files without size limits.

- **Error messaging**: The error message is descriptive and actionable, e.g.:
  ```
  content too large to store inline (262144 bytes, limit 131072 bytes); use type:"url" to stream a file from /home/user/cas_upload/ (no size limit)
  ```

- **Documentation updates**: 
  - Updated tool descriptions for both `cas_add` and `cas_get` to clearly state the 128 KiB inline limit
  - Added a new "Inline-content size limit" section in README.md explaining the limit and the `type: 'url'` alternative
  - Updated error documentation to list this new failure mode

- **Test coverage**: Added `addOversizedInlineContentIsError` test to verify that oversized inline content is properly rejected with the expected error message.

## Implementation Details

- The size check uses `bitVecLength(value) > maxLength` to validate the resolved content before writing to storage
- The limit mirrors the existing inline-fetch guard in `cas_get`, creating consistent behavior across both operations
- The `type: 'url'` path remains unlimited, allowing users to stream large files from `$HOME/cas_upload/` without restrictions

https://claude.ai/code/session_01Sf4iq4UyP2enh6ghFDdpVC

### DIFF
--- a/fs/cas/mcp/module.f.ts
+++ b/fs/cas/mcp/module.f.ts
@@ -201,7 +201,7 @@ const casToolRegistry =
     ),
     toolEntry(
         'cas_get',
-        'Inspect a blob by hash. Always returns JSON {length,mime_type,type[,url]} where type is "text" or "base64". Pass content:true to also include the inline content string.',
+        'Inspect a blob by hash. Always returns JSON {length,mime_type,type[,url]} where type is "text" or "base64". Pass content:true to also include the inline content string, but content is capped at 128 KiB (131072 bytes) — a larger blob is rejected with an error. To download a blob, prefer the url field returned in the result instead of requesting inline content.',
         casGetArgs,
         r => {
             const key = cBase32ToVec(r.hash)
--- a/fs/cas/mcp/module.f.ts
+++ b/fs/cas/mcp/module.f.ts
@@ -27,12 +27,6 @@
  *   moves the file via the streaming move-hash-move pipeline (no size limit).
  *   Paths outside `$HOME/cas_upload/` or containing `..` are rejected for security.
  *
- * **Inline content (`text`/`base64`) is capped at `maxLength` (128 KiB).** Both
- * resolve `content` into a single `Vec`, which cannot exceed `maxLength` bits, so a
- * larger payload is rejected with a descriptive *"too large"* error (carrying the
- * byte size) that points at `type: 'url'`. This mirrors the `content: true` guard in
- * `cas_get`: `type: 'url'` is the size-independent path for blobs above the limit.
- *
  * ## `cas_get` output
  *
  * Always returns a JSON object `{ length, mime_type, type[, url][, content] }`.
@@ -76,8 +70,6 @@
  * `mcpStep`. *Tool* failures come back as a normal `tools/call` result with
  * `isError: true` and a text explanation, so the model can read and react:
  * - `type: 'base64'` with malformed content (base64 `decode` → `null`) → `isError`
- * - `cas_add` with inline `content` (`text`/`base64`) larger than `maxLength`
- *   (128 KiB) → `isError` ("too large" message, pointing at `type: 'url'`)
  * - malformed `hash` (`cBase32ToVec` → `null`) → `isError`
  * - `cas_get` on an absent hash (`c.read` → `undefined`) → `isError`
  * - `cas_get` with `content: true` on a blob larger than `maxLength` → `isError`
@@ -198,24 +190,13 @@ const casToolRegistry =
                     x = pure(utf8(content))
                     break
             }
-            return x.step(value => {
-                if (typeof value === 'string') {
-                    return pure(errorResult(value))
-                }
-                // Inline content (text/base64) is resolved into a single `Vec`, which
-                // caps at `maxLength` bits (`maxLengthBytes` bytes, 128 KiB). Reject a
-                // larger payload with a descriptive error pointing at the size-independent
-                // alternative (`type: 'url'`), mirroring the inline-fetch guard in
-                // `cas_get`, instead of silently storing an over-`maxLength` blob.
-                if (bitVecLength(value) > maxLength) {
-                    return pure(errorResult(
-                        `content too large to store inline (${bitVecLength(value) / 8n} bytes, limit ${maxLengthBytes} bytes); use type:"url" to stream a file from ${casUploadDir}/ (no size limit)`))
-                }
+            return x.step(value => typeof value === 'string'
+                ? pure(errorResult(value))
                 // The resolved content fits in one chunk; feed it as a single-item stream.
-                return c.write(nonEmpty(ok(value), elEmpty<never, Ok<Vec>>())).step(([tag, hash]) => pure(tag === 'error'
+                : c.write(nonEmpty(ok(value), elEmpty<never, Ok<Vec>>())).step(([tag, hash]) => pure(tag === 'error'
                     ? errorResult('write')
                     : okResult(vecToCBase32(hash))))
-            })
+            )
         },
     ),
     toolEntry(
--- a/fs/cas/mcp/module.f.ts
+++ b/fs/cas/mcp/module.f.ts
@@ -27,6 +27,12 @@
  *   moves the file via the streaming move-hash-move pipeline (no size limit).
  *   Paths outside `$HOME/cas_upload/` or containing `..` are rejected for security.
  *
+ * **Inline content (`text`/`base64`) is capped at `maxLength` (128 KiB).** Both
+ * resolve `content` into a single `Vec`, which cannot exceed `maxLength` bits, so a
+ * larger payload is rejected with a descriptive *"too large"* error (carrying the
+ * byte size) that points at `type: 'url'`. This mirrors the `content: true` guard in
+ * `cas_get`: `type: 'url'` is the size-independent path for blobs above the limit.
+ *
  * ## `cas_get` output
  *
  * Always returns a JSON object `{ length, mime_type, type[, url][, content] }`.
@@ -70,6 +76,8 @@
  * `mcpStep`. *Tool* failures come back as a normal `tools/call` result with
  * `isError: true` and a text explanation, so the model can read and react:
  * - `type: 'base64'` with malformed content (base64 `decode` → `null`) → `isError`
+ * - `cas_add` with inline `content` (`text`/`base64`) larger than `maxLength`
+ *   (128 KiB) → `isError` ("too large" message, pointing at `type: 'url'`)
  * - malformed `hash` (`cBase32ToVec` → `null`) → `isError`
  * - `cas_get` on an absent hash (`c.read` → `undefined`) → `isError`
  * - `cas_get` with `content: true` on a blob larger than `maxLength` → `isError`
@@ -165,7 +173,7 @@ const casToolRegistry =
     return [
     toolEntry(
         'cas_add',
-        'Store content and return its hash (cBase32). Pass type:"base64" for binary; type:"url" to stream a file from $HOME/cas_upload/ (no size limit); omit or pass type:"text" for UTF-8 text (default).',
+        'Store content and return its hash (cBase32). Pass type:"base64" for binary; omit or pass type:"text" for UTF-8 text (default). Inline content (text/base64) is capped at 128 KiB (131072 bytes) — larger content is rejected. For larger content use type:"url" to stream a file from $HOME/cas_upload/ (no size limit).',
         casAddArgs,
         ({ type, content }): Effect<FileCasOperation | Rm, ToolsCallResult> => {
             // type:'url' — stream the file into cas, then delete the source on success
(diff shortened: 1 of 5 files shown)


fs/cas/mcp/module.f.ts
```diff
@@ -190,18 +198,29 @@ const casToolRegistry =
...
+            return x.step(value => {
+                if (typeof value === 'string') {
+                    return pure(errorResult(value))
+                }
+                // Inline content (text/base64) is resolved into a single `Vec`, which
+                // caps at `maxLength` bits (`maxLengthBytes` bytes, 128 KiB). Reject a
+                // larger payload with a descriptive error pointing at the size-independent
+                // alternative (`type: 'url'`), mirroring the inline-fetch guard in
+                // `cas_get`, instead of silently storing an over-`maxLength` blob.
+                if (bitVecLength(value) > maxLength) {
```
sergey-shandar: This's too late, `bun` will already fail on either `utf8(content)` or `base64decode(content)`. Proposed fix: Both of these functions should return a special error. For this specific PR: revert changes in code (except `cas_get` and `cas_add` descriptions) and create `todo` file with the proposed fix. Don't fix the logic in the PR.

sergey-shandar: Done in e8c604c. Reverted the `cas_add` guard, docstring, and proof-test changes — kept only the `cas_add`/`cas_get` tool description strings documenting the 128 KiB inline limit and recommending `type:"url"`. Filed the proposed fix (have `utf8` and base64 `decode` return a special "too large" error so the handler can react before `bun` throws on the over-`maxLength` bigint) as `fs/cas/todo/66q-cas-add-inline-size-error.md`.

---

fs/cas/todo/66q-cas-add-inline-size-error.md
```diff
@@ -0,0 +1,69 @@
...
+
+- `type: 'text'` → `utf8(content)` (`fs/text/module.f.ts`)
+- `type: 'base64'` → `decode(content)` (`fs/base64/module.f.ts`)
+
+Both build an over-`maxLength` `bigint`, which **`bun` throws on** (see the
+`maxLength` notes in `fs/types/bigint/module.f.ts`). So a size guard placed in the
+`cas_add` handler after the conversion (e.g. `if (bitVecLength(value) > maxLength)`)
+is too late — `bun` has already failed in `utf8` / `decode`. On `node` the same input
+does not throw and would instead be **silently stored** as an over-`maxLength` blob,
+which then cannot be read back inline via `cas_get` `content: true`.
```
sergey-shandar: This sentence doesn't make much sense since the size limitation is only for the specific MCP commands, and only when they are used with inlined content. For example, `cas_add({ type: url, ...})` should work with any content. The sentence is correct in the sense that Node and Deno just have another limit for bigint (bit vec), but we would like to have the same size limit on any JS engine, so our program has the same behavior on all JS engines. Another reason, 128 KiB is a good size to limit LLM waste tokens on file transfer. From a library perspective, we should discourage the use of large bit vectors, and users should consider partitioning and streaming for big data instead of loading it in memory as one piece.

fs/cas/todo/66q-cas-add-inline-size-error.md
```diff
@@ -0,0 +1,69 @@
...
+behaviour at the limit is inconsistent across runtimes and not a clean tool error.
+
+### Proposal
+
+Make the conversion functions report the size failure as data instead of throwing,
+so the `cas_add` handler can turn it into a normal `isError` tool result that names
+the byte size and points at `type: 'url'`.
+
+1. **`fs/base64/module.f.ts` `decode`** — already returns `Nullable<Vec>`. Extend the
+   contract so input that would exceed `maxLength` bits returns a distinguished error
```
sergey-shandar: Should we use `Result<Vec, DecodeErrorType>` where `DecodeErrorType` is a union of string errors = "..." | "..."`?

fs/cas/todo/66q-cas-add-inline-size-error.md
```diff
@@ -0,0 +1,69 @@
...
+   rather than throwing in `bun`. Either a dedicated error variant (preferred, so the
+   handler can tell "too large" from "malformed base64") or, at minimum, a non-throwing
+   `null`.
+2. **`fs/text/module.f.ts` `utf8`** — currently returns `Utf8` (`Vec`) and cannot fail.
+   Give it a fallible variant that returns a distinguished "too large" error when the
```
sergey-shandar: Make it `Nullable<Vec>` or `Result<Vec, DecodeErrorType>` make it consistent with `base64decode`?

fs/cas/todo/66q-cas-add-inline-size-error.md
sergey-shandar: Reframed the Problem section in 98ba058. It now states the 128 KiB inline cap is intentional and applies only to inlined MCP content (`type: 'text'`/`'base64'`) — `type: 'url'` handles any size — and lists the three reasons you gave: uniform behaviour across JS engines (a `bigint` over `maxLength` throws on bun but is accepted on node/deno, so the effective limit is engine-dependent today), LLM token economy, and discouraging large bit vectors in favour of partitioning/streaming. The "defect" is narrowed to: the cap isn't enforced as a clean error because the conversion builds the over-`maxLength` `Vec` before the handler can react.

---

fs/cas/todo/66q-cas-add-inline-size-error.md
sergey-shandar: Yes — adopted in 98ba058. The proposal now uses `Result<Vec, DecodeErrorType>` with a shared string union `type DecodeErrorType = 'too-big' | 'invalid'`: `decode` returns `error('invalid')` for malformed base64 and `error('too-big')` when the decoded `Vec` would exceed `maxLength` (checked before constructing the bigint). The `cas_add` handler matches `'too-big'` → size/limit error recommending `type:"url"`, `'invalid'` → the existing base64 error. Left an open question on where `DecodeErrorType` should live (`fs/base64` vs a shared `fs/types` spot).

---

fs/cas/todo/66q-cas-add-inline-size-error.md
sergey-shandar: Made it consistent in 98ba058: the fallible `utf8` variant also returns `Result<Vec, DecodeErrorType>` (it only ever yields `error('too-big')`, since every string is valid to UTF-8-encode), matching `decode`. Also noted an open question: whether the size check should instead live in a low-level `bit_vec` constructor so any over-`maxLength` `Vec` construction fails uniformly and both `utf8` and `decode` inherit it — generalizing the "discourage large bit vectors" guidance — at the cost of a wider API change.

---

fs/cas/todo/66q-cas-add-inline-size-error.md
```diff
@@ -0,0 +1,103 @@
...
+desirable**, not an accident to be lifted. It applies only to *inlined* MCP content
+(`type: 'text'` / `type: 'base64'`); `cas_add({ type: 'url', … })` and reading via the
+`cas_get` `url` field already handle content of any size. We keep the inline cap on
+purpose:
+
+- **Uniform behaviour across JS engines.** A single `Vec` is a `bigint`, and a
+  `bigint` over `maxLength` bits *throws on `bun`* but is *accepted on `node`/`deno`*
+  (see the `maxLength` notes in `fs/types/bigint/module.f.ts`). So today the effective
+  inline limit is engine-dependent. We want the *same* explicit 128 KiB limit and the
+  *same* behaviour on every JS engine.
```
sergey-shandar: It is worth mentioning that other JS engines also have bigint size limitations, but `bun` (WebKit) has the smallest one.

fs/cas/todo/66q-cas-add-inline-size-error.md
```diff
@@ -0,0 +1,103 @@
...
+type, so both conversions report failures consistently and the handler can tell the
+two cases apart:
+
+```ts
+// fs/base64 (or a shared location both modules import)
+type DecodeErrorType = 'too-big' | 'invalid'
+```
+
+1. **`fs/base64/module.f.ts` `decode`** — change from `Nullable<Vec>` to
+   `Result<Vec, DecodeErrorType>`: `error('invalid')` for malformed base64,
```
sergey-shandar: We should specify where `DecodeErrorType` will live, some common place.

fs/cas/todo/66q-cas-add-inline-size-error.md
```diff
@@ -0,0 +1,103 @@
...
+     `cas_get` `content: true` path);
+   - `'invalid'` → the existing `invalid base64 content` error.
+
+Reuse the byte-aligned limit constants already exported from
+`fs/types/bit_vec/module.f.ts` (`maxLength`, `maxLengthBytes`).
+
+### Open questions
+
+- Where should `DecodeErrorType` live — in `fs/base64` and imported by `fs/text`, or in
+  a shared `fs/types` location? Pick whichever keeps the dependency direction clean.
```
sergey-shandar: What about `types/result/module.f.ts`? Unless you have a better idea.

fs/cas/todo/66q-cas-add-inline-size-error.md
sergey-shandar: Added in 21e92ab. The "uniform behaviour" bullet now says every JS engine caps `bigint` size, but the cap differs: `bun` (WebKit/JSC) has the smallest and throws first, while `node`/`deno` (V8) allow much larger `bigint`s and so don't throw at 128 KiB.

---

fs/cas/todo/66q-cas-add-inline-size-error.md
sergey-shandar: Specified in 21e92ab: `DecodeErrorType` lives in `fs/types/result/module.f.ts` (per your suggestion below), so both `fs/base64` and `fs/text` import it without either depending on the other. Updated the code block, the task, and removed it from the open questions.

---

fs/cas/todo/66q-cas-add-inline-size-error.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+1. **`fs/base64/module.f.ts` `decode`** — change from `Nullable<Vec>` to
+   `Result<Vec, DecodeErrorType>`: `error('invalid')` for malformed base64,
+   `error('too-big')` when the decoded `Vec` would exceed `maxLength` bits (checked
+   *before* constructing the over-`maxLength` `bigint`). The only in-`fs` caller is the
+   CAS MCP handler (`fs/cas/mcp/module.f.ts`), so the migration is contained.
+2. **`fs/text/module.f.ts` `utf8`** — add a fallible variant returning
+   `Result<Vec, DecodeErrorType>` for consistency with `decode` (it only ever yields
+   `error('too-big')`, since every string is valid to UTF-8-encode). Keep the existing
```
sergey-shandar: Renamed to `'tooBig'` throughout in 21e92ab (`DecodeErrorType = 'tooBig' | 'invalid'`). Good distinction — `too_big`/`too-big` for filesystem names, camelCase for JS identifiers.

---

fs/cas/todo/66q-cas-add-inline-size-error.md
sergey-shandar: `fs/types/result/module.f.ts` works well — adopted in 21e92ab. It's already the common `Result` home that both `fs/base64` and `fs/text` can import without creating a cross-dependency, so no better idea needed.

---

fs/cas/todo/66q-cas-add-inline-size-error.md
```diff
@@ -0,0 +1,106 @@
...
+
+Reuse the byte-aligned limit constants already exported from
+`fs/types/bit_vec/module.f.ts` (`maxLength`, `maxLengthBytes`).
+
+`DecodeErrorType` lives in `fs/types/result/module.f.ts` (the common `Result` home),
+so both `fs/base64` and `fs/text` import it without either depending on the other.
+
+### Open questions
+
+- Should the size check live in a low-level `bit_vec` constructor (so *any* over-
```
sergey-shandar: Agreed — revised in a392a97. The proposal now *adds* new size-checked operations (working names `tryDecode` in `fs/base64` and `tryUtf8` in `fs/text`, both returning `Result<Vec, DecodeErrorType>`) and keeps the existing `decode: Nullable<Vec>` and total `utf8` unchanged, so no current caller is affected. The CAS MCP handler switches to the new ops. Left final names and "size check in each op vs. a shared `bit_vec` helper" as open questions.

---

sergey-shandar:
@codex review

PR #1192, opened 2026-06-29
Optimize bit-vector concatenation from O(n²) to O(n log n)

sergey-shandar:
## Summary

Implements a binary-counter accumulator pattern for bit-vector concatenation, reducing the algorithmic complexity from O(n²) to O(n log n). This eliminates a significant performance bottleneck in operations that build vectors from many small chunks, such as hex/base-N string decoding.

## Key Changes

- **Extracted `unpackListToVec` helper** in `fs/types/bit_vec/module.f.ts`: A reusable binary-counter accumulator that was previously private to `u8ListToVec`. Each slot `i` holds a pre-combined run of `2^i` elements; arriving elements "carry" upward like binary number increments, ensuring every merge joins comparable-sized runs.

- **Reimplemented `BitOrder.listToVec`** to use the new `unpackListToVec` helper instead of a naive left fold, making the public API O(n log n) for all callers.

- **Simplified `u8ListToVec`** to delegate to `unpackListToVec`, eliminating duplicated binary-counter logic.

- **Refactored `fs/base_n/module.f.ts` `stringToVec`** to collect all character indices first, validate them in one pass, then build the vector via `listToVec` instead of folding `concat` per character. Removed the now-unused `concat` and `empty` imports.

- **Re-enabled the `big` proof** in `fs/base_n/proof.f.ts`: The 1 Mbit hex decode test now runs in well under a second (previously ~13 s on Node, ~43 s on Bun).

- **Removed the TODO document** `fs/types/bit_vec/todo/vec-builder.md` as the work is now complete.

## Implementation Details

- **Order preservation**: The binary counter merges with the earlier run on the left (`unpackConcat(old)(cur)`), and the final reduce prepends higher slots in front, preserving left-to-right element order.
- **Empty list handling**: Still yields `unpackEmpty` as the reduce seed.
- **Validation in `stringToVec`**: Early `null` return on invalid characters is preserved by checking all indices before building.

https://claude.ai/code/session_01DMW9J2zdcMZeYTsoBuTn45


fs/base_n/proof.f.ts
```diff
@@ -58,11 +59,11 @@ export const proof = {
...
         if (cb32.stringToVec('u') !== null) { throw 'unknown char should return null' }
     },
-    // Disabled until bit_vec concatenation is O(n log n); the naive per-chunk
-    // `concat` makes this 1 Mbit decode O(n^2) (~13 s node / ~43 s bun).
-    // See fs/types/bit_vec/todo/vec-builder.md.
-    // big: () => {
-    //     const x = hex.stringToVec(bigSampleHex)
-    //     assertEq(length(x!), maxLength)
-    // }
+    // Decodes a 1 Mbit hex string. With the O(n log n) `listToVec` builder this
```
sergey-shandar:
```suggestion
    // Decodes a 1 Mibit hex string. With the O(n log n) `listToVec` builder this
```

sergey-shandar:
@codex review

fs/base_n/proof.f.ts
sergey-shandar: Done in 070a4b0 — also fixed the matching "1 Mbit" in the `bigSampleHex` comment above for consistency.

---

fs/base_n/module.f.ts
```diff
@@ -66,13 +66,9 @@ export const baseN = (
...
         stringToVec: s => {
-            let result: Vec = empty
-            for (const c of s) {
-                const index = toIndex(c)
-                if (index < 0) { return null }
-                result = concat(result)(vecN(BigInt(index)))
-            }
-            return result
+            const indices = [...s].map(toIndex)
+            if (indices.some(index => index < 0)) { return null }
```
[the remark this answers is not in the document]
sergey-shandar: Good catch — fixed in 4eb834a. `stringToVec` now builds a reversed chunk list in a loop and bails at the first invalid character, so malformed input is rejected in O(prefix) time and `normalize` is never run past it; the valid case still concatenates via the O(n log n) `listToVec`. This also matches the TODO's "preserve the early null return on the first out-of-alphabet character" requirement.

---

sergey-shandar:
@codex review

fs/types/bit_vec/module.f.ts
```diff
@@ -292,6 +292,47 @@ type Base = {
...
+ *
+ * Slot `i` of `result` holds an already-combined run of the most recent
+ * `2 ** i` elements. Each arriving element "carries" upward, merging only with
+ * runs of comparable size — exactly like incrementing a binary number — so
+ * every merge joins two runs of similar length. Left-to-right element order is
+ * preserved: `unpackConcat(old)(cur)` keeps the earlier run on the left, and
+ * the final reduce prepends higher (earlier) slots in front of accumulated
+ * later runs. An empty list yields `unpackEmpty`.
+ *
+ * This is the bit-vector analogue of a `StringBuilder` / `strings.Builder`.
```
sergey-shandar: `StringBuilder` is a term used in some PLs, we should mention these common PLs.

sergey-shandar: Done in 1973f32 — the comment now names the languages: `StringBuilder` (Java, C#), `strings.Builder` (Go), `String::push_str` (Rust).

---

sergey-shandar: Agreed — `String::push_str` mutates a `String` in place, so the `String` is the result rather than a builder you finalize with a separate step. Dropped it in 856ebb8; the comment now lists only the accumulate-then-materialize builders: `StringBuilder` (Java, C#) and `strings.Builder` (Go).

---

PR #1196, opened 2026-06-30
Add tryUtf8, a non-throwing maxLength-safe UTF-8 encoder

sergey-shandar:
## Summary
- Add `tryUtf8`, a `Nullable`-returning sibling of `utf8` that reports `null` instead of throwing when the encoded result would exceed `maxLength`, following the `tryListToVec`/`tryU8ListToVec` convention from #1195.
- Rebuild `utf8` as `mapUnwrap(tryUtf8)` so it shares the conversion logic instead of duplicating it.
- Add proof cases: `tryUtf8` round-trips with `utf8`/`utf8ToString`, handles the empty string, and returns `null`/throws appropriately one byte past `maxLengthBytes`.

## Test plan
- [x] `npm test` — 2042/2042 passing


sergey-shandar:
@codex review

PR #1198, opened 2026-07-01
cas_add: graceful isError on oversized or malformed inline content

sergey-shandar:
## Summary

- Switch `cas_add` text branch from throwing `utf8` to `tryUtf8`, so an over-`maxLength` string yields `null` instead of crashing the MCP server process on Bun/JSC.
- Collapse both `null` paths (`tryUtf8` and `base64Decode`) into a single static `isError` message: `'too large or malformed — use type:"url" for large content'`. No content echo, no cause branching.
- Remove unused `utf8` / `Cas` / `FileCas` imports left over from the old switch/case.
- Add four proof cases covering the boundary: one byte over for both encodings, exactly at the limit for text (succeeds), and exactly at the limit for base64 (currently fails due to `fs/base64/todo/decode-rejects-max-size-input.md` — asserts current error behavior with a comment to flip once that fix lands).
- Add `base64OfA(n)` helper that computes a valid base64 string for `n` zero bytes directly from arithmetic — avoids calling `base64Encode` on a Vec near `maxLength` (which would trigger the encode-padding-overflow bug) and runs in O(n) string allocation instead of O(n²) bigint shifts.
- Update README: remove stale "tracked in cas-add-inline-size-error" note; update error-cases list to cover both malformed and oversized inline content.

## Test plan
- [x] `npm test` — 2046/2046 passing
- [x] `addTextOverLimitIsError` — one byte past `maxLengthBytes` → `isError`
- [x] `addBase64OverLimitIsError` — 8 bits past `maxLength` in base64 → `isError`
- [x] `addTextAtLimitSucceeds` — exactly `maxLengthBytes` ASCII bytes stores cleanly
- [x] `addBase64AtLimitIsError` — exactly `maxLengthBytes` base64 → current `isError` (gated on `decode-rejects-max-size-input.md`)


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1199, opened 2026-07-02
cas/mcp: collapse duplicated detectStream error handling in cas_get

sergey-shandar:
## Summary
- Removes the duplicated `detectStream(c.read(key)).step(...)` call and error-handling branch in the `cas_get` handler in [fs/cas/mcp/module.f.ts](fs/cas/mcp/module.f.ts) by threading the metadata-only (`content !== true`) return through the single shared `.step()` callback instead of a separate early-return branch.
- No behavior change — same metadata/content responses and error cases, just one code path instead of two.

## Test plan
- [x] `npm test` — 2046/2046 passing


sergey-shandar:
@codex review

PR #1201, opened 2026-07-02
mcp/stdio: fail soft on oversized JSON-RPC responses instead of crashing

sergey-shandar:
## Summary
- `fs/mcp/stdio/module.f.ts` `writeResponse`: switch from the throwing `utf8` to the non-throwing `tryUtf8`. When an encoded response line would exceed `maxLength` (128 KiB), the transport now writes a fixed JSON-RPC `Internal error` (`-32603`) response — carrying the *original request's* `id` — instead of throwing and crashing the MCP server process.
- Fixes `fs/cas/mcp/todo/cas-get-response-overflow.md`: `cas_get`'s own `length > maxLengthBytes` guard only checks the *raw* blob size, which cannot see base64 inflation (~4/3×) or JSON double-escaping (a literal quote can expand 1 byte → 4 bytes across the inner CAS-JSON pass and the outer JSON-RPC pass). Only the transport's real encode attempt can catch the final, actual line size, so the fix lives in `writeResponse`, not in `cas_get`.
- Removed the now-dead `writeMaybe` helper (folded into `handleLine`, which now threads the original `Response` through so the fallback can read `resp.id`) and the unused `utf8` import.

## Test plan
- [x] `npm test` — 2051/2051 passing
- `fs/mcp/stdio/proof.f.ts`: `oversizedResponseWritesInternalError`, `loopContinuesAfterOversizedResponse` (transport keeps draining stdin after the fallback fires)
- `fs/cas/mcp/proof.f.ts`: `getContentBase64InflationOverflowWritesInternalError` + `getContentBase64NearBoundarySucceeds` (base64-inflation boundary pair), `getContentDoubleEscapedOverflowWritesInternalError` (quote-heavy text triggering double-JSON-escape overflow)


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1202, opened 2026-07-02
base_n: fix O(n²) vecToString, restore base64-inflation boundary tests

sergey-shandar:
## Summary

- **Root cause fix**: `baseN.vecToString` (`fs/base_n/module.f.ts`, shared by `base64.encode` and `cbase32`) was O(n²) on large inputs. The old loop popped one `bits`-wide chunk off the front at a time via `popFront`, and each `popFront` re-masked the *entire remaining* bigint through `vec()`'s `m & ui` — cost proportional to remaining length, not chunk size — so encoding an `n`-bit vector cost O(n²) instead of O(n log n).
- Rewrote `vecToString` as a balanced recursive split (`unpackToString`), mirroring the `chunkList`/`unpackChunkList` divide-and-conquer already used by `u8List` (byte-unpacking) and by concatenation (`msb.concat`/`unpackListToVec`, PR #1192). Each half is explicitly masked to its own length before recursing, so every node's `uint` is genuinely bounded by its own claimed length at every level — true O(n log n), not just "correct but relies on a subtle argument about how often a stale spine gets touched."
- Restores the `fs/cas/mcp/proof.f.ts` base64-inflation boundary pair (`getContentBase64InflationOverflowWritesInternalError` / `getContentBase64NearBoundarySucceeds`) that PR #1201 removed because they blew `bun test`'s 5s per-test timeout (12-27s observed). They're back and comfortably fast now.
- Adds `encodeLargeVecIsSlow` to `fs/base64/proof.f.ts` — calls `encode()` directly on a 90,000-byte `Vec`, isolating the cost at the call site with no CAS/stdio machinery. No timing assertion (duration varies by engine/machine); it's a regression guard relying on the test runner's own per-test timing.
- Updates the CHANGELOG (`Unreleased`) and touches up comments that described the pre-fix timings as current.

## Numbers (90,000-byte `Vec`, `base64.encode`)

| | Before | After |
|---|---|---|
| Node | ~6.2s | ~50ms |
| Bun | ~18-27s (times out under `bun test`'s 5s limit) | ~225ms |

## Test plan
- [x] `node ./fs/fjs/module.ts t` — 2053/2053 passing
- [x] `bun test --coverage` (exact CI command) — 1942/1942 passing, 0 timeouts


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1203, opened 2026-07-02
crypto/sha2: fix append's O(n²) block-extraction loop

sergey-shandar:
## Summary

- **Root cause**: `append` (`fs/crypto/sha2/module.f.ts`) popped one fixed-size SHA-2 block off the front of `remainder` at a time via `popFront`, and each `popFront` re-masked the *entire remaining* bigint through `vec()`'s `m & ui` — cost proportional to the remaining length, not the block size. Looping that once per block is O(n²) instead of O(n log n). Same shape as `baseN.vecToString`, fixed in #1202.
- This is hotter than the `base64`/`cbase32` case: `fileCas.write` (`fs/cas/module.f.ts`) passes the whole content `Vec` (up to `maxLengthBytes`, 128 KiB) as a single `append` call, so it runs on *every* `cas_add`, not just `content:true` reads.
- Rewrites `append` as a fold over `chunkList(msb)(chunkLength)(...)` — the same pattern used for the `base_n` fix: reuses `State`'s own shape as the fold's accumulator (only the last, possibly-partial chunk becomes the new `remainder`; every other chunk is exactly `chunkLength` and gets compressed). Hoists `chunkListMsb` (module scope, no `chunkLength`/`v` dependency) and `chunkListChunkLength`/`foldChunks` (`base(...)` scope, no `v`/`state` dependency), per the new `AGENTS.md` rule about scoping curried partial applications to their dependencies.
- Adds `appendLargeVecIsFast` to `fs/crypto/sha2/proof.f.ts` — hashes a 100,000-byte `Vec` directly and checks the resulting hash against an independently-computed value, no timing assertion (relies on the test runner's own per-test timing to catch a regression), mirroring `fs/base64/proof.f.ts`'s `encodeLargeVecIsSlow`.
- Implements and deletes `fs/crypto/sha2/todo/sha2-append-quadratic.md`.

## Numbers

| bytes | before | after |
|---|---|---|
| 80,000 | 70.9ms | 43.4ms |
| 160,000 | 458.9ms | 70.4ms |
| 320,000 | 3444.1ms | 115.8ms |

Scaling per 2x input: was 6.5–7.5x (worse than O(n²)); now 1.4–1.8x (near-linear, allowing for the fixed per-block `compress` cost).

## Test plan
- [x] `node ./fs/fjs/module.ts t` — 2054/2054 passing
- [x] `bun test --coverage` (exact CI command) — 1943/1943 passing, 0 timeouts
- [x] All SHA-2 variants (sha256/224/512/384/512x256/512x224) and known-vector/padding-boundary tests pass unchanged


sergey-shandar:
@codex review

PR #1204, opened 2026-07-02
Add 15 todo issues for DRY refactoring and code deduplication

sergey-shandar:
## Summary

This PR adds 15 new todo issues documenting opportunities for DRY (Don't Repeat Yourself) refactoring across the codebase. These issues identify duplicated logic, shared patterns, and code that should be unified into single implementations.

### BNF Parser Infrastructure
- **nullable-analysis-shared**: Move the nullable/empty-tag analysis computation from both parser backends into a shared `bnf/data` module, resolving disagreement on sequence handling between descent and ll1 implementations
- **rule-visitor**: Extract the repeated `Rule` ADT three-way discrimination pattern into a shared visitor in `bnf/data`
- **data-tosequence-reuse**: Eliminate duplicate `toSequence` implementation in `bnf/data` by importing from the parent module

### Virtual Filesystem
- **dir-spine-descend**: Share the immutable `Dir`-spine descent pattern used in three helpers (`operation`, `extractEntity`, `insertEntityAt`)
- **resolve-file-helper**: Extract the repeated single-segment file resolver preamble used in four operation handlers
- **is-proper-prefix-path**: Move the pure `isProperPrefix` path predicate from the virtual FS interpreter to `fs/path` module

### Type System & Bit Operations
- **bit-set-factory**: Unify `byte_set` and `nibble_set` bitmask-set algebra implementations via a parameterized factory
- **id-prefix-tag-factory**: Consolidate the repeated prefix-tag scheme in `sul/id` using a factory pattern

### Text & Encoding
- **lead-byte-classifier-dedup**: Extract the duplicated UTF-8 lead-byte classifier from `utf8ByteToCodePointOp`
- **error-tag-layout-constants**: Name the partial-state/error-tag bit layout constants once instead of hardcoding them in two places
- **single-signature-table**: Eliminate duplicate magic-byte signature declarations in `fs/mime` by deriving from a single source of truth

### VM & Runtime
- **bigint-shift-decode**: Extract the shared shift-amount decode prelude from `Shl` and `Shr` operations
- **primitive-coercion-dispatch**: Route `any_to_number`/`any_to_string` through the existing `Unpacked::dispatch` instead of hand-rolling 5-arm matches
- **string-utf16-from-impls**: Move UTF-16 string conversion impls from `impls/from.rs` to `vm/string/`

### CLI
- **positional-arity-check**: Extract positional-argument arity validation from individual command handlers into a shared combinator

## Priority Distribution
- P3 (3 issues): nullable-analysis-shared, bit-set-factory, single-signature-table, resolve-file-helper
- P4 (8 issues): dir-spine-descend, rule-visitor, bigint-shift-decode, primitive-coercion-dispatch, string-utf16-from-impls, is-proper-prefix-path, error-tag-layout-constants, lead-byte-classifier-dedup
- P5 (2 issues): positional-arity-check, data-tosequence-reuse
- P3 (1 issue): id-prefix-tag-factory

## Implementation Notes

Each issue includes:
- Clear problem statement with code citations
- Concrete proposal with example signatures
- Explicit task checklist
- Cross-references to related work and dependencies

The issues follow established patterns in the codebase (e.g., `fs/base_n` factory precedent, `fs/types/rtti/common` visitor pattern) and are scoped to be actionable without requiring architectural changes.

https://claude.ai/code/session_01GWpx4aCDeEBiMrKu4PMhMz

### DIFF
--- a/fs/bnf/todo/data-tosequence-reuse.md
+++ b/fs/bnf/todo/data-tosequence-reuse.md
@@ -1,0 +1,48 @@
+## data-tosequence-reuse. `bnf/data` re-implements `toSequence`
+
+**Priority:** P5
+**Status:** open
+
+### Problem
+
+`fs/bnf/module.f.ts:104-107` exports the string → terminal-range-sequence
+conversion:
+
+```ts
+const mapOneEncode = map(oneEncode)
+export const toSequence = (s: string): readonly TerminalRange[] =>
+    toArray(mapOneEncode(stringToCodePointList(s)))
+```
+
+`fs/bnf/data/module.f.ts:102,104-107` redeclares `mapOneEncode` and inlines
+the identical expression:
+
+```ts
+const mapOneEncode = map(oneEncode)
+const data = (dr: DataRule): NewRule => {
+    switch (typeof dr) {
+        case 'string': {
+            return sequence(toArray(mapOneEncode(stringToCodePointList(dr))))
+```
+
+`bnf/data` already imports `oneEncode` from `../module.f.ts`, so it depends
+on that module anyway — it re-imports the primitive and re-derives the
+composite instead of importing the composite.
+
+### Proposal
+
+Import `toSequence` in `fs/bnf/data/module.f.ts`, replace the `'string'`
+case body with `sequence(toSequence(dr))`, and delete the local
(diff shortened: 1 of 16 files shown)


sergey-shandar:
@codex review

PR #1205, opened 2026-07-03
fs/text/utf8: reject overlong 3-/4-byte UTF-8 encodings

sergey-shandar:
## Summary

- `utf8ByteToCodePointOp` accepted any first-continuation byte after a 3-/4-byte lead, so **overlong** encodings (e.g. `E0 80 80` → U+0000, `F0 80 80 80` → U+0000) decoded successfully instead of erroring. This is a classic filter-bypass vector (overlong NUL, overlong `/`).
- Per Unicode Table 3-7 (*Well-Formed UTF-8 Byte Sequences*), a lead `E0` requires a first continuation `>= 0xA0`, and a lead `F0` requires `>= 0x90`. Bytes below that bound now route through the decoder's existing maximal-subpart error-recovery path (the same one already used for other invalid leads/continuations).
- Since the fix lives in the shared decoder (`fs/text/utf8`), it fixes all consumers with no duplicated logic: `fromVec` now rejects overlong blobs, and `fs/mime`'s `detectStream`/`detectVec` (which rides the same decoder) will correctly fall back to `application/octet-stream` for such blobs.
- Scoped intentionally to overlongs only — the raw decoder still doesn't reject surrogates or code points above `U+10FFFF` (that's handled downstream by `fromVec`'s `isValidCodePoint` pass), matching the issue's stated scope.

Resolves the `open`, `P2` issue in `fs/text/utf8/todo/utf8-decoder-accepts-overlong.md` (deleted per repo convention, with the design decision captured in the updated JSDoc on `utf8ByteToCodePointOp`).

## Test plan

- [x] Added `proof.f.ts` cases for overlong rejection (3- and 4-byte, including the widest overlong continuation byte) and the valid boundary cases (`E0 A0 80` → U+0800, `F0 90 80 80` → U+10000)
- [x] Added `fromVec` cases confirming overlong blobs decode to `null`
- [x] Full suite (`node ./fs/fjs/module.ts t`): 2060/2060 passing (was 2052 before the new cases)

---
_Generated by [Claude Code](https://claude.ai/code/session_01SHkjFxzj3bKPZhpFktbP3Q)_


sergey-shandar:
@codex review

PR #1207, opened 2026-07-03
Add writeFromStream proof test for exclusive file creation

sergey-shandar:
## Summary
Added a proof test for the `writeFromStream` function to verify error handling when attempting to write to a file that already exists with the `createExclusive` flag.

## Key Changes
- Imported `writeFromStream` and `IoResult` type from the node effects module
- Imported `empty` from the list module for creating empty stream data
- Added `writeFromStream.createExclusiveFails` proof test that verifies:
  - When `writeFromStream` is called on an existing file, the `createExclusive` operation fails with an EEXIST error
  - The error is properly propagated without attempting to write bytes
  - The original file content remains unchanged after the failed operation

## Implementation Details
The test uses the virtual file system to set up a pre-existing file (`hello` with content `0x2A`), then attempts to write to it using `writeFromStream` with an empty stream. It validates that the operation fails at the file creation stage and that the original file is untouched.

https://claude.ai/code/session_01Ma84J7cU9rVCvoYqPKAf6B


sergey-shandar:
@codex review

PR #1214, opened 2026-07-05
effects/node: hoist randomInt's call-invariant constants out of the per-call closure

sergey-shandar:
## Summary
- Hoists `crypto.randomInt` (via `const { randomInt } = crypto`) and the `2 ** 32` bound (precomputed once as `randomMax = Number(1n << 32n)`) out of the `randomInt` effect handler in `fs/effects/node/module.ts`, so neither is recomputed on every call.
- No behavior change — same bound, same source of randomness.

## Test plan
- [x] `node ./fs/fjs/module.ts t` — 2073/2073 passing


sergey-shandar:
@codex review

PR #1220, opened 2026-07-05
Fix BigInt Sub sign bug by unifying Add/Sub signed dispatch

sergey-shandar:
Implements and deletes `nanvm-lib/todo/bigint-add-sub-mirror.md` (P2, open).

## Bug

`BigInt`'s `Sub` returned the wrong sign whenever both operands had the same sign and `|lhs| < |rhs|`:

- `3n - 5n` produced `2n` instead of `-2n`
- `(-3n) - (-5n)` produced `-2n` instead of `2n`

The `Ordering::Less` arm in `sub.rs` was mirror-copied from `add.rs`, where `(rhs_sign, …)` is correct because that arm only runs when the signs *differ*. In `sub` the signs are *equal* in that branch, so `rhs_sign == lhs_sign` and the required sign flip was lost. The signed dispatch had no unit tests on either side, which is how the copy-paste survived.

## Fix

Fix the bug *by* removing the duplication that enabled it, per the issue's design:

- **`Sign::flip`** (`nanvm-lib/src/sign.rs`) — used by `Sub` and by `neg.rs`, which previously open-coded the same match.
- **`BigInt::add_signed(self, rhs, rhs_sign)`** (`nanvm-lib/src/vm/bigint/mod.rs`) — the single signed dispatch, parameterized by the effective right-hand sign (`a - b = a + (-b)`). `Add` passes `rhs`'s sign as-is; `Sub` passes it flipped. With the effective sign threaded in, the `Less` arm's `(rhs_sign, …)` is correct for both operators from one source.
- The sign is passed as a parameter instead of writing `self + (-rhs)` because `Neg` rebuilds the container (copies every word) — an allocation per subtraction in a VM hot path. Noted in `add_signed`'s doc comment so it isn't "simplified" back later.

## Tests

Adds the previously missing signed-dispatch unit tests:

- `sub.rs`: both signs of the buggy `Less` arm (`3 - 5 == -2`, `(-3) - (-5) == 2`), the `Greater` arm, cancellation to zero, mixed signs, multi-word borrow (`2^64 - 1`).
- `add.rs`: equal signs, differing signs with each magnitude larger, cancellation to zero, multi-word carry.
- `mul.rs`: all four sign combinations plus zero operands.

Also updates `nanvm-lib/todo/sign-algebra.md`, which linked to the deleted issue file.

`cargo test` (110 unit + integration + doc tests), `cargo clippy`, and `cargo fmt -- --check` are clean.

https://claude.ai/code/session_01JHZCnUkQWrvFQop4NxqKnT

---
_Generated by [Claude Code](https://claude.ai/code/session_01JHZCnUkQWrvFQop4NxqKnT)_


sergey-shandar:
@codex review

sasha-gil (approved):

PR #1221, opened 2026-07-05
todo: clarify decode-rejects-max-size-input.md design constraint

sergey-shandar:
No implementation here — reverts the earlier fix in this PR and rewrites `fs/base64/todo/decode-rejects-max-size-input.md` instead.

## Why the earlier fix was reverted

The original approach widened `bit_vec`'s `tryListToVec` cap (and `baseN`'s `stringToVec`) so `base64.decode`/`cbase32.cBase32ToVec` could accept a raw, pre-trim length above `maxLength`, then re-validate the trimmed length afterward.

That's unsafe: `maxLength` isn't an arbitrary logical limit, it exists specifically to stay under **Bun's actual `bigint` size constraint** (see its doc comment in `fs/types/bit_vec/module.f.ts`). `Vec` is a signed `bigint`, so the *intermediate* value `unpackListToVec` builds before trimming is itself subject to that same runtime ceiling — widening the cap just moves the failure from "`decode` returns `null`" to "constructing an over-`maxLength` `bigint` throws/hangs on Bun," for the same boundary-sized input. That's worse, not better, and the previous PR's fix and tests never actually exercised this on Bun.

## What changed instead

Rewrote the todo's Proposal/Tasks to:
- Explicitly rule out cap-widening as a valid fix, with the reasoning above.
- Point at approaches that never materialize a `bigint` over `maxLength` at any intermediate step — e.g. masking the last chunk's padding bits before combining, or (a bigger, separate design question) having `decode`/`stringToVec` return multiple `Vec` chunks instead of one.
- Require a Bun-exercised proof case for whichever approach is chosen, not just Node.

No code changes; `fs/base64/module.f.ts`, `fs/base_n/module.f.ts`, `fs/cbase32/module.f.ts`, `fs/types/bit_vec/module.f.ts`, and the CAS/base64/cbase32 proof files are all back to their pre-PR state.


sergey-shandar:
@codex review

PR #1223, opened 2026-07-06
todo: require one shared state machine for JSON recognizer and parser

sergey-shandar:
Strengthens `fs/json/todo/streaming-recognizer.md` with an explicit architecture requirement: **one grammar → one state machine → two builders**.

- The recognizer and `parse` must be two builder instantiations of a single state machine driven by one grammar description — sharing as much code as possible.
- `parse` itself must be refactored onto the shared, builder-parameterized machine, so no parallel copy of the transitions survives.
- A standalone recognizer that re-derives the grammar is explicitly out of scope, even if a test corpus shows it equivalent; the equivalence proof becomes a regression check on the shared machine, not the mechanism holding two implementations together.
- Adds a task item for the `parse` refactor.

Context: a previous attempt (reverted in this branch's history, PR #1222) implemented the recognizer as a standalone grammar, which is exactly the divergence this requirement forbids.

https://claude.ai/code/session_01Xbd8LtU7tKcQZvEWDDrCaj

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xbd8LtU7tKcQZvEWDDrCaj)_


sergey-shandar:
@codex review

PR #1234, opened 2026-07-07
Use assertEq for fsc's def-handler proof case

sergey-shandar:
## Summary

- Follow-up to #1230: the `defHandler` proof case added there used a hand-written `if (result !== unexpectedSymbol) { throw result }`, which itself is a branch whose false side is never exercised (the test is expected to pass) — so it left `fs/fsc/module.f.ts` short of 100% branch coverage instead of closing the gap.
- Replaces it with `assertEq(def(undefined), unexpectedSymbol)` from `fs/asserts/module.f.ts`, whose own branches are already fully covered elsewhere, so the call site adds no new uncovered branch.
- Documents the pattern in `AGENTS.md` (new bullet under Testing) so future proof code reaches for `assert`/`assertEq` instead of inline `if/throw`.

## Coverage (`npm run cov`)

| | before (inline `if/throw`, merged in #1230) | after (`assertEq`) |
|---|---|---|
| `fs/fsc/module.f.ts` branch % | 96.77% | **100.00%** |
| all files branch % | 96.96% | **96.99%** (back to pre-#1230 baseline) |
| all files funcs % | 99.68% | 99.68% (unchanged — still fixed) |

Test pass/fail counts unchanged (2058/2086 pass, same 28 pre-existing unrelated failures).

## Test plan

- [x] `node --test fs/emergent_testing/all.test.ts` — `fs/fsc/module.f.ts).proof.defHandler()` passes
- [x] `npm run cov` before/after comparison (above)

---
_Generated by [Claude Code](https://claude.ai/code/session_0159rLuKRQpUWq9MjQ3Xk2dp)_


sergey-shandar:
@codex review

PR #1237, opened 2026-07-08
Fix base64.decode rejecting exactly-maxLength-sized input

sergey-shandar:
Implements the base64 half of `fs/base64/todo/decode-rejects-max-size-input.md` (P2, open).

## Bug

`base64.decode` built the entire raw body into one `Vec` via `stringToVec` *before* trimming the zero-padding bits `encode` added to reach a 6-bit boundary. `tryListToVec`'s `maxLength` cap is checked against that raw, pre-trim length, so a payload whose *decoded* length is exactly `maxLength` — but whose raw pre-trim length is 2 or 4 bits over — was rejected with `null`, even though it's legitimate, spec-sized input.

Concretely: `encode(vec(maxLength)(...))` produces a 174,763-char string + one `=`. Decoding that string's 174,763-char body is 1,048,578 raw bits — 2 over `maxLength` — so the old code bailed before it ever got to trim those 2 padding bits back down to `maxLength`. This contradicted the documented capability that a maximally-sized (128 KiB) base64 payload can be read back, and made CAS's `cas_add` return an error for base64 content at exactly the documented size limit.

`maxLength` isn't an arbitrary logical cap — its doc comment says it tracks Bun's actual `bigint` size constraint — so the fix can't just loosen `tryListToVec`'s cap; every intermediate `bigint` built while decoding has to stay within `maxLength`, not just the final result.

## Fix

RFC 4648 padding never spans more than one base64 character, so the padding bits always live entirely inside the *last* character's 6-bit chunk. `decode` (`fs/base64/module.f.ts`) now:

1. Decodes every character but the last through `stringToVec` as before (`head`).
2. Decodes the last character on its own (always safe — a single 6-bit chunk).
3. Trims that last chunk's padding bits via `popFront`, checks the combined length against `maxLength` *after* trimming, then concatenates `head` with the trimmed tail via `msb.concat`.

No intermediate `Vec` built during decoding is ever wider than the final, post-trim result.

## Tests

- `decodeAtMaxLengthSucceeds` (`fs/base64/proof.f.ts`) — companion to the existing `encodeAtMaxLengthSucceeds`, round-trips a `maxLength`-sized vector through `encode`/`decode`.
- `addBase64AtLimitIsError` → `addBase64AtLimitSucceeds` (`fs/cas/mcp/proof.f.ts`) — flips the CAS-level proof case that previously documented the failing behavior, per the issue's own "Related" note.
- Full existing base64/cbase32/CAS test suite still passes (all other padding branches — `padChars` 0/1/2 — were already exercised by pre-existing tests and are unchanged in behavior).

## Scope

`fs/cbase32/module.f.ts`'s `cBase32ToVec` has a related but distinct exposure (sentinel-bit padding instead of zero-padding, and a bit-by-bit trim scan instead of a fixed last-character split) — the original issue flagged it as needing its own design pass. I narrowed the todo file down to just that remaining, lower-priority item (cbase32 only ever decodes short, fixed-length CAS hashes in practice, not `maxLength`-scale payloads) rather than expanding this PR's scope to cover it.

`npx tsc` and the full `node ./fs/fjs/module.ts t` suite (2079 tests) are clean.

https://claude.ai/code/session_01NoiJ2sPk2EoLCFFw2SBCxw

---
_Generated by [Claude Code](https://claude.ai/code/session_01NoiJ2sPk2EoLCFFw2SBCxw)_

### DIFF
--- a/fs/base64/module.f.ts
+++ b/fs/base64/module.f.ts
@@ -3,7 +3,7 @@
  *
  * @module
  */
-import { msb, type Vec, length, vec, empty } from "../types/bit_vec/module.f.ts"
+import { msb, type Vec, length, vec, maxLength } from "../types/bit_vec/module.f.ts"
 import type { Nullable } from "../types/nullable/module.f.ts"
 import { baseN } from "../base_n/module.f.ts"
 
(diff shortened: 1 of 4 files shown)


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1246, opened 2026-07-09
Replace evolution tag with mimeType in revision format

sergey-shandar:
## Summary

Redesign the revision format's type discriminant from a mutable URL-based `evolution` tag to a stable, RFC 6838 vendor-tree `mimeType` tag. This aligns the CAS revision format with MCP resource conventions and establishes a foundation for serving CAS objects as MCP resources.

## Key Changes

- **Rename `evolution` to `mimeType`**: Replace the URL-based format tag (`https://functionalscript.com/fs/cas/evo/README.md`) with a stable media type (`application/vnd.functionalscript.revision+json`). The new approach uses RFC 6838 vendor tree naming with RFC 6839 `+json` suffix, eliminating DNS anchoring and mutable pointers while remaining self-documenting and schema-validated.

- **Decouple versioning from the tag**: The `mimeType` field is now deliberately version-free. If format versioning is needed in the future, it becomes a separate optional field rather than a change to the tag itself.

- **Establish MCP alignment**: The field name `mimeType` matches the MCP resource contents convention, enabling CAS objects to be exposed as MCP resources with consistent terminology (`mimeType`, `uri`, `text`/`blob`).

- **Add server-side validation rules**: Document that MCP servers must never blindly echo the stored `mimeType` value. Instead, they surface it only when it matches an allowlist of known `application/vnd.functionalscript.*+json` types and the blob validates against that type's schema; all other cases fall through to the existing MIME detector.

- **Create companion spec**: Add `fs/cas/mcp/todo/cas-get-mcp-resource-response.md` to align the `cas_get` tool response with MCP resource shape, renaming fields (`mime_type` → `mimeType`, `type`/`content` → `text`/`blob`, `url` → `uri`) for consistency across tool and resource views.

## Notable Details

- The format spec location moves from the tag value to the schema/docs (in `fs/cas/evo/README.md`), simplifying the tag to a pure type identifier.
- The design preserves the deduplication principle: no registry required, no mutable pointers, stable identity across versions.
- New task added to teach the MCP server to surface stored `mimeType` values with proper validation and allowlisting.

https://claude.ai/code/session_01BJLud3DtwG3tWAnseCF8A5

### DIFF
--- a/fs/cas/mcp/todo/cas-get-mcp-resource-response.md
+++ b/fs/cas/mcp/todo/cas-get-mcp-resource-response.md
@@ -1,0 +1,46 @@
+# `cas_get` should satisfy the MCP resource response shape
+
+- **Status:** open
+
+## Problem
+
+We plan to expose CAS objects as MCP resources (see [remote-url.md](remote-url.md),
+point 3). An MCP `resources/read` result carries contents with the protocol-defined
+fields `uri`, `mimeType`, and `text` or `blob` — those names are fixed by the MCP
+schema and cannot be changed on our side.
+
+Today the `cas_get` tool returns a custom JSON payload
+`{ length, mime_type, type[, url][, content] }`. The same server would describe the
+same blob with two vocabularies: `mime_type` in the tool response versus `mimeType`
+in the resource response, `url` versus `uri`, `content` + `type: 'text' | 'base64'`
+versus `text` / `blob`.
+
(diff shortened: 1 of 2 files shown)


sergey-shandar:
@codex review

sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1248, opened 2026-07-09
Align cas_get response fields with MCP resource-contents shape

sergey-shandar:
## Summary

Standardize the `cas_get` tool response to mirror the MCP `resources/read` response shape, enabling consistent vocabulary between the tool view and future resource view of blobs. This change aligns field names and structure with MCP conventions while maintaining backward compatibility through extra fields.

## Key Changes

- **Field renames for MCP alignment:**
  - `mime_type` → `mimeType` (matches MCP resource-contents convention)
  - `url` → `uri` (matches MCP resource URI terminology)
  - `content` → `text` (for text payloads) / `blob` (for base64 payloads)

- **Response structure:**
  - Metadata-only responses now return `{ length, mimeType, type[, uri] }`
  - Content-inclusive responses return `{ length, mimeType, type[, uri][, text | blob] }`
  - The `type` field (`'text'` or `'base64'`) remains as a discriminator for which payload field is populated, even in metadata-only responses
  - `length` and `type` are extra fields alongside the MCP resource-contents shape (MCP allows extra fields)

- **Implementation details:**
  - Updated `Meta` type definition to use new field names
  - Modified `cas_get` handler to destructure `mime_type` from detector and map to `mimeType`
  - Changed inline payload encoding: `text` field for UTF-8 content, `blob` field for base64 content
  - Updated error messages to reference `uri` instead of `url`
  - Updated tool description to clarify the new field semantics and payload encoding

- **Documentation and tests:**
  - Updated `README.md` with new field names and examples
  - Updated `proof.f.ts` test assertions to verify new field names and structure
  - Updated `CasGetResult` type in proof tests
  - Clarified in documentation that `uri` points to the resource URI under which the blob will be readable via `resources/read`

## Notable Details

The change maintains the single classifier approach — the `fs/mime` detector state machine produces the same three-way verdict (`type: 'text'` or `'base64'`, with corresponding `mimeType`) on both metadata-only and content-inclusive paths, ensuring no divergent rule copies. The `type` field serves as the discriminator for agents to decide whether to fetch content without token cost for the bytes themselves.

https://claude.ai/code/session_012tACb2tZbN9DMk3SD211gD


sergey-shandar:
@codex review

PR #1264, opened 2026-07-10
Simplify dialect detection to semantic validation instead of byte-prefix matching

sergey-shandar:
## Summary

This change fundamentally shifts the dialect detection strategy from byte-level prefix matching to semantic validation. Instead of relying on entry-order assumptions and fixed byte signatures, detection now parses the full blob and validates it against the dialect's schema, making the detection more robust and format-agnostic.

## Key Changes

- **JSON dialect detection**: Removed the requirement that `dialect` be the first key. Detection now parses the entire JSON object and validates it semantically against the dialect schema, regardless of key order or whitespace.

- **CBOR dialect detection**: Replaced the byte-signature approach (parsing map headers and matching fixed key bytes) with full CBOR decoding and schema validation within the 128 KiB size-bounded path. No assumptions about entry order or byte-level patterns.

- **Canonical encoding simplification**: Removed the "dialect-first" deviation from RFC 8949 §4.2 deterministic ordering. FS-produced CBOR now uses pure RFC 8949 §4.2 core deterministic encoding (definite lengths, keys in bytewise order) without special entry-order rules, since detection no longer depends on prefix matching.

- **Encoding rules clarification**: Updated producer encoding rules to use standard RFC 8949 §4.2 deterministic encoding. The optional tag-55799 wrapper is still accepted on input (transparently unwrapped) but not produced, preventing canonical-form forking.

- **Detection contract**: Established that detection is semantic, not syntactic — any JSON or CBOR that satisfies a dialect's schema is recognized as that dialect, with no byte-level shortcuts or serialization assumptions.

- Size-bounded validation remains: both JSON and CBOR dialect detection are limited to the 128 KiB inline-content cap for buffering/parsing. Larger blobs fall back to streaming detection.

- The `dialect` key name is preserved as the vocabulary for both embedded tags and out-of-band fields, avoiding collisions with common response envelope keys like `mimeType` or `contentType`.

- Schema validation is the gate for dialect recognition — invalid blobs fall through to the existing detector rather than forcing a generic type.

- The change applies consistently across JSON, CBOR, and the revision format specification, establishing a unified detection model.

https://claude.ai/code/session_01SkojZYeji2fYugfaEnYKwy


sergey-shandar:
@codex review

sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1270, opened 2026-07-12
Refresh and expand deduplication todos for 2026-07 codebase state

sergey-shandar:
## Summary

This PR refreshes and expands the deduplication-focused todo documentation to reflect the current codebase state as of 2026-07. Several existing issues are updated with new context, and 13 new deduplication opportunities are documented across the fs/ and nanvm-lib/ modules.

### Updated Issues
- **66N-cas-get-response-shape**: Refreshed to describe the current `cas_get` handler structure (rewritten around `detectStream`/`detectDialect`). The original three-phase handler no longer exists; the duplication that remains is smaller and involves verdict→`Meta` mapping and buffered-refine arms. Priority downgraded from P3 to P4.
- **primitive-coercion-dispatch**: Extended to document the wider duplication scope across `vm/number_coercion.rs` and `vm/string_coercion.rs` — the whole "coerce `Any` → T" skeleton is repeated, not just the inner 5-arm match. The narrower `Unpacked::dispatch` approach is kept as a fallback option.

### New Deduplication Todos

**fs/types/rtti/**
- **or-handler-shared**: Share the `or` visitor handler between `validate` and `parse` (P4). Both implement identical "try each variant, return first ok, else error" logic in different styles.

**fs/js/tokenizer/**
- **tokenizer-finish-number-shared**: One resolver for "is this number token complete?" (P4). The incomplete-number classification is implemented twice with identical outcomes.
- **tokenizer-continue-string-comment**: Name the string/comment accumulate-in-place emit shape (P4). The `[empty, { …, value: appendChar(state.value)(input) }]` pattern is copy-pasted four times.

**fs/media/json/**
- **parse-text-pipeline**: A Result-returning string→value parse entry point (P3). Three consumers independently compose the same `jsonParse(jsonTokenize(stringToList(text)))` pipeline.

**fs/text/**
- **vec-to-code-point-pipeline**: Single owner for the UTF-8 `Vec` → string decode pipeline (P4). The core chain is spelled independently in two modules (unchecked and checked forms).
- **decoder-oob-sentinel**: Out-of-range guard emits a magic `0xffffffff` instead of `errorMask` (P4). UTF-16 decoder uses a hardcoded literal where UTF-8 uses the shared constant.

**fs/cas/**
- **shard-dir-helper**: Single owner for the shard directory layout (P4). `publish` re-derives the parent directory by parsing the relative path back apart instead of using the layout definition.
- **stage-lease-path**: Fresh staging-path combinator for the lease-deadline rule (P4). The "read clock, add leaseDelta, build path" invariant is spelled in two places.

**fs/types/**
- **split-last-nullable-map**: `splitLast` hand-rolls the null-projection its sibling routes through `nullable.map` (P5). Inconsistent null-handling between two similar accessors.
- **map-from-match**: Derive `map` from `match` so the null-guard lives once (P5). The `value === null ? … : f(value)` projection is duplicated in nullable's own combinators.
- **defined-values-from-entries**: `definedValues` re-implements `definedEntries`' filter (P5). The "drop undefined-valued entries" predicate is written twice with different idioms.

**nanvm-lib/**
- **bigint-operator-test-scaffolding**: Shared test helpers for the bigint operator modules (P4). Every operator file redeclares identical unit-test scaffolding and copy-pastes the same TODO comment.
- **bigint-normalized-check-reuse**: `shl` restates the normalization invariant by hand (P5). The normalized-and-nonempty postcondition duplicates logic already centralized in `assert_slice_normalized`.

**fs/text/sgr/**
- **inplace-

https://claude.ai/code/session_01PAZxq5Q5jtfTEjGkDyYGsj


sergey-shandar:
@codex review

sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1277, opened 2026-07-14
cas: add opt-in hash verification on read

sergey-shandar:
## Summary

Implements [fs/cas/todo/66g-cas-get-verify-option.md](https://github.com/functionalscript/functionalscript/blob/main/fs/cas/todo/66g-cas-get-verify-option.md) (now deleted per the repo's todo convention — done issues are removed once their design is captured in a README).

CAS reads never rehash the bytes they return, so a blob corrupted, truncated, or misnamed on disk (bit-rot, an unscrubbed copy-files sync) is silently handed back under a hash it no longer matches. This adds an opt-in read-time integrity check:

- **`verifyHash`** (`fs/cas/module.f.ts`) — rehashes a stream against a requested hash without buffering it into memory, so it scales past `maxLength`.
- **`cas get --verify`** (CLI) — verifies before writing the output file; on mismatch, nothing is written and the command exits 1 with a distinct "hash mismatch" message.
- **`cas_get { verify: true }`** (MCP) — same check, returned as a distinct `isError` ("hash mismatch") rather than conflated with "no such hash".

Both are opt-in since verification costs a full extra read.

## Design decisions (open points from the issue)

- **Default:** off, for both CLI and MCP — a rehash doubles the read cost, which isn't warranted for the common "store is already trusted" case.
- **MCP exposure:** yes, `cas_get` gets a `verify` boolean argument, mirroring the CLI flag.
- **Error shape on mismatch:** a distinct "hash mismatch" error, separate from "no such hash" (an absent hash and a present-but-corrupted hash are different failures worth telling apart).

## Test plan

- [x] `npx tsc` — clean
- [x] `node ./fs/module.ts t` — 2176/2176 passing, including new unit tests (`fs/cas/proof.f.ts`), CLI tests (`fs/cas/cli/proof.f.ts`), and MCP tests (`fs/cas/mcp/proof.f.ts`) covering both the clean-blob and corrupted-blob paths

---
_Generated by [Claude Code](https://claude.ai/code/session_013aiqKbxzmAQ7vMv2xoRY6h)_


sergey-shandar:
Closing unmerged: the CLI flag is hand-parsed (`args.includes('--verify')`) because the CLI eDSL has no option concept, and the CLI/MCP wiring duplicates logic that should live in a shared command layer. Superseded by design todos (CLI options eDSL, CAS command architecture) — the verify feature is re-filed as blocked on those. The streaming `verifyHash` prototype in this PR's commit remains a useful reference for the eventual implementation.

---

PR #1290, opened 2026-07-17
Replace if-throw with assert() in test/proof files

sergey-shandar:
## Summary
- Converts ~1400 `if (cond) { throw x }` / `if (cond) throw x` patterns to `assert(...)` calls across 77 `proof.f.ts`/`proof.ts` test files, following the new `assertNotNullish`/`assert` conventions added in #1289.
- Conditions were negated via double-negation (`assert(!(cond), x)`) for guaranteed correctness, then simplified to idiomatic form (`a !== b`, dropping a leading `!`, etc.) only when unambiguous (no top-level `&&`/`||`, single comparison operator); compound conditions stay wrapped in `!(...)`.
- Library (`module.f.ts`) files are untouched in this PR — scoped to test/proof files only.

## Test plan
- [x] `npx tsc` — clean, no errors
- [x] `node ./fs/module.ts t` — 2179/2179 tests pass (same count as before the change)


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1293, opened 2026-07-18
tokenizer-new: add path/line/column metadata, export tokenize

Trinidadec:
Compute per-code-point TokenMetadata (mirroring fs/js/tokenizer's tokenizeWithPositionOp) and thread it through the AST-flattening/token pipeline so each token carries its start position. Token becomes a 3-tuple [tag, startMetadata, codePoints]; the pipeline is concretely typed to TokenMetadata rather than generic, since passing still-generic function values through flatMap/filter hits TypeScript's unreliable higher-order generic inference.

Adds tokenize: (input: List<number>) => (path: string) =>
List<JsTokenWithMetadata>, matching fs/js/tokenizer's own shape. tokenizeString's string output is unaffected (verified against the full suite).

Closes the add-metadata.md todo (prerequisite for replace-old-tokenizer.md) and updates both docs accordingly, including a documented known limitation: no per-character error recovery yet, unlike the old tokenizer's inline error tokens.


sergey-shandar:
@codex review

sergey-shandar:
@codex review

sergey-shandar:
@codex review

sergey-shandar (approved):

PR #1295, opened 2026-07-18
Replace try/catch in proofs with structural throw property

sergey-shandar:
## Summary
- FunctionalScript has no `try`/`catch` and isn't planning to add it soon, so proof files should express throw-expectations structurally instead of imperatively catching.
- Converted the remaining `try`/`catch` throw-tests in `fs/asserts/proof.f.ts`, `fs/types/result/proof.f.ts`, `fs/text/ascii/proof.f.ts`, and `fs/cas/cli/proof.f.ts` to nest under a `throw` property key, which the runner in `fs/emergent_testing/module.f.ts` already treats as a structural throw marker (inverts the sandboxed result so a thrown error counts as a pass).
- Documented the rule in AGENTS.md: never use `try`/`catch` in `.f.ts` files, use the `throw` property key convention instead.

## Test plan
- [x] `npx tsc` passes
- [x] `fjs t` passes (2114/2114 tests), including the converted `throw.*` tests in the four touched proof files


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1303, opened 2026-07-21
Fix stack-overflow tokenization: stack-based descent matcher

sergey-shandar:
## Summary

Fully resolves `fs/djs/tokenizer/todo/stack-recursive-tokenization.md` (implemented and deleted per convention; the history is recorded in the CHANGELOG's Unreleased entry): valid input of a few KB threw `RangeError: Maximum call stack size exceeded` instead of tokenizing — both a file with many short tokens (`' '.repeat(5000)`) and a single long token (5 KB string literal, 20 KB block comment, ~2-3K-char identifier).

**The fix is in the matcher, not the grammar:** `fs/bnf/descent`'s `f` recursed natively once per grammar step, so right-recursive rules (how `repeat0Plus` encodes repetition) turned input length into JS call-stack depth. `f` is now an explicit-stack machine: two suspended-frame kinds (sequence: current item index + accumulated ASTs; variant: current branch index + best zero-consumption success) on an immutable cons-cell stack, with a loop that either starts the current rule invocation or feeds the pending result into the innermost frame. Semantics are unchanged — same AST, same success/failure/index results, including the subtle variant behavior where a zero-consumption success is remembered while later branches are still tried. Call-stack depth is now O(1) for any grammar shape and input length, for every `descentParser` consumer.

The tokenizer itself is unchanged in shape: `tokenizeString`/`tokenizeJs` still match the whole file as one `jsGrammar()` rule. (An interim per-token driver, `tokenizeToAsts`, was landed mid-PR and then removed again — with the stack-based matcher it was unnecessary, and its per-token `cpm.slice(offset)` made tokenization O(n²) where the whole-file match is linear: ~4× slower at 30,000 tokens and growing. The single-token grammar is factored into a local `buildToken()`, which `jsGrammar()` wraps with `repeat0Plus`.)

This supersedes the `repeat`-primitive proposal (`fs/bnf/todo/667-bnf-repeat-flatten.md`) *as a crash fix* — and covers shapes its conservative repeat-detection would skip, e.g. the tokenizer's hand-written terminated-list `multilineContent` rule. 667 stays open for its original data-shape/flat-AST motivation; its impact note is updated accordingly.

## Verified

| Input | Before | After |
|---|---|---|
| `' '.repeat(5000)` (reported repro) | RangeError | OK |
| 3,000 short id tokens | RangeError (from ~1,000–1,500) | OK |
| 5 KB string literal | RangeError | OK |
| 20 KB block comment (old pre-grammar tokenizer's level) | RangeError | OK |
| 10 KB identifier | RangeError | OK |
| 100 KB block comment; 30,000-token file; mid-file error position after 500 tokens | RangeError | OK (manual check) |

## Notes

- CHANGELOG updated (Unreleased) with the full story; the resolved todo file is deleted and every reference to it (code comments, sibling todos) now points at `fs/bnf/descent`, its `longInput` proof group, or the CHANGELOG entry.
- The explicit `DataRule` annotation on `rule` breaks a TS7022 control-flow inference cycle (`name` → `rule` → `task` assignments → `name`); commented in place.
- Merged with `main` (including #1298, which moved nullability analysis into `fs/bnf/data`); the stack machine uses the shared `emptyTagMap` directly.
- **The LL(1) backend has the same bug** — `fs/bnf/ll1`'s matcher recurses per grammar step and overflows at ~10,000 repetitions (confirmed repro). Nothing outside its own proofs consumes it yet, so it's recorded as `fs/bnf/ll1/todo/stack-recursive-matching.md` (P3) with this PR's descent rewrite as the template, rather than fixed here.

## Test plan

- [x] `node ./fs/module.ts t` — 2128/2128 passing
- [x] `fs/bnf/descent` `longInput` proof group: a 10,000-code-point `repeat0Plus` chain and 5,000 levels of bracket nesting — both verified to throw `RangeError` against the pre-fix matcher restored from git history, and to pass now
- [x] `fs/djs/tokenizer` `largeInputs` proof group covers both former crash shapes end-to-end: many short tokens (5,000 spaces, 3,000 ids, `tokenizeJs` entry point) and single long tokens (5 KB string, 20 KB block comment, 10 KB identifier)


sergey-shandar:
@codex review

PR #1309, opened 2026-07-22
Clarify token symbol encoding options

sergey-shandar:
## Summary

- reorganize token symbol encodings into intrinsic and registered families
- retain the existing restricted and base64-style packing designs under intrinsic encoding
- add registered hash assignment with collision validation
- clarify registered positional/index assignment
- compare stability, reversibility, collision behavior, and ordering tradeoffs
- update the open questions and implementation tasks accordingly

## Why

The previous flat list mixed two fundamentally different models: encodings derived from a string alone and encodings constructed from a complete finite alphabet. Making that distinction explicit clarifies why packing has a length limit, while registered hash and positional encodings support arbitrary-length names.

## Impact

Documentation only. No runtime behavior changes.

## Validation

Reviewed the branch diff against `main`; it contains one modified file: `fs/bnf/todo/token-symbol-encoding.md`.


sergey-shandar:
@codex review

PR #1310, opened 2026-07-22
rtti: single source of truth for Unknown/Primitive

sergey-shandar:
## Summary

Fixes the `unknown-primitive-source-of-truth` todo (P3) in `fs/types/rtti/todo/`.

The rtti value-domain types `Unknown`/`Primitive` were imported inconsistently across sibling rtti modules:

- `common` imported `Primitive, Unknown` from `djs`
- `validate` imported `Unknown` from `djs`
- `parse` imported `Unknown` from the rtti-local `rtti/ts`

So the same concept was sourced from two different modules for the same code, and rtti's runtime-validation core (`common`, `validate`) depended on `djs` — the exact dependency direction `rtti/ts`'s own doc comment says it wants to avoid (djs depends on rtti, not the reverse).

## Changes

- Point `common` and `validate` at the rtti-local `Unknown`/`Primitive` in `rtti/ts`, making it the single source of truth for the rtti value domain. `parse` was already correct and is left unchanged.
- Update the `rtti/ts` JSDoc to state the intended **subset** relationship `djs.Unknown ⊆ rtti.Unknown` rather than "currently equivalent", so a future reader does not re-introduce the djs import assuming the two types must stay identical (rtti may widen `Unknown` to admit e.g. functions independently of djs).
- Delete the resolved todo file (design rationale now captured in the `rtti/ts` JSDoc, per the repo's todo convention).

This is purely a re-pointing of type-only imports plus a doc update — no runtime behavior changes. Since `djs.Unknown ⊆ rtti.Unknown`, any `djs.Unknown` value still type-checks where rtti functions consume it. No import cycle is introduced (`common → ts → rtti root` never returns to `common`).

## Verification

- `npx tsc` — clean.
- `node ./fs/module.ts t` — all 2172 proofs pass.

https://claude.ai/code/session_01AmbUpLYa9QCZUDigywxh76

---
_Generated by [Claude Code](https://claude.ai/code/session_01AmbUpLYa9QCZUDigywxh76)_


sergey-shandar:
@codex review

PR #1354, opened 2026-07-24
Make Effect the raw value; move step out as external function

sergey-shandar:
## Summary

This is a **breaking change** that restructures the core effect system. An `Effect<O, T>` is now the **raw value** (a `Pure` thunk or `Do` node) instead of a `{ value, step }` wrapper object. Composition moves from a per-node method into an external `step(e, f)` primitive, mirroring the `compose`/`fn` pattern in `fjs/types/function`.

## Key Changes

- **Effect type redefinition**: `Effect<O, T>` is now `Pure<T> | Do<O, T>` (the raw value), not a wrapper object
  - `Pure<T>` = `() => T` (a thunk)
  - `Do<O, T>` = `[command, payload, continuation]` (a positional tuple represented as an object with numeric keys for variance)

- **Variance solution**: Introduced `Cont<out O, T>` (the continuation type with covariance annotation) to make the raw `Effect` union covariant in `O`, enabling the widening that the effect system relies on everywhere

- **External `step(e, f)` primitive**: Replaces `.step()` method calls
  - Data-first: `step(effect, f)` returns a raw `Effect`
  - Chains as `step(step(e, f), g)`
  - Thin wrapper over `decode`

- **Optional `eff()` wrapper**: For method-chaining sugar (mirrors `fn` over `compose`)
  - `eff(value)` wraps a raw effect into an `Eff<O, T>` with `.step()` and `.value` properties
  - `.step(f)` returns another `Eff` (chains as `.step(f).step(g)`)
  - `.value` unwraps back to raw `Effect` (mandatory at boundaries)

- **Single shape inspector**: `decode` remains the only function that inspects the `Pure` vs. `Do` layout; no second `typeof value === 'function'` checks allowed

- **Migration**: ~161 call sites across 25 files updated from `x.step(f)` to `step(x, f)` (or `eff(x).step(f).value` where method-chaining reads better)

## Implementation Details

- The `Do` tuple is declared as an object with numeric keys (`readonly 0`, `readonly 1`, `readonly 2`) rather than a tuple literal, because TypeScript only allows variance annotations (`out`/`in`) on object/function/mapped-type aliases, not tuple aliases
- The `out O` annotation on `Cont` and `Do` is sound: the command tag pins exactly which output the continuation receives, and every interpreter dispatches on the tag first, so the op-set can grow without mishandling continuations
- `all` and `both` retain their existing casts (pre-existing, orthogonal to this change)
- Module documentation updated to clarify that `Effect` is plain data with no methods, and composition is external

## Files Modified

- Core: `fjs/effects/module.f.ts` (type definitions, `pure`, `lazy`, `doFull`, new `step` and `eff`)
- Proofs: `fjs/effects/proof.f.ts`, `fjs/cas/proof.f.ts`, `fjs/mcp/proof.f.ts`, `fjs/effects/memory/proof.f.ts`, `fjs/effects/node/proof.f.ts`, `fjs/cas/mcp/proof.f.ts`
- Core modules: `fjs/cas/module.f.ts`, `fjs/emergent_testing/module.f.ts`, `fjs/mcp/stdio/module.f.ts`, `fjs/effects/node/module.f.ts`, `fjs/cas/evo/module.f.ts`, `fjs/cas/cli/module.f.ts`, `fjs/cas/mcp/module.f.ts`, `fjs/mcp/module.f.ts`, `fjs/dev/module.f.ts`, `fjs/djs/module.f.

https://claude.ai/code/session_017xufYaofdJKLBLz7DczzQn


sergey-shandar:
@codex review

PR #1355, opened 2026-07-24
effects/todo: propose extracting Eff into a self-contained eff/module.f.ts monad layer

sergey-shandar:
Design note only — adds `fjs/effects/todo/move-eff-out-of-core.md`. No code changes.

### What

Proposes extracting the `Eff` wrapper out of `fjs/effects/module.f.ts` into its own `fjs/effects/eff/module.f.ts`, **redesigned as a self-contained monad layer** rather than moved verbatim.

### Decided design

1. **`Eff.step` accepts a continuation returning `Eff`** (not a raw `Effect`) — making `Eff` a proper monad: `pure` = unit, `.step` = bind, `.value` = exit. A chain never leaves the wrapped world until the final `.value`.
2. **The module exports its own `Eff`-returning `pure`** (the unit), so leaf continuations (`t => pure(x)`, the common case) stay in the `Eff` world with no wrapping.

### Guiding principle

**A module picks one vocabulary — raw effects *or* the `Eff` wrapper — and does not mix them.** This is the rationale for the `t => Eff` bind signature over a polymorphic `t => Effect | Eff` (which was considered): accepting both would *encourage* mixing the two styles inside one module. The stricter signature makes a module's style visible and self-consistent, enforced by which module it imports `pure` from.

### Also covered

- **Name**: `fjs/effects/eff/module.f.ts` (matches the exported symbols; slots beside `list` / `memory` / `mock` / `node`).
- **Migration**: larger than an import repoint — continuations move into the `Eff` world, with `eff(...)` / `.value` surviving only at raw-world boundaries (runners, `Effect`-typed returns, raw core combinators like `foldStep` / `forEachStep` / `all`). `tsc`-verifiable.
- **`step` stays in core** as the foundation; only the fluent layer moves.
- **Clean move, not a re-export**; `compose`/`fn` counter-precedent acknowledged; `Eff`-flavored core combinators noted as possible follow-up.

Filed as P3, status open.

### DIFF
--- a/fjs/effects/todo/move-eff-out-of-core.md
+++ b/fjs/effects/todo/move-eff-out-of-core.md
@@ -47,6 +47,19 @@ This is why `.step` takes `t => Eff` rather than accepting *either* `Effect` or
 this convention discourages. The stricter `t => Eff` signature makes a module's
 choice of style visible and self-consistent.
 
+3. **`Eff` is optional — users may bring their own chain wrapper.** The effect
+   model's contract is the raw `Effect` type plus `step` (and `decode` /
+   `match`); `Eff` is *one* wrapper over that contract, not part of it. Keeping
+   it in a separate module makes that explicit: a downstream consumer who wants
+   different ergonomics — a different fluent API, a pipe operator, do-notation
+   codegen, or integration with their own effect utilities — writes their own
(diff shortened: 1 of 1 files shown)


sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1367, opened 2026-07-26
effects: replace Frame with a flat history tuple

sergey-shandar:
Replaces the `Frame<R, P>` record with a flat, newest-first history tuple, and splits the combinator that built it into a lift plus a step.

## Why

`Frame` chained by nesting: `Frame<C, Frame<B, A>>` for three links, so a value bound two links back read as `frame.param.param.result`. Each extra link cost another `param` hop, and the depth was load-bearing — nothing could flatten it.

The tuple form makes position an index instead of a traversal:

```ts
// before
step(b, ({ result: z, param: { result: y, param: x } }) => ...)

// after
step(b, ([z, y, x]) => ...)
```

## API

| before | after |
| --- | --- |
| `type Frame<R, P>` | `type History<O, H extends readonly unknown[]>` |
| `frameStep(e, f): Effect<O \| Q, Frame<R, P>>` | `historyStep(h, f): History<O \| Q, readonly[R, ...P]>` |
| — | `history(e): History<O, readonly[T]>` |

`historyStep` takes a history and returns one, so it composes with itself to any
depth; only the entry point needs `history`. That is the property `frameStep`
lacked — it both started and extended a chain, so it nested its predecessor's
result rather than flattening it, and link two could not be spelled the same way
as link three.

`f` now receives the history spread as arguments (`(result, param) => ...`)
rather than a single record.

## Ordering

Newest first. A position is distance back from the current link, not evaluation
order, so `([z, y, x]) => ...` binds `x` earliest. This matches the tuple `Eff`
already threaded internally, which is why `eff` gets simpler here: it called
`step(value, v => pure([v]))` inline, which is exactly `history(value)`.

## Call sites

`cas` and `emergent_testing` each used `frameStep` as a single link, so both
become `historyStep(history(e), f)`. No behavioural change.

## Docs

`Frame` carried the design rationale for the whole mechanism, so its doc was
rewritten onto `History` / `historyStep` / `history` rather than dropped —
including why a plain `step` chain cannot reach backwards, and the "collapse
into a record of named fields when the chain gets long" escape hatch, which
still applies once indices get hard to count.

Also fixes references left stale by the rename: four dead `{@link frameStep}`
links, the module header's headline example (still teaching
`({ param: x, result: y })`), the `both` → `h` parameter rename in `Eff`, and
the proof group name and comments that contradicted the code beneath them.

One non-obvious invariant is now commented rather than left to be rediscovered:
`Readonly<P>` on `historyStep`'s rest parameter is load-bearing. Inferring `P`
from a bare `(...p: P)` yields a mutable, labelled tuple (`[next: string]`),
which then rejects the `readonly` tuples every history is built from.

## Test plan

- `npm test` — 2192 passed, 0 failed; `tsc` clean.
- Proof group renamed `frameStep` → `historyStep`; `chain` now exercises
  `historyStep` twice, covering the self-composition that motivates the split.


sergey-shandar:
@codex review

sergey-shandar:
Addressed both review comments in 5378a87d.

**Update the repository guidance with the API rename** — fixed. `AGENTS.md` L101–111 prescribed `frameStep`/`Frame` in the mandatory effect-sequencing rule, including a `({ param: x, result: y })` example that would no longer type-check. The rule, its example, and the combinator rationale in the following paragraph now use `history`/`historyStep`.

**Add a breaking-change entry for removing Frame** — fixed. Added at the top of `## Unreleased` with the `**BREAKING CHANGES:**` prefix and the migration (`frameStep(e, f)` → `historyStep(history(e), f)`; `{ result, param }` → `[result, param]`; `f` now receives the history spread as arguments).

Also fixed a third stale reference not flagged in the review: `fjs/effects/eff/README.md:35` still described `frameStep` and a "`param` walk". Worth noting the point it makes got stronger — it calls the walk "a de Bruijn index into the values still live at that point", which was an analogy for a nested traversal but is literally true of a tuple position.

`npm test`: 2199 passed, 0 failed; `tsc` clean.

sergey-shandar:
@codex review

PR #1368, opened 2026-07-26
Remove decode/Decoded in favor of direct typeof checks

sergey-shandar:
## Summary

Remove the `decode` function and `Decoded` type from the effects module, replacing them with direct `typeof e === 'function'` checks at call sites. This simplifies the API by eliminating an intermediate abstraction that provided no narrowing benefit and required consumers to learn a second vocabulary for a shape they could already read.

## Key Changes

- **Removed `decode` and `Decoded`**: Deleted the helper function and type that wrapped the Pure/Do discriminant check
- **Inlined shape checks in `step` and `match`**: Both functions now directly check `typeof e === 'function'` and destructure the Do tuple using object pattern `const { 0: command, 1: payload, 2: continuation } = e`
- **Added `runPure` export**: New eliminator function returning `Option<T>` for callers expecting pure effects with no operations. Returns `[t]` for Pure, `[]` for Do. The tagged result prevents collapsing `null` values with the Do case
- **Updated module documentation**: Rewrote the invariant to name the three functions that read the layout (`step`, `match`, `runPure`) plus the layout proof, establishing an enumerable set of readers
- **Migrated all proof sites**: 
  - Consolidated duplicate `assertPure` helpers into one exported definition in `fjs/effects/proof.f.ts`
  - Updated `fjs/effects/proof.f.ts` to use `runPure` and added comprehensive proof coverage for both Pure and Do cases
  - Replaced `decode` calls in `fjs/cas/proof.f.ts` with `runPure` and added `casCommand` map for routing through `match`
  - Updated `fjs/effects/node/proof.f.ts` to use `match` with a `readHello` map instead of raw layout inspection
  - Updated `fjs/media/type/proof.f.ts` to use the exported `runPure` instead of a local copy
  - Removed duplicate `assertPure` from `fjs/effects/eff/proof.f.ts` in favor of importing the shared definition

## Implementation Details

- Object destructuring is required for Do inspection (not array destructuring) because Do is declared as an object with numeric keys, not a tuple, to work around TypeScript's `TS2637` restriction on `out` variance in tuples
- `runPure` remains generic over the operation set `O` rather than narrowing to `Effect<never, T>` because Effect is covariant in O, making the narrowed form incompatible with continuations that produce the wider type
- The layout proof in `fjs/effects/proof.f.ts` is the one place the Do representation is intentionally pinned; all other code routes through `step`, `match`, or `runPure`
- This is a **breaking change**: `decode` and `Decoded` are removed from the published `.d.ts`

https://claude.ai/code/session_01UoAuLq9sYSE5hJ74goJL44


sergey-shandar:
@codex review

PR #1373, opened 2026-07-27
effects: make foldStep and forEachStep step variants

sergey-shandar:
@
Both combinators now take the effect first, matching `step` and `historyStep`:

```ts
foldStep(items, init, f)
forEachStep(items, f)
```

with `items` an `Effect<O, List<T>>` rather than a bare `List<T>`.

## Why this order

A step variant is this modules `do` notation: reading the call top-to-bottom is reading the order the program executes in, so the effect comes first because it is what happens first. Currying `f` ahead of `items` would put the *body* of the loop above the thing it loops over — exactly the inversion `do` exists to remove.

That is also why the order deliberately differs from `fold` in `fjs/types/list`. `fold` is a data pipeline, curried `f`-first because the list is what gets threaded through, and nothing about it happens in time.

The same premise fixes the formatting rule, now documented in the module header: a step call that does not fit one line breaks after `(`, one argument per line. The argument list is a statement list, so each statement gets a line; packing the leading effect onto the `(` line hides which one runs first.

## Effect on call sites

Callers holding an effectful list no longer open a continuation just to name it:

- `fjs/cas/cli` — drops its explicit type arguments and the `Vec` import; the callback parameter now infers.
- `buildCache` (`fjs/cas/evo`) — loses an eta-expanded `eff(...).step(hashes => ...(hashes)).value` wrapper.
- The four sites that already hold a strict list lift it with `pure`, which costs a wrapper but no indentation.

`O` and `Q` are now separate type parameters: the operations needed to produce the list are rarely the ones the body performs, and the result unions them.

One annotation is load-bearing — `acc: Result<readonly Revision[], string>` at `fjs/cas/evo/module.f.ts:235`. With `init` before `f`, TypeScript fixes `S` from `init` and narrows it to `Ok<...>`, which kills the `error` branch. Two alternative argument orders were tried and rejected: `(items)(init)(f)` breaks `S` inference the same way at more sites, and `(items)(f)(init)` typechecks but keeps the currying this PR is removing.

## Follow-up

`fjs/effects/todo/effect-list-fold.md` proposes moving both combinators into `fjs/effects/list` and retyping them over the streaming `EffectList<O, T>`, so folds stop materializing the list. It notes that `fjs/types/list` is imported into `fjs/effects/module.f.ts` *only* by these two functions, and that the change would largely subsume `fold-stream-combinator` — cross-linked there, not rescoped.

## Testing

- `npx tsc` — clean
- `node ./fjs/module.ts t` — 2206 passed, 0 failed

@

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -135,3 +135,4 @@ test-results/
 index.html
 
 .cas/
+.claude
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ history.
 
 ## Unreleased
 
+- `fjs/effects`: **BREAKING CHANGES:** make `foldStep` and `forEachStep` step
+  variants — `foldStep(items, init, f)` / `forEachStep(items, f)`, with `items`
+  an `Effect<O, List<T>>`, so the effect leads as in `step` and `historyStep`
 - `fjs/types/list`: add `tryFold` and the `Accumulator<I, T, R>` type, the
   `try*`/`Nullable` sibling of `fold`; `bit_vec`'s early-exit fold driver moves
   onto it [#1370](https://github.com/functionalscript/functionalscript/pull/1370)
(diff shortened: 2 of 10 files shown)


sergey-shandar:
@codex review

PR #1374, opened 2026-07-28
effects: add `mapStep` and `Eff.map`, the effect functor map

sergey-shandar:
First PR of [`fjs/effects/todo/map-step-combinator.md`](https://github.com/functionalscript/functionalscript/blob/main/fjs/effects/todo/map-step-combinator.md).

## Problem

`fjs/effects/module.f.ts` shipped `pure` (return) and `step` (bind), plus `historyStep`, `foldStep`, `forEachStep` and the `okStep` adapter — but not the one combinator every monad also has: **`map`**, "run the effect, then apply a pure function to its result". So every call site re-derived it as `step(e, x => pure(f(x)))` — 45 times, in 15 modules.

Beyond the repetition, that spelling misreports the shape of the chain. A `step` whose continuation returns `pure` is not a link in a sequence of effects — it is where the sequence ends. `AGENTS.md` asks that a sequence read top-to-bottom with one name per link; a trailing pure projection is not a link at all, yet it costs a `step(` and a `pure(` and, when it needs a value from an earlier link, tempts exactly the nesting the rule forbids.

## Changes

**`mapStep`** in `fjs/effects/module.f.ts`, a step variant (effect first, like `step` and `historyStep`):

```ts
export const mapStep = <O extends Operation, T, R>(
    e: Effect<O, T>,
    f: (t: T) => R
): Effect<O, R> =>
    step(e, t => pure(f(t)))
```

It does **not** widen the operation set — `Effect<O, R>`, not `Effect<O | Q, R>` — because a pure projection issues no commands. That is a small typing gain over `step` at every site, not just a shorter spelling.

**`Eff.map`** in `fjs/effects/eff/module.f.ts`, for the roughly one third of sites already in the fluent world:

```ts
readonly map: <R>(f: (...tp: readonly[T, ...P]) => R) => Eff<O, R, readonly[T, ...P]>
```

The history grows exactly as it does for `.step`, which is forced rather than preferred: `.map(f)` has to be observationally identical to `.step((...tp) => pure(f(...tp)))`, because that is the rewrite the follow-up PRs perform at every fluent site. Preserving `P` unchanged would mean the mechanical conversion silently changes what *later* callbacks receive — a behavioural change disguised as a readability one. `create` therefore names its result `self` and defines `.map` as that exact `.step` call, rather than as a second copy of `.step`'s body that has to be re-checked against it.

**First consumers, converted in the same PR** so neither export is speculative:

| site | before | after |
| --- | --- | --- |
| `readUtf8File` (`fjs/effects/node`) | `step(readFile(path), r => pure(mapOk(utf8ToString)(r)))` | `mapStep(readFile(path), mapOk(utf8ToString))` |
| `awaitIfPromise` (`fjs/effects/node`) | `step(awaitPromise(p), ([x]) => pure(x))` | `mapStep(awaitPromise(p), ([x]) => x)` |
| `errorExit` (`fjs/effects/node`) | `step(error(s), () => pure(1))` | `mapStep(error(s), () => 1)` |
| `decodeRevisionBlob` (`fjs/cas/evo`) | `.step(([tag, value]) => pure(…))` | `.map(([tag, value]) => …)` |
| `Eff`'s history projection | `step(x1, ([t]) => pure(t))` | `mapStep(x1, ([t]) => t)` |

## Design notes

- **One name, not two.** A constant variant (`constStep(e, v)`) was considered for the 24 constant-projection sites and rejected: `mapStep(e, () => v)` already reads clearly, and it keeps `v`'s evaluation inside the continuation where `step` puts it. The rationale is recorded in `mapStep`'s JSDoc so it isn't re-litigated.
- **Arity hazard.** Converting a fluent `.step(y => pure(f(y)))` to `.map(f)` is only safe when `f` is genuinely unary — the lambda pins arity at one, while passing `f` point-free exposes it to the positional history. This is `['1','2','3'].map(parseInt)` with a longer tuple. Documented on `Eff`; `mapStep` is not exposed to it, since it applies `f` to exactly one argument.

## Scope

`AGENTS.md` asks one improvement per PR, and the issue is explicit that the conversion must not land as one 14-module diff. The ~41 remaining sites stay in the issue, regrouped module by module; the issue file is updated to record that both APIs now exist and that only the mechanical conversion remains.

## Testing

- `npx tsc` — clean.
- `npm test` (`fjs t`) — 2207 passed, 0 failed, including six new proofs: `mapStep.pure` / `.constant` / `.over_do`, and `map.chain` / `.grows_history` / `.over_do`. `grows_history` pins the history contract with a `.map(...).step(...)` chain asserting the second callback still sees the pre-`map` value.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VKSVts1xeMsqyTCM2REMCT)_

### DIFF
--- a/fjs/effects/proof.f.ts
+++ b/fjs/effects/proof.f.ts
@@ -117,7 +117,7 @@ export const proof = {
             // Chains as step(step(e, f), g), raw effect in and out.
             assertPure(step(step(pure(3), v => pure(v + 1)), v => pure(v * 2)), 8)
         },
-        over_do: () => {
+        overDo: () => {
             // Stepping a Do node preserves the command and threads the result
             // through the rebuilt continuation.
             const e = step(do_<AddOp>('add')(2, 3), r => pure(r * 10))
@@ -135,7 +135,7 @@ export const proof = {
             // The `() => v` shape a `constStep` would have covered.
             assertPure(mapStep(pure('ignored'), () => 1), 1)
         },
-        over_do: () => {
+        overDo: () => {
             // A projection over a `Do` node keeps the command intact and is
             // applied to the command's output when the continuation resumes.
             const r = next(mapStep(do_<AddOp>('add')(2, 3), v => v * 10))
@@ -152,7 +152,7 @@ export const proof = {
             assertEq(param, 3)
             assertEq(result, 6)
         },
-        over_do: () => {
+        overDo: () => {
             // The captured value survives a command boundary: the history is
             // rebuilt inside the continuation rather than lost when `e` is a Do.
             const c = next(historyStep(history(do_<AddOp>('add')(2, 3)), r => pure(r * 10)))
@@ -174,7 +174,7 @@ export const proof = {
                 step(b, ([z, y, x]) => pure(`${x}${y}${z}`)),
                 '123')
         },
-        f_receives_whole_history: () => {
+        fReceivesWholeHistory: () => {
             // `f` is handed the whole history spread as arguments, not just the
             // most recent value - which is what lets a later link reach back.
             const a = historyStep(history(pure(1)), x => pure(x + 1))
--- a/fjs/effects/proof.f.ts
+++ b/fjs/effects/proof.f.ts
@@ -1,4 +1,4 @@
-import { step, do_, foldStep, forEachStep, match, okStep, history, pure, runPure, type Effect, type Operation, historyStep } from './module.f.ts'
+import { step, do_, foldStep, forEachStep, mapStep, match, okStep, history, pure, runPure, type Effect, type Operation, historyStep } from './module.f.ts'
 import { error, ok } from '../types/result/module.f.ts'
 import { assert, assertEq } from '../asserts/module.f.ts'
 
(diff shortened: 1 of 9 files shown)


fjs/effects/proof.f.ts
```diff
@@ -127,6 +127,23 @@ export const proof = {
...
     },
+    mapStep: {
+        pure: () => {
+            assertPure(mapStep(pure(3), v => v + 1), 4)
+        },
+        constant: () => {
+            // The `() => v` shape a `constStep` would have covered.
+            assertPure(mapStep(pure('ignored'), () => 1), 1)
+        },
+        over_do: () => {
```
sergey-shandar: Use camelCase `overDo`.

sergey-shandar:
@codex review

fjs/effects/proof.f.ts
sergey-shandar: Done in 5e3ba29. Renamed the three keys this PR added — `mapStep.overDo`, `map.growsHistory`, `map.overDo` — and also the pre-existing snake_case siblings in the same two files (`step.over_do`, `historyStep.over_do`, `historyStep.f_receives_whole_history`, `eff/proof.f.ts`'s `over_do`), so neither file is left mixing conventions. Names only; no assertion or coverage change. `npx tsc` clean, 2208 tests passing.

48 snake_case keys remain elsewhere — mostly `fjs/sul/id` (18) and `fjs/sul/level/hash` (11), plus `nanvm-lib/tests`, `fjs/sul`, `fjs/types/bit_vec`, `fjs/types/prime_field`, `fjs/fsc`. I left those out to keep this PR to one change and filed `todo/camel-case-proof-keys.md` for the sweep, which also proposes writing the rule into `AGENTS.md` so it's checkable in review rather than inferred from neighbours. Happy to fold that in here instead if you'd rather have it in one go.

I've left the PR in draft — I saw you converted it and didn't want to flip it back.

---

PR #1391, opened 2026-07-30
Remove npm-check-updates from `npm run update`, pin @playwright/test exactly

sergey-shandar:
@
`npm run update` ran `npx npm-check-updates -u` first, which rewrote pinned
versions in `package.json` independently of `fjs/ci/config/module.f.ts`. Those
two have to agree — `fjs/ci/config/module.f.ts` is the source the CI workflow
and the Dockerfile are generated from, so a drifting `package.json` silently
puts the local install and CI on different Playwright versions.

- Drop `npx npm-check-updates -u` from `update` and run `ci-update` first, so
  the generated config leads and the installs follow it.
- Pin `@playwright/test` to `=1.62.0` so npm cannot resolve it away from the
  version `fjs/ci/config/module.f.ts` declares.
- Note the coupling next to `playwright` in `fjs/ci/config/module.f.ts`.

Version bumps become a deliberate edit to `fjs/ci/config/module.f.ts` (plus the
matching `package.json` entry) instead of a side effect of running `update`.

@

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ history.
 
 ## Unreleased
 
+- `package.json`: pin `@playwright/test` exactly (`=1.62.0`) and exclude it from
+  `npm run update`'s `npm-check-updates` pass, so the pinned test runner version
+  survives routine dependency updates
+  [#1391](https://github.com/functionalscript/functionalscript/pull/1391)
+
 ## 0.39.0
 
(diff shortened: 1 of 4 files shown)


sergey-shandar:
@codex review

sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1415, opened 2026-08-03
Filter archived subjects from `evo_list`

sergey-shandar:
Implements `fjs/cas/evo/todo/evo-list-archived-filtering.md` (P2), deleted here as the fix ships.

## What changed

`Evo.list` takes an optional status filter, forwarded unchanged from the `evo_list` MCP tool:

```ts
list(archived?: true): Effect<MemOp, readonly Subject[]>
```

- omitted → the **active** subjects: at least one current head that is not archived;
- `archived: true` → the **archived** subjects: at least one current head, all of them archived;
- a subject with **no** current heads is in neither result.

**BREAKING CHANGE:** `list()` used to return every subject with at least one stored revision; it now returns only the active ones. There is deliberately no all-subjects mode — nothing needs one yet, and only this direction is reversible: adding a third value to the filter later is a compatible extension, removing one would break callers.

## Design notes

**Why the status comes from the heads.** `archived` is a flag on a *revision*, but "is this thing still being worked on" is a question about a *subject*, and the heads are exactly the revisions anything new would be built on. Concurrent heads can disagree, and both rules resolve that the same way: a single unarchived head keeps the whole subject active. That also makes archiving reversible in the ordinary way — a new unarchived revision on top of an archived head demotes it out of the head set and the subject is active again, with no separate "unarchive" operation and no rewriting of stored revisions.

**Why `archived` is cached per revision hash, not per subject.** `SubjectState` gains a third hash set alongside `hashes`/`parents`, and the flag is applied only to the heads `headsOf` derives at read time. A per-subject flag maintained during the fold would be wrong for the same reason a running head set is (documented in the module doc and `README.md#head-resolution`): CAS hash order is not ancestry order, so folding a child before its parent would keep revising the answer. Deriving from the complete graph at read time has no ordering dependency.

**The no-heads case.** A hash-consistent store cannot produce a subject with no heads — that would need a cycle among content hashes — but nothing in this layer verifies that a blob hashes to the key it sits under, so a corrupt or hand-crafted store can. Such a subject gets "neither status" rather than an arbitrary one; the proof reproduces it with the existing `fixedCas` helper, which a real `fileCas` cannot.

## Proofs

Five new tests, all four `fjs/cas/evo` ones covering a distinct rule of the classification:

- `listPartitionsSubjectsByHeadArchivedFlag` — an archived root and an unarchived one land in opposite results;
- `listTreatsDisagreeingHeadsAsActive` — two concurrent heads, one archived, the subject stays active;
- `listIgnoresArchivedRevisionsThatAreNoLongerHeads` — an unarchived child reactivates a subject archived at its root;
- `listExcludesSubjectWithNoCurrentHeads` — neither result;
- `fjs/mcp/evo`: `evoListForwardsTheArchivedFilter` — the tool argument is a pass-through.

`fjs/cas/evo/module.f.ts` branch coverage goes 99.19% → 99.25% (the same single pre-existing uncovered branch, now out of a larger total), and `fjs/mcp/evo/module.f.ts` stays at 100%.

## Checks

- `npx tsc` — clean
- `node ./fjs/module.ts t` — 2282 pass, 0 fail
- `npm run cov` — as above
- `npm run ci-update` — no regenerated-workflow changes
- No Rust touched

Docs updated: `fjs/cas/evo/README.md` (new **Subject status** section + API block), `fjs/mcp/evo/README.md`, `fjs/mcp/README.md`, and both module docs. The `fjs/mcp` module-doc table also had a stale `evo_list` result column ("subjects, one per line") corrected to the JSON array it actually returns.

https://claude.ai/code/session_011gbjiceavPurLpa6WyikKH

---
_Generated by [Claude Code](https://claude.ai/code/session_011gbjiceavPurLpa6WyikKH)_


sergey-shandar:
@codex review

PR #1421, opened 2026-08-04
Fix prototype pollution vulnerability in match handler lookup

sergey-shandar:
## Summary

This PR fixes a security vulnerability in the `match` function where untrusted command strings could dispatch to inherited `Object.prototype` members instead of actual operation handlers. The fix uses safe own-property lookup via the existing `at` utility instead of direct property access.

## Changes

- **Updated `fjs/effects/module.f.ts`**: Modified the `match` function to use `at()` for safe handler lookup instead of `map[command]`, which could resolve to inherited properties like `constructor` or `toString`. Added comprehensive documentation explaining the vulnerability and the fix.

- **Added safety assertion**: The handler lookup now asserts that a valid handler exists, treating unknown/inherited commands as invariant violations that should throw rather than silently dispatching to prototype members.

- **Added regression tests in `fjs/effects/proof.f.ts`**: 
  - New `AnyOp` type to simulate externally-sourced commands (where `command` is any string)
  - `ownCommand` test verifies legitimate own properties still dispatch correctly
  - `constructorCommand` and `toStringCommand` tests verify that commands naming `Object.prototype` members throw instead of dispatching

- **Removed obsolete issue document**: Deleted `match-prototype-property-access.md` as the issue is now resolved.

## Implementation Details

The fix leverages the existing `at` utility from `fjs/types/object/module.f.ts`, which uses `Object.getOwnPropertyDescriptor` to safely read only own properties. This prevents the type confusion where a value promised by the type system as `(...payload) => R` could actually be an arbitrary inherited function invoked with attacker-controlled arguments.

The explicit type argument `at(command)<OperationMap<O, R>[O[0]]>(map)` is required to help TypeScript's type inference work correctly with the mapped type structure of `OperationMap`.

https://claude.ai/code/session_01QZDEEU8uExMFr68ViJdWzB


sergey-shandar:
@codex review

o2alexanderfedin (approved):
Verified the mechanism, the proofs, and the reasoning. Approving.

**The fix is right.** `at` reads through `getOwnPropertyDescriptor`, which only ever sees own properties, so an inherited name yields `null` instead of a callable. Confirmed behaviourally against the current implementation:

```
plain read  map["constructor"] -> function   (callable — would be invoked with the payload)
at("constructor")              -> null
at("toString")                 -> null
at("add")                      -> function   (own key still dispatches)
```

**The proof design is what makes this trustworthy.** `ownCommand` asserting that `'add'` still dispatches to `42` sits alongside the two `throw` cases, so the test cannot pass with a map that dispatches nothing — the failure mode a refusal-only test would miss. Using `AnyOp` with a `string` command to express "a `Do` node decoded from external input" avoids an `as` cast, which keeps the proof honest about what it is modelling.

**Throwing rather than widening `MatchResult` is the right call**, and the docstring's argument holds: every `O1 extends O` the signature admits has its command in `map`, so a `null` handler means the node's `command` was never the `O1[0]` it claimed to be. A runner cannot resume a command it has no handler for, so a recovery branch would have nothing to do. Widening the result type would push a case onto every caller that no type-correct one could observe.

**One note for downstream, not a change request.** `fjs-dev/finance` is building a restricted runner whose whole purpose is bounding what a stored, agent-authored program may do — this PR is what stops that bound being escapable by naming an inherited property, which matters a lot there. But since `match` throws rather than returning a refusal, that project still needs its own guard at the dispatch boundary to turn an out-of-vocabulary command into a reported `operation not permitted: <command>` rather than an exception. That is already how its spec plans it, so the division of responsibility looks correct: fjs closes the escape, the caller shapes the error.

CI is green across all 19 checks — ubuntu/macos/windows on intel and arm, wasm, deno, bun, node22/24/26, nix-flakes, CodeQL.

Deleting `fjs/effects/todo/match-prototype-property-access.md` in the same PR is the right move: the report existed to track the gap, and the gap is closed here.

PR #1422, opened 2026-08-04
Recognize `.f.mjs` in proof discovery and coverage tooling

sergey-shandar:
Implements the tooling half of
[`fjs/emergent_testing/todo/f-mjs-test-and-coverage.md`](https://github.com/functionalscript/functionalscript/blob/main/fjs/emergent_testing/todo/f-mjs-test-and-coverage.md)
(P1).

Authored `.f.mjs` is the extension a FunctionalScript module moves to once the
current parser and compiler accept it. Before the first rename, the test and
coverage tooling has to treat it like `.f.ts` — otherwise migrating a module
silently drops its proofs and its coverage rows while CI stays green.

## Changes

- **Proof discovery** — `shouldLoad` (`fjs/dev/module.f.ts`) matches `.f.mjs`,
  so a migrated `module.f.mjs` and its internal `proof` export stay in
  bulk-load discovery. Ordinary `.mjs` files remain opt-in through the existing
  `proof.mjs` convention. Covered by an extended `shouldLoad` proof and a new
  virtual-filesystem `allFiles` proof.
- **Node coverage** — `npm run cov` now passes `--test-coverage-include` for
  both `**/module.f.ts` and `**/module.f.mjs`.
- **Deno coverage** — `deno task cov` and the generated CI step use
  `.*module\.f\.(ts|mjs)`.
- **CI generator** — `fjs/ci/deno/module.f.ts` exports `coverageInclude` as the
  single source of that filter. New `fjs/ci/deno/proof.f.ts` fails if `.f.mjs`
  support is removed from it, and `fjs/ci/proof.f.ts` now asserts against the
  same constant instead of a duplicated literal. `.github/workflows/ci.yml` is
  regenerated from the generator.
- **Policy** — `AGENTS.md` §2/§3.1/§3.2 and the matching summary in
  `CONTRIBUTING.md` apply mandatory 100% proof coverage to both authored
  extensions and document the mixed `module.f.mjs` / `proof.f.ts` layout used
  during incremental migration, plus the dependency-closure rule a
  `proof.f.mjs` must satisfy.

## What is not in this PR, and why

The issue also asked for two runtime fixtures. Both turned out to depend on
[`fjs/ci/todo/f-mjs-package-support.md`](https://github.com/functionalscript/functionalscript/blob/main/fjs/ci/todo/f-mjs-package-support.md):

1. A `proof.f.ts` importing a `module.f.mjs` fails `npx tsc` with `TS7016`
   while `allowJs` is off. Turning `allowJs`/`checkJs` on makes `npx tsc` pass,
   but `npm run prepack` (`tsc --NoEmit false`) then fails with `TS5055`
   because authored `.mjs` becomes both an input and a JavaScript emit target —
   exactly the repeatable-emission work owned by the package-support issue.
2. An internal `proof` inside a `.f.mjs` module has no assert helper it may
   import: `AGENTS.md` §3.3 requires `assert`/`assertEq` from
   `fjs/asserts/module.f.ts`, but the dependency-closure rule forbids authored
   `.f.mjs` from importing a relative `.f.ts` module.

Rather than duplicate the package-support work here, the issue file is rewritten
to cover just the remaining fixtures, marked `blocked` on package support, with
both obstacles and a proposed resolution for the assert-helper question recorded.
`f-mjs-package-support.md` and `fjs/fsc/README.md` are updated so the ordering
between the two prerequisites is stated in one direction everywhere.

## Verification

- `npx tsc` — clean.
- `npm test` — 2290 pass, 0 fail.
- `npm run ci-update` — regenerated workflow matches the committed one.
- `node --test --experimental-test-coverage` with both include flags — accepted,
  report unchanged for `.f.ts` modules.

Deno is not installed in this environment, so `deno task cov` was not executed;
the filter change is covered by the generator proof instead.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JimQDf1MdP9oPskJj7nfrn)_

### DIFF
--- a/fjs/dev/module.f.ts
+++ b/fjs/dev/module.f.ts
@@ -31,24 +31,28 @@ export type ModuleMap = StringMap<string, Module>
 /**
  * Returns `true` if the file should be loaded for proof discovery.
  *
- * All FunctionalScript modules are safe to bulk-load by construction — they
- * have no import side effects. That covers the two authored extensions,
- * `.f.ts` and `.f.mjs`, plus the `.f.js` generated from `.f.ts`. For vanilla
- * TS/JS the load gate stays opt-in by filename: any file ending in
- * `proof.ts`, `proof.js`, `proof.mts`, or `proof.mjs` is included.
+ * Two symmetrical rules, each covering all four TS/JS module extensions:
  *
- * `.f.mjs` is matched by extension like any other FunctionalScript module, so
- * a migrated `module.f.mjs` keeps its internal `proof` export in discovery.
- * A `proof.f.mjs` matches the same `.f.mjs` rule, not the vanilla
- * `proof.mjs` one.
+ * - **FunctionalScript modules** — anything ending in `.f.ts`, `.f.mts`,
+ *   `.f.js`, or `.f.mjs`. They are safe to bulk-load by construction, since
+ *   they have no import side effects, so the whole module is loaded and its
+ *   internal `proof` export (if any) is discovered. `.f.ts` and `.f.mjs` are
+ *   the authored extensions; `.f.js` is generated from `.f.ts`.
+ * - **Impure JavaScript/TypeScript proofs** — anything ending in `proof.ts`,
+ *   `proof.mts`, `proof.js`, or `proof.mjs`. Outside FunctionalScript a module
+ *   may have import side effects, so the load gate stays opt-in by filename.
+ *
+ * A `proof.f.mts` / `proof.f.mjs` matches the FunctionalScript rule by its
+ * `.f.` infix, not the vanilla proof rule.
  *
  * Whether a loaded module actually _contains_ a proof is determined at
  * runtime by checking for an exported `proof` property.
  */
 export const shouldLoad = (s: string): boolean =>
-    s.endsWith('.f.ts')    || s.endsWith('.f.js')    || s.endsWith('.f.mjs') ||
-    s.endsWith('proof.ts') || s.endsWith('proof.js') ||
-    s.endsWith('proof.mts')|| s.endsWith('proof.mjs')
+    s.endsWith('.f.ts')    || s.endsWith('.f.mts')   ||
+    s.endsWith('.f.js')    || s.endsWith('.f.mjs')   ||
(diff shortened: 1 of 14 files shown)


fjs/dev/module.f.ts
```diff
@@ -31,16 +31,22 @@ export type ModuleMap = StringMap<string, Module>
...
+ * a migrated `module.f.mjs` keeps its internal `proof` export in discovery.
+ * A `proof.f.mjs` matches the same `.f.mjs` rule, not the vanilla
+ * `proof.mjs` one.
  *
  * Whether a loaded module actually _contains_ a proof is determined at
  * runtime by checking for an exported `proof` property.
  */
 export const shouldLoad = (s: string): boolean =>
-    s.endsWith('.f.ts')    || s.endsWith('.f.js')    ||
+    s.endsWith('.f.ts')    || s.endsWith('.f.js')    || s.endsWith('.f.mjs') ||
```
sergey-shandar: Make it symmetrical, accept:
- All FunctionalScript files: ends with:`.f.ts`, `.f.mts`, `.f.js` and `.f.mjs`,
- All unpure JavaScript/TypeScript proofs: ends with: `proof.ts`, `proof.mts`, `proof.js` and `proof.mjs`.

sergey-shandar: Done in 60a534f — `shouldLoad` is now the two symmetrical families you described, and the doc comment states them as such:

- FunctionalScript modules, loaded whatever the file name: `.f.ts`, `.f.mts`, `.f.js`, `.f.mjs`
- impure JavaScript/TypeScript, loaded only under the `proof` name: `proof.ts`, `proof.mts`, `proof.js`, `proof.mjs`

The proof now covers both families exhaustively, including the negative side for every impure extension (`module.ts` / `module.mts` / `module.js` / `module.mjs` are all rejected), and the `allFiles` discovery proof gained a `.f.mts` entry.

I left the coverage filters (`npm run cov`, `deno task cov`, the generated Deno CI step) on `module.f.ts` + `module.f.mjs`, since those are the two authored implementation extensions in `fjs/fsc/README.md`. Say the word if you'd like `.f.mts` there too.

---

sergey-shandar:
@codex review

PR #1428, opened 2026-08-04
Make dialect registry parameterizable in media.detect

sergey-shandar:
## Summary

This PR resolves the hardcoded dialect limitation in `fjs/media`'s `detect` function by making the set of recognized dialects a parameter rather than an import. The detector now accepts a list of dialect entries at call time, enabling any module to register custom dialects following the `vnd.fjs.<name>` convention.

## Key Changes

- **Parameterized dialect detection**: `detect(dialects)(bytes)` replaces `detect(bytes)`, allowing callers to specify which dialects to recognize
- **New `DialectEntry` type**: Represents a registered dialect with a `dialect` name and a `match` predicate over parsed JSON values
- **New `dialectEntry()` function**: Builds dialect entries from an rtti schema and optional refinement predicate (`extraValidate`), with the dialect name read from the schema itself
- **Removed hardcoded revision import**: `fjs/media/module.f.ts` no longer imports or knows about any specific dialect
- **Added `revisionDialect` export**: `fjs/media/revision/module.f.ts` now exports its own `DialectEntry`, combining `revisionSchema` with `checkReferences` as validation
- **Single parse, multiple validations**: JSON text is parsed once and matched against all registered entries, avoiding redundant parsing
- **Integrated `parse` function**: `fjs/media/json/module.f.ts` now exports a total `parse(text) => Result<Unknown, string>` built on the module's own tokenizer/parser pipeline; native `JSON.parse` moved to `parseNative`

## Implementation Details

- The dialect name is extracted from the schema's `dialect` member at registration time via `assert`, ensuring no mismatch between the schema and the reported name
- Media types are derived mechanically as `application/${dialect}+json` rather than supplied separately
- `extraValidate` runs after structural validation on the decoded type, allowing dialects to enforce constraints rtti cannot express (e.g., cbase32-decodable strings)
- The `Struct` constraint on `dialectEntry`'s type parameter ensures all schema members are valid rtti `Type`s, catching mistakes like function members at compile time
- No grammar check or allowlist on dialect names—the name comes from programmer-written schemas, not untrusted blob content
- Detection remains semantic (key order/whitespace independent) and size-bounded to already-buffered `Vec`s

## Breaking Changes

- `detect(bytes)` → `detect(dialects)(bytes)` — all call sites updated
- `fjs/media/json` `parse` signature changed from `(text: string) => Unknown` (throwing) to `(text: string) => Result<Unknown, string>` (total); old behavior available as `parseNative`

All in-repo importers updated in this PR: `fjs/mcp/cas/module.f.ts`, `fjs/media/proof.f.ts`, `fjs/dev/package_json/module.f.ts`, `fjs/protocol/mcp/stdio/module.f.ts`, and `fjs/ci/proof.f.ts`.

https://claude.ai/code/session_01JvvayENXU54Se7PTZ6UbrT


sergey-shandar:
@codex review

PR #1429, opened 2026-08-04
Tests: add JavaScript multiplication coverage

sasha-gil:
## Summary
- Add native-JS multiplication proofs matching NaNVM's Rust cases.
- Update the test coverage map for multiplication and strict number/string equality.

## Testing
- npm test
- cargo test --manifest-path nanvm-lib/Cargo.toml


o2alexanderfedin (commented):
Reviewed the diff and verified the README reclassification against `test.rs` rather than taking it on trust. CI is green across bun/deno/macos-arm/CodeQL. The `assert`-instead-of-`throw` cleanup in `ois`/`nanRes` is a clear improvement — those were the two remaining hand-rolled throws in the file.

**The `string.number` reclassification is correct.** I checked: `old_eq` really does carry the cross-type assertion at `nanvm-lib/tests/test.rs:282` — `assert_ne!(number_p0, string0.clone())`, i.e. `0` vs `"0"`. So flipping that row to `[x]` and dropping it from the gap list is accurate, not optimistic.

One consequence of that is worth writing down, and three coverage gaps I think are worth closing before this lands.

### The new `[x]` depends on a test the same document proposes deleting

`string.number`'s Rust coverage now rests entirely on `old_eq`. Two rows further down, the same README says:

> \| `old_eq` in `test.rs` \| Redundant — overlaps with `nullish_eq`/`bool_eq`/`number_eq`/etc.; consider removing \|

If someone takes that advice, the `string.number` cross-type assertion disappears with it and the table still reads `[x]` — a coverage hole that the coverage map itself asserts does not exist. `old_eq` is *mostly* redundant, which is exactly why the one assertion that isn't will not be noticed on the way out.

Cheapest fix: move `assert_ne!(number_p0, string0)` into `string_eq` (or `number_eq`) in this PR, then the row is backed by a test nobody is proposing to remove. Failing that, the Notes cell should say `old_eq` **and** the removal row should note the dependency.

### `mul` never produces `-0`, and `ois` is `Object.is`

`multiplyEq` compares with `Object.is`, which distinguishes `-0` from `+0` — and the README specifically calls out `number_eq` as covering `±0`. But no case in the new `mul` table produces a negative zero. Verified:

```
0 * -1   = -0     Object.is(0 * -1, 0) === false
-0 * -0  = 0
```

For a NaN-boxing VM, sign-of-zero on multiply is precisely the kind of thing that breaks, and the harness is already equipped to catch it. Suggest adding to `number`:

```ts
multiplyEq(0)(-1)(-0)
multiplyEq(-0)(-0)(0)
```

Note the first would *pass* against a VM that returned `+0` if the comparison were `===`; it only bites because `ois` is `Object.is`. That is the test earning its keep.

### No `±Infinity` cases

`Infinity * 0 === NaN` is the classic multiply edge case and is absent, as are the sign-propagation ones. Verified: `Infinity * 0 = NaN`, `Infinity * 2 = Infinity`, `-1 * Infinity = -Infinity`. The `eq` table already covers `±Inf`, so `mul` is the outlier here.

### The one throwing case of `*` is untested

`1n * 1` is a `TypeError` ("Cannot mix BigInt and other types") — the only way `*` fails rather than returning `NaN`. The `bigint` group covers bigint×bigint only, so the mixed case is missed. Given the proof suite has a structural convention for throw-tests, this seems worth pinning: a VM that silently coerced instead of throwing would pass everything here.

### Minor

`const multiply = (a: any) => (b: any)` uses two bare `any`s. Coercion testing genuinely needs them — you cannot write `*` over `unknown` — so I am not suggesting a change, but a one-line comment saying *why* would stop it reading as an oversight, particularly with #1430 currently arguing that `any` at a call site is a defect worth a breaking change.

`multiplyEq` asserting both `a*b` and `b*a` is a nice property to pin. Worth knowing it stops being safe the moment a case involves an object with a side-effecting `valueOf`, since JS evaluates the left operand first — none of the current cases do, so this is a note for whoever extends the table rather than a problem here.

sergey-shandar:
@codex review

sergey-shandar (approved):

PR #1430, opened 2026-08-04
Add issue: remove JSON.parse native export

sergey-shandar:
## Summary

This PR adds a tracking issue documenting the plan to remove the `parseNative` export from the JSON module and migrate all call sites to use the module's own `parse` function instead.

## Changes

- Added `fjs/media/json/todo/remove-json-parse.md` — a comprehensive issue documenting:
  - **Problem statement**: The repository maintains two JSON parsing APIs (`parse` returning `Result` and `parseNative` wrapping `JSON.parse`), but all 21 call sites use only the native version
  - **Rationale for removal**: Four key reasons:
    1. Violates the principle of avoiding duplicate APIs for the same concept
    2. `JSON.parse` throws exceptions, which are panics in FunctionalScript with no recovery path
    3. Returns `any`, forcing unsafe `as` casts at every call site (17 casts in one file alone)
    4. Depends on host `JSON` global rather than FunctionalScript's own tokenizer/parser pipeline
  - **Detailed migration strategy**: Replacement pattern using `unwrap(parse(text))` and schema-based validation instead of hand-written types and casts
  - **Concrete examples**: Shows how to replace unsafe `as` casts with proper RTTI schemas
  - **Call site inventory**: Maps all 21 uses across 7 files with specific line numbers
  - **Implementation tasks**: Checklist for migration, including two decision points requiring design input
  - **Performance considerations**: Notes to monitor large-input proofs during migration

## Notes

This is a tracking/planning document, not a code change. It establishes the scope and approach for a breaking change to a published export that will be executed in follow-up PRs.

https://claude.ai/code/session_01UiCc5ATPUadPqfrgusVEwV


o2alexanderfedin (commented):
Reviewed against `origin/main` rather than the description — the four reasons, the inventory, and the proposed replacement.

**The change is right and I'd merge it.** Reasons 1, 3 and 4 each independently justify removing `parseNative`, and reason 4 is the strongest: nothing in `fjs/` defines `JSON` and NaNVM has no such object, so a `.f.ts` reaching for it depends on the host rather than on the language. Verified there is no production call site — the only non-proof file that mentions `JSON.parse` is `fjs/types/nibble_set/module.f.ts:12`, and that is a docstring about what round-trips through the host globals, not a call. The inventory's "nothing in production code parses JSON natively" holds.

One substantive point on the rationale, and one on the counts.

### Reason 2 does not survive the proposed replacement

> **It throws.** A `throw` in FunctionalScript is a panic (§3.4): nothing can catch it, so a malformed input reaching `parseNative` kills the program with no recoverable failure value.

The mechanical replacement is `unwrap(parse(text))`, and `unwrap` is:

```ts
export const unwrap = <T, E>([kind, v]: Result<T, E>): T => {
    if (kind === 'error') { throw v }
    return v
}
```

So at 24 of the sites the migration relocates the panic from `JSON.parse` to `unwrap` rather than removing it. By §3.4's own logic the replacement is equally a panic and equally uncatchable.

The doc half-anticipates this — "making the failure a `Result` the runner reports rather than an unlabeled host exception" — but that reads as though the failure stops being a throw, and it doesn't. What actually improves is the *quality* of the panic: `parse` returns `Result<Unknown, string>`, so `unwrap` throws a described string instead of a `SyntaxError` about position 47.

That is a real improvement and worth stating; it just isn't "no longer a panic." I'd rewrite reason 2 as *"it throws a host exception nobody labelled, where the replacement throws the parser's own message"* — otherwise reason 2 invites a reader to check whether the migration delivers what it promises, find it doesn't, and discount the other three reasons that are airtight.

Worth noting alongside it: `unwrap` throws the raw error **value**, so what lands in a catch is a bare string, not an `Error` — `e.message` is `undefined` and `e instanceof Error` is `false`. That is the same ergonomics trap `assert` has, and it bit us in `finance` when a handler branched on `instanceof Error` and missed every refusal. If the migration lands, the test runner's reporting path is worth a look, since a bare-string throw prints differently from an `Error`.

### The inventory undercounts

The prose says 21 call sites and the table sums to 23; grepping `origin/main` for non-comment lines gives 26 occurrences across the same 7 files, ~24 of them consumers. `fjs/media/json/proof.f.ts` is listed as 1 and has 3.

Only worth fixing because the value of this document *is* the inventory — the task checklist is derived from it, and "migrate `fjs/media/json/proof.f.ts`" reads as smaller than it is.

### Smaller notes

- The two flagged decision points (`Event` schema vs. asserting on text; dropping the tokenizer round-trip entirely) are the right two to flag, and the `fjs/djs/tokenizer/proof.f.ts` suggestion — count tokens from the list the test already has, no JSON step — is strictly better than a migration.
- Scoping `JSON.stringify` out is correct: it is total, and `stringify` already sits beside it.
- No impact on `fjs-dev/finance` — we never adopted `parseNative`. This did surface one of ours, though: `fjs/server/module.f.js` has a test helper decoding JSON-RPC envelopes into `any[]`, which is the same shape as the 17 casts in `fjs/mcp/proof.f.ts`. We'll follow whatever schema pattern lands here.

o2alexanderfedin (approved):
Approving. Docs-only, the direction is right, and the widened scope (covering `JSON.stringify`, not just `parseNative`) made it a substantially better document — reason 4 in particular now lands harder, since `fjs/media/json/serializer/module.f.ts:9` means the FunctionalScript serializer bottoms out in the host.

**Correcting my earlier comment on this PR:** I claimed the inventory undercounts. It doesn't — I was wrong. Recounting against `main` and excluding docstrings and the import line:

| | in the doc | actual |
| --- | --- | --- |
| `JSON.parse` / `parseNative` | 23 (table) | **23** — 17 `fjs/mcp/proof.f.ts` + 1 each `mcp/evo`, `emergent_testing`, `djs/tokenizer`, `ci/proof`, `media/json/proof`, + the export itself |
| `JSON.stringify` | 117 | 118 |

The `JSON.parse` table is exact. My "26 occurrences" counted raw grep hits including `fjs/types/nibble_set/module.f.ts:12` (a docstring, which the doc correctly excludes) and `fjs/media/json/proof.f.ts:1` (the import). The stringify side is off by one, in the `fjs/djs/tokenizer/proof.f.ts` row: the file has 40 non-comment uses, the table allocates it 39 (31 messages + 8 expected-output). `fjs/protocol/mcp/stdio/proof.f.ts:102` is a comment and correctly excluded.

So the only real inconsistency left is the prose in **Problem** — "all 21 call sites go through `parseNative`" — against the table's 23 just below it. Worth a one-character fix since the task checklist is derived from the table.

Everything else in the plan reads well:

- Splitting `numberSerialize` into its own issue under `fjs/types/bigfloat/todo/` is the right call — shortest-round-trip decimal is a numeric algorithm, not a JSON concern, and folding it in here would have made this issue unclosable. Naming the `JSON.stringify`-emits-`null`-for-`NaN` vs `String`-emits-`NaN` divergence is exactly the detail that would otherwise bite whoever picks it up.
- The lone-surrogate requirement on `stringSerialize` (match ES2019 well-formed escaping, because the migrated proofs compare against host-produced literals) is the kind of constraint that is invisible until a test fails on one platform. Good catch.
- Deferring the source-text-quoting row to `66c-emit-literals-via-owner-modules.md` rather than re-deciding it is right.
- The five phases are genuinely separable, and gating phase 4 on 2 and 3 is the correct dependency.

One non-blocking wording point I raised before and still think holds — reason 2:

> **`JSON.parse` throws.** A `throw` in FunctionalScript is a panic (§3.4) ... `parse` returns a `Result`.

Both sentences are true, but phase 1's replacement is `unwrap(parse(text))`, and `unwrap` throws too. At the proof sites that is fine — panicking is what you want in a test — so the honest framing is that the migration improves *what* the panic says (the parser's own message, not a `SyntaxError` about position 47) rather than removing it. As written, a reader who checks reason 2 against phase 1 finds a gap, and reasons 1, 3 and 4 are airtight enough not to want that.

Related note for whoever does phase 4: `unwrap` and `assert` both throw the raw value, so a bare string lands in any catch — `e.message` is `undefined`, `e instanceof Error` is `false`. We hit exactly that in `fjs-dev/finance` with a handler branching on `instanceof Error`. The doc already flags that the reporter renders payloads with `String(v)` (`fjs/emergent_testing/module.f.ts:384`), which is the same seam.

No impact on `fjs-dev/finance` — we never adopted `parseNative`. We do have the same shape as the 17 `as CasGetResult` casts in `fjs/server/module.f.js`, and we'll follow whatever schema pattern phase 1 settles on.

Still a draft, so this approval is agreement with the plan rather than a merge signal — flip it to ready when you're happy with it.

PR #1431, opened 2026-08-04
Add function parameter dispatch example to RTTI validate proofs

sergey-shandar:
Adds a `funcParam` proof to [`fjs/types/rtti/validate/proof.f.ts`](fjs/types/rtti/validate/proof.f.ts) demonstrating how RTTI validators can drive overload-style dispatch on function parameters.

The example builds two parameter tuple types (`["hello", bigint]` and `["goodbye", string, 43]`), combines them with `or`, and wraps two implementations behind a single function whose argument type is the union. At call time each validator is tried in turn and the first `ok` result selects the implementation, so both the static types and the runtime dispatch come from the same RTTI description.

Test plan: the proof asserts `x("hello", 42n) === 42` and `x("goodbye", "world", 43) === 13`, and runs as part of the existing proof suite.

### DIFF
--- a/fjs/types/rtti/validate/proof.f.ts
+++ b/fjs/types/rtti/validate/proof.f.ts
@@ -340,7 +340,7 @@ export const proof = {
         type F0<T> = (...args: Param0) => T
         type F1<T> = (...args: Param1) => T
 
-        const func = <T>(f0: F0<T>, f1: F1<T>) =>(...args: Param01): T => {
+        const func = <T>(f0: F0<T>, f1: F1<T>) => (...args: Param01): T => {
             {
                 const [t, r] = v0(args)
                 if (t === 'ok') {
@@ -353,7 +353,7 @@ export const proof = {
                     return f1(...r)
                 }
             }
-            throw new Error('Invalid arguments')
+            throw 'unreachable: args did not match any parameter set'
         }
 
         const f0 = (a: 'hello', b: bigint): number => 42
--- a/fjs/types/rtti/validate/proof.f.ts
+++ b/fjs/types/rtti/validate/proof.f.ts
@@ -324,4 +324,46 @@ export const proof = {
             assertError(v({ a: 42 }))
         },
     },
+    funcParam: () => {
+        const paramSet0 = ['hello', bigint] as const
+        const paramSet1 = ['goodbye', string, 43] as const
+
+        const paramSet01 = or(paramSet0, paramSet1)
+
+        type Param0 = Ts<typeof paramSet0>
+        type Param1 = Ts<typeof paramSet1>
+        type Param01 = Ts<typeof paramSet01>
+
+        const v0 = validate(paramSet0)
+        const v1 = validate(paramSet1)
+
+        type F0<T> = (...args: Param0) => T
+        type F1<T> = (...args: Param1) => T
+
+        const func = <T>(f0: F0<T>, f1: F1<T>) =>(...args: Param01): T => {
+            {
+                const [t, r] = v0(args)
+                if (t === 'ok') {
+                    return f0(...r)
+                }
+            }
+            {
+                const [t, r] = v1(args)
+                if (t === 'ok') {
+                    return f1(...r)
+                }
+            }
+            throw new Error('Invalid arguments')
+        }
+
+        const f0 = (a: 'hello', b: bigint): number => 42
+        const f1 = (a: 'goodbye', b: string, c: 43): number => 13
+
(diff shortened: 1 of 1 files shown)


o2alexanderfedin (commented):
Nice example — deriving the static type and the runtime dispatch from one RTTI description is the genuinely useful idea here, and `Param01 = Ts<typeof paramSet01>` alongside `validate(paramSet0)` makes it concrete in a way prose could not. CI is green and all the symbols used (`or`, `bigint`, `string`, `Ts`, `validate`, `assertEq`) are already imported in this file, so it lands clean.

One thing I checked before commenting, and three suggestions.

### The `return () => { … }` is load-bearing, and worth a comment

My first reaction was that the assertions never run — the leaf returns a closure instead of asserting. That is wrong, and I verified it rather than guessing: `registerModule` walks a leaf's **return value** through `collectTests` and registers what it finds as sub-tests (`fjs/emergent_testing/module.f.js`, the `const sub = collectTests([...path, null], false, resolved)` branch). I confirmed empirically with a probe module whose leaf returned `() => assert(false, …)` — `node --test` reported `tests 136, pass 135, fail 1`, so the sub-test really does run and really can fail.

So the pattern is correct. But it is also non-obvious, and the file's other leaves all assert inline. Since this is an *example* — the entire point is that people read and copy it — one line saying why the closure is returned (setup in the parent, assertions in the registered sub-test) would prevent both the "these never run" misreading and someone helpfully flattening it.

### First-match dispatch is order-dependent

`func` tries `v0`, then `v1`, and the first `ok` wins. That is correct here because the two parameter sets are disjoint on their first literal (`'hello'` vs `'goodbye'`), so no argument list validates against both.

An example teaches its technique, though, and this one does not show the constraint that makes it safe. If someone copies it with overlapping alternatives — say `['x', string, 43]` and `['x', string, number]` — dispatch silently resolves by declaration order, and reordering the `or` changes behaviour with no error anywhere. Worth a sentence: alternatives must be disjoint, or ordered most-specific-first.

### `throw new Error('Invalid arguments')` is unreachable and off-convention

It cannot fire for a well-typed caller, since the parameter type is `Param01` — so it is also untested. Two smaller things about it:

- Everything else in this file signals failure through `assert`, which throws a **bare string**, not an `Error`. A caught value from `assert` has `e.message === undefined` and `e instanceof Error === false`; mixing the two shapes in one file means a handler cannot treat them uniformly. (We hit exactly this in `fjs-dev/finance` — a handler branching on `instanceof Error` missed every `assert` refusal.)
- By #1430's own §3.4 argument, a `throw` here is a panic. That is defensible for "the validators disagree with the type, which is a bug" — but if it is defensible, it is worth saying so in a comment, because #1430 is currently arguing the opposite about `parseNative`.

`assert(false, ['invalid arguments', args])` would be consistent with the file and would name the offending arguments.

### Smaller

The unreachable branch is also the interesting one: it fires exactly when `Ts<typeof paramSet01>` and `validate(paramSet01)` disagree — a divergence between the type-level and runtime halves of RTTI, which is the property this example implicitly claims holds. If there is a cheap way to assert that positively (e.g. a case that validates against neither and is expected to throw, nested under a `throw` key), it would turn the dead branch into the example's strongest claim.

o2alexanderfedin (approved):
Approving. I verified this locally rather than relying on the green CI badge: checked out the PR head, `npm install`, `npx tsc --noEmit` clean, `node ./fjs/module.ts t` → **pass: 2303, fail: 0**.

I also settled the one thing I flagged before — whether the `return () => { … }` leaf actually asserts anything. It does:

```
proof.funcParam(): ok, 0.0320 ms
proof.funcParam()(): ok, 0.0710 ms
```

Both the leaf and the closure it returns register as tests. Negative probe: changing `assertEq(x('hello', 42n), 42)` to `999` turns the second line into `error` and the run into `pass: 2302, fail: 1`. So the sub-test is genuinely failable, not a no-op. Pattern confirmed correct.

The three notes from my earlier comment stay as optional follow-ups, none blocking:

- a one-line comment on why the closure is returned (setup in the parent, assertions in the sub-test) — this is an *example*, so the non-obvious bit is worth naming;
- a sentence that first-match dispatch requires the `or` alternatives be disjoint, or ordered most-specific-first — true here (`'hello'` vs `'goodbye'`), but a copier with overlapping sets gets order-dependent behaviour and no error;
- `assert(false, ['invalid arguments', args])` instead of `throw new Error(...)`, for consistency with the rest of the file and to name the offending arguments.

The idea itself — one RTTI description driving both the static type and the runtime dispatch — is the right thing to have an example for. Good addition.

PR #1433, opened 2026-08-05
Remove parseNative; migrate reading sites to parse with rtti schemas

sergey-shandar:
This PR completes the first phase of removing native JSON dependencies from `fjs/media/json`. It removes the `parseNative` export and migrates all reading sites to the total `parse` with rtti schema validation instead of `JSON.parse` plus `as` casts.

## Summary

The `fjs/media/json` module now has a complete, self-hosted JSON pipeline for reading. `JSON.parse` appears nowhere in `fjs/` outside prose.

## Key Changes

- **Removed `parseNative`** from `fjs/media/json/module.f.ts`, along with its proof test and the module-header mentions.
- **`fjs/mcp/proof.f.ts`** (17 sites): the hand-written `CasGetResult` type and every `JSON.parse(textOf(resp)) as CasGetResult` are replaced by a `casGetResult` rtti schema and a `casGetResultOf` helper. A response that drifts from the schema now fails at the parse rather than silently at an assertion. `uri` is required in the schema, matching `Meta` in `fjs/mcp/cas/module.f.ts:152-157`.
- **`fjs/emergent_testing/proof.f.ts`**: the proof-local `Event` type is now derived from an rtti schema (`Ts<typeof event>`). The reporter's `SandboxResult` payload is no longer written — no assertion reads it, and its `undefined` values have no JSON representation to round-trip through.
- **`fjs/mcp/evo/proof.f.ts`**: `as readonly string[]` → `array(string)` schema.
- **`fjs/ci/proof.f.ts`**: `parseGitHubAction` already validates, so only the `Result` needed unwrapping.
- **`fjs/djs/tokenizer/proof.f.ts`**: the 6000-token case no longer round-trips its dump through JSON. It compares the whole dump against a reconstructed literal — stronger than the entry count the old `JSON.parse` could check, since it also pins every token's kind and value — and keeps a token-list count as a cross-check.

## What this does and doesn't change

Every migrated call site is `unwrap(parse(text))`, and `unwrap` throws, so **these sites still panic on malformed input** — which is correct in proof code, where a malformed line should fail the test. What the change buys is that the panic is now the caller's decision instead of the parser's, and that narrowing to a domain type goes through a schema rather than an `as` cast.

Pattern:

```ts
unwrap(parse(text))                  // untyped Unknown
unwrap(schema(unwrap(parse(text))))  // narrowed by an rtti schema, no `as`
```

## Found along the way

`fjs/media/json`'s `parse` **panics** — `RangeError: Maximum call stack size exceeded` — on roughly 5000 or more sibling container values, on well-formed input:

| input | `parse(text)[0]` |
| --- | --- |
| 12000 numbers / 12000 short strings in an array | `ok` |
| 3000 `{}` (or `[]`) in an array | `ok` |
| 6000 `{}` (or `[]`) in an array | **`RangeError`** |
| nesting depth 5000 | `ok` |

Not a depth limit, not an element-count limit, and not the tokenizer — it's `fjs/media/json/parser`, and only for container-valued siblings. This is why the tokenizer proof above compares text instead of round-tripping. Filed as `fjs/media/json/todo/parser-sibling-container-overflow.md` with the mechanism and a proposed fix.

## Scope

Reading only. The 117 `JSON.stringify` sites are untouched; `remove-native-json.md` now covers just those four remaining phases, renumbered, with its write-site line references refreshed. Phase 2 there (shortest-round-trip number formatting) is deliberately deferred to its own issue as a numeric-algorithm task.

## Checks

`npx tsc` clean, `fjs t` 2310/2310, `npm run cov` unchanged from baseline (99.94 / 97.88 / 99.72), `npm run ci-update` produces no drift.

https://claude.ai/code/session_01EwpxtZ1pCGoDssmLHUHKjM


sergey-shandar:
@codex review

o2alexanderfedin (commented):
Reviewed the full diff and ran the suite against this head locally: **2310 pass, 0 fail**. Confirmed no `parseNative` references survive outside CHANGELOG/todo prose, and no `JSON.parse` remains in any `.f.ts` source except a doc comment in `fjs/types/nibble_set/module.f.ts`.

The core change is right. `parseNative` was `JSON.parse` under another name — it returned `Unknown` typed loosely enough that every call site reached for an `as` cast, and 17 of those disappear in `fjs/mcp/proof.f.ts` alone. Replacing hand-written response types with rtti schemas means a drifted response now fails at the parse rather than silently at an assertion, which is a real improvement in what these proofs actually prove. Only one production-facing change (the removed export), correctly marked breaking in the CHANGELOG; everything else is test code.

Three things worth settling before this comes out of draft — one is a genuine coverage regression, the other two are smaller.

### Verified as safe, for the record

- Dropping `SandboxResult` from the mock reporter's `result` event is fine. The only reads of `parseEvents` output are `events.length` and `const [, pass, fail] = events[events.length - 1]` (the summary arm) — nothing reads index 3, so the justification in the comment holds.
- `Event`'s `array(or(rttiString, null))` correctly matches `Path = readonly (string | null)[]` (`fjs/emergent_testing/module.f.ts:286`).
- `option(rttiString)` on `text`/`blob` round-trips absent keys to `undefined`, which the existing `assertEq(result.text, undefined)` assertions depend on.

fjs/djs/tokenizer/proof.f.ts
```diff
@@ -935,9 +935,10 @@ export const proof = {
             const src = Array.from({ length: 3000 }, (_, i) => `a${i % 10}`).join(' ')
             const result = tokenizeString(src)
             assert(result !== 'error', result)
-            const parsed = JSON.parse(result)
-            // 3000 ids + 2999 separating ws + a trailing eof token
-            assertEq(parsed.length, 3000 + 2999 + 1)
+            // 3000 ids + 2999 separating ws + a trailing eof token. Counted off
+            // the token list rather than off `result`: the dump is a string, and
+            // re-parsing it would only measure the serializer.
+            assertEq(toArray(tokenizeJs(stringToList(src))('a.js')).length, 3000 + 2999 + 1)
```
o2alexanderfedin, line 949: This assertion now tests a different function than the one it replaced, and the old coverage is gone.

`tokenizeString` (`module.f.ts:484`) does not go through `tokenizeJs` — it runs `descentParser(jsGrammar())` + `getTokensFromAstSequence` and serializes the result with `stringify`. The old test parsed that serialized output and counted its entries, so it validated that `tokenizeString`'s dump actually contained 6000 tokens. This version counts a freshly-tokenized list from `tokenizeJs` instead, and `result` is now only checked for `!== 'error'` two lines up.

Net effect: a `tokenizeString` that silently dropped or duplicated tokens during serialization would still pass. The comment argues that re-parsing "would only measure the serializer" — but measuring the serializer on the 3000-token input was the point of this case, and there's already an adjacent case covering the metadata-aware entry point.

Cheapest fix is to keep both: assert on the re-parsed dump *and* on the direct token list, so the round-trip and the tokenizer are each covered.

fjs/mcp/proof.f.ts
```diff
@@ -24,19 +26,26 @@ import type {
...
+// `cas_get`'s result JSON. As an rtti schema it both describes the shape and
+// checks it, so reading a field needs neither a hand-written type nor an `as`
+// cast — a response that drifts from this fails the test at the parse, not
+// silently at the assertion. `uri` is always emitted; only the payload
+// (`text` for `type: 'text'`, `blob` for `type: 'base64'`) is conditional.
+const casGetResult = {
+    length: rttiNumber,
+    mimeType: rttiString,
+    type: rttiString,
+    uri: rttiString,
```
o2alexanderfedin, line 43: `uri` was optional in the type this replaces (`uri?: string`) and is required here. The comment above states `uri` is always emitted, and every `cas_get` path exercised in this file agrees — the suite passes — so this is very likely correct and is a stronger test than the old type if so.

Still worth a confirming look at the `cas_get` implementation before it lands: this promotes an observed behavior into an enforced schema, and if any branch does omit `uri`, the failure surfaces later as an opaque parse error rather than as the `undefined` field the old type permitted. If it is guaranteed, no change needed — the tightening is a good one.

fjs/media/json/module.f.ts
```diff
@@ -128,18 +127,13 @@ export const stringify
 
 /**
  * Parses `text` as JSON with this module's own pure tokenizer and parser,
- * reporting failure as a `Result` rather than throwing — the total,
- * errors-as-values counterpart of {@link parseNative}.
+ * reporting failure as a `Result` rather than throwing. Malformed input is a
+ * value the caller destructures, so nothing in FunctionalScript has to survive
+ * a panic to read JSON.
```
o2alexanderfedin: This claim doesn't match what the migrated call sites do. Every one of them is `unwrap(parse(text))`, and `unwrap` throws — so the panic behavior at each reading site is unchanged from `parseNative`. The same overstatement is in the PR description ("Malformed JSON is a value the caller destructures, not a panic").

Unwrapping is the right call in proof code: a malformed line *should* fail the test. The issue is only that the doc now promises a property the module's own callers don't take up. What this PR actually buys is the removal of `as` casts and the addition of schema validation, which is a strong enough result on its own.

Suggest narrowing it to what `parse` offers rather than what callers do — e.g. that malformed input is *available* as a value, leaving whether to unwrap to the caller.

fjs/djs/tokenizer/proof.f.ts
sergey-shandar: You're right, and thanks — I checked `module.f.ts:483-505` and `tokenizeString` really does run its own pipeline (`getTokensFromAstRule` → `filter`/`stateScan` → `toJsTokens` → `stringify`), sharing only the grammar with `tokenizeJs`. My version asserted nothing about it.

Fixed, but not by re-parsing. I first tried exactly that and hit a wall: `fjs/media/json`'s `parse` **panics** on this dump. It throws `RangeError: Maximum call stack size exceeded` at roughly 5000 sibling container values, and the 3000-token case has 6000. Isolated it away from this PR:

| input | `parse(text)[0]` |
| --- | --- |
| 12000 numbers / 12000 short strings in an array | `ok` |
| 3000 `{}` (or `[]`) in an array | `ok` |
| 6000 `{}` (or `[]`) in an array | **`RangeError`** |
| nesting depth 5000 | `ok` |

Not depth, not element count, not the tokenizer (`tokenize` materializes all 18002 tokens fine) — it's `fjs/media/json/parser`, and only for container-valued siblings. Filed as [`fjs/media/json/todo/parser-sibling-container-overflow.md`](https://github.com/functionalscript/functionalscript/blob/claude/remove-native-json-c8wrqq/fjs/media/json/todo/parser-sibling-container-overflow.md) with the mechanism (`addToArray`'s `concat` spine forced by `endArray`/`endObject` mid-fold) and a proposed fix.

So instead of a count, the test now compares the **whole dump** against a reconstructed literal:

```ts
const ids = Array.from({ length: 3000 }, (_, i) => `a${i % 10}`)
const src = ids.join(' ')
const result = tokenizeString(src)
assert(result !== 'error', result)
assertEq(
    result,
    `[${ids.map(v => `{"kind":"id","value":"${v}"}`).join(',{"kind":"ws"},')},{"kind":"eof"}]`)
assertEq(toArray(tokenizeJs(stringToList(src))('a.js')).length, 3000 + 2999 + 1)
```

That is strictly stronger than what it replaced — the old JSON round-trip could only check the entry count, whereas this pins every token's kind and value as well — and it keeps the token-list count as the cross-check you asked for. The expectation is built from `ids`, not from the tokenizer, so it can't move with the code under test.

---

fjs/mcp/proof.f.ts
sergey-shandar: Checked, and it's guaranteed by a type rather than by observation — so no change.

`fjs/mcp/cas/module.f.ts:152-157` declares `uri` as required:

```ts
type Meta = {
    readonly length: number
    readonly mimeType: string
    readonly type: 'text' | 'base64'
    readonly uri: string
}
```

`cas_get` binds `const uri = c.url(key)` before any branching, and all five success paths spread a `Meta`/`refinedMeta` built from it (`:220`, `:228`, `:230`, `:260`, `:266`). There is no `okResult(toJson(...))` in that handler that doesn't. The old `uri?: string` was the hand-written type being looser than the value it described, which is the kind of drift the schema is there to stop.

---

fjs/media/json/module.f.ts
sergey-shandar: Agreed — the doc promised a property none of the callers take up. Every migrated site is `unwrap(parse(…))`, so the panic is still there; what actually changed is that the panic is now the caller's choice rather than the parser's. Narrowed to that:

```
 * Parses `text` as JSON with this module's own pure tokenizer and parser,
 * reporting failure as a `Result` rather than throwing: malformed input is
 * *available* as an `error` to branch on. Whether to branch or to `unwrap` it
 * back into a panic is the caller's decision — the parser no longer makes it
 * for them.
```

Also fixed the same overstatement in the PR description.

---

sergey-shandar:
@codex review

fjs/media/json/todo/parser-sibling-container-overflow.md
```diff
@@ -0,0 +1,104 @@
...
+| `[{}, {}, … ]`, 3000 empty objects | `ok` |
+| `[{}, {}, … ]`, 6000 empty objects | **`RangeError`** |
+| `[{"k":1}, … ]`, 6000 one-field objects | **`RangeError`** |
+| `{"k0":{}, "k1":{}, … }`, 6000 objects in an object | **`RangeError`** |
+| `[[[[…]]]]`, nesting depth 5000 | `ok` |
+
+Three things this **is not**:
+
+- **Not a depth limit.** 5000 levels of nesting parse fine; the failing cases
+  are flat, depth 2.
```
o2alexanderfedin: This bullet is engine-dependent but stated absolutely, and it flips on a newer Node.

The table is honestly labelled "Measured on Node 22" (line 15), so this isn't a mistake so much as a result that doesn't travel. On **Node v23.11.0** I get:

| case | 3000 | 4000 | 5000 | 6000 | 12000 | 50000 |
| --- | --- | --- | --- | --- | --- | --- |
| nesting depth | ok | ok | **RangeError** | — | — | — |
| sibling `{}` | ok | ok | **RangeError** | RangeError | — | — |
| sibling numbers | — | — | — | ok | ok | ok |

So `nesting depth 5000` (line 27, listed as `ok`) throws, and depth lands in the *same* ~5000 band as sibling containers rather than comfortably above it.

What this does **not** touch: the asymmetry that makes this a bug is intact on 23.11.0 — 50,000 sibling primitives parse fine while ~5,000 sibling containers overflow. Deep nesting consuming stack is expected of a recursive-descent parser; flat siblings consuming it is not. Direction 1 (fix the accumulator) is still the right target.

What it does touch is the framing. "5000 levels of nesting parse fine" reads as *depth is safe, siblings are the problem*, and on 23.11.0 depth is not safe — it fails at the same size. Suggest either re-measuring the depth row on the Node you're releasing against, or qualifying the bullet the way line 39–40 already qualifies the sibling threshold ("varies with the engine and with how much stack…"). That caveat currently attaches only to the sibling number, not to this claim.

Worth adding the exact Node version to the depth row too, since the two rows evidently disagree across majors.

fjs/media/json/todo/parser-sibling-container-overflow.md
```diff
@@ -0,0 +1,104 @@
...
+
+Whichever lands, the contract to establish is that `parse` **never panics on
+input the tokenizer accepts**: any limit it cannot honor must come back as
+`error`, not as a `throw`.
+
+### Tasks
+
+- [ ] Reproduce as a proof in `fjs/media/json/parser/proof.f.ts`: 6000 sibling
+      `{}` and 6000 sibling `[]` both return `ok`, plus the 12000-primitive and
+      5000-deep cases as the already-working baselines.
```
o2alexanderfedin: Small wording snag: this sits under **Reproduce** but states the post-fix expectation. Today 6000 sibling `{}` and `[]` both throw — a proof asserting they "return `ok`" would fail on `main` until the accumulator fix lands.

Probably worth splitting into the two things it's actually describing: a repro proof pinning the current `RangeError` (so the bug can't silently regress after being fixed), and the `ok` assertion as an item under the fix. Also worth flagging the 5000-deep case as a baseline only if you re-confirm it on the target Node, per the other comment.

o2alexanderfedin (commented):
Re-reviewed at `13ed83d4`. Verified locally: `npx tsc --noEmit` clean, `npm test` → **2310 pass, 0 fail**.

All three points are resolved, and two of them better than what I proposed:

1. **Tokenizer coverage** — stronger than the fix I suggested. Comparing the whole dump against a literal rebuilt from `ids` pins every token's kind *and* value, where my suggestion would only have restored the count. Building the expectation from `ids` rather than from the tokenizer is the right call — it can't drift with the code under test. Keeping the `tokenizeJs` count alongside as a cross-check is a nice touch.
2. **`uri`** — you're right and I was wrong to doubt it. `Meta` in `fjs/mcp/cas/module.f.ts:152-157` declares it required, and `const uri = c.url(key)` binds before any branching, so it's guaranteed by a type rather than by observation. The old `uri?: string` was the loose one.
3. **The doc claim** — narrowed to exactly the right thing.

Finding the parser overflow while doing (1) is a good catch, and the todo file is a thorough writeup. I reproduced it independently and the central asymmetry holds on my machine too: sibling **primitives** parse fine at 50,000, sibling **containers** die around 5,000. The bug and the proposed fix direction both stand.

One of the supporting claims doesn't survive a newer engine, though — details inline. It doesn't change the diagnosis, but it would mislead whoever picks this up.

fjs/media/json/module.f.ts
```diff
@@ -128,18 +127,14 @@ export const stringify
 
 /**
  * Parses `text` as JSON with this module's own pure tokenizer and parser,
- * reporting failure as a `Result` rather than throwing — the total,
- * errors-as-values counterpart of {@link parseNative}.
+ * reporting failure as a `Result` rather than throwing: malformed input is
+ * *available* as an `error` to branch on. Whether to branch or to `unwrap` it
+ * back into a panic is the caller's decision — the parser no longer makes it
```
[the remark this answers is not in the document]
sergey-shandar: The defect is real — I found and filed it — but I don't think either proposed remedy is right, and I don't believe this PR is what exposes anyone to it.

**This PR doesn't remove a path anyone had.** `parseNative` was added in #1428 and its CHANGELOG entry is still in `## Unreleased`, so it has never been published — no consumer can lose it. All 21 of its call sites were in proofs; the repository has no production JSON reads at all. And `parse` has carried this bug since #1428 regardless: deleting `parseNative` doesn't create the panic, it only removes a workaround that nothing was using. None of the migrated sites parse documents anywhere near the threshold — the one that would have, the 6000-token tokenizer dump, is deliberately not routed through `parse` for exactly this reason.

**"Keep a safe fallback" is the thing this issue exists to undo.** AGENTS.md §5.2 calls keeping the old API beside the new one a last resort — "two shapes for one concept doubles what a reader has to understand and, in practice, the old one never leaves." That is not a general preference here; it is the stated motivation of `remove-native-json.md`, whose whole diagnosis is that `parse` landed in #1428 and nothing moved because `parseNative` stayed. Re-adding it as a fallback would reproduce the exact situation being fixed.

**"Fix the parser first" folds two changes into one PR.** §8.1 asks a PR to implement one improvement. The parser fix is not a drive-by either: my first-pass mechanism was wrong, and I've corrected the issue rather than ship a guess. New measurements —

| input | result |
| --- | --- |
| array of 6000 primitives | `ok` |
| object of 6000 primitive-valued fields | `ok` |
| array of 6000 `[]` / `{}` / `[1]` / `{"k":1}` | **`RangeError`** |

— rule out the accumulator I'd blamed: 12000 primitives walk a longer spine of the same shape and are fine. The trigger is the *number of completed child containers*, empty ones included, not the size of anything. That needs localizing before a fix, not folding into a proof migration.

What I did take from your comment: **raised it P2 → P1**, since `parse` is now the only JSON reader in the tree and a `Result` return that can still panic gives callers a guarantee the code doesn't keep. The issue now records the ruled-out suspect so the next person doesn't re-derive it, and makes localizing the recursion task one. `0e7313c`.

On the AGENTS.md citation — §3.4 (L208-L214) says a `throw` is a panic nothing can catch, which is why this is a bug worth P1. It doesn't argue for retaining a second reader; §5.2 argues against it.

---

sergey-shandar:
@codex review

o2alexanderfedin:
Chased the overflow down to a root cause and it turned out to be a one-liner. Fix + regression proofs are up as #1435; details here so this thread has them.

## Root cause

`endArray`/`endObject` pop the enclosing container with `first(null)(state.stack)` / `drop(1)(state.stack)` (`parser/module.f.ts:93` and `:110`). `drop` is `apply(dropStep(n))` — it returns a **`Thunk`**, not a forced list. So every closed container replaces the stack with *another unforced thunk wrapping the previous one*. After N containers the stack is `drop(1)(drop(1)(…))` nested N deep — logically still depth 1, structurally a chain — and forcing it recurses `apply` → `next` → `trampoline`, one frame per link.

That is exactly the shape you measured. The threshold tracks **containers**, not values, because primitives never push or pop the stack — hence 12000 numbers fine, ~5000 sibling `{}` fatal. The stack trace is the tell:

```
at next       (fjs/types/list/module.f.ts:62:24)
at            (fjs/types/list/module.f.ts:104:15)   <- apply's thunk
at trampoline (fjs/types/list/module.f.ts:58:49)
at next       (fjs/types/list/module.f.ts:65:16)
… repeating
```

So it is not the `addToArray` `concat` spine — `next` walks that iteratively at lines 69–71 (`[head, tail] = [head.head, concat(head.tail)(tail)]; continue`), which is why appending 50000 primitives is fine. The `concat` accumulator is not the thing that needed fixing.

## Fix

The stack is written only by `startArray`/`startObject`, always as a literal cons `{ first, tail }`. Popping it needs no list combinator — force one `next` and take the parts:

```ts
// Pops the enclosing container off `stack`. `next` is forced here rather than
// left as a `drop(1)` thunk: the stack is written only by startArray/startObject,
// always as a literal cons, and a lazy pop would leave one unforced thunk per
// closed container — a chain that overflows the stack when it is finally forced.
const popStack
    : (stack: JsonStack) => StateParse
    = stack => {
        const ne = next(stack)
        return ne === null
            ? { status: '', top: null, stack: null }
            : { status: '', top: ne.first, stack: ne.tail }
    }
```

then both `endArray` and `endObject` become `const newState = popStack(state.stack)`, and `first`/`drop` drop out of the imports in favour of `next`. O(1) per pop, nothing accumulates.

## Results

| | before | after |
| --- | --- | --- |
| 6000 sibling `{}` / `[]` | `RangeError` | `ok` |
| 6000 one-field objects | `RangeError` | `ok` |
| 6000 objects in an object | `RangeError` | `ok` |
| nesting depth 5000 | `RangeError` | `ok` |
| **100000** sibling containers | `RangeError` | `ok` |
| **nesting depth 20000** | `RangeError` | `ok` |
| 12000 primitives (baseline) | `ok` | `ok` |

`tsc --noEmit` clean, full suite **2316 pass / 0 fail**. The five new `siblingContainers` proofs in `fjs/media/json/parser/proof.f.ts` are a real regression test — reverting the parser change while keeping them yields **4 failures**, all `RangeError`, with the primitives baseline still green.

## What this means for this PR

Two knock-ons, neither blocking:

- Most of the **Tasks** list in `parser-sibling-container-overflow.md` is now done. Once both land, that file wants updating or deleting — I deliberately left it untouched in #1435 so it would not conflict with this branch.
- The `largeInputs[1]` note about revisiting the tokenizer test once `parse` can read the dump is now unblocked: the 3000-token dump parses. Your literal-comparison assertion is still the better test, so I would keep it either way.

The **depth** correction from my earlier comment stands and is the same bug — deep nesting was failing for this identical lazy-pop reason, not a separate limit. That is also why the fix lifts depth to 20000 for free.

One thing I did not check, which your todo raises: whether `fjs/djs`'s parser shares the lazy-pop shape.

fjs/media/json/todo/parser-sibling-container-overflow.md
sergey-shandar: You're right, and it's worse than "doesn't travel" — the claim was wrong on my own Node too. I'd sampled depth at 5000 and called it safe; binary-searching the largest parsable input of each shape on Node 22.22.2 gives:

| shape | largest that parses |
| --- | --- |
| sibling numbers | >200000 (no limit found) |
| sibling strings | >200000 (no limit found) |
| primitive-valued object fields | >200000 (no limit found) |
| sibling `{}` | **5687** |
| sibling `[]` | **5687** |
| nesting depth | **5689** |

Depth and siblings fail **two frames apart**. My 5000 sample just happened to sit under the cliff, so "5000 levels of nesting parse fine, therefore not a depth limit" was an artifact of where I measured, not a property — exactly as your 23.11.0 numbers imply. The depth-vs-breadth framing was wrong.

Your reframing is the correct one and I've adopted it: the anomaly is the **primitive/container asymmetry**, not depth vs breadth. Deep nesting costing stack is ordinary for a recursive-descent-shaped parser; 200,000 sibling primitives costing none while 5,688 empty `{}` overflow is not — an empty object is not more work than a number. The issue is retitled **`parser-container-stack-cost` — `parse` spends a call-stack frame per container**, and renamed to match, since the defect isn't specific to siblings.

Two more things I took from this:

- Every number is now explicitly a measurement of a moving limit, with the engine and version attached and your 23.11.0 data point recorded alongside mine — the "varies with the engine" caveat now governs the whole table rather than one row.
- The near-identical 5687/5689 is itself a finding: it says look for **one** cause, not two. That's now in the proposal, along with the note that a *depth* bound may be legitimate to keep (refusing a 100,000-deep document is reasonable) as long as it comes back as `error` rather than a panic — whereas the flat-sibling cost has no such defence.

I also flagged the superseded claim inline so nobody re-derives it. `35f48df`.

---

fjs/media/json/todo/parser-sibling-container-overflow.md
sergey-shandar: Correct — that task asserted the post-fix behavior while sitting under "Reproduce", so a proof written from it would have failed until the fix landed. Split as you suggested:

```
- [ ] Localize the recursion: what does completing a container leave on the
      stack that completing a primitive does not?
- [ ] Pin today's behavior first, so the bug cannot silently return: a proof in
      `fjs/media/json/parser/proof.f.ts` with the sibling-container and
      deep-nesting cases under a `throw` key (§3.4), plus the
      200,000-primitive array and object cases as `ok` baselines. Size the
      throwing cases well past any plausible engine threshold — they are
      measurements of a moving limit, not of a fixed one.
- [ ] Fix, then flip those `throw`-keyed cases to `ok` assertions and confirm
      sibling containers scale like primitives.
```

The repro goes under a `throw` key rather than a hand-written check, since §3.4 makes that the way to assert a call panics.

On your last point — the 5000-deep case is no longer listed as a baseline at all. Re-measuring showed it fails at 5689 on Node 22.22.2, so it belongs with the throwing cases, not with the working ones. The only `ok` baselines left are the primitive ones, which are the rows that actually hold across engines.

---

o2alexanderfedin:
The rewritten `parser-container-stack-cost.md` is a clear improvement — binary-searching the thresholds was the right call, and 5687 siblings vs 5689 depth landing **two frames apart** is much stronger evidence for a single cause than my one-off samples were. Reframing the anomaly as primitive-vs-container rather than depth-vs-breadth is correct.

Our edits crossed, though — it went in ~46s after my comment above, so its **Tasks** list still reads as unfixed. Two things it should probably absorb:

**It is already fixed.** #1435 is open and green (18/18 CI, suite 2316 pass / 0 fail). Root cause is the lazy pop: `endArray`/`endObject` pop with `drop(1)`, which is `apply(dropStep)` and returns a `Thunk`, so each closed container leaves one unforced thunk wrapping the stack. N containers ⇒ an N-deep chain that recurses `apply` → `next` → `trampoline` when forced. That is exactly why the cost is per *container* and why arrangement is irrelevant — nesting and siblings both push and pop.

**"Direction 1" is not needed.** The accumulator is not the bug. `next` walks the `addToArray` `concat` spine iteratively at `list/module.f.ts:69-71` (`[head, tail] = [head.head, concat(head.tail)(tail)]; continue`), which is why 200,000 sibling primitives cost nothing — they go through that same accumulator. Swapping it for an O(1)-prepend would be a real improvement for other reasons, but it would not have moved this threshold. The fix in #1435 is to pop eagerly instead: force one `next`, take `first`/`tail`. The stack is only ever written by `startArray`/`startObject` as a literal cons, so no list combinator is needed there at all.

After it: 100,000 sibling containers and nesting depth 20,000 both parse.

The proofs you describe in Tasks already exist there in the form you asked for — sibling `{}`/`[]`, objects-in-object, deep nesting, plus a primitive baseline. I sized them at 6000 rather than "well past any plausible engine threshold"; if you would rather they sit at 200,000 so they cannot quietly pass on a roomier engine, say so and I will raise them. I kept them as `ok` assertions rather than `throw`-keyed ones since the fix ships in the same PR.

I deliberately did not touch the todo in #1435 to avoid conflicting with this branch — it wants updating or deleting once both land.

o2alexanderfedin (approved):
Approving at `22893706`. Verified at that head: `npx tsc --noEmit` clean, full suite **2312 pass / 0 fail**, 18/18 CI green.

All three of my original findings are resolved, two of them better than what I proposed:

- **Tokenizer coverage** — comparing the whole dump against a literal rebuilt from `ids` pins every token's kind *and* value, where restoring the count would only have pinned the total. Building the expectation from `ids` rather than from the tokenizer is the right instinct.
- **`uri`** — my concern was unfounded; `Meta` in `fjs/mcp/cas/module.f.ts:152-157` guarantees it by type, and `const uri = c.url(key)` binds before any branching. The old `uri?: string` was the loose one.
- **The panic claim** — narrowed to exactly what `parse` offers.

The core change stands on its own: `parseNative` was `JSON.parse` renamed, and replacing it with schema-validated `parse` removes 17 `as` casts in `fjs/mcp/proof.f.ts` alone, so a drifted response now fails at the parse rather than silently at an assertion. Breaking change correctly recorded in the CHANGELOG.

The parser-overflow writeup that came out of this is a genuinely good side effect — binary-searching to 5687 siblings vs 5689 depth is much stronger evidence than my coarse sampling, and the primitive-vs-container reframing is right.

One thing to reconcile after merge, not a blocker: `parser-container-stack-cost.md`'s Tasks list still reads as unfixed and proposes replacing the accumulator. #1435 already fixes it by popping eagerly, and the accumulator is not the cause — `next` walks the `concat` spine iteratively (`list/module.f.ts:69-71`), which is why 200,000 sibling primitives cost nothing through that same accumulator. Details in my comment above; that file wants updating or deleting once both land.

PR #1434, opened 2026-08-05
Add object parameter dispatch example to RTTI validate proofs

sergey-shandar:
Adds a `funcObj` proof to [`fjs/types/rtti/validate/proof.f.ts`](fjs/types/rtti/validate/proof.f.ts), the object-shaped counterpart to the existing `funcParam` proof from #1431.

Where `funcParam` dispatches on a positional argument tuple, `funcObj` describes each overload as a record (`{ a: 'hello', b: bigint }` and `{ a: 'goodbye', b: string, c: 43 }`), combines them with `or`, and wraps two implementations behind a single function taking the union. At runtime each validator is tried in turn and the first `ok` result selects the implementation, so the static type and the dispatch both come from the same RTTI description.

Test plan: `npx tsc` is clean and the proof runner passes (2317 pass, 0 fail).

### DIFF
--- a/fjs/types/rtti/validate/proof.f.ts
+++ b/fjs/types/rtti/validate/proof.f.ts
@@ -398,8 +398,8 @@ export const proof = {
             throw 'unreachable: args did not match any parameter set'
         }
 
-        const f0 = (args: Ts<Param0>): number => Number(args.b) + args.a.length
-        const f1 = (args: Ts<Param1>): number => args.c
+        const f0 = (args: Param0): number => Number(args.b) + args.a.length
+        const f1 = (args: Param1): number => args.c
 
         const x: (args: Param01) => number = func(f0, f1)
 
--- a/fjs/types/rtti/validate/proof.f.ts
+++ b/fjs/types/rtti/validate/proof.f.ts
@@ -399,6 +399,13 @@ export const proof = {
         }
 
         const f0 = (args: Ts<Param0>): number => Number(args.b) + args.a.length
-        const f1 = (args: Ts<Param1>): number => 13
+        const f1 = (args: Ts<Param1>): number => args.c
+
+        const x: (args: Param01) => number = func(f0, f1)
+
+        return () => {
+            assertEq(x({ a: 'hello', b: 42n }), 47)
+            assertEq(x({ a: 'goodbye', b: 'world', c: 43 }), 43)
+        }
     }
 }
--- a/fjs/types/rtti/validate/proof.f.ts
+++ b/fjs/types/rtti/validate/proof.f.ts
@@ -365,5 +365,40 @@ export const proof = {
             assertEq(x('hello', 42n), 42)
             assertEq(x('goodbye', 'world', 43), 13)
         }
+    },
+    funcObj: () => {
+        const param0 = { a: 'hello', b: bigint }
+        const param1 = { a: 'goodbye', b: string, c: 43 }
+
+        const param01 = or(param0, param1)
+
+        type Param0 = Ts<typeof param0>
+        type Param1 = Ts<typeof param1>
+        type Param01 = Ts<typeof param01>
+
+        const v0 = validate(param0)
+        const v1 = validate(param1)
+
+        type F0<T> = (args: Param0) => T
+        type F1<T> = (args: Param1) => T
+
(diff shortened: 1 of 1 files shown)


fjs/types/rtti/validate/proof.f.ts
```diff
@@ -365,5 +365,47 @@ export const proof = {
...
+                const [t, r] = v1(args)
+                if (t === 'ok') {
+                    return f1(r)
+                }
+            }
+            throw 'unreachable: args did not match any parameter set'
+        }
+
+        const f0 = (args: Ts<Param0>): number => Number(args.b) + args.a.length
+        const f1 = (args: Ts<Param1>): number => args.c
```
o2alexanderfedin: `Ts` is applied twice here. `Param0` is already `Ts<typeof param0>` (line 375) — the TypeScript projection — so `Ts<Param0>` re-runs the schema→type transform on something that is no longer a schema. Same on line 402 with `Ts<Param1>`.

It's also internally inconsistent: `F0`/`F1` two lines up (382–383) take `Param0`/`Param1` directly, and the sibling `func` proof writes its handlers with plain concrete parameters (`const f0 = (a: 'hello', b: bigint): number` at line 359) — no `Ts` at all. So the double application is new here rather than an existing habit.

I verified both halves of why this is worth fixing rather than leaving:

1. **It's redundant.** Replacing `Ts<Param0>`/`Ts<Param1>` with `Param0`/`Param1` compiles clean — the two types are identical for this schema.
2. **It compiles only by coincidence.** `Ts<T extends Type>` requires its argument to be a schema, and `Param0` satisfies that only because this particular projection happens to round-trip: `bigint` is a member of rtti's `Primitive`, so `ConstTs` falls through to identity. That doesn't generalize. Probing the same double application against other schemas: a nested `array(string)` survives, but an `unknown` field fails outright with `TS2589: Type instantiation is excessively deep and possibly infinite`.

Since this file exists to demonstrate the dispatch pattern, `Ts<Ts<schema>>` is the part a reader is most likely to copy — and it will break the moment their schema has an `unknown` in it.

```suggestion
        const f0 = (args: Param0): number => Number(args.b) + args.a.length
        const f1 = (args: Param1): number => args.c
```

o2alexanderfedin (commented):
Checked this out and verified locally: `npx tsc --noEmit` clean, `npm test` → **2313 pass, 0 fail** (+3 over base, as expected). All 18 CI checks green.

The proof itself is a faithful mirror of the `func` case above it — tuple parameters swapped for an object parameter, same two-schema dispatch, same structure. `throw 'unreachable: args did not match any parameter set'` matches line 356 verbatim, and the deferred `return () => { … }` shape matches lines 364/406, so it's consistent with the file and with the wider repo idiom (~20 bare `throw`s across `.f.ts` sources, e.g. `fjs/types/btree/set/module.f.ts:88`). The arithmetic checks out too: `Number(42n) + 'hello'.length === 47`.

One substantive thing, on lines 401–402 — a second `Ts` application that shouldn't be there.

o2alexanderfedin (commented):
Re-reviewed at `d32eb823` — the suggestion is applied and this looks good to me.

Verified locally: `npx tsc --noEmit` clean, `npm test` → **2313 pass, 0 fail**. `f0`/`f1` now take `Param0`/`Param1` directly, which lines them up with `F0`/`F1` above and with the sibling `func` proof, and drops the `Ts<Ts<schema>>` shape a reader might otherwise have copied.

Nothing further from me. Note the `github-code-quality` bot's "unused variable" hits on `func`, `f0`, and `f1` are false positives — all three are consumed at line 404 via `func(f0, f1)`.

o2alexanderfedin (approved):
Approving. Re-verified at `d32eb823`: `npx tsc --noEmit` clean, full suite **2313 pass / 0 fail**, and all 18 CI checks green.

The one thing I raised — the double `Ts` application on `f0`/`f1` — is fixed, and the proof now reads consistently: `f0`/`f1` take `Param0`/`Param1` exactly as `F0`/`F1` declare them, matching the sibling `func` proof above. The added `funcObj` case is a useful complement to it — object-parameter dispatch through `or(param0, param1)` with `validate` doing the narrowing, no `as` casts anywhere.

Nothing else from me.

PR #1436, opened 2026-08-05
Fix the same per-container stack cost in the djs parser; retire the issue file

o2alexanderfedin:
Follow-up to #1435, which fixed `fjs/media/json`. `fjs/djs`'s parser carried the **identical** defect, so the issue file that #1435 could not fully close is retired here.

## The same bug

`fjs/djs/parser/module.f.ts:292,316` popped the enclosing container with `first(null)(state.stack)` / `drop(1)(state.stack)` — character-for-character what `fjs/media/json` had. `drop` is `apply(dropStep(n))` and returns a `Thunk`, so each closed container left one unforced thunk wrapping the stack; forcing that chain cost a call-stack frame per container.

Measured on the djs parser before this change:

| input | before | after |
| --- | --- | --- |
| 6000 sibling `{}` | `RangeError` | `ok` |
| 6000 sibling `[]` | `RangeError` | `ok` |
| nesting depth 6000 | `RangeError` | `ok` |
| 20000 primitives | `ok` | `ok` |

Same signature as the json parser: the cost is per **container**, arrangement irrelevant, primitives unbounded.

Same fix — the stack is written only by `startArray`/`startObject`, always as a literal cons, so popping needs no list combinator: force one `next`, take `first`/`tail`.

## Retiring `parser-container-stack-cost.md`

AGENTS.md §"After fixing" says to delete the issue file in the same PR and capture the design first. #1435 could not do that honestly — the issue named `fjs/djs` as an open question, and it was still broken. Both parsers are fixed as of this PR, so it is deleted here:

- The mechanism now lives in `popStack`'s comment in each parser and in the two regression proof blocks.
- Its P1 contract — *`parse` never panics on input the tokenizer accepts* — is met. After #1435, `fjs/media/json`'s `parse` handles **nesting depth 1,000,000** and **1,000,000 sibling containers**; the recursion was entirely the lazy-pop chain, and the fold plus explicit stack were already iterative.
- Its one cross-reference, from `remove-native-json.md`, is retargeted at #1435 rather than left dangling.

Also fixes a dead path Codex flagged on #1435: `fjs/media/json/parser/proof.f.ts` referenced the issue under its pre-rename name `parser-sibling-container-overflow.md`. That context is inline now, so it cannot rot again.

## Verification

`npx tsc --noEmit` clean. Full suite **2321 pass / 0 fail**.

The four `containerStackCost` proofs are a genuine regression test: reverting only `fjs/djs/parser/module.f.ts` while keeping them gives **3 failures**, all `RangeError`, with the 20000-primitive baseline still passing. Sized at 20000 rather than 6000 so they cannot quietly pass on a roomier engine — the thresholds are a moving target, as the issue file noted.

https://claude.ai/code/session_01V9JGBQacxZPLJGmuHUaEzV


sergey-shandar:
@codex review

sergey-shandar (approved):

PR #1438, opened 2026-08-05
Serialize JSON strings in FunctionalScript, not `JSON.stringify`

sergey-shandar:
Phase 1 of [`fjs/media/json/todo/remove-native-json.md`](https://github.com/functionalscript/functionalscript/blob/main/fjs/media/json/todo/remove-native-json.md).

The reading half of `fjs/media/json` stopped depending on the host when `parseNative` went away (#1433). The writing half still bottomed out in `JSON.stringify`: `stringSerialize` and `numberSerialize` were the host under a different name, and `fjs/djs/serializer` imports both — so every string this repository serialized, JSON and DJS alike, was formatted by the host.

This PR replaces the string leaf. `numberSerialize` stays as it is; shortest-round-trip number formatting is phase 2 and belongs next to the numeric code, per the issue.

## What it does

`stringSerialize` decodes its input with `stringToCodePointList` and escapes each code point:

- the named escapes `\b \t \n \f \r \" \\`, as a table keyed by the `fjs/text/ascii` constants;
- every other code point below `space` as `\u00XX`, lowercase hex, matching ECMAScript's `UnicodeEscape`;
- unpaired surrogates as `\ud800` — this is well-formed `JSON.stringify` (ES2019) behavior, and it falls out for free because `stringToCodePointList` already tags lone surrogates with `errorMask`;
- everything else through unchanged.

The output is still a single chunk, like `numberSerialize`, `nullSerialize`, and `boolSerialize`, so no consumer sees a shape change.

## Correctness

The contract is "identical to the host", so it was checked that way before the proof was written: every single code unit `0x0000`–`0xFFFF`, every surrogate × surrogate pair, surrogates adjacent to BMP characters, and 50k random strings — all byte-identical to `JSON.stringify`. The proof itself pins the host's literals (rather than calling the host at test time) so a future divergence fails a test instead of being papered over.

## `codePointToString`

`fjs/text/utf16` gains `codePointToString`, the scalar counterpart of `codePointListToString`. The escape loop holds one code point at a time, and wrapping it in a single-element list to reach the list version cost about half of the loop's total time. It is a primitive the module was missing, not a fast path in the serializer.

## Cost

Replacing a C++ builtin with an interpreted decode pipeline is not free. On a 20k-key object (~1.3 MB of JSON) end-to-end `stringify` goes from ~130 ms to ~700 ms; the full FunctionalScript suite goes from ~67–69 s to ~72 s. The naive version — per-code-point chunks and `codePointListToString` per character — was ~1520 ms on the same benchmark, so this is already 2× better than the first working shape. If the remaining gap ever matters, the generic fix is a faster scan in `text/utf16`, not a special case here; the issue file records the measurement.

## Checks

- `npx tsc` — clean
- `node ./fjs/module.ts t` — 2340 pass, 0 fail
- `npm run cov` — `fjs/media/json/serializer/module.f.ts` at 100% line / branch / function
- `bun fjs/module.ts t` — 2338 pass, 2 fail; both failures (`fjs/fsc/proof.f.ts` `f1`/`f2`) reproduce unchanged on `main`

Rust untouched.

https://claude.ai/code/session_0199JaXVTAGvky4bhP6c1eJZ

---
_Generated by [Claude Code](https://claude.ai/code/session_0199JaXVTAGvky4bhP6c1eJZ)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ history.
 
 ## Unreleased
 
+- `fjs/media/json`: `stringSerialize` escapes in FunctionalScript instead of
+  calling the host's `JSON.stringify`, matching it exactly down to `\uXXXX` for
+  unpaired surrogates; `fjs/text/utf16` gains `codePointToString`
+  [#1438](https://github.com/functionalscript/functionalscript/pull/1438)
 - `fjs/djs`: the parser no longer spends a call-stack frame per container —
   the same lazy `drop(1)` pop that #1435 fixed in `fjs/media/json`
   [#1436](https://github.com/functionalscript/functionalscript/pull/1436)
--- a/fjs/media/json/serializer/module.f.ts
+++ b/fjs/media/json/serializer/module.f.ts
@@ -1,19 +1,79 @@
 /**
  * JSON serializer for deterministic string output.
  *
+ * `stringSerialize` is FunctionalScript, not the host's `JSON.stringify`: it
+ * escapes over this repository's own UTF-16 decoder and reproduces the
+ * ECMAScript `QuoteJSONString` result exactly, lone surrogates included.
+ *
  * @module
  */
-import { flat, reduce, empty, type List } from '../../../types/list/module.f.ts'
+import { flat, map, reduce, empty, type List } from '../../../types/list/module.f.ts'
 import { type Reduce } from '../../../types/function/operator/module.f.ts'
+import { concat } from '../../../types/string/module.f.ts'
+import {
+    codePointToString,
+    stringToCodePointList,
+    type CodePoint,
+} from '../../../text/utf16/module.f.ts'
+import { errorMask } from '../../../text/code_point/module.f.ts'
+import {
+    backspace,
+    cr,
+    digit0,
+    ff,
+    ht,
+    latinSmallLetterA,
+    lf,
+    quotationMark,
+    reverseSolidus,
+    space,
+} from '../../../text/ascii/module.f.ts'
 
 const jsonStringify = JSON.stringify
+
+const { fromCharCode } = String
+
+/**
+ * The code points JSON gives a two-character escape. Every other code point
+ * below `space` has no short form and goes through `unicodeEscape` instead.
+ */
+const escapeTable = {
+    [backspace]: '\\b',
+    [ht]: '\\t',
+    [lf]: '\\n',
+    [ff]: '\\f',
+    [cr]: '\\r',
+    [quotationMark]: '\\"',
+    [reverseSolidus]: '\\\\',
+} as const
+
+const hexDigit = (value: number): string =>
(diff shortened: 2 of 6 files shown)


sergey-shandar:
@codex review

o2alexanderfedin (approved):
Approving. I verified the "identical to the host" contract independently rather than taking the PR description's word for it.

**Verification** (worktree at `eb9b69d0`, clean `npm ci`):

- `npx tsc --noEmit` — clean
- `node ./fjs/module.ts t` — 2340 pass, 0 fail
- Differential test against `JSON.stringify`: every code unit `0x0000`–`0xFFFF` alone, embedded (`a<c>b`), and doubled; ~11k surrogate pairs; every surrogate adjacent to a BMP character and to an astral character; 50k random strings. **281,424 inputs, 0 mismatches** — lone-surrogate `\uXXXX` behavior included.

The design choice that makes this work is worth calling out: leaning on `errorMask` from `stringToCodePointList` means unpaired surrogates fall out of the existing decoder instead of becoming a special case in the serializer, and pinning the host's literals in the proof (rather than calling the host at test time) means a future divergence fails a test instead of being papered over. Single-chunk output keeps the leaf shape consistent with `numberSerialize` / `nullSerialize` / `boolSerialize`, so no consumer sees a change.

Two non-blocking notes:

1. **`codePointToString`'s out-of-range contract isn't in the JSDoc.** The proof documents that `codePointToString(0x110000)` keeps the low 16 bits, but the doc comment only says "the one- or two-code-unit string encoding it". Since this is newly *exported* API, silent truncation of invalid input deserves a sentence at the declaration, where callers will read it. The behavior itself is right: it matches `codePointListToString`, and consistency beats a new error path here.

2. **The ~5× cost is a floor now, not just a benchmark number.** `fjs/djs/serializer` imports `stringSerialize`, so every DJS string in every downstream consumer pays it. ~2 MB/s is fine for current workloads and the todo file records the measurement honestly, but it is worth knowing that the generic fix (a faster `text/utf16` scan) is the one to reach for if it ever bites — not a special case back in the serializer.

sergey-shandar:
Thanks for the independent differential run — 281k inputs is a stronger check than mine.

**Note 1 — fixed** in 878cda2. `codePointToString`'s JSDoc now states the out-of-range contract at the declaration: a surrogate, a value above `0x10FFFF`, or a negative one is truncated to its low 16 bits and encoded as that code unit, rather than rejected, and says why — malformed input is reported through the decoder's `errorMask` on the way *in*, not through a failure on the way out. The example gained the `0x110000` case so the behavior is visible without opening the proof. `npx tsc` clean, 2340 pass / 0 fail.

**Note 2 — agreed, no change.** The `fjs/djs/serializer` path is exactly why the measurement is in the issue file rather than only in this PR description, and the file already says the generic fix is a faster `text/utf16` scan and not a special case in the serializer. Worth adding that phase 2 will put `numberSerialize` on the same footing, so the floor is not final until that lands and the two leaves can be measured together.

---

PR #1440, opened 2026-08-05
Upgrade Node.js to 26.6.0 and update nixpkgs commit

sergey-shandar:
## Summary
This PR updates the default Node.js version from 26.5.1 to 26.6.0 across the project configuration and CI workflows, along with updating the nixpkgs commit reference.

## Key Changes
- Updated default Node.js version from 26.5.1 to 26.6.0 in `fjs/ci/config/module.f.ts`
- Updated Node.js version in all CI workflow job configurations (`.github/workflows/ci.yml`)
- Updated nixpkgs commit reference from `531670d871c0e29724a02f3cbcac170adc65b58c` to `04607e1165ac22c5fde6dcc54c9e0b3c0487c555` in:
  - `fjs/ci/config/module.f.ts`
  - `nix/generated/node22/flake.nix`
  - `nix/generated/node24/flake.nix`
  - `nix/generated/node26/flake.nix`
- Updated Node.js version assertion in CI workflow test from v26.5.1 to v26.6.0

## Implementation Details
The changes maintain consistency across all CI job configurations and Nix flake files, ensuring that all environments use the same Node.js version and nixpkgs commit. The nixpkgs update aligns with the Node.js version upgrade to ensure compatibility.

https://claude.ai/code/session_015qNmRRuZrsaNUEKYdVJ243

PR #1437, opened 2026-08-05
Add o2alexanderfedin to CONTRIBUTORS.md

sergey-shandar:
Added o2alexanderfedin to the contributors list.

PR #1441, opened 2026-08-05
0.42.0

sergey-shandar:


o2alexanderfedin (approved):
Approving.

Checked the things a release PR can quietly get wrong:

- **Version sync** — `package.json` and `package-lock.json` both move to `0.42.0`. Nothing else needed syncing: `deno.json` carries no version field, and `nanvm-lib/Cargo.toml` (`0.1.0`) versions independently of the npm package.
- **CHANGELOG shape** — `## 0.42.0` inserted below an emptied `## Unreleased`, entries preserved in place.
- **Bump size** — minor is right by this file's own stated convention: two `BREAKING CHANGES` entries (#1433, #1428) are in this release, and pre-`1.0` the convention uses a minor bump for those.
- **Completeness** — #1440's Node `26.5.1` -> `26.6.0` bump has no entry, which reads as consistent rather than missed: this changelog records Node changes when they are user-visible (dropping Node 22, further down the file), not for routine CI/nix patch bumps. #1437 and #1439 correctly have none either.

PR #1443, opened 2026-08-05
0.43.0

sergey-shandar:


o2alexanderfedin (approved):
Approving.

**Release mechanics**, checked the same way as 0.42.0:

- **Version sync** — `package.json` and `package-lock.json` both to `0.43.0`. Nothing else carries one: `deno.json` and the root `Cargo.toml` have no version field, and `nanvm-lib/Cargo.toml` versions independently (`0.1.0`, untouched).
- **CHANGELOG** — `## 0.43.0` inserted under an emptied `## Unreleased`, two lines added, entry unchanged from where #1442 left it.
- **Completeness** — `git log a44bd402..main` is exactly one commit, #1442, and it is exactly the one entry. Nothing merged since 0.42.0 is missing.
- **Bump size** — minor for a breaking change is what this file's own convention prescribes pre-`1.0`.

**Verified on the PR head:** `npx tsc --noEmit` clean, `node ./fjs/module.ts t` — **2341 pass, 0 fail**. All CI checks green.

---

Worth recording, since 0.43.0 ships it: #1442 got materially better after I approved it. Four commits landed on that branch afterwards, and one of them — "Make the object type proofs actually check" — fixed something I reviewed and did not catch.

The old idiom was `true as _Predicate`, with `_Predicate` a conditional type resolving to `true` or `false`. I checked the new AGENTS.md §3.4 claim that this proves nothing rather than taking it on trust:

```ts
type E<A, B> = A extends B ? B extends A ? true : false : false
type _Wrong = E<string, number>   // resolves to `false`
const a = true as _Wrong          // compiles, zero errors
```

It compiles clean. So a type-level proof written that way passes whatever its predicate resolved to, and the entry is inert twice over — the runner only invokes functions, so a boolean leaf in a `proof` object is never counted as a test either. `Assert<Equal<…>>` turns a wrong claim into TS2344.

That means the type proofs I signed off on in #1442 were not checking anything, and the ones shipping here are. Good catch, and §3.4 is the right place for it — the rule generalizes well beyond that one file.

PR #1448, opened 2026-08-07
0.43.1

sergey-shandar:


o2alexanderfedin (approved):
Approving. `npx tsc --noEmit` clean, `node ./fjs/module.ts t` — **2354 pass, 0 fail** on the release head.

**Release mechanics:**

- **Version sync** — `package.json` and `package-lock.json` both to `0.43.1`; nothing else carries a version (`deno.json` and the root `Cargo.toml` have no field, `nanvm-lib` versions independently at `0.1.0`, untouched).
- **CHANGELOG** — `## 0.43.1` under an emptied `## Unreleased`, entry unchanged from where #1447 left it.
- **Completeness** — three PRs merged since `0.43.0`, and exactly one of them (#1447) changes shipped code; #1444 and #1446 are `todo/` documents, which this file has consistently not given entries.

**The patch bump is right, and this is the first release where the convention actually bites.** The rule at the top is explicit — `0.Y` is bumped **only** by a release containing `**BREAKING CHANGES:**`, and *"every other release — new features included — is a patch bump."* #1447 adds a schema field and two exports, which under a conventional reading of SemVer would be a minor; under this file's stated pre-1.0 rule it is a patch, because it carries no breaking marker. `0.42.0` and `0.43.0` both contained `BREAKING CHANGES` entries, so they were minor and never exercised the distinction. Worth noting only because the bump *looks* wrong at a glance and is not.

---

One note, carried forward rather than re-argued.

The convention's justification for patch is that *"a patch upgrade is always safe."* #1447's serialization change is the one place that claim is worth watching: a consumer that re-serializes an already-stored revision now gets a different CAS address for the same logical value. #1447's own reasoning — parsed value and field semantics unchanged, old blobs still readable — is why I accepted it there, and I am not reopening it here.

What is new is that this release is the moment it reaches consumers automatically. `fjs-dev/finance` pins `^0.43.0` and writes revisions through Evo (`fjs/server/module.f.js`, `fjs/server/fjs_run/snapshot/module.f.js`), so it picks `0.43.1` up on its next `npm ci` with no PR and no review. Its `.cas` store is project-local and gitignored, so whether any revision has actually been written is a question about users' machines rather than about the repository — which is the same "has the design window closed" question I raised on #1447, now with a concrete delivery date rather than a hypothetical one.

Nothing to change in this PR for it. If the answer turns out to be "records exist", the useful artifact is a line in the `0.43.1` entry saying so, since that is what someone debugging a shifted address will read.

PR #1450, opened 2026-08-07
Expand documentation for compile command and CAS features

sergey-shandar:
## Summary
This PR significantly expands the documentation for FunctionalScript's compiler and CLI, with a focus on explaining the `fjs compile` command and its capabilities. The changes reorganize and enhance the README files to provide clearer guidance for users.

## Key Changes

- **Enhanced main README.md**: 
  - Added note about running CLI with `npx` without installation
  - Moved compiler documentation to a dedicated "Compiling a module" section with concrete examples showing input/output for both JavaScript and JSON formats
  - Explained the difference between JavaScript output (preserves object graph with `const` sharing) and JSON output (tree format with expanded values)
  - Added details about supported language features and limitations
  - Reorganized CLI commands table with improved formatting, added documentation links, and clarified command descriptions
  - Added reference to `fjs help` command

- **Enhanced fjs/README.md**:
  - Updated commands table with documentation column linking to relevant README files
  - Added new "fjs compile" section with usage examples and format selection explanation
  - Clarified that output format is determined by file extension

- **Minor cleanup**:
  - Removed trailing semicolon in example file (`fjs/djs/examples/input.f.ts`) for consistency

## Notable Details

The documentation now clearly explains the compiler's dual-output capability: JavaScript modules that maintain object identity through `const` declarations, versus JSON trees where shared values are expanded. This helps users understand when to use each format based on their needs.

https://claude.ai/code/session_01MbDEXSgXRuNhGDWzHdhqDH


fjs/README.md
```diff
@@ -20,15 +20,27 @@ fjs <command> [args]
...
+| `mcp`    | `m`   | Run an MCP server over stdio exposing the CAS and Evo as tools | [mcp](mcp/README.md) |
+| `ci`     | `i`   | Generate the GitHub Actions CI workflow | [ci](ci/README.md) |
+| `run`    | `r`   | Run a FunctionalScript module as a program | [below](#fjs-run--running-a-module-as-a-program) |
+| `help`   | `h`, `?` | Print available commands | |
+
+## `fjs compile` — compiling a module
+
+```sh
+fjs compile <input> <output>
+fjs c       <input> <output>
```
sergey-shandar:
```suggestion
```
DRY

o2alexanderfedin (commented):
Reviewed by checking out the branch and running the things it documents. The `fjs compile` section is excellent — I ran the example verbatim and the output matches the README byte for byte, both formats. All seven new documentation links resolve. Two problems with the alias→full-name change itself.

### 1. `npx fjs <command>` runs somebody else's package

`README.md:31`:

> or run the CLI without installing it, with `npx fjs <command>`.

`npx` resolves by **package** name, not bin name. `fjs` is a real, unrelated package on npm:

```
$ npm view fjs name version description repository.url
name = 'fjs'
version = '0.5.0'
description = 'Utilities for Functional Javascript'
repository.url = 'git@github.com:idottv/fjs.git'
```

So `npx fjs test` downloads and executes that stranger's package rather than FunctionalScript. Every other `npx` line in the repo gets this right — `AGENTS.md:70`, `fjs/cas/evo/README.md:242`, `fjs/mcp/README.md` all say `npx functionalscript ...`. Suggest the same here.

The same bug already exists at `fjs/README.md:12` (`npx fjs <command> [args]`) — it predates this PR, but since this PR is precisely about how the CLI is invoked in the docs, it seems like the right place to fix both.

### 2. Docs and the CLI now disagree about aliases

The commit is titled *"Use full command names in documentation, drop CLI aliases"*, but the aliases are not dropped. `fjs/module.f.ts` still declares them:

```
names: ['test', 't'] / ['compile', 'c'] / ['cas', 's'] / ['mcp', 'm'] / ['ci', 'i'] / ['run', 'r']
```

and `node ./fjs/module.ts t` still runs the full suite (2354 passing). Meanwhile `fjs/README.md` drops the Alias column entirely, and `README.md` adds "Run `fjs help` for the same list" — but `fjs help` on this branch prints:

```
Available commands:
  test, t     Run the FunctionalScript test suite
  compile, c  Compile a FunctionalScript module to JavaScript
  ...
  help, h, ?  Print this help message
```

That is not the same list. Either drop the aliases in `fjs/module.f.ts` / `fjs/cli/module.f.ts` so the help output matches, or keep documenting them. As it stands `fjs help` is a live counter-example to the README's own claim.

Happy to approve once the `npx` line is corrected; the alias question is your call on which direction to take it.

fjs/README.md
sergey-shandar: Done in 424d19fc — that `fjs c <input> <output>` line is gone, and the same pass removed every other alias from the documentation: the Alias columns in both command tables, `fjs t`/`fjs r`/`fjs i` in prose and examples, and `npx functionalscript m` → `npx functionalscript mcp`.

Two spots intentionally still show `fjs t`: `fjs/ci/README.md` lines 75 and 79, which describe what the generated workflow runs. `fjs/ci/node/module.f.ts:34` and `:40` emit `test({ run: 'fjs t' })`, so spelling it out there would make the docs wrong. Happy to change the generator (2 lines, plus the assertions in `fjs/ci/proof.f.ts:89` and `:138`, a regenerated `ci.yml`, and a CHANGELOG entry) if you want that in this PR.

---

o2alexanderfedin (commented):
Re-checked at `a9377224`. The update since my last review is the merge from `main` — it's clean, no conflict markers, and #1449's new JSDoc/variance section in `AGENTS.md` survived intact. The documentation content itself is unchanged, so both earlier points are still open:

1. `npx fjs <command>` — still at `README.md:31` and `fjs/README.md:12`. `npx` resolves by package name, and `fjs` is a real unrelated package on npm (`fjs@0.5.0`, "Utilities for Functional Javascript"), so this line tells users to run someone else's code. This is the one I'd call blocking; it's a one-word fix to `npx functionalscript <command>`.
2. Aliases — `fjs/module.f.ts` still declares `['test','t']`, `['compile','c']`, etc., and `fjs help` still prints them, while the tables no longer do and `README.md` says `fjs help` gives "the same list".

No new problems. Everything else I verified before still holds on this head.

sergey-shandar:
Both points addressed — thanks, the `npx` one was a real bug.

**1. `npx fjs` (db11844f).** Confirmed independently: `npm view fjs` returns `fjs@0.5.0`, "Utilities for Functional Javascript" (`idottv/fjs`), so `npx fjs test` really would fetch and run that package. `README.md:31` and `fjs/README.md:12` now both say `npx functionalscript <command>`, and `fjs/README.md` states why, so the next person doesn't reintroduce it:

> `npx` resolves the **package** name, not the bin name: `fjs` is an unrelated package on npm, so `npx fjs …` would run somebody else's code.

The remaining `npx fjs js t` in `fjs/emergent_testing/todo/661-sample-repo-ts.md` is left alone — it's an unimplemented issue file describing a sample repo, not user-facing docs.

**2. Aliases (3c6cd4b9).** You were right that the commit title overstated it: the aliases were never dropped, only undocumented. The maintainer's call is to keep them in the CLI, so the READMEs no longer pretend otherwise — the false "Run `fjs help` for the same list" claim is gone, replaced by an explicit note in both files that commands accept short aliases and that `fjs help` prints them, while the documentation uses full names. `fjs/module.f.ts` and `fjs/cli/module.f.ts` are untouched, so `fjs t` keeps working and no version bump is needed.

Head is `3c6cd4b9`, with `main` merged in (#1451). `npx tsc` clean, 2354/2354 proofs pass, and `npm run ci-update` leaves the tree clean.

---

o2alexanderfedin (approved):
Approving at `c3ff1173`. Both points are fixed, and the second one you took further than I suggested.

**`npx`** — now `npx functionalscript <command>` in both places, and `fjs/README.md:16` explains *why* rather than just changing the string:

> `fjs` is a different package on npm, so `npx fjs …` would run somebody else's code.

That note is the part that keeps it from regressing.

**Aliases** — instead of picking one side, you removed the false claim and made the docs state the actual arrangement: aliases exist, `fjs help` prints them, the documentation uses full names throughout. That's more accurate than either option I offered.

**`ci: generate full command names`** — this was the right root-cause fix and it's the one I checked hardest, since it changes a generator rather than prose. Verified:

- `npm run ci-update` regenerates `.github/workflows/ci.yml` and leaves the working tree **clean** — the committed workflow is exactly what the generator produces, so generated CI and docs can't drift apart.
- `npx tsc` clean; `npm test` → `pass: 2354, fail: 0, total: 2354`.
- No `t`-alias remnants left in the generated workflow or `deno.json`.

Catching that `fjs/ci/README.md` was documenting `fjs t` only because it was quoting real generator output — and fixing the generator instead of the quote — is the correct direction.

PR #1454, opened 2026-08-07
TS => MJS. Iteration 3

sergey-shandar:
## Summary

(#1449) started in #1452/#1453. Migrates `fjs/types/array`,
`fjs/types/function`, `fjs/types/nullable`, and `fjs/types/option` from
authored TypeScript (`.f.ts`) to JSDoc-typed JavaScript (`.f.mjs`), and
updates all importers to the new specifiers.

- `fjs/types/array`: fixed-arity type exports (`Array1`..`Array16`,
  `Index1`..`Index16`, `Tuple2`, `Tuple3`) and `isArray2` are replaced by
  the generic `Tuple<N, T>`, `Index<N>`, and `isTuple(n)`
- `fjs/types/function`, `fjs/types/nullable`, `fjs/types/option`: same
  runtime exports, now authored as `.f.mjs` with JSDoc types
- All internal importers (`btree`, etc.) updated to import from the new
  `.f.mjs` paths

Any code importing these modules by explicit `.f.ts` path, or using the
removed `array` type aliases, needs to update.

## Test plan

- [ ] `deno check` passes across the repo
- [ ] existing proofs for `array`, `function`, `nullable`, `option`, and
      their consumers still pass


o2alexanderfedin (commented):
Reviewed at `ef95a3ef`. The generic redesign is the good kind of breaking change and I verified it holds up — details at the bottom. Two things to fix first.

### 1. `KeyOf` silently changed behaviour for variable-length arrays

The old `KeyOf` ended with an explicit fallback:

```ts
    T extends readonly (infer _)[] ? number :
    never
```

so `KeyOf<readonly T[]>` was `number`. The new `KeyOf<T> = Index<T['length']>` loses that, because `Index<number>` collapses to `never`: with `N = number`, `IndexX<number, readonly[]>` hits `0 extends number` on the first step, which is **true**, and returns `never`.

Verified both directions:

```
KeyOf<readonly string[]>  →  main: number   this PR: never
Index<number>             →  never
```

This one is unpleasant because it fails *quietly*. Nothing errors at the definition site; instead any `(index: KeyOf<T>) => …` over a variable-length array becomes uncallable, at the call site, possibly in someone else's code. The repo only uses `KeyOf` on fixed tuples in `btree/find`, so the suite stays green and CI can't catch it.

The PR description and CHANGELOG list the removed fixed-arity exports carefully, but neither mentions this — and it's the one change a downstream consumer can't discover by a rename.

A one-line fix restores it; I checked that it keeps every fixed arity exact:

```js
/**
 * @template {number} N
 * @typedef {number extends N ? number : N extends number ? IndexX<N, readonly[]> : never} Index
 */
```

→ `KeyOf<readonly string[]>` and `Index<number>` are `number` again, while `Index<2>`, `Index<5>`, `Index<16>` and `KeyOf<Tuple<3,string>>` all stay exactly as they are now. (If dropping the fallback is deliberate, then it just needs a CHANGELOG line — but it should be a decision, not a side effect.)

### 2. Debug leftovers

`fjs/types/array/module.f.mjs:43-45`:

```js
/**
 * @typedef {KeyOf<Tuple<2, number> | Tuple<1, number>>} X
 */
```

`fjs/types/btree/find/module.f.ts:47-48`:

```ts
        type N = typeof node
        type X = KeyOf<Tuple<2, number>>
```

None of the three is referenced anywhere. The `X` in the array module is the worse of the two: as we established on #1453, a JSDoc `@typedef` is exported whether you want it or not, so this ships a public type named `X` in the package's API.

Worth noting these won't be caught for you — `noUnusedLocals` is commented out in `tsconfig.json:91`, and `npx tsc` is clean with all three present.

### Verified good

The `ArrayN`/`IndexN` → `Tuple<N,T>`/`Index<N>` collapse is genuinely correct, which is the part I most expected to find a hole in. Using a strict `Equal<A,B>` check, the new generics are **exactly** the old types — not merely assignable:

- `Tuple<N, string>` ≡ `ArrayN<string>` for N = 1, 2, 3, 4, 5, 8, 16
- `Index<N>` ≡ `IndexN` for N = 1, 2, 3, 4, 5, 8, 16
- `KeyOf` unchanged on fixed tuples, including the union case (`KeyOf<Tuple<1,T> | Tuple<2,T>>` → `0 | 1`)

`isTuple(2)` infers `N` as the literal `2` rather than widening to `number`, narrows identically to the old `isArray2`, and is directly assignable to the old signature.

`npm test` → `pass: 2354, fail: 0, total: 2354`; `npx tsc` clean.

Replacing 30-odd hand-written arity aliases with two recursive generics that are provably identical is a real improvement — the two items above are the only things standing between this and an approval.

o2alexanderfedin (approved):
Approving at `fa51799a`. Both items from my last review are fixed, and I re-verified them rather than taking the diff at face value.

**`KeyOf` fallback restored.** `Index` now guards the non-literal case:

```js
@typedef {number extends N ? number : N extends number ? IndexX<N, readonly[]> : never} Index
```

Confirmed `KeyOf<readonly string[]>` and `Index<number>` are `number` again, while every fixed arity stays *exactly* as it was — `Tuple<1|2|5|16, T>`, `Index<2|5|8|16>`, `KeyOf<Tuple<3,T>>`, and the union case all still pass a strict `Equal<A,B>` check. `isTuple(2)` remains assignable to the old `isArray2` signature.

**Debug leftovers gone.** The `X` typedef in `array/module.f.mjs` and `type N` / `type X` in `btree/find/module.f.ts` are all removed.

`npm test` → `pass: 2354, fail: 0, total: 2354`; `npx tsc` clean.

### New in this round: `fjs/types/ts`

The two new commits also migrate `fjs/types/ts`, which wasn't in the version I reviewed before, so I gave it the same treatment. The translation is faithful and the types are genuinely enforced — five deliberately-wrong uses each produce the right diagnostic:

| mutation | diagnostic |
|---|---|
| `printer().tuple('notanarray')` | TS2345 `string` → `readonly string[]` |
| `primitive({})` | TS2345 `{}` not assignable to the primitive union |
| `union([1, 2])` | TS2322 `number` → `string` |
| `printer(false)` | TS2345 `false` → `true` |
| `struct([['k','v','extra']])` | TS2322 `[string,string,string]` → `readonly [string,string]` |

`Equal<A,B>` still discriminates correctly (`Equal<string,string>` is `true`, `Equal<string,number>` is `false`), and `Printer` survives as a usable exported type.

### One thing to add before merge

`fjs/types/ts` is missing from the CHANGELOG entry, which currently names only `array`, `function`, `nullable`, and `option`. It belongs there — the breakage is identical and consumer-visible. Running the emit on `main` shows why:

```
already-migrated (function):   module.f.mjs  module.f.d.mts
still-TypeScript (ts):         module.f.js   module.f.d.ts   module.f.ts
```

Migration removes `module.f.js` from the published package, so anyone importing `functionalscript/fjs/types/ts/module.f.js` breaks exactly as they would for the four listed modules. One line to the existing entry covers it.

### Nit

`printer = (mut = undefined)` reintroduces the no-op default you removed from `assertNotNullish` in #1452 — `mut` is already `undefined` when omitted. Harmless, just inconsistent with the convention you settled on.

Neither is a correctness problem, so I don't want to hold the PR on them — the generic redesign is provably equivalent and the migration is clean.

PR #1456, opened 2026-08-08
TS => MJS. Iteration 4

sergey-shandar:
## Summary

(#1449) started in #1452/#1453/#1454. Migrates
`fjs/types/function/compare` from authored TypeScript (`.f.ts`) to
JSDoc-typed JavaScript (`.f.mjs`), and updates all importers to the new
specifier.

- `fjs/types/function/compare`: same runtime exports (`index3`, `index5`,
  `cmp`, `min`, `max`, `bsearch`) and same types (`Sign`, `Compare<T>`,
  `Cmp<T>`, `Cmp1`, `Cmp2<A, B>`), now authored as `.f.mjs` with JSDoc
  types
- Importers updated to `./module.f.mjs`: `fjs/asn.1`, `fjs/types/bigint`,
  `fjs/types/bit_vec`, `fjs/types/btree/{find,remove,set}`,
  `fjs/types/number`, `fjs/types/ordered_map`, `fjs/types/range_map`,
  `fjs/types/sorted_list`, `fjs/types/sorted_set`, `fjs/types/string`,
  and the `compare` proof

Any code importing this module by explicit `.f.ts` path needs to update.

## Test plan

- [x] `npm test` (`tsc` + full proof suite): 2354 pass, 0 fail


o2alexanderfedin (approved):
Approving at `0d28a42f`. Clean iteration, and the CHANGELOG entry is there this time.

All five exported types (`Sign`, `Compare`, `Cmp`, `Cmp1`, `Cmp2`) were already `export type` in the `.ts`, so the typedef-visibility issue from #1453/#1454 doesn't arise here — nothing goes from private to public.

Verified rather than eyeballed:

- Every type is unchanged under a strict `Equal<A,B>` check, including the conditional `Cmp2<A,B>` (`Cmp2<number,number>` → `number`, `Cmp2<string,number>` → `never`).
- The self-referential constraint `<A extends Cmp1>(a: A) => <B extends Cmp2<A, B>>(b: B) => …` survives translation, which was the part most likely to break. The rejection the docstring promises still happens: `min(1)("a")`, `max(1)("a")`, `cmp(1)("a")`, `cmp(1)(2n)` all fail with `TS2345: … not assignable to parameter of type 'never'`, and `cmp({})` fails against `Cmp1`.
- `index3` / `index5` still return `Index<3>` / `Index<5>` — `index3(...)` assigned to `3` correctly errors with `Type '0 | 1 | 2' is not assignable to type '3'`.
- `npm test` → `pass: 2354, fail: 0, total: 2354`; `npx tsc` clean.

### Follow-up: stale `.f.ts` paths are accumulating across the series

Not from this PR, and not blocking it — but it's now big enough to be worth one sweep. On `main`, **30 lines across 13 files** still reference the old `.f.ts` path of a module that has already migrated:

```
AGENTS.md                                   2
fjs/types/object/todo/structurally-same.md  2
fjs/types/ordered_map/todo/at-nullable-map.md 2
… 10 more todo files                        1 each
```

The two in `AGENTS.md` are the ones I'd fix first, since agents act on that file:

- `AGENTS.md:205` tells you to import `Assert` from `fjs/asserts/module.f.ts` and `Equal` from `fjs/types/ts/module.f.ts`. Both are now `.f.mjs`, so following that snippet literally produces a broken import — and it's exactly the snippet someone copies when writing a type-level proof.
- `AGENTS.md:183` uses `fjs/asserts/module.f.ts` as its example of *an unmigrated helper*, which is now the opposite of true — `asserts` migrated first, in #1452.

Each iteration adds a few more, so a single `grep`-and-fix pass at the end of stage 1 would probably be cheaper than doing it per-PR. Might be worth a line in the migration todo so it isn't forgotten.

sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1463, opened 2026-08-09
TS => MJS: basen/base128

sergey-shandar:
Continues the TypeScript-to-`.mjs` migration.

- Renames `fjs/basen/base128/module.f.ts` to `module.f.mjs`, converting type annotations to JSDoc (`@param`/`@returns`) and the `Vec` type import to `@import`.
- Updates importers (`fjs/asn.1/module.f.ts`, `fjs/basen/base128/proof.f.ts`) to the new `.mjs` path.

No behavior changes.


o2alexanderfedin (approved):
Approving. Small, clean migration — proper rename, CHANGELOG entry present, two importers updated.

**Signatures are preserved exactly**, checked with a strict `Equal<A,B>` against the pre-migration types rather than by reading:

```
typeof encode === (uint: bigint) => Vec                    ✓
typeof decode === (v: Vec) => readonly[bigint, Vec]        ✓
```

and they're genuinely enforced — `encode(5)`, `encode(5n)` assigned to `string`, and `decode(5n)` each produce the right specific error, so nothing collapsed to `any`.

`npx tsc` clean, `npm test` → `pass: 2354, fail: 0, total: 2354`, CI 19/19.

No typedefs in this module, so the new `_` convention has nothing to apply to here — worth noting only because this is the first migration since #1462 landed it.

One small thing done well: keeping `@param` / `@returns` rather than collapsing to a single `@type` preserves the parameter names and descriptions the original TypeScript JSDoc carried. A `@type` annotation would have type-checked identically but dropped "The bigint to encode" and "The MSB Base128 vector to decode" from the emitted declarations.

PR #1464, opened 2026-08-09
TS => MJS: asn.1

sergey-shandar:
Continues the stage-1 TypeScript-to-`.mjs` migration.

- Renames `fjs/asn.1/module.f.ts` to `module.f.mjs`, converting type annotations
  to JSDoc (`@type`, `@typedef`, `@template`) and type imports to `@import`.
- Adds doc comments to the exported tag constants, encoders/decoders, and
  typedefs that previously had none.
- Updates `fjs/asn.1/proof.f.ts` to import from the new `.f.mjs` path.

The exported API is unchanged; implementation-only typedefs (`_ClassPc`,
`_ParsedTag`, `_Tag`, `_Round8`) keep their private `_`-prefixed names.


sergey-shandar:
@codex review

sergey-shandar:
@codex review

o2alexanderfedin (commented):
Two small things to fix; the migration work itself is clean.

### 1. Two module migrations are missing from the CHANGELOG

The entry covers the `base_n` → `basen` move and `fjs/asn.1`, but this PR also migrates two more public modules:

```
fjs/basen/base64/module.f.ts   -> module.f.mjs     (2 exports:  encode, decode)
fjs/basen/cbase32/module.f.ts  -> module.f.mjs     (4 exports:  cBase32ToVec, cBase32ToVec5x,
                                                                vec5xToCBase32, vecToCBase32)
```

Both are consumer-visible and both have cross-module importers already — `cbase32` alone is imported from `fjs/cas`, `fjs/cas/evo` and `fjs/cas/cli`. Their importers face exactly the same `.f.mjs` specifier break the entry describes for `asn.1`, so they belong in the same list.

(The proof-file renames — `asn.1`, `asserts`, `base128`, `base64` — correctly need no entry; those aren't public API.)

### 2. The `base_n` → `basen` move breaks a live reference in `AGENTS.md`

```
AGENTS.md:746
Example (`fjs/base_n/module.f.ts`): `chunkList(msb)` depends on neither `bits`
```

That path is valid on `main` and doesn't exist after this PR, so unlike the stale-reference cases I've raised before, this one is a regression the PR introduces rather than something it inherited. It's a live instruction example, so an agent following it looks for a file that isn't there. Should be `fjs/basen/module.f.ts`.

The historical `base_n` mentions further down `CHANGELOG.md` (lines 303, 328, 441) are correct as-is — those describe past releases and shouldn't be rewritten.

### Verified good

- **Every export survives all three migrations.** I checked by importing all 29 `asn.1` names, both `base64` names and all four `cbase32` names from the new `.mjs` and compiling — clean. (Worth noting because a pattern-scan of the source suggested `Record` had gone missing from `asn.1`; it hadn't. The typedef form just doesn't match a naive grep, which is why I stopped trusting that method for export comparisons.)
- **The `_` convention is respected.** `asn.1` already used `_ClassPc`, `_ParsedTag`, `_Tag` and `_Round8` in TypeScript, and the migration preserves all four prefixed — no private type leaks into the public surface.
- `npx tsc` clean, `npm test` → `pass: 2354, fail: 0, total: 2354`. CI was still mid-run when I looked (14 pending), which I'd expect to settle.

Happy to approve once the CHANGELOG covers the two modules and `AGENTS.md:746` points at the new path.

sergey-shandar:
@codex review

sergey-shandar:
@codex review

o2alexanderfedin (commented):
Both earlier points are fixed — the CHANGELOG now covers `basen`, `base64` and `cbase32`, and `AGENTS.md:746` points at `fjs/basen/module.f.mjs`, which exists. Thanks.

The scope grew in the seven commits since, and two things haven't kept pace with it.

### 1. The CHANGELOG is behind again — two more migrated modules

The PR now also migrates:

```
fjs/text/code_point/module.f.ts -> .f.mjs    8 exports,  2 external importers
fjs/types/range/module.f.ts     -> .f.mjs    2 exports,  8 external importers
```

Neither appears anywhere in the `Unreleased` section (zero hits for `code_point` and `types/range`). Both were public `.f.ts` on `main`, so their importers hit the same specifier break already documented for the other five.

This is the same thing I flagged last round, recurring because the scope expanded after the entry was written. Since these PRs keep growing mid-review, it might be worth deriving the list mechanically at the end — every `module.f.ts -> module.f.mjs` rename in the diff is exactly one CHANGELOG line — rather than maintaining it by hand as modules are added.

### 2. Ten backticked paths now point at files that no longer exist

The renames left prose and comment references behind. Comparing against `main` to isolate what this PR causes (main has its own set, mostly *proposed* modules, so only the delta is meaningful):

```
fjs/asn.1/module.f.ts          fjs/base_n/module.f.ts        fjs/basen/cbase32/module.f.ts
fjs/asserts/proof.f.ts         fjs/base_n/proof.f.ts         fjs/basen/cbase32/proof.f.ts
fjs/basen/base64/module.f.ts   fjs/basen/base64/proof.f.ts
fjs/text/code_point/module.f.ts                              fjs/types/range/module.f.ts
```

Spread across about a dozen files. Three are in **source**, not just todo docs:

```
fjs/crypto/sha2/proof.f.ts        -> `fjs/basen/base64/proof.f.ts`
fjs/mcp/proof.f.ts                -> `fjs/basen/base64/proof.f.ts`
fjs/basen/base64/proof.f.mjs:125  -> `fjs/base_n/module.f.ts`
```

That last one is the one I'd fix regardless of the rest: it's a file this PR itself migrated, and its comment still cites the pre-move path.

None of these break the build — they're comments and prose, which is why `tsc` is clean and CI is green. Purely a documentation-accuracy matter, and I only raise it as a batch because I named a single instance last time and that's what got fixed; the sweep is the useful version.

### Verified good

- Every export survives all six migrations, checked by importing each name and compiling rather than pattern-matching the source.
- `fjs/types/range` is a proper add+delete pair and the rest are clean renames — no add-without-delete.
- `npx tsc` clean, `npm test` → `pass: 2354, fail: 0, total: 2354`, CI now 19/19.

sergey-shandar:
@codex review

sergey-shandar:
@codex review

o2alexanderfedin (approved):
Approving. Both points are resolved, and the second one you handled more carefully than I'd asked.

**CHANGELOG now covers all six migrated modules** — `fjs/text/code_point` and `fjs/types/range` are in, alongside `asn.1`, `basen`, `base64`, `cbase32` and the move.

**Stale paths: 8 of the 10 rewritten, and the 2 left are correctly left.** Those are on the completed checkbox in `fjs/todo/group-fs-subdirectories-by-concern.md:158`:

> - [x] Move the shared codec factory `fjs/base_n/module.f.ts` to `fjs/basen/module.f.mjs` **together with `fjs/base_n/proof.f.ts`** …

Naming the pre-move path is exactly right there — rewriting it would turn the record of the move into "move X to X". I'd flagged those two mechanically in my batch; you were right to treat them as a description of history rather than a stale pointer.

Net effect across the tree: broken backticked paths go **47 on `main` → 46 here**, so the PR leaves the docs more accurate than it found them despite renaming six modules.

### Verified

- `npx tsc` clean, `npm test` → `pass: 2354, fail: 0, total: 2354`, CI 19/19.
- The `AGENTS.md` anchor you added resolves — `#### Private JSDoc typedefs` exists at `fjs/fsc/README.md:61`.

### Two notes, neither blocking

**AGENTS.md and the typedef rule.** #1461 removed this policy from `AGENTS.md` as a duplicate; this PR adds 11 lines of it back. The difference is real — it's now a summary plus a working link rather than a full restatement — so it may well be the right balance, since `AGENTS.md` does need enough for an agent to know the rule exists. Just worth a deliberate look given you made the opposite call three PRs ago; the sentence "renaming or removing a `_`-prefixed alias is not by itself a breaking change" is the one bit that's now stated in two places and could drift.

**Overlap with #1465.** I opened a PR sweeping the same class of stale `.f.ts` references for modules already on `main`. It touches `AGENTS.md` and `fjs/basen/cbase32/todo/178.md`, which this PR also touches. I checked: my branch merges onto yours **cleanly**, no conflicts. Whichever lands first, the other needs no manual fixup — but if you'd rather not have two docs sweeps in flight, mine is the one to drop or defer.

PR #1466, opened 2026-08-09
Add test for DFA with missing states fallback behavior

sergey-shandar:
## Summary
Added a test case to verify that the `run` function correctly handles DFAs with missing states by falling back to an empty transition table.

## Key Changes
- Added a new test case in the proof suite that validates `run` accepts any `Dfa` (a `StringMap`), not only those built by the `dfa()` constructor
- The test verifies that when a state is absent from the DFA map, it falls back to an empty transition table, resulting in no transitions being available
- Test case passes an empty object `{}` as the DFA and verifies it produces an empty array `[]` when processing input

## Implementation Details
- The test uses the existing test infrastructure (`stringifyIdentity`, `toArray`, `run`, `stringToList`, `assertEq`)
- Demonstrates that the `run` function is flexible enough to work with incomplete DFA definitions, which is useful for partial or dynamically constructed state machines

https://claude.ai/code/session_017WMju3jw7xPWz59gktNGDm

### DIFF
--- a/fjs/fsm/proof.f.ts
+++ b/fjs/fsm/proof.f.ts
@@ -131,6 +131,16 @@ export const proof = {
             ]
             const expectedResult = stringifyIdentity(expectedOutput)
             assertEq(result, expectedResult)
+        },
+        () => {
+            // `run` accepts any `Dfa` (a `StringMap`, so entries may be
+            // missing), not only one built by `dfa()`. A state absent from
+            // the map falls back to the empty transition table.
+            const result = stringifyIdentity(toArray(run({})(stringToList('a'))))
+
+            const expectedOutput = ['[]']
+            const expectedResult = stringifyIdentity(expectedOutput)
+            assertEq(result, expectedResult)
         }
     ]
 }


o2alexanderfedin (approved):
Approving. I ran this against the repo's own standard — *"a proof is not known to work until you have watched it fail"* — and it holds.

The branch under test is the fallback at `fjs/fsm/module.f.ts:94`:

```ts
= dfa => input => s => get(dfa[s] ?? [])(input)
```

I mutated it so a missing state yields `initialStateStringify` instead of the empty table, and ran the full suite on both sides:

| | result |
|---|---|
| `main` + mutation | `pass: 2354, fail: 0` — **not caught** |
| this PR + same mutation | `pass: 2354, fail: 1` — **caught** |
| this PR, unmutated | `pass: 2355, fail: 0` |

So the `?? []` fallback was genuinely uncovered before, and this test is what closes it. That's the meaningful check for a test-only PR — a new case that passes proves nothing on its own, and this one earns its place.

Two things I liked:

The comment says *why* the case exists — that `run` accepts any `Dfa` (a `StringMap`, so entries may be missing), not only one built by `dfa()`. That's the actual contract being pinned, and it's the part a reader six months from now needs; without it, `run({})` looks like an arbitrary edge case rather than a statement about the public type.

The expectation is independent of the code under test — `['[]']` is written out literally rather than derived from `run` or from `dfa()`. That matters here specifically: the four defects listed in `AGENTS.md` were all proofs whose expected side was computed from the thing being tested, and all four stayed green. This one avoids that shape.

CI 19/19.

PR #1465, opened 2026-08-09
docs: point prose references at migrated .f.mjs modules

o2alexanderfedin:
## Problem

Eleven modules have migrated from authored TypeScript to JSDoc-typed JavaScript, but **55 prose and comment references across 22 files** still cite their pre-migration `.f.ts` paths.

None break the build — they're documentation and comments, not imports, which is why `tsc` and CI have stayed green — but each sends a reader to a file that no longer exists. `AGENTS.md` was among them, so an agent following its type-proof example got a broken import path.

This is one pass of the sweep tracked in [`todo/migrate-typescript-to-mjs.md`](../blob/main/todo/migrate-typescript-to-mjs.md), **not its completion** — it will need repeating as further modules migrate.

## What it changes

Rewrites `X/module.f.ts` → `X/module.f.mjs` for the eleven migrated modules:

```
fjs/asserts          fjs/types/array     fjs/types/function   fjs/types/result
fjs/basen/base128    fjs/types/bigint    fjs/types/list       fjs/types/ts
fjs/common/monoid    fjs/types/bit_vec   fjs/types/nullable
```

## What it deliberately does not change

The rewrite is gated on a path being **provably** stale — the cited `.f.ts` is absent *and* the sibling `.f.mjs` exists. That test excludes two categories that a blunter search would have damaged:

- **Proposed modules.** `main` cites ~35 paths that don't exist because a TODO proposes creating them (`fjs/media/cbor/…`, `fjs/bnf/byte/…`, `fjs/media/json/common/…`). Neither file exists, so they're untouched.
- **`fjs/types/rtti/ts/module.f.ts`** — a *different* module that has not migrated. An earlier basename-matching attempt of mine collapsed `types/ts` and `types/rtti/ts` together and would have corrupted this; matching full paths avoids it.

`CHANGELOG.md` is untouched. Its historical entries describe past releases under the old `fs/` layout and should keep naming paths as they were.

## Verification

- `npx tsc` clean; `npm test` → `pass: 2354, fail: 0, total: 2354`.
- Documentation only — no source file changed.
- Dangling relative markdown links: **49 → 45**. Four fixed, **zero introduced** (verified by diffing the full link sweep against `main` rather than counting).
- No remaining `.f.ts` citation of any migrated module.

PR #1467, opened 2026-08-09
Update test command and documentation references

sergey-shandar:
## Summary
Updated documentation to reflect current testing practices and reorganized design documentation references for the nanvm-lib project.

## Key Changes
- **docker/README.md**: Updated the Bun test command from `bun ./dev/test/module.ts` to `bun test` and added a reference to AGENTS.md for the complete list of equivalent test suite commands
- **nanvm-lib/README.md**: Updated design documentation references to point to the new `todo/fjs-nanvm-integration.md` file and the `nanvm-lib/todo/` directory instead of the legacy `issues/082-nanvm.md`

## Details
These changes improve documentation accuracy by:
1. Simplifying the Bun test command to use the standard `bun test` convention
2. Directing users to comprehensive testing documentation in AGENTS.md
3. Updating nanvm-lib to reference current design documentation locations in the todo directory structure

https://claude.ai/code/session_01SLtc9FVEwkAMjuJa3W4pb7


o2alexanderfedin (approved):
Approving. Both changes fix documentation that was actively wrong, not just stylistically dated — I checked each rather than taking the description at face value.

**`docker/README.md` — the old bun command was broken.** `./dev/test/module.ts` does not exist in the tree, so anyone following that line got a failure. `bun test` is right and matches what `AGENTS.md:68` already lists as Bun's native runner, so the docker doc now agrees with the canonical table instead of carrying its own drifting copy.

The added pointer resolves too: `../AGENTS.md` exists from `docker/`, and the anchor `#14-ways-to-run-the-functionalscript-test-suite` matches the real heading `### 1.4 Ways to run the FunctionalScript test suite` at `AGENTS.md:54`. Linking to §1.4 rather than duplicating the list is the right call — that table has changed twice recently (the alias→full-name pass in #1450), and a second copy here would have drifted again.

**`nanvm-lib/README.md` — this was a dangling link.** `issues/082-nanvm.md` doesn't exist; the `issues/` directory is gone. Both replacements are real: `todo/fjs-nanvm-integration.md` and `nanvm-lib/todo/`. It's one of the entries in the broken-link inventory I catalogued while preparing #1465, so this closes it at the source.

Documentation only, CI 19/19. No overlap with #1465 — different files, so no conflict either way.

PR #1468, opened 2026-08-09
TS => MJS: text/utf16

sergey-shandar:
## Summary

(#1449). Migrates `fjs/text/utf16` from authored TypeScript (`.f.ts`) to
JSDoc-typed JavaScript (`.f.mjs`), and updates all importers
(`fjs/bnf/data`, `fjs/bnf/descent`, `fjs/bnf/ll1`, `fjs/bnf/module`,
`fjs/djs/parser`, `fjs/djs/tokenizer`, `fjs/djs/transpiler`,
`fjs/effects/node`, `fjs/fsm`, `fjs/js/tokenizer`, `fjs/media/html`,
`fjs/media/json`, `fjs/text`, `fjs/text/utf8`) to the new `.f.mjs`
specifier.

- `fjs/text/utf16`: same runtime exports (`fromCodePointList`,
  `toCodePointList`, `stringToList`, `stringToCodePointList`,
  `listToString`, `codePointListToString`, `codePointToString`), now
  authored as `.f.mjs` with JSDoc types; `U16`/`CodePoint` type-only
  exports become JSDoc typedefs
- `fjs/bnf/data`: implementation-only typedefs (`TerminalRange`,
  `EmptyTagMap`, `FRuleMap`, `NewRule`) renamed to their private
  `_`-prefixed forms, matching the convention from #1462

Any code importing `fjs/text/utf16` by explicit `.f.ts` path needs to
update to `.f.mjs`.

## Test plan

- [ ] existing proofs for `utf16` and its consumers still pass


sergey-shandar:
@codex review

o2alexanderfedin (approved):
Approving. The migration is clean and the convention is applied correctly. One wording point on the CHANGELOG.

**`fjs/text/utf16` verified:** all nine exports on `main` (`CodePoint`, `U16`, `codePointListToString`, `codePointToString`, `fromCodePointList`, `listToString`, `stringToCodePointList`, `stringToList`, `toCodePointList`) import cleanly from the new `.mjs`. Checked by compiling the imports rather than pattern-matching the source.

**The `_` convention is applied exactly right here.** `Utf16State` was the one unexported type in the `.ts`, and it's now `_Utf16State`; `CodePoint` and `U16` keep public names. That's the first migration since #1462 where the rule actually had something to bite on, and it landed without prompting.

`npx tsc` clean, `npm test` → `pass: 2355, fail: 0, total: 2355`.

### The `fjs/bnf/data` half isn't a breaking change

The entry files both halves under `**BREAKING CHANGES:**`, but the second one can't break anyone. `EmptyTagMap`, `FRuleMap` and `NewRule` are declared without `export` in `module.f.ts`, and that file is still authored TypeScript — so they were never reachable. Importing them from `main` fails:

```
TS2724  '"./module.f.ts"' has no exported member named 'EmptyTagMap'
TS2459  Module '"./module.f.ts"' declares 'FRuleMap' locally, but it is not exported
TS2459  Module '"./module.f.ts"' declares 'NewRule' locally, but it is not exported
```

The `_` prefix exists because a JSDoc `@typedef` is exported whether you want it or not; TypeScript's `type` already isn't. So prefixing a non-exported `.f.ts` type changes nothing observable today — it's pre-work for when `bnf/data` migrates, which is a fine reason to do it, just not a breaking one.

Worth separating because #1461 drew this line deliberately: renaming a private alias is only breaking when it alters the expanded contract of a public declaration. Here there's no public declaration involved at all. Marking non-breaking changes as breaking has a cost in the other direction — it trains readers to skim the label, and this repo's entries are otherwise precise enough that the label carries weight.

The utf16 half genuinely is breaking, so the entry needs to stay — just worth splitting the `bnf/data` rename into its own non-breaking line.

PR #1469, opened 2026-08-09
TS => MJS: types/btree/types

sergey-shandar:
## Summary

(#1449). Migrates `fjs/types/btree/types` from authored TypeScript
(`.f.ts`) to JSDoc-typed JavaScript (`.f.mjs`), and updates all importers
(`fjs/types/btree`, `fjs/types/btree/find`, `fjs/types/btree/remove`,
`fjs/types/btree/set`, `fjs/types/ordered_map`, `fjs/types/string_set`,
plus their proofs) to the new `.f.mjs` specifier.

- `fjs/types/btree/types`: same public type vocabulary (`Leaf1`, `Leaf2`,
  `Branch3`, `Branch5`, `TNode`, `Tree`, `Branch1`, `Branch7`) and the same
  `collapseRoot` runtime export, now authored as `.f.mjs` with JSDoc
  `@typedef`/`@template` types — none of the type names needed the private
  `_`-prefix since all are consumed cross-module

Any code importing `fjs/types/btree/types` by explicit `.f.ts` path needs
to update to `.f.mjs`.

## Test plan

- [x] `npm test` — 2355/2355 proofs pass


o2alexanderfedin (approved):
Approving. Clean single-module migration, nothing to flag.

**All nine exports preserved** — `collapseRoot` plus the eight generic types (`Tree`, `TNode`, `Leaf1`, `Leaf2`, `Branch1`, `Branch3`, `Branch5`, `Branch7`). Verified by importing every name from the new `.mjs` and instantiating each generic at `<1>`, so the type parameters carry over too, not just the names.

**The `_` convention has nothing to apply to here** — the `.ts` had no unexported types, so every typedef correctly keeps its public name. Worth stating explicitly since the last two migrations did have private types to prefix; the absence here is right, not an oversight.

Proper add + delete pair, CHANGELOG entry present and accurate — this one genuinely is breaking, since all nine names are public and importers must move to the `.f.mjs` specifier.

`npx tsc` clean, `npm test` → `pass: 2355, fail: 0, total: 2355`. CI was still mid-run when I checked (17 pending), which I'd expect to settle given the local suite is green.

PR #1470, opened 2026-08-09
TS => MJS: types/btree/find

sergey-shandar:
## Summary

(#1449). Migrates `fjs/types/btree/find` from authored TypeScript
(`.f.ts`) to JSDoc-typed JavaScript (`.f.mjs`), and updates all importers
(`fjs/types/btree`, `fjs/types/btree/remove`, `fjs/types/btree/set`,
`fjs/types/ordered_map`, `fjs/types/string_set`, plus their proofs) to
the new `.f.mjs` specifier.

- `fjs/types/btree/find`: same runtime exports (`find`, `isFound`,
  `value`) and public type vocabulary (`First`, `PathItem`, `Path`,
  `Result` — consumed by `btree/remove` and `btree/set`), now authored as
  `.f.mjs` with JSDoc `@typedef`/`@template` types
- Implementation-only typedefs (`FirstLeaf1`, `FirstBranch3`,
  `FirstLeaf2`, `FirstBranch5`, `PathItem3`, `PathItem5` — only ever used
  to build `First`/`PathItem`, never consumed elsewhere) are renamed to
  their private `_`-prefixed forms, matching the convention from #1462
- The three `as` casts in the original module become inline
  `/** @type {X} */ (expr)` casts; the generic type parameters that don't
  carry across a JSDoc function boundary use the
  `@typedef {typeof param extends Generic<infer T> ? T : never} T`
  recovery idiom already established in `fjs/types/list`'s `equal`

Any code importing `fjs/types/btree/find` by explicit `.f.ts` path, or
relying on the now-private type names, needs to update.

## Test plan

- [x] `npm test` — 2355/2355 proofs pass


o2alexanderfedin (approved):
Approving. Clean migration, and the `BREAKING CHANGES` label is correct here — worth saying explicitly, because it's the opposite call from #1468 and both are right.

There, the renamed types were declared without `export` in a still-TypeScript file, so nothing could observe the rename. **Here all six genuinely were public:**

```ts
export type FirstLeaf1<T> = readonly[Index<3>, Leaf1<T>]
export type FirstBranch3<T> = readonly[1, Branch3<T>]
export type FirstLeaf2<T> = readonly[Index<5>, Leaf2<T>]
export type FirstBranch5<T> = readonly[1|3, Branch5<T>]
export type PathItem3<T> = readonly[0|2, Branch3<T>]
export type PathItem5<T> = readonly[0|2|4, Branch5<T>]
```

so demoting them to `_`-prefixed removes names a consumer could have imported. Breaking, and labelled as such.

**The demotion itself is justified.** I checked whether anything outside `btree/find` depends on them, since a `_`-prefixed name that another module imports would contradict the convention. Nothing does — the only hits in `fjs/types/btree/set/module.f.ts` are lines 52-56, a commented-out block illustrating the type shapes, not real imports. `btree/set` imports only `find`, `First`, `PathItem` and `Result`, all of which stay public.

That's the right cut: `First` remains the public union, and the four constituent shapes plus the two `PathItem` variants become private detail. A consumer that needs to name a branch of `First` can still narrow on it structurally.

**Retained exports verified** — `find`, `isFound`, `value`, `First`, `Path`, `PathItem`, `Result` all import cleanly from the new `.mjs`, with each generic instantiated at `<1>` so the type parameters are checked too.

Proper add + delete pair. `npx tsc` clean, `npm test` → `pass: 2355, fail: 0, total: 2355`, CI 19/19.

Very minor, not worth a commit on its own: the commented block at `btree/set/module.f.ts:52-56` still transcribes `First` as `FirstLeaf1<T> | …`, which is now `_FirstLeaf1<T> | …`. It reads as an explanation of the shape rather than a pointer to importable names, so it's still accurate as prose.

PR #1471, opened 2026-08-10
TS => MJS: types/btree/set

sergey-shandar:
Migrates `fjs/types/btree/set` from authored TypeScript (`.f.ts`) to JSDoc-typed JavaScript (`.f.mjs`) under the stage-1 TypeScript-to-mjs migration.

- Renames `module.f.ts` -> `module.f.mjs` and updates the importers (`fjs/types/btree/proof.f.ts`, `fjs/types/btree/find/proof.f.ts`, `fjs/types/btree/remove/proof.f.ts`, `fjs/types/btree/set/proof.f.ts`, `fjs/types/ordered_map/module.f.ts`, `fjs/types/string_set/module.f.ts`).
- Converts TypeScript type syntax to JSDoc: value imports plus `@import` tags, `@type` annotations on `b57`, `reduceOp`, `nodeSet`, and `set`.
- Renames the implementation-only `Branch1To3` alias to its private `_Branch1To3` form, per the private JSDoc typedef convention.
- Recovers the type parameter inside `nodeSet` with the `typeof c extends Compare<infer T>` pattern already used by `fjs/types/btree/find`.
- Refreshes stale comments that quoted `Index2` / `Index3` / `Index5` and TypeScript `export type` spellings removed by the earlier `find` and `types` migrations.

Runtime semantics are unchanged, including both TSGO `throw ''unreachable''` workarounds. The emitted declaration keeps the public contract identical:

```ts
export declare const set: <T>(c: Compare<T>) => (f: (value: T | null) => T) => (tree: Tree<T>) => TNode<T>;
```

`proof.f.ts` stays TypeScript for now: it still imports unmigrated `.f.ts` dependencies.

## Verification

- `npx tsc` - exit 0
- `npm start test` - 2355 pass, 0 fail
- `npm run update` - no generated-file changes


sergey-shandar:
@codex review

o2alexanderfedin (commented):
Migration itself checks out — one stale doc citation that this PR invalidates.

## Finding: `fjs/types/todo/193.md` now cites a file that no longer exists

`193.md` quotes `set`'s fold engine to motivate the shared-`Path` refactor. On the base commit that citation was exact — `const reduceOp` really was at `module.f.ts:17`, and lines 13–14 reproduce the base source verbatim. After this PR all four anchors are wrong:

- **:12** — `// fjs/types/btree/set/module.f.ts:17` → the path is now `module.f.mjs`, and `reduceOp` moved to **:25**.
- **:13–14** — the quoted TypeScript signature (`const reduceOp\n    : <T>(i: PathItem<T>) => (a: Branch1To3<T>) => Branch1To3<T>`) is the syntax this PR removes; the file now reads `/** @type {<T>(i: PathItem<T>) => (a: _Branch1To3<T>) => _Branch1To3<T>} */`.
- **:71** — prose "`set` threads `Branch1To3<T>`" → the type is now spelled `_Branch1To3<T>`.

Same class of thing this PR already fixes *inside* the module (the `Index2`/`Index3`/`Index5` comment refresh) — it just stops at the module boundary. Consistent with #1464, where the migration also carried the doc-reference updates.

Unrelated and pre-existing, so out of scope here: the `// :107` / `// :108` anchors further down that snippet were already stale on the base (`collapseRoot(reduceBranch(f())(tail))` sits at `:110` there, and the quoted `r.length === 1 ? r[0] : r` shape predates the `collapseRoot` extraction).

## Everything else verified clean

- **Runtime semantics unchanged.** Diffing the base `.f.ts` against the new `.f.mjs` with type syntax and comments stripped leaves only the import specifiers — every expression is character-identical.
- **Public contract identical.** Emitted declaration is `export declare const set: <T>(c: Compare<T>) => (f: (value: T | null) => T) => (tree: Tree<T>) => TNode<T>`, matching the previous annotation exactly.
- **All importers moved.** No `btree/set/module.f.ts` specifier remains anywhere in the tree — source, config, or generated. The three value importers plus the four `proof.f.ts` files are the complete set.
- **`_Branch1To3` naming is right.** It was a non-exported TS `type` and JSDoc `@typedef` exports unconditionally, so the `_` prefix is exactly the #1461 convention.
- **The refreshed `First<T>` comment is accurate** against the current `find/module.f.mjs` — `[Index<3>, Leaf1]` / `[1, Branch3]` / `[Index<5>, Leaf2]` / `[1|3, Branch5]` match `_FirstLeaf1`…`_FirstBranch5` at :16–:36.
- **The `typeof c extends Compare<infer T>` recovery is load-bearing, not vacuous.** Changed `/** @type {Result<T>} */` to `Result<string>` → `TS2322: Type 'Result<T>' is not assignable to type 'Result<string>'`, so the recovered `T` really is the caller's type parameter and not `never`/`any`. Two more mutations confirm the annotations still constrain the implementation: dropping the array wrapper in `set` (`f(null)` for `[f(null)]`) → `TS2322: Type 'T | TNode<T>' is not assignable to type 'TNode<T>'`; dropping it in `b57` → `TS2322: … not assignable to type '_Branch1To3<T>'`. Baseline is 0 errors, so each of those is genuinely caught rather than lost in noise.

One thing worth knowing for future migration PRs, unrelated to this one: if a worktree has leftover `.d.mts` from `npm run prepack`, those shadow the `.f.mjs` sources and drop them out of the tsc program entirely — `tsc` then reports 0 errors on a file with a blatant `/** @type {string} */ const x = 5`. They're gitignored so CI is unaffected, but locally "npx tsc — exit 0" can be vacuous for exactly the files a migration PR changes. Deleting stray `.d.mts` before checking restores real coverage.

o2alexanderfedin (approved):
The `193.md` citation is fixed and `btree/remove` came along for the ride. Reviewed the whole delta since `c875b572` — clean.

## `193.md` re-anchoring is exact

I checked every line number the doc now claims against the actual source, and all of them land:

| citation | resolves to |
| --- | --- |
| `set/module.f.mjs:24` | `/** @type {<T>(i: PathItem<T>) => (a: _Branch1To3<T>) => _Branch1To3<T>} */` |
| `remove/module.f.mjs:108` | `/** @type {<A, T>(ms: Tuple<2, _Merge<A, T>>) => …} */` |
| `remove:124` / `:126` | `const reduce = fold(reduceX([reduceValue0, reduceValue2]))` / `const initReduce = …` |
| `remove:115` | `return n.length === 3 ? ra(n) : [...ra([n[0], n[1], n[2]]), n[3], n[4]]` |

Same for `66f-btree-remove-mirror-merge.md` — `:48-60`, `:62-74`, `:76-88`, `:90-100` each open on the `@type` line of the right helper and close on its final brace. The `Array2<Merge>` → `Tuple<2, _Merge>` prose fix in `193.md:46` is also correct against the current signature.

## `btree/remove` migration

- **Runtime semantics unchanged.** Stripping type syntax and comments from the base `.f.ts` and the new `.f.mjs` leaves only the import specifiers — every expression is character-identical, including both `null` casts and the `Branch5` split.
- **Public contract identical.** `nodeRemove: <T>(c: Compare<T>) => (node: TNode<T>) => Tree<T>`, `remove: <T>(c: Compare<T>) => (tree: Tree<T>) => Tree<T>`, and `proof.throw.reduceValue0DefaultBranch: () => void` all match the previous annotations exactly.
- **All four private typedefs correctly prefixed** — `_Leaf01`, `_RemovePath`, `_Branch`, `_Merge` were all non-exported TS `type`s, and JSDoc `@typedef` exports unconditionally, so the `_` form is right per #1461.
- **No `.f.ts` specifier for either module remains anywhere** in the tree.
- **Dropping the explicit `<string>` on `reduceValue0` in the proof is safe** — type arguments are erased, so the throw path is unchanged, and the suite still exercises it.

**Both new `T`-recovery typedefs are load-bearing**, which is the part worth checking since they're doing real work here. Against a 0-error baseline:

- `reduceX`'s `/** @typedef {(typeof n)[1]} T */` — changing `f`'s return to `_Branch<string>` → `TS2322: Type '(a: A) => … | _Branch<string>' is not assignable to type '(a: A) => _Branch<T>'`. The recovery through `PathItem<T>`'s tuple slot really does yield the caller's `T`.
- `nodeRemove`'s `typeof c extends Compare<infer T>` — `_RemovePath<string>` → `TS2322: … is not assignable to type '(node: TNode<T>) => Tree<T>'`.
- Two implementation mutations also caught: dropping `n[4]` from the `Branch5` split → `TS2322` on `_Branch<T>`; dropping the `nullable` `map` in `remove` → `TS2322: '(node: TNode<T>) => Tree<T>' is not assignable to '(tree: Tree<T>) => Tree<T>'`.

**`npm test`** — tsc clean, 2355 pass / 0 fail. CHANGELOG correctly folds both modules into the one `#1471` entry, and `BREAKING CHANGES` is right since the specifier changes for importers.

Approving.

---

Non-blocking, entirely optional, and the same pre-existing drift I flagged as out of scope last time — `193.md:23-24` and `:40-41` still quote a two-line root collapse (`const r = reduceBranch(f())(tail)` / `return r.length === 1 ? r[0] : r`, anchored `:107`/`:108` and `:137`/`:138`). Both modules have used `collapseRoot(…)` on a single line since before this PR — `set/module.f.mjs:113` and `remove/module.f.mjs:164`. Not yours to fix, but you're already in the file if you want to close it out.

PR #1474, opened 2026-08-10
Add stat function import and test coverage for directory stat

sergey-shandar:
## Summary
This PR adds test coverage for the `stat` operation on directories to improve code coverage of the virtual file system implementation.

## Key Changes
- Added `stat` to the imports from the effects module in the proof file
- Added a new test case `statOnDirectory` that verifies `stat` operation on a directory returns an error, covering the `!Array.isArray(file)` branch in the `statOp` implementation

## Implementation Details
The new test creates a virtual file system with a directory and calls `stat` on it, asserting that the operation returns an error result. This ensures proper error handling when attempting to stat a directory rather than a file.

https://claude.ai/code/session_01UAemArpsaRDc9BDkbDdjVj

### DIFF
--- a/fjs/effects/node/virtual/proof.f.ts
+++ b/fjs/effects/node/virtual/proof.f.ts
@@ -174,11 +174,13 @@ export const proof = {
         const [, result] = virtual({ ...emptyState, root })(writeBytes('file', 5, vec8(0x43n)))
         assert(result[0] === 'error')
     },
-    statOnDirectory: () => {
-        // stat on a directory covers the !Array.isArray(file) branch of statOp
-        const root: Dir = { 'mydir': {} }
-        const [, result] = virtual({ ...emptyState, root })(stat('mydir'))
+    statOnJsModule: () => {
+        // stat on a JsModule entry (neither an array nor a descendable
+        // directory) covers the !Array.isArray(file) branch of statOp.
+        const root: Dir = { 'a.f.ts': (() => ({})) as JsModule }
+        const [, result] = virtual({ ...emptyState, root })(stat('a.f.ts'))
         assert(result[0] === 'error')
+        assertEq(result[1], `'a.f.ts' is not a file`)
     },
     largeFileReadBytes: () => {
         // A file stored as two 128 KiB chunks is larger than maxLengthBytes.
--- a/fjs/effects/node/virtual/proof.f.ts
+++ b/fjs/effects/node/virtual/proof.f.ts
@@ -1,5 +1,5 @@
 import { assert, assertEq } from '../../../asserts/module.f.mjs'
-import { access, awaitIfPromise, fetch, rm, writeFile, readFile, readdir, import_, rename, readBytes, writeBytes } from '../module.f.ts'
+import { access, awaitIfPromise, fetch, rm, writeFile, readFile, readdir, import_, rename, readBytes, writeBytes, stat } from '../module.f.ts'
 import { maxLengthBytes, vec, vec8 } from '../../../types/bit_vec/module.f.mjs'
 import { emptyState, virtual, type Dir, type JsModule } from './module.f.ts'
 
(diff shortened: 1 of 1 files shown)


o2alexanderfedin (commented):
The new proof passes, but it doesn't test what it says it does — and it can't fail.

## `statOnDirectory` never reaches the `!Array.isArray(file)` branch

I instrumented `statOp` to print what it actually receives for `stat('mydir')` with `root = { 'mydir': {} }`:

```
PROBE statOp path= [] dirKeys= []
result: ["error",{"code":"ENOENT"}]
```

The path resolver **descends into** the directory, so `statOp` is handed `path = []` and `dir = {}`. It returns from the very first guard — `if (path.length !== 1) { return enoent }` — and never evaluates `dir[path[0]]`, let alone `!Array.isArray(file)`. The comment on the test names a branch two guards further down than the one that fires.

## The assertion can't distinguish them, so the mutation survives

`assert(result[0] === 'error')` holds for every failure path here, so deleting the branch changes nothing the test can see. Deleting `if (!Array.isArray(file)) { … }` and running the full suite:

- **on this PR:** `statOnDirectory(): ok` — 2356 pass, **0 fail**
- **on `main`** (same mutation, without the new test): 2355 pass, 0 fail

Identical. The proof does not detect removal of the branch it was written for. For completeness I also deleted the `path.length !== 1` guard, and the directory case still returns `ENOENT` — so the test doesn't pin that branch either. Whatever line coverage the run reports, nothing here is actually held in place.

## What does reach it

The branch needs an entry that is neither an array nor a descendable directory — a `JsModule`, the same shape `readFileJsModule` already uses:

```ts
statOnJsModule: () => {
    const root: Dir = { 'a.f.ts': (() => ({})) as JsModule }
    const [, result] = virtual({ ...emptyState, root })(stat('a.f.ts'))
    assertEq(result, ['error', `'a.f.ts' is not a file`])
},
```

Confirmed against the real module:

| input | guard present | guard deleted |
| --- | --- | --- |
| `{ 'mydir': {} }`, `stat('mydir')` | `["error",{"code":"ENOENT"}]` | `["error",{"code":"ENOENT"}]` — unchanged |
| `{ 'a.f.ts': fn }`, `stat('a.f.ts')` | `["error","'a.f.ts' is not a file"]` | **throws** `chunks.reduce is not a function` |

So the `JsModule` version does fail when the branch goes. The `assertEq` on the exact message rather than `result[0] === 'error'` is the other half — that's what stops a different error path from satisfying it, which is exactly what happened here.

If the directory case is worth keeping as its own test, it's a fine test of the descend-then-empty-path behaviour — it just wants a name and comment describing that, and an assertion on `ENOENT` specifically.

Everything else is fine: the import addition is correct and `npm test` is green at 2356/0. The head moved to `c8f55ae9` while I was checking; that's a merge of `main` bringing in #1472, nothing on the branch side, and I re-confirmed the finding against it.

sergey-shandar:
Good catch — thanks for the mutation-testing verification.

Fixed in 750b402: replaced the directory-based `statOnDirectory` case with `statOnJsModule`, using a `JsModule` entry (neither an array nor a descendable directory, same shape `readFileOnJsModule` already uses), which is the only way to reach `statOp`'s `!Array.isArray(file)` guard — the path resolver descends into directories before `statOp` sees them, so the original case only ever hit the `path.length !== 1` guard.

Also switched the assertion from `assertEq(result, [...])` to checking `result[0]` and `result[1]` separately, since `assertEq` is `===`-based and would never match a fresh array literal by reference.

Re-verified locally: `npm test` is green at 2356/2356, and deleting the `!Array.isArray(file)` guard now makes `statOnJsModule` fail (confirming the branch is actually pinned).

---

o2alexanderfedin (approved):
Fixed, and now it actually fails when the branch goes.

Re-ran the same mutation as before — deleting `if (!Array.isArray(file)) { … }` from `statOp`:

| | baseline | guard deleted |
| --- | --- | --- |
| **before** (`statOnDirectory`) | 2356 pass / 0 fail | 2356 pass / **0 fail** |
| **now** (`statOnJsModule`) | 2356 pass / 0 fail | 2355 pass / **1 fail** |

```
import("./fjs/effects/node/virtual/proof.f.ts").proof.statOnJsModule(): error
```

The `JsModule` entry reaches the guard (it's neither an array nor a descendable directory, so the resolver leaves `path.length === 1` and hands `statOp` the function), and `assertEq(result[1], "'a.f.ts' is not a file")` is what makes the difference detectable — with only `result[0] === 'error'` the mutation would still slip through, since removing the guard makes the call throw and the throw also surfaces as an error result. Keeping both assertions is right.

The comment now describes what the fixture actually exercises, and `JsModule` was already imported so nothing else changed. `npm test` green at 2356/0.

Approving.

PR #1475, opened 2026-08-10
Convert btree module to JSDoc type annotations

sergey-shandar:
This PR converts the btree module from TypeScript type syntax to JSDoc type annotations, enabling it to be written in plain JavaScript (.mjs) rather than TypeScript (.ts).

## Summary
The btree module's type definitions have been migrated from inline TypeScript syntax to JSDoc `@type` and `@import` annotations. This allows the module to use the `.mjs` extension while maintaining full type safety through JSDoc comments.

## Key changes
- Converted `import type` statements to `/** @import ... */` JSDoc comments for type-only imports
- Replaced TypeScript function signature syntax with `/** @type {...} */` JSDoc annotations
- Renamed `fjs/types/btree/module.f.ts` to `fjs/types/btree/module.f.mjs`
- Updated all import paths in dependent modules (`ordered_map`, `string_set`, `proof`) to reference the new `.mjs` file instead of `.ts`

## Implementation details
- Type annotations are preserved as JSDoc comments above variable declarations
- Generic type parameters are maintained in the JSDoc syntax: `<T>(node: TNode<T>) => Thunk<T>`
- All dependent modules that imported from the btree module have been updated to use the new file extension

https://claude.ai/code/session_018cCFhVohQxcmqYdZb8Kg1d


o2alexanderfedin (commented):
Migration itself is clean — one thing to fix, and it's a one-liner that the plan doc now on this branch already calls out.

## The `@module` header is dropped from the emitted declaration

`fjs/types/btree/module.f.mjs` has no blank line between its header and the first `import` statement:

```js
/**
 * Core persistent B-tree construction and traversal helpers.
 *
 * @module
 */
import { flat } from '../list/module.f.mjs'   // <- header attaches here
```

so `module.f.d.mts` opens straight at `import type { List } from '../list/module.f.mjs';` with no `@module` anywhere in it. `flat` and `map` are runtime-only imports that declaration emit drops, and the header goes with the statement it's attached to.

This is exactly the case #1472 documented — and since that merged into this branch, `todo/migrate-typescript-to-mjs.md` here now carries both the rule and the checklist item naming the three modules to fix. Landing as-is adds a fourth. Adding one blank line after `*/` fixes it; I verified that the header then appears at the top of `module.f.d.mts`.

Worth flagging that I got a false negative on this check first time round: I tested the fix without deleting the previously emitted `.d.mts`, and the stale file made it look like the blank line changed nothing. Deleting the emitted declarations first and re-running shows it working.

## Everything else verified

- **Runtime semantics unchanged** — the only edits are type-syntax removal; every expression is untouched.
- **Public contract preserved**: `empty: null` and `values: <T>(tree: Tree<T>) => List<T>`, matching the previous annotation. `nodeValues` stays module-private and is a `const`, not a typedef, so no `_` prefix question arises.
- **All importers moved** — `btree/proof.f.ts`, `types/ordered_map`, `types/string_set`. No `btree/module.f.ts` specifier remains anywhere in the tree.
- **CHANGELOG** entry present with the right `**BREAKING CHANGES:**` label — the specifier does change for importers.
- **Annotations are load-bearing**: against a 0-error baseline, retyping `values` as `Thunk<T>` → `TS2322: '(value: Nullable<…>) => Nullable<Thunk<T>>' is not assignable to '<T>(tree: Tree<T>) => Thunk<T>'`, and dropping the `nullable` `map` → `TS2322: '<T>(node: TNode<T>) => Thunk<T>' is not assignable to '<T>(tree: Tree<T>) => List<T>'`. (Widening `nodeValues` from `Thunk<T>` to `List<T>` correctly does *not* error, since `Thunk` is one of `List`'s constituents — that one was a bad probe on my part, not a gap.)

The head moved to `dd8e44e9` while I was checking; that's a merge of `main` bringing in #1472, nothing on the branch side, and I re-confirmed the header loss against it.

sergey-shandar:
Fixed in 2705570 — added a blank line after the `@module` header so it attaches to the module instead of the `import` statement. Verified `module.f.d.mts` now carries the header (after deleting the stale emitted `.d.mts` and re-running declaration emit). Merged `main` and pushed; `npm test` is 2355/2355 passing on the merged branch.

---

o2alexanderfedin (approved):
Header fix confirmed. `fjs/types/btree/module.f.d.mts` now opens with

```ts
/**
 * Core persistent B-tree construction and traversal helpers.
 *
 * @module
 */
import type { List } from '../list/module.f.mjs';
```

which was the only thing outstanding. Everything else I checked last round still holds — contract unchanged (`empty: null`, `values: <T>(tree: Tree<T>) => List<T>`), all three importers moved, no `btree/module.f.ts` specifier anywhere, CHANGELOG labelled correctly, `npm test` green, and the two annotations still load-bearing under mutation.

Approving.

---

Unrelated to this PR, but I noticed it while sweeping the tree for the same problem: the checklist item in `todo/migrate-typescript-to-mjs.md` names three modules that lose their `@module` header, and on current `main` it is **five**. `fjs/types/btree/set` and `fjs/types/btree/remove` also have their header adjacent to the first `import` statement — they landed in #1471, which merged before #1472 established the rule, so nothing caught them at the time. After this PR the full list is `common/monoid`, `types/list`, `types/nullable`, `types/btree/set`, `types/btree/remove`. Same one-line fix each; just flagging so the checklist isn't worked from an undercount.

PR #1477, opened 2026-08-10
TS => MJS: types/number, types/nibble_set

sergey-shandar:
## Summary

- Migrate `fjs/types/number/module.f.ts` -> `module.f.mjs` under the stage-1
  TypeScript-to-mjs migration
  All of the module's runtime dependencies (`list`, `function/operator`,
  `function/compare`, `common/monoid`) are already `.f.mjs`, so it qualifies as
  a dependency leaf.
  - Also migrate `proof.f.ts` -> `proof.f.mjs`: its only dependency (`asserts`)
    is already migrated.
  - Generic function instantiation (`uMin<number>`, `uMax<number>`) has no
    direct JSDoc equivalent, so each is bound to a separately
    `@type {Reduce<number>}` annotated `const` instead of an inline cast, per
    the `AGENTS.md` guidance on avoiding unsafe casts.
  - Update the three remaining importers of the old `.f.ts` specifier
    (`fjs/types/range_map/module.f.ts`, `fjs/types/sorted_list/proof.f.ts`,
    `fjs/types/sorted_set/proof.f.ts`) and the stray prose reference in
    `fjs/common/monoid/todo/balanced-fold.md`.
- Migrate `fjs/types/nibble_set/module.f.ts` -> `module.f.mjs` — another
  dependency leaf with no authored runtime dependencies.
  - Also migrate `proof.f.ts` -> `proof.f.mjs`: its dependencies (`list`,
    `asserts`) are already `.f.mjs`.
  - Re-anchor the byte_set/nibble_set line-number comparison table in
    `fjs/types/todo/bit-set-factory.md` to the new source.
- Confirmed no open/draft PR touches either module before starting each (no
  other open PRs existed at the time).

**BREAKING CHANGES:** importers of `fjs/types/number/module.f.ts`,
`fjs/types/number/proof.f.ts`, `fjs/types/nibble_set/module.f.ts`, and
`fjs/types/nibble_set/proof.f.ts` must use the `.f.mjs` specifier instead.

## Test plan

- [x] `node ./fjs/module.ts t` — full proof suite passes (2356/2356), including
      the migrated `types/number` and `types/nibble_set` proofs


o2alexanderfedin (approved):
Verified independently — clean, and the `uMin`/`uMax` handling is the right call.

## Contract is unchanged

Diffed the emitted declaration against a declaration emit at `main`. `sum`, `min`, `max`, `cmp` and `countOnes` are byte-identical; the only additions are the echoed `@type` JSDoc, `import type` spelling, and `export type _MaskOffset = readonly [number, number]`. `MaskOffset` was a non-exported `type` on the base, so the `_` prefix is right.

The `@module` header keeps its blank line and survives into `module.f.d.mts`, and the `countOnes` doc comment survives too (it sits on an `export const`, which declaration emit does carry).

## The generic-instantiation rewrite

Binding `uMin`/`uMax` to separately annotated `const`s rather than casting inline is the right reading of the `AGENTS.md` rule you added in #1476, and I checked that the declaration form really is *checked* rather than overriding: `/** @type {Reduce<number>} */ const minReduce = countOnes` → `TS2322: Type '(n: number) => number' is not assignable to type 'Reduce<number>'`. So the annotation constrains the initializer the way `satisfies` would.

Two of my probes came back clean and are worth reporting as non-findings rather than silently dropping:

- `const minReduce = uCmp` type-checks, because `Reduce<number>` is `Fold<number, number>` and `Sign` is assignable to `number`. Legitimate, not a checking hole.
- `const maxReduce = uMin` also type-checks — `min` and `max` share the signature `<A extends Cmp1>(a: A) => <B extends Cmp2<A, B>>(b: B) => A | B`, so nothing at the type level can tell them apart. That's the exact copy-paste hazard this refactor introduces, so I ran it as a behavioural mutation instead: **2355 pass / 1 fail**, with `proof.max(): error`. The proofs pin what the types can't.

## Everything else

- **All three importers moved** — `types/range_map/module.f.ts`, `types/sorted_list/proof.f.ts`, `types/sorted_set/proof.f.ts`. That's the complete set; no `number/module.f.ts` or `number/proof.f.ts` specifier survives anywhere in the tree.
- **The migrated proof is collected and runs** — all eight appear as `import("./fjs/types/number/proof.f.mjs").proof.…`, including the nested `min.empty` / `min.multi` and the indexed `countOnes[0]` / `countOnes[1]`.
- **`npm test`** — tsc clean, 2356 pass / 0 fail.
- **Annotations load-bearing** against a 0-error baseline: dropping `| null` from `min` → `TS2322`; retyping `mo` as `readonly number[]` → `TS2322: '[number, number]' is not assignable to 'number'`; adding a third tuple element → `TS2322: '[number, number, number]' is not assignable to '_MaskOffset'`; binding `cmp` to `uMin` → `TS2322` on the `Sign` return.
- **CHANGELOG** entry present and correctly labelled.
- **`balanced-fold.md:129`** repointed correctly, and the citation carries no line number so there's nothing further to re-anchor. The other `.f.ts` path in that file, `fjs/types/string/module.f.ts`, still exists — correctly left alone.

Approving.

sergey-shandar:
@codex review

PR #1478, opened 2026-08-10
TS => MJS: patricia_trie, sorted_list, range_map, range_set, crypto/pow, bigfloat, prime_field, crypto/secp

sergey-shandar:
## Summary

Eight independent dependency-leaf modules migrated from authored TypeScript
to JSDoc-typed JavaScript under the stage-1 TypeScript-to-mjs migration
([`todo/migrate-typescript-to-mjs.md`](https://github.com/functionalscript/functionalscript/blob/main/todo/migrate-typescript-to-mjs.md)),
each becoming eligible only once its dependencies were already `.f.mjs`:

- `fjs/types/patricia_trie/module.f.ts` -> `module.f.mjs` (+ `proof.f.mjs`).
  No authored runtime deps.
- `fjs/types/sorted_list/module.f.ts` -> `module.f.mjs`. `proof.f.ts` stays
  TypeScript (still imports unmigrated `media/json`, `object`).
- `fjs/types/range_map/module.f.ts` -> `module.f.mjs`. Became a leaf once
  `sorted_list` migrated. `proof.f.ts` stays TypeScript.
- `fjs/types/range_set/module.f.ts` -> `module.f.mjs` (+ `proof.f.mjs`).
  Became a leaf once `range_map` migrated.
- `fjs/crypto/pow/module.f.ts` -> `module.f.mjs`. No external importers.
  `proof.f.ts` stays TypeScript (still imports unmigrated `text`).
- `fjs/types/bigfloat/module.f.ts` -> `module.f.mjs` (+ `proof.f.mjs`).
- `fjs/types/prime_field/module.f.ts` -> `module.f.mjs` (+ `proof.f.mjs`).
- `fjs/crypto/secp/module.f.ts` -> `module.f.mjs` (+ `proof.f.mjs`). Became
  a leaf once `prime_field` migrated. No generics in this module — every
  export is concrete (`Point`, `bigint`, `Curve`) — so the conversion is
  mechanical `@type`/`@param`/`@returns` annotation.

Every dependent `.f.ts` importer across the repo was updated to the new
`.f.mjs` specifier in the same commit as its dependency's migration (18
importer files total across the eight modules), along with the doc/todo
files that quoted pre-migration source or line numbers.

For `sorted_list`/`range_map`, the curried, doubly-generic helpers
(`genericMerge`, `merge`, `intersect`, `get`, `fromRange`, `rangeMap`) use
the per-arrow `@template`/`@param`/`@returns` JSDoc style already
[...]

PR #1480, opened 2026-08-10
TS => MJS: text/utf8, text, types/uint8array, nullable/range proofs

sergey-shandar:
## Summary

- Migrate `fjs/text/utf8/module.f.ts` -> `module.f.mjs` under the stage-1
  TypeScript-to-mjs migration
  Dependency leaf; declaration emit checked clean.
- Migrate `fjs/text/module.f.ts` -> `module.f.mjs` — became a fresh leaf
  once `text/utf8` migrated; updated 21 remaining `.f.ts` importers found
  via an exhaustive resolved-path scan (a literal-substring grep missed
  `fjs/text/sgr/module.f.ts`'s `../module.f.ts`).
- Migrate `fjs/types/uint8array/module.f.ts` -> `module.f.mjs` (+
  `proof.f.mjs`) — became a fresh leaf once `text` migrated.
- Migrate `fjs/types/nullable/proof.f.ts`, `fjs/types/range/proof.f.ts`,
  `fjs/types/function/proof.f.ts`, `fjs/types/result/proof.f.ts`, and
  `fjs/types/function/compare/proof.f.ts` -> `proof.f.mjs`. Module-level
  leaves are exhausted for now (only the blocked `phantom` hard case
  remains as a `module.f.ts` leaf), so this picks up ready proof-only
  migrations instead, per the migration doc's rule that a proof moves to
  `.f.mjs` as soon as it's JS/JSDoc-ready and its dependencies are already
  migrated, independent of when its module migrated — all five modules
  predate this PR entirely. `range/proof.f.ts` had no TypeScript-only
  syntax at all, so it's a byte-identical rename;
  `function/compare/proof.f.ts`'s commented-out compile-error
  demonstrations are left as-is, since JSDoc/`tsc` still enforces the same
  behavior at those call sites under `checkJs`. No other module imports a
  proof file, so none of the five is a breaking change.
- Confirmed no open/draft PR touches any of these before starting each (no
  other open PRs existed at the time).

**BREAKING CHANGES:** importers of `fjs/text/utf8/module.f.ts`,
`fjs/text/module.f.ts`, `fjs/types/uint8array/module.f.ts`, and
`fjs/types/uint8array/proof.f.ts` must use the `.f.mjs` specifier instead.

## Test plan

- [x] `node ./fjs/module.ts t` — full proof suite passes (2356/2356)
- [x] Declaration emit checked clean (0 `any`/`elided`) for every migrated
      module


o2alexanderfedin (commented):
The migration itself is solid — I ran my own resolved-path scan and it agrees with yours. Two line-range citations need another look.

## Two doc anchors

**`error-tag-layout-constants.md:9` — `utf8StateToError` (`:148-173`)**

`export const utf8StateToError` is at **`:173`** and its body closes at **`:198`**, so the cited range now *ends* exactly where the function begins. This is the one citation in the set written as a bare `(:148-173)` with no path attached, which is presumably why a path-driven sweep didn't reach it — it sits in the same sentence as the `codePointToUtf8` reference you did update.

**`vec-to-code-point-pipeline.md:13` — `fjs/text/module.f.mjs:69-70`**

Off by one, and this one was correct before. On `main`, `text/module.f.ts:61-62` is exactly the two lines the block quotes:

```js
export const utf8ToString = (msbV: Utf8): string =>
    codePointListToString(toCodePointList(u8List(msb)(msbV)))
```

After the migration those same two lines are at **`:70-71`** — `:69` is the closing `*/` of the preceding JSDoc. The `+8` shift landed one short.

For contrast, the sibling citation in the same block is right: `:283-290` → `:293-300` lands exactly on `export const fromVec = v => {` … `}`. Same for `:270-273` (`utf8EofToCodePointOp`), `:217-220` (`utf8ByteToCodePointOp`), `:92-96` (`restart`) and `:293-300`.

Worth noting so it doesn't read as a bigger problem than it is: `codePointToUtf8`'s `:107-128` starts on a `*/` rather than the declaration at `:108`, but that range was already imprecise on `main` (base `:107` is a bare `}`), so it's pre-existing drift rather than something this PR shifted.

## The importer sweep holds up

This was the part I most wanted to check independently, given 30 importers and your note about `fjs/text/sgr/module.f.ts`'s `../module.f.ts` escaping a substring grep. I resolved every relative `.f.ts` specifier in the tree against its importing file and compared against the six migrated paths: **zero stale references**, across sources, `.ts`, `.mjs` and docs. The one remaining textual mention, `CHANGELOG.md:408`, is inside the released `## 0.34.0` section and correctly records the path as it was then — history, not a stale pointer.

## Everything else verified

- **All three contracts preserved.** `text/utf8` and `types/uint8array` differ from base only in `import type` spelling. `text` additionally renames `ItemArray`/`ItemThunk` to `_ItemArray`/`_ItemThunk` — both non-exported at base, so correct — and `utf8`'s declared return goes from `Vec` to `Utf8`, which is the same type under its existing exported alias (`export type Utf8 = Vec`), so nothing moves for consumers.
- **0 `any` / 0 `elided`** in all three emitted declarations — the `@returns` discipline from #1478 is holding.
- **All three `@module` headers survive** into their declarations. `types/uint8array`'s long `@deprecated` block had me briefly reading it as a casualty; a full-tree sweep confirms it's fine, and the set of modules losing their header is unchanged at seven — this PR adds none.
- **`npm test`** — tsc clean, 2356 pass / 0 fail.
- **CHANGELOG** entry present with the right label.

sergey-shandar:
Fixed both in 0670228, thanks for the independent scan.

- `error-tag-layout-constants.md`: `utf8StateToError`'s bare `(:148-173)` citation — the one without a leading path in that sentence, which is why my path-driven sweep missed it — is now `(:173-198)`, matching the current declaration/close.
- `vec-to-code-point-pipeline.md`: `utf8ToString`'s citation is now `:70-71` (was off by one at `:69-70`, landing on the preceding JSDoc's closing `*/`).

Left `codePointToUtf8`'s `:107-128` and the individual bit-flag sub-citations alone per your note — that drift predates this PR.

`npx tsc` clean, `node ./fjs/module.ts t` 2356/2356.

---

o2alexanderfedin (approved):
Both anchors fixed and verified exact:

| citation | lands on |
| --- | --- |
| `utf8/module.f.mjs:173-198` | `export const utf8StateToError = state => {` … `}` |
| `text/module.f.mjs:70-71` | the two quoted `utf8ToString` lines |

Four proof files also migrated since (`types/nullable`, `types/range`, `types/function`, `types/result`). The thing worth checking there is whether they're still collected — a proof that quietly stops being discovered is a green suite that proves nothing. All four are, by name:

```
import("./fjs/types/function/proof.f.mjs").proof()
import("./fjs/types/nullable/proof.f.mjs").proof[0..3]()
import("./fjs/types/range/proof.f.mjs").proof()
import("./fjs/types/result/proof.f.mjs").proof.example() / .invertTest() / .mapOkTest()
import("./fjs/types/result/proof.f.mjs").proof.throw.unwrapError()   # EXPECTED TO THROW
```

including the nested array form and the `throw`-marked case. Assertion counts are unchanged across all four (nullable 12, range 6, result 8), and the only edits are type-syntax — the `types/function` one shows as add/delete rather than a rename, so I diffed it against `main` directly and it is purely `: T` annotations becoming `@type`. Suite still 2356/0, so nothing was dropped from discovery.

Treating the proof migrations as non-breaking is right, and the reason given — no module imports a proof file — is the correct one; I confirmed no `proof.f.` specifier is imported anywhere outside the runner's own discovery.

Approving.

---

Non-blocking, pre-existing, and in the file you just edited — the bullet list three lines below the anchor you fixed has the same class of bare path-less range, and all four are stale:

| doc says | actually at |
| --- | --- |
| `0b1000_0000_0000_0000` at `:108` and `:166` | `:131` and `:191` |
| `0b0000_0100_0000_0000` at `:115` and `:158` | `:138` and `:183` |
| `0b0000_0010_0000_0000` at `:121` and `:159` | `:144` and `:185` |
| `0b0000_0000_1000_0000` at `:127` | `:150` |

I checked `main` before attributing these: `:108` there is `if (input >= 0x0080 && input <= 0x07ff) {` and `:166` is a bare `*/`, so they were already drifted before this PR and it neither introduced nor shifted them. Same root cause as the `(:148-173)` you fixed, though — a range with no path in it is invisible to a path-driven sweep. Worth a pass whenever that todo is next touched.

sergey-shandar:
Thanks for the approval and the thorough proof-discovery check.

Fixed the non-blocking bit-flag anchor drift in 192d760 anyway, since it's the same file already touched in this PR and you'd already worked out the correct current numbers:

- `codePointToUtf8`'s error branch: `:107-128` → `:130-151`
- `0b1000_0000_0000_0000`: `:108`/`:166` → `:131`/`:191`
- `0b0000_0100_0000_0000`: `:115`/`:158` → `:138`/`:183`
- `0b0000_0010_0000_0000`: `:121`/`:159` → `:144`/`:185`
- `0b0000_0000_1000_0000`: `:127` → `:150`

All verified against the current source before applying. `npx tsc` clean, `node ./fjs/module.ts t` 2356/2356.

---

o2alexanderfedin (approved):
All four bare bit-flag anchors fixed, and you went further than I did on the fifth — every citation in the doc now lands exactly:

| doc | lands on |
| --- | --- |
| `:130-151` | `if ((input & errorMask) !== 0) {` … `}` |
| `:131` / `:191` | both `0b1000_0000_0000_0000` sites |
| `:138` / `:183` | both `0b0000_0100_0000_0000` sites |
| `:144` / `:185` | both `0b0000_0010_0000_0000` sites |
| `:150` | `0b0000_0000_1000_0000` |
| `:173-198` | `utf8StateToError` |

Worth calling out that `:107-128` → `:130-151` is a better anchor than the one I was implicitly asking for. I'd noted the old range started on a `*/` rather than the declaration at `:108`, but the doc's claim is `codePointToUtf8`'s **error branch**, not the whole function — and `:130-151` is exactly that branch. That's more precise than either the old range or the declaration-start I had in mind.

`types/function/compare/proof.f.ts` → `.mjs` also came along: collected and running (`import("./fjs/types/function/compare/proof.f.mjs").proof(): ok`), assertion count unchanged at 13, and the only edits are type-syntax. No stale specifier for it anywhere — the one remaining textual mention is the CHANGELOG entry describing the migration itself, which correctly names the old path.

`npm test` — 2356 pass / 0 fail.

Approving.

PR #1486, opened 2026-08-11
Migrate fjs/crypto/vdf to .f.mjs

sergey-shandar:
## Summary

- Migrates `fjs/crypto/vdf/module.f.ts` (and `proof.f.ts`) from authored TypeScript to JSDoc-typed `.f.mjs`, per `todo/migrate-typescript-to-mjs.md`.
- Splits the public `Sloth` type into a sibling `types.ts`.
- A dependency leaf: only depends on already-migrated `types/prime_field`, `types/nullable`, and `types/bigint`; nothing else in the repo imports `crypto/vdf`.
- Updates the stale `module.f.ts` line references in `fjs/crypto/vdf/todo/iterate-combinator.md` to match the new file and line numbers.

## Test plan

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@ history.
 
 ## Unreleased
 
+- **BREAKING CHANGES:** `fjs/bnf/token_symbol` migrates from authored
+  TypeScript (`.f.ts`) to JSDoc-typed JavaScript (`.f.mjs`), splitting
+  the `Encoding<T>` type into a sibling `types.ts` — importers must use
+  the `.f.mjs` specifier for runtime values and the `types.ts`
+  specifier for types
+  [#1486](https://github.com/functionalscript/functionalscript/pull/1486)
 - **BREAKING CHANGES:** `fjs/crypto/vdf` migrates from authored
   TypeScript (`.f.ts`) to JSDoc-typed JavaScript (`.f.mjs`), splitting
   the `Sloth` type into a sibling `types.ts` — importers must use the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@ history.
 
 ## Unreleased
 
+- **BREAKING CHANGES:** `fjs/crypto/vdf` migrates from authored
+  TypeScript (`.f.ts`) to JSDoc-typed JavaScript (`.f.mjs`), splitting
+  the `Sloth` type into a sibling `types.ts` — importers must use the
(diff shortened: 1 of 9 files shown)


o2alexanderfedin (approved):
Approving. Reviewed at `37df907` against `main` (`e9e34c6`). Two modules — `fjs/crypto/vdf` and `fjs/bnf/token_symbol`. Nothing to flag.

## The 1024-bit modulus is intact

`crypto/vdf` is the one place in this migration where a silent digit change would be both catastrophic and invisible, so I did not settle for a token diff. Against `main`'s compiled output the module looks different, but the *only* difference is numeric separators: TypeScript's emitter strips `_` from `0xf2346eae06a23388_2814ff16…`, while the authored `.mjs` keeps the grouping exactly as `main`'s **source** wrote it. Discount separators and the two are identical.

Then I checked the values rather than the text — imported both module versions and compared:

```
p equal:                 true   (1024 bits)
sloth.p equal:           true
quadRes agree:           true
modSqrt agree:           true
eval(50)(12345) equal:   true
verify agrees:           true  ->  true
```

So the modulus, the quadratic-residue test, the modular square root, and a full eval/verify round-trip all match `main`. `bnf/token_symbol` is token-identical outright.

## Everything else

- **Public API unchanged on both axes** — 0 removed / 0 added / 0 changed for exported type aliases *and* for exported const signatures. (I started checking the second of those after `eof` slipped past the alias-only check on #1485; nothing here.)
- Markdown links: 144 broken on `main`, 144 here — **none added** — and 0 label/target mismatches. `crypto/vdf`'s own `todo/` doc was updated as part of the change rather than left for review.
- Both new `types.ts` files carry the `@module` header *with* the blank line, so it survives into the emitted `types.js` instead of being swallowed by the following import.
- Both proofs import only `.f.mjs` runtime modules, so `proof.f.mjs` is legitimate under §2.
- CHANGELOG has a correctly-scoped `**BREAKING CHANGES:**` entry for each, in the short §8.3 form.

One nice detail worth naming: the updated line references in `todo/iterate-combinator.md` are accurate, not approximate. `module.f.mjs:29-37` lands exactly on `repeatSeq`'s JSDoc and body, and `51-57` covers both call sites (`squareLoop` and `modSqrtLoop`). Stale line numbers are the easiest thing to hand-wave in a rename; these were actually recomputed.

CI is still running on `37df907`; I verified locally at that commit.

sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1487, opened 2026-08-11
Migrate fjs/bnf/data to .f.mjs

sergey-shandar:
## Summary

- Migrates `fjs/bnf/data/module.f.ts` from authored TypeScript to JSDoc-typed `.f.mjs`, per `todo/migrate-typescript-to-mjs.md`.
- Splits the public type-level API (`TerminalRange`, `Sequence`, `Variant`, `Rule`, `RuleSet`, `EmptyTag`) into a sibling `types.ts`.
- `proof.f.ts` stays TypeScript for now — it has a runtime dependency on `../testlib.f.ts` and `../../media/json/module.f.ts`, both still unmigrated.
- Updates the two dependents (`fjs/bnf/descent`, `fjs/bnf/ll1` — module + proof each) to import runtime values from `.f.mjs` and types from `types.ts` separately.
- Fixes stale `module.f.ts` references (paths and, where cited, line numbers) across `fjs/bnf/data`'s `todo/` docs.

## Test plan

---


o2alexanderfedin (approved):
Approving. Reviewed at `28199f7` against `main` (`03126e4`). Five modules — `bnf/data`, `bnf/ll1`, `bnf/descent`, `effects`, `effects/list` — plus ~45 dependents updated.

## Two modules genuinely restructure, and both check out

`bnf/data`, `bnf/descent`, and `effects/list` are equivalent to `main`'s compiled output. The other two each carry one real code change, both the same idiom: JSDoc has no call-site type-argument syntax, so an inline argument gets extracted into a named `const` that can be annotated.

```js
// ll1:  rangeMap<DispatchResult>({…})   ->
/** @type {Properties<_DispatchResult>} */
const dispatchProps = { … }
const dispatchOp = rangeMap(dispatchProps)

// effects:  fold<T, Effect<O|Q, S>>(item => acc => step(acc, f(item)))   ->
/** @type {Fold<T, Effect<O | Q, S>>} */
const op = item => acc => step(acc, f(item))
return step(items, fold(op)(pure(init)))
```

I enumerated every normalized difference in each module to confirm these are the *only* ones — one hunk in `ll1`, one in `effects` — then checked behaviour rather than arguing from the shape. `effects` is the core combinator module, so it got a direct comparison against `main`:

```
foldStep sum 1..5    main: [15]   pr: [15]
foldStep empty list  main: [42]   pr: [42]      (init passes through)
forEachStep          main: [null] pr: [null]
okStep ok / error    main: [42] [["error","boom"]]
                     pr  : [42] [["error","boom"]]
```

Both `okStep` branches, the empty-list edge, and the folding path all match. For `ll1` I leaned on the repo's own coverage instead — my ad-hoc `parser` harness was calling the API wrong (it failed identically against `main`, so it was my mistake, not a finding), and the three migrated modules' proofs contribute 164 passing cases, with another 222 across `effects`.

`effects/list` shows up as differing only because my normalizer rewrites `'.f.js'` in single quotes and that file uses double quotes — the sole difference is the import specifier, and `empty`/`nonEmpty` produce identical results.

## Everything else

- **Public API: 0 removed, 0 non-`_` additions.** All 10 new types are `_`-prefixed. The 3 changed types and 3 changed const signatures are entirely those renames (`AstRule`→`_AstRule`, `DispatchRule`→`_DispatchRule`, `DispatchMap`→`_DispatchMap`) plus one `effects/module.f.ts`→`effects/types.ts` import path. I confirmed all three renamed types were non-exported on `main`, so nothing public was renamed.
- Markdown links: 144 broken on `main`, 144 here, **none added**, 0 label/target mismatches. No bare `.f.ts` code-span references to migrated files either — the shape that slipped through on #1486.
- All four new `types.ts` files keep the `@module` header through emit.
- CHANGELOG entries for all five, including the explicit "`proof.f.ts` stays TypeScript for now" note.

**The `proof.f.ts` decision is right and I verified the reason.** All three `bnf` proofs import runtime values from `../testlib.f.ts`, and `data`'s also from `media/json/module.f.ts` — both still `.f.ts` on disk, so §2 does not yet permit `proof.f.mjs`. Stating that in the PR body *and* the CHANGELOG rather than leaving a reviewer to work it out is the right call.

CI still finishing on `28199f7`; verified locally at that commit.

sergey-shandar:
@codex review

sergey-shandar:
Checked this directly against the actual declaration emit (`npx tsc --noEmit false --declaration --emitDeclarationOnly --outDir <scratch>`). `match`'s emitted `.d.mts` is:

```ts
export declare const match: <O extends Operation, R>(map: OperationMap<O, R>) => <O1 extends O, T>(
/**
 * @template {O} O1
 * @template T
 * @param {Effect<O1, T>} e
 * @returns {MatchResult<O1, T, R>}
 */
e: Effect<O1, T>) => MatchResult<O1, T, R>;
```

`O1` and `T` **are** declared, in `<O1 extends O, T>(...)` — the JSDoc block just lands as a positional comment ahead of the `e` parameter, which is legal TS syntax (a comment can appear anywhere whitespace can, including inside a parameter list). This is the same "noisy but not broken" per-arrow-JSDoc emit pattern already flagged as pre-existing on 14 other declarations in an earlier review round on this PR (the `option` example) — not new breakage.

I verified two ways:
1. Compiled the exact emitted snippet standalone (`tsc --noEmit` on a `.d.ts` reproducing it, with a matching `Operation`/`Effect`/`MatchResult`/`OperationMap`) — clean, no errors, in particular no `TS2304`.
2. Wrote a small consumer importing `match` from the real emitted `module.f.d.mts` and calling it — it type-checks and dispatches correctly; the only diagnostic I hit was an unrelated `R` inference note from my test's loosely-typed `map`, not anything about `O1`/`T` being undeclared.

No `TS2304` anywhere in either check. I don't see the four errors described — happy to look again with a concrete repro if you have one, but as written this looks like a false positive.

---

sergey-shandar:
@codex review

PR #1488, opened 2026-08-11
Migrate fjs/effects/memory, mock, eff, node, and fjs/text/sgr to .f.mjs

sergey-shandar:
## Summary

Five modules migrate from authored TypeScript to JSDoc-typed `.f.mjs`, per `todo/migrate-typescript-to-mjs.md`, each splitting its public type-level API into a sibling `types.ts`:

- `fjs/effects/memory/module.f.ts` — `Key`, `MemCreate`, `MemRead`, `MemWrite`, `MemOp`, and the private `MemKeyHash` → `_MemKeyHash`. Updates 13 dependents (`cas`, `mcp`, `protocol/mcp`, `effects/node` and its `virtual`/`memory` submodules), and fixes two markdown link hrefs (`cas/evo/module.f.ts`'s docstring, `cas/evo/README.md`) that pointed at the old `.f.ts` path.
- `fjs/effects/mock/module.f.ts` — `MemOperationMap`, `RunInstance`. Updates 4 dependents.
- `fjs/effects/eff/module.f.ts` — the `Eff<O,T,P>` type. Updates 4 dependents plus 3 doc references.
- `fjs/effects/node/module.f.ts` (566 lines) — the Node.js effect operations module: filesystem, networking, exec, console, sandbox, test registration, and `NodeOp`/`NodeProgram`. Updates 26 dependents across the repo, plus a batch of stale `module.f.ts` doc references (including several with stale line-number citations that now split across `module.f.mjs` and `types.ts`).
- `fjs/text/sgr/module.f.ts` — `Stdout`, `WriteText`, `CsiConsole`. Updates the `emergent_testing` dependent and stale line-number citations across three todo docs.

All five `proof.f.ts` files stay TypeScript for now (their remaining runtime dependencies: `effects/proof.f.ts`, `effects/node/virtual/module.f.ts` — both still unmigrated).

## Test plan


o2alexanderfedin (commented):
Reviewed at `91bd993`. Four modules — `effects/memory`, `effects/mock`, `effects/eff`, `effects/node`. All four are equivalent to `main`'s compiled output, `npx tsc` is clean, 2356 pass / 0 fail, links are 144/144 with none added and 0 label-target mismatches, and the CHANGELOG covers all four. One thing to fix.

Two apparent differences turned out to be artifacts of comparing against compiled output, and both are worth naming because they are the *good* outcome:

- `effects/node` looks like it changes `(lenV & 0b111n)` to `(lenV & 7n)`. It doesn't — `main`'s **source** at line 216 already reads `0b111n`, and TypeScript's emitter is what normalizes it to `7n`. The branch preserves the binary spelling, so the "low three bits" intent of the mask survives.
- `effects/memory`'s `asBase`/`asNominal` declarations gain a named `_MemKeyHash` where `main` inlined the raw hash literal `"3f114fa6…"`. That is a readability improvement, not a change: the alias was non-exported on `main` so the emitter had to inline it, and now that it is exported (correctly `_`-prefixed) it can be named.

Also noticed `read`'s parameter was quietly fixed from `ket` to `key`.

## `log` and `error` dropped the `Console` annotation

`main`:

```ts
export const log: Console = writeString('stdout')
export const error: Console = writeString('stderr')
```

this branch:

```js
/** Writes a line to `stdout`. Replaces the retired `Log` effect. */
export const log = writeString('stdout')
```

With no annotation the type is inferred from `writeString('stdout')`, so the emitted declarations expand to `(s: string) => Effect<Write, void>` instead of naming `Console`. `Console` is `(s: string) => Effect<Write, void>`, so the types are identical and no caller can tell — nothing is weakened.

What makes it worth fixing is the second-order effect: `Console` is still exported from `types.ts`, unchanged, and after this change **nothing in the emitted surface references it**. An exported alias that no declaration mentions reads as dead, and the next person to touch this file has no signal that `log`/`error` are meant to be the same shape.

This is the same case as `eof` losing `TerminalRange` on #1485. Fix is `/** @type {Console} */` above each, plus adding `Console` to the `@import` list in `module.f.mjs` — it isn't currently imported there. I swept the rest of the diff for this shape (an export whose `main` declaration was a bare alias that still exists as an exported type): these two are the only ones.

## Minor — a stale line in the description

> `proof.f.ts` stays TypeScript for now — it has a runtime dependency on `../mock/module.f.ts`, still unmigrated.

`mock` is migrated by the third commit of this same PR, so at this head every runtime import in `effects/memory/proof.f.ts` is already `.f.mjs`. Leaving the proof as `.f.ts` is fine either way — §2 says renaming an implementation never requires renaming its proof — so this is only the stated reason having gone stale mid-PR, not a problem with the outcome. (`eff/proof.f.ts` genuinely is still blocked, on `../proof.f.ts`.)

---

No correctness objection; happy to approve once the two `@type {Console}` annotations are back.

o2alexanderfedin (approved):
Approving. Re-reviewed at `d6cf1b1`. The `Console` annotation is back and `fjs/text/sgr` joins as a fifth module. CI green.

## The fix works

`log` and `error` emit as `Console` again, matching `main`:

```ts
export declare const log: Console;
export declare const error: Console;
```

They no longer appear in my declaration diff against `main` at all.

One note, and I checked before raising it so it stays a note rather than a request. The fix uses an inline cast — `/** @type {Console} */ (writeString('stdout'))` — where the declaration-annotation form would also work:

```js
/**
 * Writes a line to `stdout`. Replaces the retired `Log` effect.
 * @type {Console}
 */
export const log = writeString('stdout')
```

I assumed the cast would be the weaker of the two and tested it: substituting a deliberately wrong value (`writeString` instead of `writeString('stdout')`) produces exactly one error under **both** forms. So there is no safety difference here, and the emitted declaration is identical either way. The only argument for the annotation is consistency with the "annotated const over inline cast" change made on #1486 — genuinely minor, and not worth a commit on its own.

## `fjs/text/sgr` verified

- Token-identical to `main`'s compiled output.
- `csi` and `sgr` change only by the `End` → `_End` / `Csi` → `_Csi` prefix renames; both were non-exported on `main`, so nothing public was renamed.
- `csiWrite`'s signature differs only in the parameter *name* (`{ std }` destructuring became `options`) — same type.
- `proof.f.ts` correctly stays TypeScript: it imports `effects/node/virtual/module.f.ts`, still unmigrated.

## Whole-PR state at this head

- Five modules; four token-identical, and `effects/node` differs only by the `0b111n` → `7n` normalization TypeScript's emitter applies — `main`'s **source** already spells it `0b111n` and this branch preserves it.
- **Public API: 0 removed, 0 non-`_` additions.** The changed signatures are all prefix renames plus two parameter-name differences.
- Links 144/144 with none added, 0 label/target mismatches.
- CHANGELOG covers all five.

## Still stale, from last round

The summary's line about `fjs/effects/memory`:

> `proof.f.ts` stays TypeScript for now — it has a runtime dependency on `../mock/module.f.ts`, still unmigrated.

`mock` migrated in this same PR, so every runtime import in that proof is `.f.mjs` now. Leaving the proof as `.f.ts` is fine — §2 never requires moving it — so this is only the stated reason, not the outcome. Worth a touch-up before merge since the summary is what describes this in history.

PR #1490, opened 2026-08-11
Add initValue1DefaultBranch test case to btree remove module

sergey-shandar:
## Summary
Added a new test case to the btree remove module's proof object to verify the default branch behavior of the `initValue1` function.

## Key Changes
- Added `initValue1DefaultBranch` test case that validates `initValue1` with a null initial value and a nested array structure containing three elements with separators

## Implementation Details
The new test case follows the existing pattern in the proof object and tests the `initValue1` function with:
- A null initial value
- A complex nested structure: `[[['a'], 'b', ['c']], 'd', ['e']]`

This complements the existing `reduceValue0DefaultBranch` test case and ensures comprehensive coverage of default branch scenarios in the btree removal logic.

https://claude.ai/code/session_015FZ8jonsa3mBCwrJaB3FQJ

### DIFF
--- a/fjs/types/btree/remove/module.f.mjs
+++ b/fjs/types/btree/remove/module.f.mjs
@@ -177,5 +177,8 @@ export const proof = {
         reduceValue0DefaultBranch: () => {
             reduceValue0([['leaf']])([['x'], 's', ['y']])
         },
+        initValue1DefaultBranch: () => {
+            initValue1(null)([[['a'], 'b', ['c']], 'd', ['e']])
+        },
     },
 }


o2alexanderfedin (approved):
Approving. Reviewed at `fb07366`. Three lines, and they do exactly what the title says. `npx tsc` clean, 2357 pass / 0 fail — one more than `main`'s 2356, which is the new case.

## The test isn't vacuous

A case under `proof.throw` passes when its body throws, so the thing worth checking is whether it would still pass if the branch it targets stopped throwing. I mutated `initValue1`'s default arm to return instead:

```diff
-            default: { throw 'invalid node' }
+            default: { return [[...n0, v1]] }
```

```
before:  proof.throw.initValue1DefaultBranch(): ok  # EXPECTED TO THROW   → 2357 pass, 0 fail
after:   proof.throw.initValue1DefaultBranch(): error                     → 2356 pass, 1 fail
```

So it genuinely pins the guard. The inputs are right for it too: `a === null` with `n0.length === 3` is the one shape that reaches `default`, since `case 1` and `case 2` take the other arms and a non-null `a` skips the switch entirely.

Not asserting the thrown payload is correct here, per §3's note that whether it threw is normally the part of the contract that matters.

The comment above the proof is the part I'd single out — it explains that these arms are unreachable through the public `remove` API because the sibling is always a Branch in any reachable tree, which is exactly the question a reader has when they see a guard being exercised directly rather than through the API.

## A natural follow-up, not a request

All four merge helpers carry the same guard:

| function | `default: throw 'invalid node'` | covered |
| --- | --- | --- |
| `reduceValue0` | line 55 | yes (pre-existing) |
| `reduceValue2` | line 69 | **no** |
| `initValue0` | line 83 | **no** |
| `initValue1` | line 97 | yes (this PR) |

This takes the set from 1/4 to 2/4, and the two remaining ones have no proof case anywhere in the repo. The comment already describes all four as a group, so closing the other two would be the same two lines each. Out of scope for a PR that says it covers `initValue1` — just flagging it so the gap is visible rather than implied to be closed.

PR #1491, opened 2026-08-11
Migrate fjs/dev/update to .f.mjs

sergey-shandar:
## Summary

- Migrates `fjs/dev/update/module.f.ts` from authored TypeScript to JSDoc-typed `.f.mjs`, per `todo/migrate-typescript-to-mjs.md`.
- No local types to split — it reuses `Mkdir`/`NodeProgram`/`ReadFile`/`WriteFile` from `effects/node/types.ts` and `Effect` from `effects/types.ts`, so no `types.ts` companion is needed.
- `proof.f.ts` stays TypeScript for now — it has a runtime dependency on `effects/node/virtual/module.f.ts`, still unmigrated.
- A dependency leaf: nothing else in the repo imports `dev/update`.

## Test plan

---


o2alexanderfedin (approved):
Approving. Reviewed at `a0bab54` against `main` (`dff17b0`). Two modules — `fjs/dev/update` and `fjs/website`. CI green, nothing to flag.

## The `package.json` fix is the interesting part

`dev-update` and `index-html` pointed at `module.f.ts` paths this PR renames. That is a class neither `tsc` nor the test suite can catch — the scripts are only exercised when someone runs them — so it would have sat broken until it bit somebody. Both run now:

```console
$ npm run dev-update   → OK
$ npm run index-html   → OK
```

I swept the rest for the same shape and found none remaining: every `.f.ts`/`.f.mjs` path in `package.json` and the other config files resolves, and `.github` has no such references at all.

## Everything else

- `npx tsc` clean; 2357 pass / 0 fail, matching `main`.
- **Both modules are token-identical** to `main`'s compiled output.
- **Public API completely untouched** — 0 removed / 0 added / 0 changed for exported type aliases *and* exported const signatures. That is what a leaf migration with no local types should look like, and it matches the summary's claim that no `types.ts` companion was needed.
- Links 144/144, none added, 0 label/target mismatches, and no stale `.f.ts` references to either migrated module outside the CHANGELOG entries themselves.
- CHANGELOG covers both.

**The `proof.f.ts` rationale is accurate this time**, which I checked rather than assumed given it went stale on #1488: both proofs import `virtual` and `emptyState` from `effects/node/virtual/module.f.ts`, and that module is still TypeScript on disk. So §2 genuinely blocks `proof.f.mjs` for both, and the summary says so correctly.

sergey-shandar:
@codex review

sergey-shandar:
@codex review

sergey-shandar:
Checked both findings.

**`@module` blank-line issue — confirmed and fixed** (`d791a69`). Verified via the actual declaration emit (`npx tsc --noEmit false --declaration --emitDeclarationOnly --outDir <scratch>`): before the fix, `fjs/cli/module.f.d.mts` started directly with the `import` statements and the module doc comment was gone entirely; after adding a blank line following `@module` in all three files (`cli`, `dev/update`, `website`), the doc comment is emitted correctly in all three.

**`fjs/cas/README.md:13` link — false positive.** That link, `[\`cas\` CLI](cli/module.f.ts)`, is relative to `fjs/cas/README.md`, so it resolves to `fjs/cas/cli/module.f.ts` — a different, still-unmigrated module (this PR only touched the top-level `fjs/cli`). That file still exists at that path, so the link is unaffected by this PR.

---

sergey-shandar:
@codex review

o2alexanderfedin (approved):
Re-approving at `d791a69`. `fjs/cli` joins, and the two follow-up commits are both pre-empting things that came up earlier in this migration rather than waiting for review to raise them.

## `Handler` stays private — and this is the better fix

`3098a08` keeps `Handler` non-exported in `fjs/cli/types.ts`. Verified against `main`: it emits as a bare `type Handler<O extends NodeOp> = …` on both sides, no `export`, so the privacy is exactly preserved.

Worth naming why this is cleaner than the `_Byte`/`_NodeList`/`_Type0` outcomes earlier: those types had to move into JSDoc `@typedef`s, which have no non-exported form, so the `_` prefix was the only marker available. `Handler` moved into a `types.ts` — real TypeScript — where a plain non-exported `type` is expressible. Reaching for the stronger option when the file format allows it is the right instinct.

## The `@module` blank lines are load-bearing

`d791a69` adds a blank line after the `@module` block in `fjs/cli/types.ts`, `fjs/dev/update/module.f.mjs`, and `fjs/website/module.f.mjs`. All three have an `import` as the next statement, which is precisely the shape where TypeScript folds the header into the import's leading comment and drops it on elision. I confirmed the header survives emit in all three (`types.js` for `cli`, the `.d.mts` for the other two). Fixing that at authoring time rather than discovering it in a later sweep is the outcome I was hoping for after the 18-file case.

## Everything re-verified at this head

- `npx tsc` clean; 2357 pass / 0 fail, matching `main`.
- **All three modules token-identical** to `main`'s compiled output.
- **Public API untouched: 0 removed / 0 added / 0 changed**, on both exported type aliases and exported const signatures.
- Links 144/144, none added, 0 label/target mismatches.
- Every `.f.ts`/`.f.mjs` path in `package.json` still resolves — including `fjs/cli`'s arrival not stranding one.
- CHANGELOG covers all three.

sergey-shandar:
@codex review

PR #1493, opened 2026-08-11
Migrate fjs/ci/nix to .f.mjs

sergey-shandar:
## Summary

- Migrates `fjs/ci/nix/module.f.ts` from authored TypeScript to JSDoc-typed `.f.mjs`, per `todo/migrate-typescript-to-mjs.md`.
- Splits the public `NixJob` type into a sibling `types.ts`.
- `proof.f.ts` stays TypeScript for now — it has runtime dependencies on `effects/node/virtual/module.f.ts` and `ci/node/module.f.ts`, both still unmigrated.
- Updates the two dependents (`fjs/ci`, `fjs/ci/node`) to import runtime values from `.f.mjs` and types from `types.ts` separately, and fixes stale `module.f.ts` doc references across three files, including two with stale line-number citations.

## Test plan

---


o2alexanderfedin (approved):
Approving. Reviewed at `5c07521` against `main` (`caa5894`). Three modules — `fjs/ci`, `fjs/ci/nix`, `fjs/ci/node`.

## The generated CI workflow doesn't move

`fjs/ci` is the generator for this repo's own workflow, so behaviour preservation is checkable at the output rather than only at the source: I ran `npm run ci-update` on the branch and `git status` came back empty. The workflow the migrated generator produces is byte-for-byte what is committed.

## Source-level equivalence

`fjs/ci` and `fjs/ci/nix` are token-identical to `main`'s compiled output. `fjs/ci/node` differs by exactly one thing — a parameter rename inside the **non-exported** `installNode` (`version` → `v`) — which cannot be observed from outside the module.

The two changed declarations are both inert:

- `ci`: `({ nodeExtra }: Setup)` → `(setup: Setup)` — parameter name only, same type.
- `nodeMainSteps`: `typeof platformNodeSteps` → `(version: string) => readonly MetaStep[]`. I checked these are the same type and they are: `main` emits `platformNodeSteps: (version: string) => readonly MetaStep[]` and `nodeMainSteps: typeof platformNodeSteps`, so the branch just spells the expansion. Unlike an alias being dropped, nothing is orphaned here — `typeof` queries a value, not a named type, and `platformNodeSteps` is still exported and still referenced.

## Everything else

- `npx tsc` clean; 2357 pass / 0 fail, matching `main`.
- **Public API: 0 removed, 0 added**, and the two changed entries are the ones above.
- Links 144/144, none added, 0 label/target mismatches.
- Both new `types.ts` files keep their `@module` header through emit.
- CHANGELOG covers all three.
- `fjs/ci` and `fjs/ci/nix` proofs correctly stay TypeScript — both import `effects/node/virtual/module.f.ts`, still unmigrated. `fjs/ci/node`'s proof had no such dependency and moved to `.f.mjs`, which is the right discrimination rather than a blanket rule.

sergey-shandar:
@codex review

sergey-shandar:
@codex review

PR #1495, opened 2026-08-12
Document native JavaScript syntax for eDSLs

sergey-shandar:
## Summary

Document a general FunctionalScript eDSL design principle in `AGENTS.md`:

- reuse ordinary JavaScript literals and structures when their JavaScript meaning already matches the DSL meaning;
- introduce constructors/tags only for concepts that JavaScript cannot express directly or where the ordinary shape would be ambiguous;
- allow the ergonomic authoring representation to normalize to a different internal tagged AST when useful;
- apply the same principle to the future FunctionalScript function AST.

This captures the direction already visible in RTTI and the shared NaNVM operator-test data, where ordinary values are more readable than tagging every primitive and collection.

## Impact

Documentation/design guidance only. No runtime or public API changes, so no CHANGELOG entry is needed.

## Validation

Compared the branch against `main`: only `AGENTS.md` changes, with 24 added lines and no deletions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -423,6 +423,30 @@ have the caller check the `null` result rather than a precomputed bound.
 CLI parameters are preferred over environment variables when adding new
 features.
 
+### 5.8 Reuse JavaScript syntax in embedded DSLs
+
+**An embedded DSL should reuse ordinary JavaScript values and constructions when
+their JavaScript meaning already matches the DSL meaning.** Prefer a number,
+string, boolean, `null`, array, or object literal directly over wrapping the same
+value in a tagged representation such as `['number', 3.14]` or
+`['array', [...]]`. The host language already supplies syntax, tooling, type
+inference, and reader intuition for those concepts; duplicating that vocabulary
+adds ceremony without adding semantics.
+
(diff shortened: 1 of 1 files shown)


o2alexanderfedin (commented):
Reviewed at `023c7347dd707e097e05ae0ae754f2a7c73b424c`.

This and #1496 are the same change written twice — both insert `### 5.8` at the same
point in `AGENTS.md` from the same pre-image (`f4164b61d`), so they conflict by
construction and only one should land. The text here is accurate and §8.3 does exempt
`AGENTS.md` from a CHANGELOG entry, so this is purely a question of which draft to keep.

I put the comparison on #1496, which is the one I'd keep — it carries the RTTI and NaNVM
evidence in the added text rather than only in the PR body, and it states which layer the
public API should be optimized for rather than only permitting the two layers to differ.

The one substantive thing this version does that #1496 doesn't is name existing eDSLs —
"HTML, BNF, RTTI, generated-code/test descriptions" — as targets for the principle. That
cuts both ways: BNF's tagged encoding is deliberate and settled, so naming it reads as an
invitation to revisit a closed design. If you keep this one instead, that list is the part
I'd narrow.

Still a draft, so no approval either way.

PR #1497, opened 2026-08-12
Migrate fjs/media/json to .f.mjs

sergey-shandar:
## Summary
- Migrates `fjs/media/json/module.f.ts` from authored TypeScript to JSDoc-typed JavaScript (`.f.mjs`), continuing the `.f.ts` → `.f.mjs` migration series
- Splits the rtti schemas (`primitive`, `unknown`, `object`, `array`) into a sibling `rtti.f.mjs`, and the derived TypeScript types (`Primitive`, `Unknown`, `Object`, `Array`, `Entry`, `MapEntries`) into a sibling `types.ts`
- Updates all ~25 dependents across the tree to import runtime values (`parse`, `stringify`, `setProperty`) from `module.f.mjs`/`rtti.f.mjs` and types from `types.ts`
- Removes a stale CHANGELOG entry for `fjs/fsc/json.f.ts`/`fjs/fsc/bnf.f.ts` that had already landed on `main` via #1494

## Test plan
- [x] `npm test` (tsc typecheck + full proof suite): 2357/2357 passing


o2alexanderfedin (commented):
Reviewed at `f10c9f7f2ce1a38b967ccd2d1badfb9a97a34232`. The redo is much better placed than
the version #1494 reverted — the derived types go to `types.ts`, where ordinary TypeScript
visibility applies, so the unprefixed-`@typedef` problem that blocked that attempt is gone.
Three things to fix before this can land, one of them CI-breaking.

## 1. `npm run prepack` fails on a clean tree; CI runs it

`ci.yml` runs on `pull_request` and its last step is `npm pack`, which triggers
`prepack` — so this fails CI. From a freshly cleaned tree:

```
main         npm run prepack → exit 0, 0 errors
f10c9f7f     npm run prepack → exit 2, 5 errors
```

`npx tsc --noEmit` passes, which is why it isn't visible from the usual check. The
difference is that `prepack` is two passes:

```
tsc --noEmit false --emitDeclarationOnly && tsc --noEmit false --declaration false
```

Pass 1 writes the declarations; pass 2 then resolves `./rtti.f.mjs` through the emitted
`rtti.f.d.mts` rather than the authored `.mjs`. And declaration emit cannot express the
self-referential schema — the emitted `unknown` comes out as

```ts
export declare const unknown: () => readonly ['or', Or<[…]>,
    Type1<"record", () => readonly ['or', Or<[…]>,
        Type1<"record", /*elided*/ any>, Type1<"array", /*elided*/ any>]>, …]
```

so `Ts<typeof unknown>` computed from that is not the hand-written `Unknown`, and the pin
at `fjs/media/json/types.ts:16` fails:

```
fjs/media/json/types.ts(16,24): error TS2344: Type 'false' does not satisfy the constraint 'true'.
```

The other four errors are that failure propagating —
`fjs/protocol/json_rpc/module.f.ts:117`, `fjs/protocol/mcp/module.f.ts:245` and `:401`,
`fjs/protocol/mcp/stdio/module.f.ts:61`, all ending in `Type 'undefined' is not assignable
to type 'Unknown'`.

Worth being precise about what changed, because the elision itself is not new: main's
emitted `fjs/media/json/module.f.d.mts` contains exactly the same `/*elided*/ any`. What
is new is that anything *evaluates a pin against it*. On main the schemas and
`Assert<Equal<Unknown, Ts<typeof unknown>>>` live in the same authored file, so the pin is
only ever checked against precisely-inferred source. Splitting the schemas into
`rtti.f.mjs` turns that into a cross-file import, and on pass 2 that import goes through
the lossy declaration. Keeping the schemas and the pin in one authored file is the shape
that works today; which way to resolve it is your call.

## 2. The CHANGELOG loses an entry and doesn't gain one

The diff is five deletions and no additions.

The deleted entry is the `fjs/fsc/json.f.ts` / `fjs/fsc/bnf.f.ts` migration. The PR
description calls it stale because it "had already landed on `main` via #1494" — but that
is what makes it correct, not stale. `fjs/fsc/json.f.mjs` and `fjs/fsc/bnf.f.mjs` are on
main, main's CHANGELOG carries that entry exactly once, and this branch deletes it. That
drops the only release note for a breaking change that has already shipped.

Separately, this PR migrates `fjs/media/json/module.f.ts` to `.f.mjs`, moves the schemas to
`rtti.f.mjs` and the types to `types.ts`, and updates ~25 importers — a breaking change with
no entry of its own (`grep -c 'pull/1497' CHANGELOG.md` → 0). Every sibling migration in
this series has one.

## 3. Three markdown references still point at the removed `module.f.ts`

Resolving every relative link against the filesystem: 144 broken on main, 147 here, and the
three added are all the renamed file.

```
fjs/effects/node/todo/readjsonfile-writejsonfile-helpers.md → ../../../media/json/module.f.ts
fjs/media/json/todo/number-edge-cases.md                    → ../module.f.ts
fjs/media/json/todo/standard-parse-serialize.md             → ../module.f.ts
```

The ~25 code importers were all updated; this is just the doc sweep coming up short.

## Non-blocking

The #1494 entry for `fjs/emergent_testing` says its `proof.f.ts` "stays TypeScript for now
(it depends on `media/json/module.f.ts`, still unmigrated)". After this PR that reason no
longer holds — every runtime import in that proof is `.f.mjs`, including
`../media/json/module.f.mjs` — so under §2 it becomes eligible to migrate. Not something
this PR has to do, but the parenthetical will be wrong once this lands.

## What checks out

`npx tsc --noEmit` exits 0. `npm test` gives 2357 pass / 0 fail, matching main exactly.
The public type surface adds only `MapEntries`; it was module-private on main and is
exported here because `module.f.mjs` `@import`s it, and publishing it under an ordinary
name is a legitimate call under `fjs/fsc/README.md` — types split into `types.ts` use
ordinary TypeScript visibility rather than the `_` workaround, and a formerly-private type
may be published deliberately. The four schema consts move to `rtti.f.mjs` with byte-identical
signatures. Comparing main's compiled module against this branch's `module.f.mjs` +
`rtti.f.mjs` gives 8 declarations on each side with matching bodies; the only deltas are the
intended split, an import-alias rename in `primitive` (`rttiBoolean`→`boolean`, same values),
and a `const de = definedEntries` binding — the usual annotation extraction.

o2alexanderfedin (commented):
Reviewed at `e4e820949b7016fdb6da73a2050841fc6a2e0f05`. Moving the pin into `rtti.f.mjs` is
the right structural instinct — `typeof unknown` is now inferred from the authored source
rather than through the emitted declaration, and keeping `primitive` as the only cross-file
import from `types.ts` is safe, since `primitive` isn't self-referential and its emitted
signature carries no elision. But `npm run prepack` still fails from a clean tree:

```
main       exit 0, 0 errors
f10c9f7f   exit 2, 5 errors
e4e8209    exit 2, 4 errors     ← the pin error is gone; the other four are unchanged
```

The one error that disappeared is the pin itself. That's worth pausing on, because it means
the pin was the *detector*, not the defect — silencing it removed the diagnostic without
changing what it was diagnosing. The remaining four bottom out at the same place they did
before:

```
Type 'fjs/types/rtti/ts/types'.Unknown is not assignable to type 'fjs/media/json/types'.Unknown
  Type 'undefined' is not assignable to type 'Unknown'
```

So on prepack's second pass, `Ts<>` over the JSON schemas still degrades to rtti core's
`Unknown` — the one that admits `undefined` — instead of the JSON `Unknown` that excludes it,
exactly as the `/*elided*/ any` in the emitted `rtti.f.d.mts` would predict. The elision is
still the mechanism; it now bites at the three consumers rather than at the assert.
`fjs/protocol/json_rpc/module.f.ts:117`, `fjs/protocol/mcp/module.f.ts:245` and `:401`,
`fjs/protocol/mcp/stdio/module.f.ts:61`.

Which suggests the split itself is the thing to reconsider, not where the pin sits: as long
as any file outside `rtti.f.mjs` derives a type from those schemas through the emitted
declaration, it gets the degraded answer. On main nothing does, because the schemas and
every type derived from them share one authored file.

`npx tsc --noEmit` still exits 0, so this remains invisible to the usual check and visible to
CI's `npm pack` step.

The other two items from my previous comment are untouched, which I assume is just sequencing:

- `grep -c 'pull/1497' CHANGELOG.md` → 0, and the deleted `fjs/fsc/json.f.ts` / `bnf.f.ts`
  entry is still absent (`grep -c 'fjs/fsc/json.f.ts'` → 0) even though those files are on main.
- Broken relative markdown links: still 147 here against 144 on main — the same three
  references to the removed `fjs/media/json/module.f.ts`.

PR #1500, opened 2026-08-12
42

sergey-shandar:


o2alexanderfedin (approved):
Reviewed at `822953d0ce2acdb12e0d9e9643dfb3fdbabd6efa`. Approving — this is a pure
reorganization and I was able to prove that rather than assume it.

**It converges on the existing majority style rather than inventing one.** `main` already has
13 `.f.mjs` modules carrying their `@import`s inside the `@module` header block, including
`fjs/media/json/parser/module.f.mjs` from #1499. AGENTS.md documents `@import` usage but not
placement, so there was no rule to check against — the relevant question was which pattern the
repo already leans on, and this moves toward it.

**The change is semantically inert.** The risk worth taking seriously with this edit is a
`@import` quietly failing to resolve in its new position and the affected type silently
widening to `any`, which `tsc` will not always shout about. Diffing the emitted declarations
against `main` across the whole tree gives **0 type-alias differences and 0 const-signature
differences**, so every published signature is exactly what `main` produces. Nothing degraded.

Backing that up at the source level: of the 21 changed files, **15 are token-identical** to
`main` once comments are normalized away — as they must be, since only comment blocks moved.
The other 6 are the ones where imports are real syntax rather than comments (five `.f.ts`
proofs plus `crypto/sign/proof.f.mjs`), and normalizing import order and quote style shows all
six are **unchanged in substance** — no statement added, removed, or altered.

`crypto/sign/proof.f.mjs` also switches its specifiers from double to single quotes, which
matches both the rest of that file and the repo at large; no double-quoted specifiers remain in
any changed file.

**Everything else.** `npx tsc --noEmit` 0, `npm run prepack` 0 errors from a clean tree,
`npm test` 2368 pass / 0 fail against main's 2368, broken markdown links 144 on both sides,
branch level with `origin/main`. No CHANGELOG entry, which I think is right — the surface is
provably byte-identical, so there is nothing to tell a consumer.

One aside, explicitly not a request: `crypto/sign/proof.f.mjs` has no `@module` header. That
predates this PR — it previously opened straight into `import { utf8 }` — so nothing regressed
here, but it is now the one file in this set whose leading block is `@import`-only.

PR #1503, opened 2026-08-12
Migrate fjs/dev/package_json to .f.mjs

sergey-shandar:
## Summary
- Migrates `fjs/dev/package_json/module.f.ts` and `proof.f.ts` to JSDoc-typed JavaScript (`.f.mjs`) under the stage-1 TypeScript-to-mjs migration.
- `PackageJson` and `JsonTextError` stay as JSDoc typedefs directly in the module — no sibling `types.ts` needed, since nothing outside this directory imports the types.
- No importers elsewhere in the repo reference this module, so no other files needed updating.

## Test plan
- [x] `npm run prepack` (declaration-emit + plain `tsc` checks) — clean
- [x] `npm test` — 2376/2376 passing
- [x] CHANGELOG.md `## Unreleased` entry added with `**BREAKING CHANGES:**` marker

---


o2alexanderfedin (commented):
Reviewed at `291f43c`. I started at `58bca9a`, and the head moved to `291f43c` (adding
`8684183` "Migrate fjs/cas/evo to .f.mjs" and `291f43c`) while I was working, so I re-ran the
whole battery on `291f43c` — everything below is from that head, not the earlier one.

Mostly clean, and the migration itself is exact. Three things worth fixing, one of them a
convention regression that no gate will catch.

## What passed

- `npm run prepack` from a freshly cleaned tree — exit 0, both passes.
- `npm test` — 2368/2368 pass, 0 fail. `origin/main` at `ede2a96` is also 2368/2368, so the
  count matches exactly, which is what a behaviour-preserving migration should do. (The PR
  body's "2376/2376" is stale — it predates the rebase onto current `main`.)
- **Token equivalence.** I compared `main`'s *compiled* `.f.js` against this PR's *authored*
  `.f.mjs` for all 13 migrated files (`fjs/cas/evo`, `fjs/dev/package_json`, `fjs/fsm`,
  `fjs/media`, `fjs/media/revision`, `fjs/protocol/json_rpc`, `fjs/protocol/mcp/stdio`,
  modules and proofs), normalized and with the `.f.js`/`.f.mjs` specifier difference
  neutralized. Ten are byte-for-byte identical after normalization. The only three
  differences in the entire set are:
  - `fjs/cas/evo/proof`: `vec(5n)(21n)` → `vec(5n)(0b10101n)` — same value, binary literal;
  - `fjs/media/revision/module`: `const lock` → `export const _lock`;
  - `fjs/protocol/json_rpc/module`: `const id` → `export const id`.

  That is a very strong result — the migration changed nothing else at all.
- **Public surface, both axes.** Exported type aliases: two additions, `_Rule` and `_Dfa` in
  `fjs/fsm`, both correctly `_`-prefixed per §6.2 since a JSDoc `@typedef` has no
  non-exported form; nothing removed; the three "changed" entries are just those renames
  flowing through (`Grammar = List<_Rule>`, `LockMap = Ts<typeof _lock>`,
  `RpcError = Ts<typeof errorSchema>`). Exported const signatures: nothing removed, nothing
  widened to `any`. Every `any` I grepped in the emitted declarations is English prose in a
  doc comment ("any JSON that satisfies…"), not a TypeScript `any`, and there is no `elided`
  anywhere.
- **Stage 1 discipline (§2).** No migrated `.f.mjs` — module or proof — has a relative runtime
  `.ts` / `.f.ts` import left. Checked all 13.
- **Link check.** Broken-link *set* compared against `main`, not just the count: byte-identical.
  Given how many `todo/*.md` specifier references this PR rewrites, that is worth stating —
  nothing was stranded.
- **Module header blank line (§4).** All six migrated `module.f.mjs` files have the required
  blank line after the header block, and `grep -c '@module'` is 1 on each emitted
  `module.f.d.mts`, so no header was folded into import trivia.
- **CHANGELOG.** Seven entries, one per migration, all with the `**BREAKING CHANGES:**` prefix,
  all linking only `/pull/1503`, released sections untouched.
- `291f43c`'s explicit `Effect<MemOp, Result<Hash, string>>` cast is type-only — `fjs/cas/evo/module`
  is still token-identical to `main`'s compiled output, so it changes no runtime behaviour.

## 1. The four new `types.ts` files have no `@module` header (§4)

`fjs/cas/evo/types.ts`, `fjs/media/types.ts`, `fjs/media/revision/types.ts` and
`fjs/protocol/json_rpc/types.ts` all begin directly with an `import type` line. There is no
leading `/** … @module */` block on any of them.

This is not a judgement call about whether `types.ts` counts as an "implementation module" —
the repo has already answered it. On `origin/main`, **75 of 75** tracked `*/types.ts` files
carry a `@module` header. At this head there are 79, and the only four without are exactly the
four this PR adds:

```
fjs/cas/evo/types.ts
fjs/media/revision/types.ts
fjs/media/types.ts
fjs/protocol/json_rpc/types.ts
```

The consequence is visible in the emitted output: `fjs/media/revision/types.d.ts`,
`fjs/media/types.d.ts` and `fjs/protocol/json_rpc/types.d.ts` all emit with `grep -c '@module'`
= 0 and no module documentation at all, while every other `types.d.ts` in the package ships one.
`types.ts` is precisely where §4 warns this is most dangerous, because every import is
`import type` and can be elided entirely.

Worth noting the contrast: #1489 in flight right now went the other way and spent a whole
commit (`a0a8985`) bringing its headers into line with §4, and its `fjs/nanvm/types.ts` does
carry one.

## 2. `id` is published unprefixed while the identical case became `_lock`

This PR hits the same situation twice — a schema const that was module-private but now has to
cross a file boundary so the sibling `types.ts` can say `typeof` on it — and resolves it two
different ways:

- `fjs/media/revision`: `const lock` → `export const _lock`, and `types.ts` uses
  `Ts<typeof _lock>`. The `_` marks it private-by-convention. Good.
- `fjs/protocol/json_rpc`: `const id` → `export const id`, plain. `types.ts` uses
  `Ts<typeof id>`.

So `id` becomes a genuine new public runtime export of `fjs/protocol/json_rpc` — on `main` it is
`declare const id`, module-private. It is the only unprefixed addition to the exported const
surface in the whole PR. `_id` would match both §6.2's intent and what this same PR did one
directory over. If publishing it *is* deliberate, the CHANGELOG entry should say so, since a
new public export is exactly the kind of thing that entry is for; right now it mentions only
the type moves.

## 3. The `fjs/media/module` CHANGELOG entry's stated reason is stale at the tip

That entry ends:

> `fjs/media/proof.f.ts` stays TypeScript for now: it still depends on `fjs/media/revision`,
> which is not yet migrated

A later commit in this same PR (`12b7ddb`) migrates `fjs/media/revision`. At `291f43c` every
runtime import in `fjs/media/proof.f.ts` is already `.f.mjs` — `../asserts/module.f.mjs`,
`../types/bit_vec/module.f.mjs`, `./module.f.mjs`, `./revision/module.f.mjs`,
`../types/rtti/module.f.mjs` — so the blocker the entry names no longer exists by the time the
PR merges. Either drop the clause, or say what actually holds it back now. (Leaving the proof
as TypeScript is fine on its own; §2 permits it. It is only the stated reason that is wrong.)

## Not verified

Coverage. `npm run cov` reports a vacuous `100.00` over 0 tests in this Node v23 environment
and does the same on `main`, so it says nothing about this PR either way.

o2alexanderfedin (commented):
Re-reviewed at `4c8f14d`, previously reviewed at `291f43c`. All three earlier findings are
addressed — thank you. One of them has come back in two files added since.

## The three previous findings

1. **`types.ts` `@module` headers — fixed for the four files named.** `fjs/cas/evo/types.ts`,
   `fjs/media/types.ts`, `fjs/media/revision/types.ts`, and `fjs/protocol/json_rpc/types.ts` all
   carry a header now, and it survives declaration emit (`grep -c '@module'` = 1 on each emitted
   `.d.ts`), so the blank line is in the right place.
2. **`id` → `_id` — fixed.** `fjs/protocol/json_rpc/module.f.mjs` exports `_id`, `types.ts`
   imports it as `typeof _id`, and it now matches `fjs/media/revision`'s `_lock` exactly. Both
   show up in the const surface as new `_`-prefixed exports that main did not have at all
   (neither `lock` nor `id` was exported on main), which is the intended §6.2 shape.
3. **Stale `fjs/media/module` CHANGELOG clause — fixed.** The claim that `fjs/media/proof.f.ts`
   stays TypeScript is gone, `proof.f.ts` is migrated, and the correction is called out
   explicitly in the `_id` entry.

## One finding: the `@module` header regressed in the two new `types.ts` files

`4326af5` adds two `types.ts` files, and neither has a module header:

```
fjs/djs/types.ts       — no @module
fjs/djs/ast/types.ts   — no @module
```

Every other `types.ts` in the tree has one — 76/76 on `main`, and 80/82 here, the two misses
being exactly these. It shows up in declaration emit the same way the earlier four did:

```
fjs/djs/types.d.ts               @module=0
fjs/djs/ast/types.d.ts           @module=0
fjs/media/json/types.d.ts        @module=1
fjs/protocol/json_rpc/types.d.ts @module=1
```

`fjs/djs/ast/types.ts` does open with a JSDoc block, but it documents `AstModule`, not the
module, so there is no header for the emitted declaration to carry.

Worth noting because the CHANGELOG entry for this commit describes `fjs/djs/types.ts` as
"hand-authored, following the pattern `fjs/media/json/types.ts` established" — and that
exemplar leads with a full `@module` block. Following it the rest of the way would close this.

Nit, same commit family: the `fjs/djs/serializer` entry says it updates "five proof files";
`4c8f14d` updates six (`fjs/djs/ast`, `fjs/djs/parser`, `fjs/djs/tokenizer`,
`fjs/djs/transpiler`, `fjs/js/tokenizer`, `fjs/media/json/tokenizer`), on top of renaming
`fjs/djs/serializer/proof`.

## Everything else verified clean

- `npm run prepack` from a freshly cleaned tree — exit 0 here **and** on `main`. Both passes.
- `npm test` — **2495 pass / 0 fail**, byte-identical to `main` (2495/0). Clean tree before each
  run, so no `.f.js`/`.f.mjs` double-discovery.
- **Public surface, dual axis.** Every added type is `_`-prefixed — `_Dfa`, `_FoldObjectState`,
  `_GetConstsState`, `_RefCounter`, `_RefLookup`, `_Refs`, `_Rule`, `_RunState`. The previously
  public `MapEntries`/`Dfa` positions narrow to `_`-prefixed ones. No unprefixed additions,
  nothing widened to `any`, no `elided` in any migrated module's declarations. The two
  `_MapEntries` live in different modules (`fjs/djs/serializer`, `fjs/media/json`) and do not
  collide.
- **Token equivalence** (main's compiled `.f.js` vs the authored `.f.mjs`): `fjs/djs/ast/module`,
  `fjs/djs/serializer/module`, and `fjs/djs/serializer/proof` are all identical under `norm5`.
  `fjs/media/proof` differs only in the migrated specifiers and in `1_000n` vs `1000n` — and
  `1_000n` is what `main`'s own source already says, so that one is `tsc` stripping the
  separator on the way to `.f.js`, not a change.
- **§2 pairing** holds for both newly migrated proofs: every runtime import in
  `fjs/djs/serializer/proof.f.mjs` and `fjs/media/proof.f.mjs` is already `.f.mjs`.
- **Link check** — 163 broken links here, 163 on `main`, identical sets. The `todo/*.md`
  renames stranded nothing.
- Rust gates and the generator round-trip were not run and are not needed: relative to
  `origin/main` this PR touches no `nanvm-lib/` or `fjs/nanvm/` file. Those showed up only
  because `9aa4685` merged `main` in.

Merge base is `origin/main`'s tip, so the diff above is genuinely the PR's own 71 files.

o2alexanderfedin (commented):
Re-reviewed at `1f93485` (previous review was at `4c8f14d`). The head moved twice while I was working — first pass at `b91f3f0`, re-run at `2d0aceb`, and a final full re-run at `1f93485`. Everything below is measured at `1f93485`.

## Previous round's two items

**1. `@module` on the `djs` `types.ts` files — fixed.** `fjs/djs/types.ts` and `fjs/djs/ast/types.ts` both carry the header now, and their emitted declarations each show `grep -c '@module'` = 1.

**2. "five proof files" → "six" — fixed.** The serializer entry now says "six proof files" and lists six.

## Verified

- `npx tsc --noEmit` → 0.
- `npm run prepack` from a freshly cleaned tree → 0.
- `npm test` → `pass: 2495, fail: 0, total: 2495`, exactly matching `origin/main` (`b76018c`) in the baseline worktree. A behaviour-preserving migration should match, and it does.
- Public surface (both axes, `prepack`ed in both trees): 32 exported type aliases added, 0 removed, 6 changed; the 6 changes are only re-pointing at renamed `_`-prefixed neighbours (`Grammar`→`_Rule`, `Id`→`typeof _id`, `LockMap`→`typeof _lock`, `RpcError`→`typeof errorSchema`) plus the two typedefs that gained a second declaration site. Const signatures: 2 added (`_lock`, `_id`, both documented), 0 removed, 7 changed — all `_`-prefix renames. **No signature widened to `any`.**
- Token equivalence for every migrated module, comparing main's compiled `.f.js` against the PR's authored `.f.mjs` with a real JS tokenizer (my usual `norm5`/`norm6` scripts strip comments with a regex and mangle the `'/'`, `'//'` and `'/*'` string literals that the djs tokenizer and parser are full of, so they were unusable here). Modulo redundant parentheses the streams are identical everywhere; the only residual differences are trailing commas, dropped `import {} from '…'` shells that the old `.f.ts` compile left behind, one unused `_` binding, and two numeric literals respelled (`21n`→`0b10101n`, `1000n`→`1_000n` — same values). Negative-controlled: the comparator flags a one-token mutation.
- Link check: broken-link *set* identical to main (163 lines both); the single textual difference is one entry's own source label moving from `djs/parser/module.f.ts:73` to `djs/parser/module.f.mjs:66`.
- Rust gates and the generator round-trip: **still justifiably skipped** — I re-checked rather than assuming. Against the merge-base the PR touches exactly one path under `nanvm-lib/`, and it is `nanvm-lib/todo/mvp-roadmap.md`. No `.rs` file and nothing under `fjs/nanvm/`.
- `@module` on the `types.ts` added by the latest commits (`fjs/protocol/mcp/types.ts`): present, emits 1.

## Two things

**1. `fjs/djs/tokenizer/types.ts` has no `@module` header** — same class as the item fixed last round, on a file added in the same push. It starts straight at `import type { … }` on line 1, so `fjs/djs/tokenizer/types.d.ts` emits `@module` = 0. The near-miss is visible in the commit order: the file arrived in "Migrate fjs/djs/tokenizer to .f.mjs", and the follow-up "Address review feedback" commit right after it added headers to `fjs/djs/types.ts` and `fjs/djs/ast/types.ts` but not to this one; every `types.ts` added since (`parser`, `protocol/mcp`) does have it.

Repo-wide this is the only new gap. Scanning every emitted `types.d.ts`/`types.d.mts`:

- `origin/main`: 2 missing — `fjs/types/rtti/types.d.ts`, `fjs/types/rtti/ts/types.d.ts` (both pre-existing, not yours)
- this PR: those same 2, plus `fjs/djs/tokenizer/types.d.ts`

**2. `UnknownConst` becomes a public unprefixed export** (`fjs/media/json/schema/module.f.mjs:62`). On main it is a plain non-exported `type UnknownConst` in `module.f.ts`, and main's declaration emits it as `type UnknownConst = {` — no `export`. As a JSDoc `@typedef` it emits as `export type UnknownConst = {` (`module.f.d.mts:48`), so a name that was module-private is now part of the published surface. It is implementation-only — its sole use is the `$out` annotation on `unknown` inside the same file, so it does not even cross a file boundary.

This is exactly the case AGENTS §6.2 covers, and this PR already answers it that way everywhere else: the serializer entry says its five types "stay as `_`-prefixed JSDoc typedefs", and `id`→`_id` and `lock`→`_lock` got renamed for precisely this reason with a CHANGELOG note explaining it. `_UnknownConst` would keep it consistent. It is also the one unprefixed addition in the whole type-surface diff, and the schema CHANGELOG entry doesn't mention it — it says only that "`Unknown` stays as a JSDoc typedef in the module … no external importers of the type", which reads as if nothing new became public.

## Not a finding, recorded so it isn't re-raised

The `Unknown` typedef's description no longer sits above `export type Unknown` in the emitted declaration, unlike main. I chased this and it is not a regression you introduced: TypeScript never attaches a JSDoc description to the `export type X` line it synthesises from an `@typedef`, under any block layout — I tested both the split blocks as written and a single merged block on a clean tree, plus a probe typedef as a negative control. The text is still in the `.d.mts`, carried verbatim in the copied JSDoc block lower in the file. Nothing to change.

Everything else — CHANGELOG entries present for all migrations in this push and linking only the PR, `@module` on every migrated `module.f.mjs`, proofs correctly carrying no header as on main — checks out.

o2alexanderfedin (commented):
Re-review at `ac7d420` (previous review was at `1f93485`). The head moved twice while I was
working — `23746fe` → `43e7787` → `ac7d420` — so I re-ran the gates at each; everything below
is from the final run at `ac7d420` against `origin/main` @ `b76018c` (the merge base, so the
branch is not behind).

## Both findings from the last review are fixed

1. **`fjs/djs/tokenizer/types.ts` module header** — resolved. The file now opens with a
   `/** … @module */` block followed by a blank line, and the emitted `fjs/djs/tokenizer/types.d.ts`
   carries `@module` (count 1). Repo-wide the `types.d.ts` gap set is now exactly main's two
   pre-existing ones (`fjs/types/rtti/types`, `fjs/types/rtti/ts/types`) — no new gap. Across
   *all* emitted declarations the PR introduces zero new `@module` misses and actually closes
   five that exist on main (`fjs/dev/package_json/module.f`, `fjs/djs/tokenizer/module.f`,
   `fjs/mcp/cas/module.f`, `fjs/mcp/evo/module.f`, `fjs/media/json/schema/module.f`).

2. **`UnknownConst` → `_UnknownConst`** — resolved, and the CHANGELOG entry for
   `fjs/media/json/schema` was amended to record the rename. The type-surface diff now has
   **32 added type aliases, all `_`-prefixed, zero unprefixed additions, zero removals**.

## What I verified at `ac7d420`

- `npm test` — 2495 pass / 0 fail, exactly matching `origin/main` (2495 / 0).
- **Type surface**: 32 added, all `_`-prefixed; 0 removed; nothing widened to `any`.
  The six "changed" entries are internal-reference renames only, with identical type identity
  (`Grammar`→`_Rule`, `Id`→`typeof _id`, `LockMap`→`typeof _lock`, `RpcError`→`typeof errorSchema`,
  and two same-name-different-file collisions).
- **Const surface**: the only "additions" are exports whose declaration moved from `module.d.ts`
  to `module.d.mts` as part of the `.ts` → `.mjs` renames, with byte-identical signatures — except
  `asyncRun`, which drops its unused inner type parameter
  (`<T, E extends Effect<O, T>>` → `<T>`). `E` appeared in no parameter position, so it could only
  ever resolve to its constraint; I grepped every `asyncRun` call site and none supplies inner type
  arguments, so this is dead weight being removed, not a behaviour change.
- **Token equivalence** for all six modules migrated since the last review
  (`fjs/mcp/module`, `fjs/mcp/proof`, `fjs/module`, `fjs/proof`, `fjs/effects/module`,
  `fjs/types/result/module`, `fjs/effects/node/memory/module`, `fjs/effects/node/memory/proof`):
  main's *compiled* output vs the PR's authored source is identical modulo parentheses and
  trailing commas. Every paren delta is accounted for — TS emit writes `(x) =>` where the JSDoc
  source writes `x =>`, and the JSDoc cast form requires its own parens where `as T` did not
  (e.g. `(unwrap(x).main as NodeProgram)` → `(/** @type {NodeProgram} */ (unwrap(x).main))`).
  Trailing commas are argument lists TS emit collapses onto one line.

  **Method note:** I did *not* use regex comment-stripping for this — I tokenized both sides with
  a real JS lexer (acorn's `tokenizer`), dropping semicolons and neutralizing module-specifier
  extensions. I negative-controlled it first: a one-token mutation of a source file produced a
  diff, confirming the comparison is not vacuously passing.
- **Link check** — 163 broken links in both trees; the sets are identical apart from one
  pre-existing broken link whose containing file was renamed
  (`djs/parser/module.f.ts:73` → `djs/parser/module.f.mjs:66`).
- **CHANGELOG** — additions only, all inside `## Unreleased`, all correctly prefixed
  `**BREAKING CHANGES:**`, all linking only `/pull/1503`. No released section touched, nothing
  removed.
- **Rust gates / generator round-trip: still skippable.** I re-checked rather than assumed — the
  only `nanvm-lib/` path in the whole PR diff is `nanvm-lib/todo/mvp-roadmap.md`, there are no
  `.rs` or `Cargo.*` changes, and nothing under `fjs/nanvm/`.
- **§2 pairing** — no `.f.mjs` in the tree imports a `.f.ts` runtime specifier.

## One thing to fix

Two module-header doc comments still point at paths this PR renames away, so after merge they
reference files that no longer exist:

- `fjs/mcp/cas/module.f.mjs:8` — ``* `fjs/mcp/module.f.ts` composes into the FJS MCP server``
  (this PR renames it to `fjs/mcp/module.f.mjs`)
- `fjs/protocol/mcp/stdio/module.f.mjs:5` — ``from `fjs/protocol/mcp/module.f.ts` ``
  (renamed to `fjs/protocol/mcp/module.f.mjs`)

This is the same class of reference the PR already updates elsewhere — `fjs/README.md`, the
`todo/*.md` files, and `fjs/mcp/proof.f.mjs:144`, where
``fjs/protocol/mcp/stdio/module.f.ts`` was correctly rewritten to `.f.mjs`. These two just
got missed. Doc-comment only: no effect on emit, types, or tests. I swept every path the PR
renames against the whole tree and these are the only two stale references.

Separately, and explicitly **not** a finding: the newly authored `fjs/effects/module.mjs` and
`fjs/types/result/module.mjs` have no `@module` header, but neither did the `.ts` files they
replace, so this is unchanged from main rather than a regression — noting it only as an
opportunity, since the PR is already closing five such gaps elsewhere.

## Not re-raised

The `Unknown` typedef description no longer sitting directly above `export type Unknown` in the
emitted declaration was chased down in an earlier review and disproved — it is TypeScript's
`@typedef` emit behaviour, not a regression.

o2alexanderfedin (commented):
Re-reviewed after the head moved three times during this pass — the branch is being pushed to actively. I started at `8e61631` and re-ran at `64bdb57`, `88c24ae`, and finally **`261ac5c`**. Everything below was verified at `261ac5c` except where noted: the full public-surface diff and link check were measured at `88c24ae`, and the only change since is one further proof migration (`fjs/types/bigint/proof.f.ts`) that I verified separately — `tsc` clean, token-equivalent to main’s compiled output modulo redundant arrow parens, no new type assertions.

## Previous findings: both resolved ✅

The two stale module-header doc references are fixed:

- `fjs/mcp/cas/module.f.mjs:8` — now cites `fjs/mcp/module.f.mjs`
- `fjs/protocol/mcp/stdio/module.f.mjs:5` — now cites `fjs/protocol/mcp/module.f.mjs`

I also swept **all 39 paths this PR renames** for residual references anywhere in the tree (excluding `CHANGELOG.md` and `package-lock.json`): **zero hits**. This class is fully closed.

(For the record, `fjs/types/result/module.f.mjs:9` and `fjs/types/bigint/module.f.mjs:9` still carry doc examples importing `'./module.f.ts'`, but both are byte-identical on `origin/main` and untouched here — pre-existing, out of scope for this PR.)

## Verification battery — all green

| Check | Result |
|---|---|
| `npm pack` (runs `prepack`, both passes) from a freshly cleaned tree | exit 0 |
| `npm test` | **2495 pass / 0 fail**, exactly matching `origin/main` (`b76018c`) at 2495 (re-run at `261ac5c`) |
| Public type surface (at `88c24ae`) | 507 vs 472: **35 added, 0 removed**; 32 of the additions are `_`-prefixed per §6.2 |
| Exported const signatures | 8 "removals" are all `.d.ts` → `.d.mts` key renames of the same names, 1:1; **0 signatures widened to `any`** |
| `/*elided*/` in emitted declarations | 1 occurrence, identical count on `main` — pre-existing |
| Link check (at `88c24ae`) | broken-link **sets** identical (163 = 163); the one delta is a pre-existing broken link whose source line moved because of an earlier rename in this PR |
| `npm run ci-update` + `git add -A && git diff --cached --exit-code` | clean — which also proves the new `.mjs` CLI entry point runs under Node exactly as CI invokes it |
| Rust gates | skipped — the only `nanvm-lib/` path is `todo/mvp-roadmap.md` (docs) |
| Generator round-trip | no `fjs/nanvm/` or `generated.rs` changes; ran `ci-update` + drift gate anyway since the CLI entry point moved |

**`bin` change validated end-to-end.** Since `bin.fjs` moves off the compiled `fjs/module.js` onto the authored `fjs/module.mjs`, I packed the tarball, installed it into a scratch project, and ran it: `node_modules/.bin/fjs` symlinks to `functionalscript/fjs/module.mjs`, is executable, and prints the command table. No `.ts` files ship, and the shipped runtime graph contains no `.ts` specifiers. `files` already covers `**/*.mjs`, so nothing is missing from the package.

**Token equivalence** — 14 migrated files, main's *compiled* output vs the PR's authored source. Method: I did **not** use `bin/norm5.mjs` / `norm6.mjs` / `normjs.mjs`; their regex comment-stripping mangles `'/'`, `'//'` and `'/*'` string literals and can produce a false "identical". I used acorn's real JS tokenizer (comments dropped by the lexer, `;` ignored for ASI, quotes normalized, module-specifier extensions neutralized), sanity-checked on exactly those literal traps plus `1 / 2` vs a regex, and **negative-controlled** with a one-token rename that it correctly flagged.

- Token-identical: `fjs/module.mjs`, `fjs/common/monoid`, `fjs/text/ascii`, `fjs/text/code_point`, `fjs/types/array`, `fjs/djs/ast` proofs.
- Differ only by **redundant grouping parentheses** (single arrow param, or an arrow body wrapped by an inline `@type` cast): `fjs/text/sgr`, `fjs/types/function/operator`, `fjs/crypto/pow`, `fjs/crypto/sha2`, `fjs/dev/update`, `fjs/fsc` proofs. Semantically identical.
- Every inline `@type` cast in the five newest proof migrations maps **1:1** to an `as` / `as const` in the `.ts` original — the §6.2 carry-over exception applies cleanly. (`crypto/pow` and `crypto/sha2` each grep-match ` as ` only in English prose inside comments; I checked the matches.)
- `fjs/effects/node/module.mjs` differs by the `__rewriteRelativeImportExtension` compiler prologue (an emit artifact), redundant parens, one parameter-list change, and two type assertions — the last of which is the finding below.

`mkdir: (...p) => mkdir(...p)` becoming `mkdir: (path, options) => mkdir(path, options)` is fine: `Mkdir` is `(path, options?) => IoResult<void>`, and Node's `fsPromises.mkdir` normalizes a missing `options` via `options || kEmptyObject`, so passing an explicit `undefined` is identical. The new form is also more precisely typed. No objection.

---

## Finding 1 — two new type assertions in `fjs/effects/node/module.mjs` that the TypeScript original did not have, neither of which is needed

Every other assertion in this migration carries over an `as` from the `.ts` source, which §6.2 explicitly permits. These two do not — they are new, and they weaken checking on the Node effect operation map, which is about the most safety-critical bridge in the repo.

**`fjs/effects/node/module.mjs:222`**

```js
/** @type {EffectToPromise} */
const runNodeEffect = asyncRun(/** @type {import('../types.ts').ToAsyncOperationMap<NodeOp>} */ ({
    ...memoryOperationMap(),
    // ~31 operations
}))
```

`main`'s `fjs/effects/node/module.ts:197` had no assertion at all:

```ts
const runNodeEffect: EffectToPromise = asyncRun({ ... })
```

There, the object literal is **contextually typed** by `asyncRun`'s parameter, so every one of the 31 operation implementations is checked against `NodeOp`. The inline cast replaces that check with an override: if an operation's implementation drifts from its declared `NodeOp` shape, the cast absorbs it silently.

**`fjs/effects/node/module.mjs:313`** — `main:287` used `satisfies`, which is a *check*:

```ts
return asNominal(createServer(nodeRl)) satisfies EffectServer   // main
return /** @type {EffectServer} */ (asNominal(createServer(nodeRl)))   // PR
```

`satisfies` verifies assignability while preserving the inferred type; `@type` asserts and discards it. JSDoc has `@satisfies` and it works here.

**Neither is necessary, and the first one causes the second.** Applying both of these:

```diff
-const runNodeEffect = asyncRun(/** @type {import('../types.ts').ToAsyncOperationMap<NodeOp>} */ ({
+const runNodeEffect = asyncRun(({
-        return /** @type {EffectServer} */ (asNominal(createServer(nodeRl)))
+        return /** @satisfies {EffectServer} */ (asNominal(createServer(nodeRl)))
```

`npx tsc --noEmit` → **exit 0**.

The dependency runs the other way from how it looks. Dropping *only* the `:313` cast while keeping `:222` fails, and the error lands on line **222**:

```
fjs/effects/node/module.mjs(222,43): error TS2352: Conversion of type '{ all: ...; mkdir: ...; ... }'
  to type 'ToAsyncOperationMap<NodeOp>' may be a mistake ...
  The types returned by 'createServer(...)' are incompatible between these types.
    Type 'Promise<Nominal<string, "1608...", Server>>' is not comparable to type 'Promise<Server>'.
```

That is the `:222` cast stripping the contextual type off the literal, so `asNominal(...)` at `:313` infers `Nominal<string, …, Server>` instead of `EffectServer` — which then *requires* the `:313` cast to compile. Remove the cast that created the problem and both disappear.

All probes were run on a freshly cleaned tree (no emitted `.d.mts` in the resolution path) and negative-controlled: injecting `const __probe = /** @type {number} */ ('str')` into the same file produced `TS2352` on that line and nothing else, confirming `tsc` was reading the mutated file and that the removed cast genuinely was not load-bearing.

Suggested fix: drop the `:222` cast, and use `@satisfies` at `:313`.

## Finding 2 — seven types published under ordinary public names while the CHANGELOG calls them "internal-only"

The CHANGELOG entry for this commit says:

> no local complex types to split (`Server`, `Readable`, `IncomingMessage`, `ServerResponse`, `RequestListener`, `EffectToPromise`, `FrameworkRegister` are all **internal-only**).

All seven were non-exported `type` declarations in `main`'s `module.ts` and appear in no declaration file. In `module.mjs` they are unprefixed `@typedef`s, and `fjs/effects/node/module.d.mts` now emits all seven as `export type`. Per `fjs/fsc/README.md` § *Private JSDoc typedefs*, the `_` prefix — not declaration-level visibility — is precisely what marks a JSDoc typedef private, so as emitted these read as public API rather than as the internal-only types the CHANGELOG describes.

I'm not raising this mechanically: the README does permit publishing a formerly-private type under an ordinary name as a design decision, and I'd have left it alone if the CHANGELOG had claimed that. It's the contradiction that makes it worth fixing, and the rest of this PR gets it right — 32 of the 35 new public type names are `_`-prefixed (`_CompileOp`, `_Dfa`, `_Refs`, `_Rule`, `_UnknownConst`, …).

There's also a concrete consequence. Four of the seven now collide by name with genuinely public, **differently shaped** types exported from the sibling `fjs/effects/node/types.ts` in the same directory:

| name | `fjs/effects/node/types.ts` (public) | `module.d.mts` (new) |
|---|---|---|
| `Server` | `Nominal<'server', '1608…', unknown>` | `{ readonly listen: (port: number) => void }` |
| `IncomingMessage` | `{ method, url, headers }` | `Readable & { method, url, headers }` |
| `ServerResponse` | `{ status, headers, body }` | `{ writeHead, end }` |
| `RequestListener` | `<O extends Operation>(_: IncomingMessage) => Effect<O, ServerResponse>` | `(req, res) => Promise<void>` |

Two public `RequestListener`s with unrelated signatures in one directory is a genuine trap for a consumer. Prefixing all seven with `_` resolves both the contradiction and the collision, and per the README that rename is not itself a breaking change.

## Minor — `package-lock.json` is stale against the new `bin`

`package.json` now has `"bin": { "fjs": "fjs/module.mjs" }`, but `package-lock.json:12` still records `"fjs": "fjs/module.js"`. Confirmed by running `npm install --package-lock-only` on a copy — it rewrites exactly that one line. It won't fail CI (`npm ci` doesn't validate the root `bin`, and the `node26` drift gate only covers `ci-update` output), but it means the next contributor's `npm install` — or `npm run update`, which shells out to `npm install` — produces an unrelated diff.

## Note — not a finding, pre-existing

`fjs/effects/node/module.mjs` has no blank line between its `@module` header and the first import, so the header is folded into the import's leading trivia and does not reach `module.d.mts` (§4). This is **byte-identical to `main`'s `module.ts`**, so it is not a regression — but the migration touched this exact block, and it'd be a cheap fix while you're here. The six proof files without `@module` headers match their predecessors, which had none either.

---

Overall this remains a careful, well-sequenced migration: test count is unchanged at 2495, no public type is removed, no signature widens to `any`, the link-check set is untouched, and the `bin` change genuinely works from an installed tarball. Findings 1 and 2 are both confined to the single `fjs/effects/node/module.ts` → `.mjs` commit and are straightforward to address.

o2alexanderfedin (commented):
Re-reviewed at `8c955a8d`. The head moved four times while I was verifying (`35f70d4` → `ccdbf52` → `d60f38f4` → `8c955a8d`); everything below is a full re-run at `8c955a8d`, not carried forward.

## The three previous findings are all fixed — verified, not just read

**1. The two new type assertions in `fjs/effects/node/module.mjs`.** Both gone, and I checked the fix is load-bearing rather than cosmetic.

- The `ToAsyncOperationMap<NodeOp>` cast around the 31-operation literal is removed, so the literal is contextually typed by `asyncRun` again. Negative control: replacing `forever: () => new Promise(() => {})` with `forever: (/** @type {string} */ x) => x` now produces `error TS2322: Type '(x: string) => string' is not assignable to type '() => Promise<never>'` at that key. Under the previous cast that mismatch was absorbed silently.
- `/** @type {EffectServer} */ (...)` → `/** @satisfies {EffectServer} */ (...)`, which is the faithful translation of main's `return asNominal(createServer(nodeRl)) satisfies EffectServer`. Negative control: substituting `(42)` for the expression errors, so the position is genuinely checked (via the contextual return type — `EffectServer` is `Nominal<'server', …, unknown>`, so the `@satisfies` itself is a weak check, but nothing is being asserted away any more).

The AGENTS.md §"Prefer `@satisfies` over `@type` when checking, not overriding" and the matching `todo/migrate-typescript-to-mjs.md` checklist item are a good addition — that generic-argument-contextual-typing trap is exactly the kind of thing worth writing down once.

**2. The seven `internal-only` types.** All seven are now `_Server`, `_Readable`, `_IncomingMessage`, `_ServerResponse`, `_RequestListener`, `_EffectToPromise`, `_FrameworkRegister`, and the CHANGELOG entry now describes what actually happens instead of claiming they are unexported. The name collisions with `fjs/effects/node/types.ts` are gone.

**3. `package-lock.json`.** Now records `"fjs": "fjs/module.mjs"`, matching `package.json`'s `bin`. `npm install --package-lock-only` in the PR tree leaves the file byte-identical, so it is genuinely in sync.

## One thing left, and it is the same class as #2

The dual-axis surface diff (`prepack` in both trees, then the exported-type extractor) shows **41 type aliases added and 0 removed** relative to `origin/main`. 39 of the 41 carry the `_` prefix. Two do not:

- `MemoryState` — `fjs/effects/memory/proof.f.mjs:15`. This one arrived in `723e931` (after the fix commit). In main it is a file-local `type MemoryState` in `proof.f.ts`; as a `@typedef` it becomes a public `export type`. The sibling `fjs/protocol/mcp/proof.f.mjs:26` declares the *same* type in this PR and correctly names it `_MemoryState` — so the two proofs now disagree with each other on the same name.
- `Benchmark` — `fjs/types/bigint/proof.f.mjs:142`, same story (main: local `type Benchmark` in `proof.f.ts`). My earlier passes missed this one; it was present before the last push and I should have caught it then rather than now.

For context on why this reads as an oversight rather than a decision: these are the only two unprefixed `@typedef`s in any `proof.f.mjs` in the tree. Every other one — `_TestOp`, `_H`, `_P`, `_Result`, `_StringMapIsOptional`, `_NodeList`, `_Tests`, `_Entries`, `_Op`, `_StepResult`, `_Merge`, … — is prefixed. Renaming both to `_MemoryState` / `_Benchmark` is a mechanical change and neither is referenced outside its own file.

## Battery, at `8c955a8d`

| Check | Result |
|---|---|
| `npx tsc --noEmit` | exit 0 (negative-controlled: an injected `@type {string}` cast produces TS2352/TS2345) |
| `npm run prepack`, clean tree | exit 0, both passes |
| `npm test` | 2495 pass / 0 fail — exactly matches `origin/main` (`b76018cc`), also 2495 |
| Public type surface | 41 added / 0 removed / 6 changed; changes are all `Rule`→`_Rule`, `id`→`_id`, `lock`→`_lock`, `error`→`errorSchema` renames in `typeof` positions. No signature widened to `any`. |
| Broken-link sets | identical, 141 either side; the only textual delta is one listed entry retargeting `djs/parser/module.f.ts:73` → `module.f.mjs:66` |
| CHANGELOG | every migrated file has an entry, `**BREAKING CHANGES:**` prefixed, PR-linked only; `AGENTS.md` and `todo/` changes correctly have none |
| Rust gates / generator round-trip | skipped — re-checked the file list rather than assuming: `nanvm-lib/` is touched only in `todo/mvp-roadmap.md`, and neither `fjs/nanvm/` nor `nanvm-lib/tests/test/generated.rs` appears in the diff at all |

**Token equivalence** for every proof migrated since my last review — `text`, `types/bit_vec`, `types/sorted_list`, `types/sorted_set`, `website`, `effects/memory`, `types/btree`, `types/byte_set`, `types/btree/find`, `types/list` — plus a re-check of `effects/node/module`. Method: an **acorn tokenizer** over main's *compiled* `.f.js` vs the PR's authored `.f.mjs`, comparing `(type, value)` pairs with `;` dropped and only module-specifier extensions neutralized inside string literals, then an LCS diff so a token insertion does not cascade into a false mismatch. I did not use `bin/norm5.mjs`/`norm6.mjs`/`normjs.mjs`. Negative control: mutating a single identifier in one input yields exactly one reported difference.

Result: `text`, `sorted_list`, `sorted_set`, `website` are token-identical outright. `bit_vec`, `effects/memory`, `btree`, `byte_set`, `btree/find`, `list` differ **only in parentheses** — `(a) =>` → `a =>` from TS's arrow-parameter emit, and the extra pair JSDoc inline casts require (`(asBase(v) as bigint)` → `(/** @type {bigint} */ (asBase(v)))`). `effects/node/module` differs only in parentheses, the `__rewriteRelativeImportExtension` compile helper, and `mkdir: (...p) =>` → `mkdir: (path, options) =>`; the op type is `(path: string, options?: MakeDirectoryOptions) => IoResult<void>` and `fsPromises.mkdir(path, undefined)` is equivalent to `fsPromises.mkdir(path)`, so that one is behaviour-preserving too.

Nothing here blocks CI — this is a naming-convention consistency fix, not a correctness one. Happy to approve on the two renames.

o2alexanderfedin (approved):
Re-reviewed after the head moved. The one item left open at `8c955a8d` is **resolved**, and I
re-ran the battery at the current head. Approving.

## The open item is closed

`a7d0d897` renames both remaining unprefixed public `@typedef`s — `MemoryState` → `_MemoryState`
in `fjs/effects/memory/proof.f.mjs` (now matching its sibling `fjs/protocol/mcp/proof.f.mjs`)
and `Benchmark` → `_Benchmark` in `fjs/types/bigint/proof.f.mjs` — and amends the two CHANGELOG
entries to record the renames. A multi-line-aware scan (JSDoc blocks collapsed first, so
continuation lines are not missed) over every `.mjs`/`.ts` file under `fjs/` finds **zero**
unprefixed `@typedef` in any `proof.f.mjs`. The remaining unprefixed ones are either
module-level public API names in `module.f.mjs` (`PackageJson`, `Grammar`, `Unknown`, `Step`,
…) or function-local aliases such as `/** @typedef {(typeof a)[0]} T */` inside a function
body, which never cross a file boundary. The emitted surface agrees: **0 unprefixed type
additions**.

## What else landed since `8c955a8d`

Eight more proof migrations — `fjs/types/range_map`, `fjs/ci/nix`, `fjs/cli`,
`fjs/types/btree/set`, `fjs/types/btree/remove`, `fjs/ci`, `fjs/text/utf16`, `fjs/text/utf8` —
plus the `todo/*.md` reference updates that go with them.

## Battery, at `1315b1a9`

| Check | Result |
|---|---|
| `npm run prepack`, clean tree | exit 0, both passes |
| `npm test` | 2495 pass / 0 fail — exactly matches `origin/main` (`b76018cc`), also 2495 |
| Public type surface | 40 added / 0 removed / 6 changed; **every** addition is `_`-prefixed. The 6 changes are the already-reviewed `Rule`→`_Rule`, `id`→`_id`, `lock`→`_lock`, `error`→`errorSchema` renames in `typeof` positions. |
| Exported const signatures | 10 added / 8 removed / 8 changed; the add/remove pairs are purely `.d.ts`→`.d.mts` path renames of the same symbols, and no signature widened to `any` |
| Broken-link sets | identical, 163 either side; the only textual delta is the same entry retargeting `djs/parser/module.f.ts:73` → `module.f.mjs:66` |
| CHANGELOG | all eight new migrations have an entry, `**BREAKING CHANGES:**` prefixed, PR-linked only, all under `## Unreleased`; `todo/` edits correctly get none |
| Rust gates / generator round-trip | skipped, re-checked rather than assumed: against **current `origin/main`** the whole PR touches `nanvm-lib/` only in `todo/mvp-roadmap.md`, and `fjs/nanvm/` and `nanvm-lib/tests/test/generated.rs` are byte-identical to main. The delta since `8c955a8d` touches neither path at all. |

## Token equivalence for the eight new proofs

Method: **acorn**, but this round an *AST* comparison rather than a token stream — parse both
sides, drop `start`/`end`/`raw`/`loc`, canonicalise key order, and neutralise only module
specifier extensions on `ImportDeclaration.source`. Parentheses are not AST nodes, so TS's
`(a) => …` arrow-parameter emit no longer cascades into thousands of phantom token
mismatches the way a positional token diff does. I did **not** use `bin/norm5.mjs`,
`bin/norm6.mjs` or `bin/normjs.mjs`. Negative controls: a whitespace-only edit compares
IDENTICAL, and renaming a single identifier (`proof` → `proofX`) compares DIFFER at exactly
that node.

Comparing main's *compiled* `.f.js` against the PR's authored `.f.mjs`:

- `ci/nix`, `cli`, `types/btree/set`, `types/btree/remove`, `types/range_map`, `text/utf16`,
  `text/utf8` — **AST-identical**. So are `effects/memory` and `types/bigint` after the
  typedef renames, confirming those touched only comments.
- `fjs/ci/proof` — one real difference, at exactly two call sites. `makeState(rust: boolean,
  packageJson?: string)` became `(/** @type {boolean} */ rust, /** @type {string | undefined} */
  packageJson)`, so `makeState(rust)` / `makeState(false)` are now `makeState(rust, undefined)` /
  `makeState(false, undefined)`. I verified this is the *only* difference by patching those two
  call sites into main's compiled file and re-running the comparison — IDENTICAL. `packageJson`
  is `undefined` either way, so it is behaviour-preserving. (An inline `@type` cast on an arrow
  parameter cannot express optionality; `@param {string} [packageJson]` could, but for a
  file-local proof helper the explicit `undefined` is fine.)

## One non-blocking nit

`fjs/media/json/todo/stringify-sorted-canonical.md` lists where each proof binds its
`stringify` alias, and this PR updated some of those line numbers but not all — the JSDoc
`@type` lines the migration inserts push each binding down 2–3 lines. Current state:
`btree/remove:17` ✓, but `btree/set` says `:13` (actual `jsonStr` is `:16`), `range_map` says
`:12` (actual `str` is `:15`), and from earlier rounds `btree:11` (actual `:14`),
`byte_set:8` (actual `:10`), `sorted_list:10` (actual `:12`), `list:37` (actual `:11`). A
`todo/` pointer, no CI impact — worth a sweep whenever that note is next touched.

sergey-shandar:
@codex review

o2alexanderfedin (commented):
Addendum. The head moved twice while I was posting the approval above, so that approval was
tested at `1315b1a9`. I re-ran the full battery at `b1774825`, and checked the docs-only commit
`b4e262a6` on top of it. The battery still passes, but `b1774825` **reintroduces the exact
issue the approval had just recorded as closed**, so this is a comment rather than a second
approval.

## The `_` prefix regressed in `b1774825`

`Migrate fjs/effects/proof.f.ts and fjs/effects/eff/proof.f.ts to .f.mjs` turns two
**non-exported** TypeScript locals into **exported** JSDoc typedefs:

```
fjs/effects/proof.f.d.mts:15:   export type AddOp = readonly ['add', (a: number, b: number) => number];
fjs/effects/proof.f.d.mts:16:   export type AnyOp = readonly [string, (a: number) => number];
fjs/effects/eff/proof.f.d.mts:1: export type AddOp = readonly ['add', (a: number, b: number) => number];
```

On `main` these are `type AddOp = …` / `type AnyOp = …` — file-local, invisible in the emitted
declarations. The public type surface goes from 40 added / 0 unprefixed at `1315b1a9` to
**42 added, 2 of them unprefixed**, and `AddOp` is now published twice under the same public
name from two different modules — the same collision shape as the seven internal-only types
earlier in this PR. Neither is referenced outside the file that declares it, and `AnyOp`'s own
doc comment describes it as a proof-local device ("this type is how a proof says so without an
`as` cast").

`_AddOp` / `_AnyOp` restores the invariant that held at `1315b1a9`. Non-blocking for CI, same
as last time.

## The two new `@type` casts are fine — I checked rather than assumed

`AGENTS.md` §"Prefer `@satisfies` over `@type` when checking, not overriding" (added by this
PR) warns that a `@type` cast around an object literal handed to a generic function strips the
contextual typing the callee relies on. The new
`match(/** @type {OperationMap<AddOp, number>} */ ({ add: (a, b) => a + b }))` is literally
that shape, so I mutation-tested it in a cleaned tree:

- handler return changed to a template string → `TS2352`, "the types returned by `add(...)` are
  incompatible … `string` is not comparable to `number`"
- key `add` renamed to `sub` → `TS2352`, "Property `add` is missing … but required in
  `OperationMap<AddOp, number>`"

Both caught. A one-key literal still has to *sufficiently overlap* the assertion target, so this
is not the absorbing case that section describes. The declaration-form
`/** @type {(command: 'add') => …} */ const doAdd = do_` is checked too — narrowing the command
literal to `'sub'` yields `TS2322` at the declaration plus `TS2345` at both call sites. No loss
of checking, and I withdraw any suggestion that this needed `@satisfies`.

## The line-citation nit is fixed

`b4e262a6` corrects `fjs/media/json/todo/stringify-sorted-canonical.md`. I re-checked all
sixteen citations in that list against the files at `b4e262a6`: `btree:14`, `set:16`,
`remove:17`, `array:6`, `byte_set:10`, `range_map:15`, `sorted_list:12`, `sorted_set:12`,
`list:11`, `ascii:6`, `utf8:8`, `utf16:18`, `mcp/stdio:20`, `djs/parser:309`,
`djs/serializer:47` all land exactly on the alias binding. One left: `btree/find/proof.f.mjs:10`
is an `import` line — the binding is `:13`. Same off-by-three as the others, just missed.

## Battery, at `b1774825`

| Check | Result |
|---|---|
| `npm run prepack`, clean tree | exit 0, both passes |
| `npm test` | 2495 pass / 0 fail — matches `origin/main` (`b76018cc`) |
| Public type surface | 42 added / 0 removed / 6 changed — **2 unprefixed additions**, `AddOp` and `AnyOp`, per above |
| Exported const signatures | unchanged from `1315b1a9`: the add/remove pairs are `.d.ts`→`.d.mts` renames of the same symbols, nothing widened to `any` |
| Broken-link sets | identical, 163 either side, same single retarget of `djs/parser/module.f.ts:73` → `module.f.mjs:66` |
| CHANGELOG | the new entry covers both files and names the generic-instantiation conversions; `**BREAKING CHANGES:**` prefixed, PR-linked only, under `## Unreleased` |
| Rust gates / generator round-trip | still skippable, re-checked: against current `origin/main` the PR touches `nanvm-lib/` only in `todo/mvp-roadmap.md`, and neither of these two commits goes near `nanvm-lib/` or `fjs/nanvm/` |

`b4e262a6` changes one `todo/*.md` and nothing else, so it cannot move any of those gates; I
verified its content directly rather than re-running them.

Token equivalence does not apply to `fjs/effects/proof` and `fjs/effects/eff/proof` — that
migration is a deliberate restructuring (`do_<AddOp>('add')` → a typed `doAdd` binding,
`assertPure`'s type parameters → `@template`), not a mechanical rename. I read the diff
instead: every call site is a substitution of `doAdd`/`doAny` for the explicit instantiation,
and the test count is unchanged. `fjs/effects/module.f.mjs`'s only change is a
`proof.f.ts` → `proof.f.mjs` reference inside a doc comment.

Everything else from the approval above still holds. Happy to approve again on `_AddOp` /
`_AnyOp`.

o2alexanderfedin (approved):
Re-reviewed at `59fcb29c`. Both items from the `b4e262a6` review are resolved, and the full battery is clean at the new head. Approving.

## The two open items

**1. `AddOp` / `AnyOp` published unprefixed — fixed.** Both are now `_AddOp` / `_AnyOp` in `fjs/effects/proof.f.mjs` and `fjs/effects/eff/proof.f.mjs`, and the CHANGELOG entry records the rename with the reason (`neither was previously exported, but a JSDoc @typedef has no non-exported form`). Neither name appears outside its own file.

**2. Stale citation `fjs/types/btree/find/proof.f.mjs:10` — fixed.** It now reads `:13`, which is `const jsonStr = stringify(sort)`. I re-verified every citation in `fjs/media/json/todo/stringify-sorted-canonical.md` mechanically against the head tree; all nineteen resolve to the `stringify(sort)` binding (including the aliased-import sites that bind it as `jsonStringify(sort)`, and the `utf16` wrapper the doc calls out).

## Public surface — the check this PR series keeps regressing on

Ran `npm run prepack` in both trees from freshly cleaned checkouts, then compared every `export type` / `export declare const` in the emitted `.d.ts` / `.d.mts` files, file-aware rather than name-grouped:

```
ADDED exported symbols: 44   (42 type, 2 const)
UNPREFIXED ADDITIONS:    0
REMOVED:                 0
```

Every one of the 44 carries the `_` prefix. Const signature changes are all the corresponding renames (`Dfa` → `_Dfa`, `MapEntries` → `_MapEntries`, `Refs` → `_Refs`, `CompileOp` → `_CompileOp`, `UnknownConst` → `_UnknownConst`); nothing widened to `any`. The four `_`-prefixed names that appear in two modules each (`_InitialState`, `_MapEntries`, `_AddOp`, `_MemoryState`) are the already-settled collisions — the prefix is what marks them internal, and none is imported across the boundary.

## Battery

| Check | Result |
| --- | --- |
| `npm run prepack` (clean tree, both passes) | exit 0 on PR and on `main` |
| `npm test` | **2495 pass / 0 fail** — identical to `main` at `b76018cc` (2495 / 0) |
| Public surface diff | 44 added, all `_`-prefixed; 0 removed; no `any` widening |
| Link check | 141 broken relative links on both PR and `main`, **identical sets**; 0 label/target extension mismatches |
| CHANGELOG | additive only (0 removed lines), all under `## Unreleased`, `**BREAKING CHANGES:**` prefixed, links only `/pull/1503` |

Tests were run on a `git clean`ed tree *before* `prepack`, so the count is not contaminated by emitted `.f.js` siblings.

**Delta equivalence.** I did not use `bin/norm*.mjs` here. I parsed `fjs/effects/proof.f.mjs` and `fjs/effects/eff/proof.f.mjs` at `b4e262a6` and at `59fcb29c` with acorn and compared position-stripped ASTs: **IDENTICAL** for both files, i.e. the rename is comment-only and cannot move runtime behaviour. Negative control: flipping `a + b` to `a - b` in the old copy makes the same comparator report `DIFFERENT`.

## Not run, and why

- **Rust gates** — the only path under `nanvm-lib/` in the whole PR is `nanvm-lib/todo/mvp-roadmap.md`. No Rust source touched, so `cargo fmt/test/clippy` cannot be affected.
- **Generator round-trip** — nothing under `fjs/nanvm/`, and `nanvm-lib/tests/test/generated.rs` is untouched.

## One non-blocking note (pre-existing, not introduced here)

Two entries in `fjs/media/json/todo/stringify-sorted-canonical.md` still point at `.f.ts` paths, and I confirmed against `origin/main` that both were already stale before this PR, so this is not a regression and I am not asking for it:

- `fjs/bnf/data/proof.f.ts` — that file is `fjs/bnf/data/proof.f.mjs` on `main` already.
- `fjs/media/json/parser/proof.f.ts:13` — the file is untouched by this PR and its `const stringify = jsonStringify(sort)` binding sits at `:14` on `main` too (off by one).

Since you are sweeping that doc's citations anyway, they would be cheap to fold in.

PR #1473, opened 2026-08-10
Update dependencies and Node.js versions

sergey-shandar:
## Summary
This PR updates various development dependencies and runtime versions across the CI configuration and Nix flakes.

## Key Changes
- **FunctionalScript**: Updated from `0.41.0` to `0.43.1` across all CI workflows and configuration
- **Node.js versions**:
  - Default Node.js: `26.6.0` → `26.7.0`
  - Node 24: `24.18.0` → `24.18.1`
- **Deno**: Updated from `2.9.4` to `2.9.5`
- **Nixpkgs**: Updated commit reference from `04607e1165ac22c5fde6dcc54c9e0b3c0487c555` to `fcb8fcd6bf2d0adecae5bd491afaaaf8311b758d` (nixos-26.05 channel)
- **Nix flakes**: Updated all generated Nix flake files (node22, node24, node26) to reference the new Nixpkgs commit

## Implementation Details
- Updates are applied consistently across all CI workflow jobs (6 separate workflow configurations)
- Changes cover npm, deno, and bun package managers
- Nix development environment flakes are regenerated with the updated Nixpkgs snapshot
- All version updates are pinned to specific releases for reproducibility

https://claude.ai/code/session_01EfbrXbhQxV9cmjp8HXBWhW

PR #1508, opened 2026-08-13
Update nixpkgs commit to 9f78f44a87948854445dae0b6bf82b2e87e4efb5

sergey-shandar:
## Summary
Updates the nixpkgs reference across the configuration to a newer commit within the nixos-26.05 channel.

## Changes
- Updated the nixpkgs commit hash in the module configuration from `fcb8fcd6bf2d0adecae5bd491afaaaf8311b758d` to `9f78f44a87948854445dae0b6bf82b2e87e4efb5`
- Regenerated flake.nix files for Node.js 22, 24, and 26 with the updated nixpkgs reference

## Details
The nixpkgs reference remains on the `nixos-26.05` channel while updating to a newer commit. The generated flake.nix files are automatically updated to reflect this change across all Node.js version configurations.

https://claude.ai/code/session_01Jx3U5vFKxd2mhXNo3XDumT

### DIFF
--- a/fjs/ci/config/module.f.mjs
+++ b/fjs/ci/config/module.f.mjs
@@ -53,7 +53,7 @@ export const node = /** @type {const} */({
 // https://channels.nixos.org/nixos-26.05/git-revision
 export const nixpkgs = /** @type {const} */({
     ref: 'nixos-26.05',
-    commit: 'fcb8fcd6bf2d0adecae5bd491afaaaf8311b758d',
+    commit: '9f78f44a87948854445dae0b6bf82b2e87e4efb5',
 })
 
 // https://github.com/bytecodealliance/wasmtime/releases
--- a/nix/generated/node22/flake.nix
+++ b/nix/generated/node22/flake.nix
@@ -1,5 +1,5 @@
 {
-    inputs.nixpkgs.url = "github:NixOS/nixpkgs/fcb8fcd6bf2d0adecae5bd491afaaaf8311b758d";
+    inputs.nixpkgs.url = "github:NixOS/nixpkgs/9f78f44a87948854445dae0b6bf82b2e87e4efb5";
     outputs = { nixpkgs, ... }: {
         devShells.aarch64-linux.default = let
             pkgs = import nixpkgs {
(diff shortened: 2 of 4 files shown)


o2alexanderfedin (approved):
Approving. Clean pin bump — I checked the generator, the channel, and the one thing a nixpkgs bump can actually break.

## Drift gate

`npm run ci-update` then `git add -A && git diff --cached --exit-code` — both exit 0 at `1ec8ff42`. So the three `nix/generated/*/flake.nix` really are the generator's output for the edited `nixpkgs.commit`, not hand-edited alongside it. `.github/workflows/ci.yml` is untouched here, which is correct: it carries no nixpkgs reference (only the `nix develop ./nix/generated/<job>` invocations), unlike the earlier combined bumps #1418/#1423/#1440/#1473 that also moved Node and wasmtime pins.

Old hash `fcb8fcd6…` has zero remaining occurrences in the tree; the new hash appears in exactly four places — `fjs/ci/config/module.f.mjs` plus the three flakes.

## The hash is the right hash

The comment above the const points at `https://channels.nixos.org/nixos-26.05/git-revision`, so I fetched it:

```
$ curl -sSL https://channels.nixos.org/nixos-26.05/git-revision
9f78f44a87948854445dae0b6bf82b2e87e4efb5
```

Exact match with `ref: 'nixos-26.05'`. The commit resolves in `NixOS/nixpkgs` and is a `release-26.05` backport dated 2026-08-12, so it is a stable-branch revision rather than something off `master`.

## The risk that matters: does the new snapshot still give CI the Node it asserts

The `nix-flakes` job asserts exact versions (`test "$(nix develop ./nix/generated/node22 --command node --version)" = v22.23.2`, and likewise 24.18.1 / 26.7.0), so a snapshot that moved a Node point release would fail CI. `nix` is not available in my environment, so instead I read the package definitions at both revisions:

| attr | `fcb8fcd6…` (old) | `9f78f44a…` (new) | ci.yml asserts |
|---|---|---|---|
| `nodejs_22` | 22.23.2 | 22.23.2 | v22.23.2 |
| `nodejs_24` | 24.18.1 | 24.18.1 | v24.18.1 |
| `nodejs_26` | 26.7.0 | 26.7.0 | v26.7.0 |

Unchanged across the bump, and all three match the assertions. Nothing in the `nix-flakes` job should move.

`npm test` at this head: `pass: 2495, fail: 0, total: 2495` — identical to `origin/main`. `tsc` exit 0 (it runs first in `npm test`).

No CHANGELOG entry, which matches every previous pin bump on these files (#1418, #1423, #1440, #1473 — none touched `CHANGELOG.md`).

What I could not verify: that the flakes actually *evaluate* under the new snapshot, since there is no `nix` here. That is precisely what the `nix-flakes` CI job exists to check, and the version data above says it has no reason to fail.

PR #1509, opened 2026-08-13
Add reduceValue2DefaultBranch test case to btree remove proof

sergey-shandar:
## Summary
Added a missing test case for the `reduceValue2DefaultBranch` function in the btree remove module's proof suite.

## Key Changes
- Added `reduceValue2DefaultBranch` test case that verifies `reduceValue2` correctly handles a leaf node with a default branch structure `[['x'], 's', ['y']]`

## Implementation Details
The new test follows the same pattern as the existing `reduceValue0DefaultBranch` test case, ensuring consistent test coverage for the reduce value functions in the btree removal proof suite.

https://claude.ai/code/session_01VAUANGNZv3GnL65EbNZ5D6

### DIFF
--- a/fjs/types/btree/remove/module.f.mjs
+++ b/fjs/types/btree/remove/module.f.mjs
@@ -177,6 +177,9 @@ export const proof = {
         reduceValue0DefaultBranch: () => {
             reduceValue0([['leaf']])([['x'], 's', ['y']])
         },
+        reduceValue2DefaultBranch: () => {
+            reduceValue2([['leaf']])([['x'], 's', ['y']])
+        },
         initValue1DefaultBranch: () => {
             initValue1(null)([[['a'], 'b', ['c']], 'd', ['e']])
         },


o2alexanderfedin (approved):
Approving. The new case is not vacuous — I mutation-tested it in both directions.

## The case does reach the branch it claims

`reduceValue2([['leaf']])([['x'], 's', ['y']])` takes `a.length === 1`, then `n0 = ['x']` has length 1, which falls through `case 3` / `case 5` to `default: { throw 'invalid node' }`. Confirmed by mutation rather than by reading:

**Kill test (this PR, `24f30a6f`).** Replaced *only* `reduceValue2`'s default arm with a non-throwing `return /** @type {any} */ (null)` (the `any` cast is needed — a plain `return [n0, v1, a]` there does not type-check, so `npm test` dies in `tsc` before the runner ever starts, which would have looked like a false negative):

```
import("./fjs/types/btree/remove/module.f.mjs").proof.throw.reduceValue2DefaultBranch(): error
Number of tests: pass: 2495, fail: 1, total: 2496
```

Exactly one failure, and it is the new case. So the case is the thing holding that arm.

**Redundancy control (`origin/main`, `3bc32ab9`).** Same mutation, same file, in a separate worktree on main:

```
Number of tests: pass: 2495, fail: 0, total: 2495
```

Nothing fails. The arm was genuinely unreached before this PR, so the case adds coverage rather than duplicating `reduceValue0DefaultBranch`.

Unmutated at this head: `pass: 2496, fail: 0, total: 2496` — exactly main's 2495 plus one. `npx tsc --noEmit` exit 0 (via `npm test`), and `npm run prepack` from a freshly cleaned tree exit 0.

## Follow-up, not a blocker: `initValue0` is still uncovered

The comment above `proof` names all four helpers — «`reduceValue0`/`reduceValue2`/`initValue0`/`initValue1` merge a Branch1's lone child…» — but the group now exercises three of them. I ran the same mutation on `initValue0`'s default arm at this head:

```
Number of tests: pass: 2496, fail: 0, total: 2496
```

No test notices, so `initValue0`'s invariant guard is still unreached. By the pattern of the three existing cases the missing one looks like `initValue0(null)([['x'], 's', ['y']])` — `a === null`, `n2 = ['y']` has length 1, so neither `case 1` nor `case 2` matches. Worth a fourth case, here or in a follow-up; it does not hold up this PR.

I could not report a coverage number for this file: at this head `npm run cov` is still the pre-#1512 script, which reports `tests 0` on my Node (v23.11.0). Once #1512 lands, this becomes directly measurable rather than mutation-inferred.

No CHANGELOG entry, which I think is right for a proof-only addition — nothing about `remove`'s behaviour changes.

PR #1515, opened 2026-08-13
Add objectIdentifier proof test case for generic encode/decode

sergey-shandar:
## Summary
Added a new proof test case for `objectIdentifier` to exercise the generic `encode`/`decode` dispatch path, complementing the existing direct `encodeObjectIdentifier`/`decodeObjectIdentifier` tests.

## Changes
- Imported `objectIdentifier` from the module
- Added `objectIdentifier` proof test that:
  - Creates a record with `objectIdentifier` type and a sample OID value `[1n, 2n, 840n, 113549n]`
  - Encodes and decodes it through the generic `encode()`/`decode()` functions
  - Verifies the round-trip produces the expected result
  - Provides test coverage for the `recordToRaw`/`rawToRecord` code paths for object identifiers

## Implementation Details
The test follows the existing proof test pattern using `ch0()` to validate that encoding a record and then decoding it returns to the original state. This ensures the generic dispatch mechanism properly handles the `objectIdentifier` type, which was previously only tested through direct function calls.

https://claude.ai/code/session_018SZWuzstBRNCbtQbLdnCtY


o2alexanderfedin (approved):
Approving. Draft status is not a blocker.

Note on head: I started on `2975869` and it moved to `7e10001` mid-review (a merge of `main` picking up #1514). Everything below was re-run from scratch at **`7e10001`** against `origin/main` = `6dd0d9a` — the PR's own change is still exactly the same +10 lines in `fjs/asn.1/proof.f.mjs` (`git diff --stat origin/main...HEAD`).

**The new case is not vacuous — I mutation-tested it rather than trusting the comment.** Mutation: make the `objectIdentifier` arm of the generic dispatch unreachable by rewriting `case objectIdentifier:` to `case 0x99n:` (edit verified present in the file before each run, and reverted after).

| mutation | this PR | `origin/main` |
|---|---|---|
| `recordToRaw` `objectIdentifier` arm | **2498 pass / 1 fail** — `proof.encodeDecode.objectIdentifier` | 2498 / **0 fail** |
| `rawToRecord` `objectIdentifier` arm | 2499 / **0 fail** | 2498 / **0 fail** |

So the encode half is genuinely newly covered: the mutation kills exactly one test, the new one, and kills nothing at all on main. Confirmed with the numbers rather than the description.

**One honest caveat about the comment, not the code.** The comment says the case "exercises the objectIdentifier case in recordToRaw/**rawToRecord**", but the `rawToRecord` half is not actually verified — I can delete that arm and the new case still passes. The reason is structural: `ch0` only compares bytes (`encode(r) === v`, `encode(r0) === v`), never `r0` against `r`. With the arm gone, `rawToRecord` falls to `default: return encodeRaw(raw)`, which reproduces the input TLV byte for byte:

```
                              unmutated                          decode arm unreachable
decode(v)[0]   ["6n",["1n","2n","840n","113549n"]]          "-9657453207173396237n"   (the raw Vec)
encode(r0)===v true                                          true
```

**This is pre-existing, not something this PR introduced.** I ran the same mutation against the sibling arms on `origin/main`: removing the `boolean`, `integer` and `octetString` arms of `rawToRecord` each also kills **0** tests (2498 / 0 in all three). `ch0` has never been able to see a decode that returns the raw vec. The new case is therefore exactly as strong as every case around it, which is the right bar for it to clear. Only flagging it so the comment is not read as a stronger guarantee than it is — narrowing it to `recordToRaw`, or having `ch0` also assert `r0` structurally against `r` (which would harden all six arms at once), would be follow-up material, not a change I would ask for here.

**Coverage effect, measured rather than asserted** (`npm run cov` now works here):

- `fjs/asn.1/module.f.mjs` branch coverage **96.61 → 100.00** (line and function already 100.00 on both).
- `all files` branch **98.07 → 98.12**; line 99.93 and function 99.73 unchanged.
- `tests 2434 → 2435, pass 2435, fail 0`.

So the case closes a real, reported branch gap. Worth noting the two numbers disagree in an instructive way: node's branch counter now reads 100.00 for `rawToRecord`'s arm because the line *executes*, while the mutation shows its result is never asserted. Executed is not verified — the mutation is the stronger signal, and it is the one that says the encode side is real.

Rest of the battery at `7e10001`, both trees `git clean`ed before testing:

- `npx tsc --noEmit` — exit 0 (0 on main too).
- `npm run prepack` from a freshly cleaned tree — exit 0, both passes (0 on main too).
- `node ./fjs/module.mjs t` — **2499 pass / 0 fail** vs **2498 / 0** on main. +1 is exactly the one added case.
- Dual-axis public surface (`prepack` in both trees, then `extract.mjs` / `consts.mjs`; outputs 45.9 KB / 64.7 KB, so non-empty): **zero differences** in exported type aliases and zero in exported const signatures. No new `export type`, so no §6.2 `_`-prefix question, and nothing widened to `any`.
- `linkcheck.mjs` — broken-link **sets** identical to main, not just counts.
- Rust gates skipped — `nanvm-lib/` untouched.

`objectIdentifier` is added to an existing import list and `SupportedRecord` already carries the `[typeof objectIdentifier, ObjectIdentifier]` arm, so the `/** @type {SupportedRecord} */` annotation type-checks with no cast. No CHANGELOG entry, which matches §8.3 and matches the precedent of #1509 (also proof-only, also no entry) — nothing shipped changes.

PR #1518, opened 2026-08-13
Add test case for const declaration missing newline

sergey-shandar:
## Summary
Added a test case to verify that the parser correctly handles and reports an error when a `const` declaration is missing a required newline at the end.

## Key Changes
- Added a new test case in the proof object that validates error handling for `const` declarations without a trailing newline
- The test verifies that parsing `'const a = 1'` (without a newline) produces an error with the message `'unexpected end'`
- Includes a comment explaining that this test differs from other `const` tests because the parser reaches EOF while in a newline-required state, rather than receiving an `nl` token

## Implementation Details
- The test uses `tokenizeString()` to tokenize the input and `parseFromTokens()` to parse it
- Asserts that the result is an error object with the expected error message
- This ensures the parser properly validates the syntax requirement for newlines after const declarations

https://claude.ai/code/session_01FzcAzw79iaDormeMYNVaYs


o2alexanderfedin (approved):
Verified at `387d76b9` against `origin/main` (`7ae03f04`, which is this branch's merge base), Node v23.11.0. The case is real, not vacuous — approving. Draft status is not a blocker.

### The case is not vacuous — mutation-tested both ways

First I pinned down which branch `'const a = 1'` actually reaches, rather than trusting the comment: I tagged all 16 `'unexpected end'` literals in `fjs/djs/parser/module.f.mjs` with distinct suffixes and re-parsed. The input lands on tag `@0`, the first occurrence — `parseNewLineRequiredOp`'s `case 'eof'` (line 107). The comment's claim is accurate.

Then the two-sided mutation, deleting that `case 'eof'` line so `eof` falls through to `default` (`'unexpected token'`):

| tree | result |
|---|---|
| this PR, `387d76b9` | **`invalidWithConst[3]` fails** — `pass 2522, fail 1, total 2523` |
| `origin/main`, `7ae03f04` | **nothing fails** — `pass 2522, fail 0, total 2522` |

So the branch was completely unexercised before and this is the only test that covers it. That is exactly the shape a coverage case should have.

`npm run cov` agrees independently: `fjs/djs/parser/module.f.mjs` branch coverage **93.23 → 93.61**, and all-files branch **98.12 → 98.15**. Line and function coverage are unchanged at 100.00, as expected for a branch-only gap.

### Baseline battery

- `npm test` — **2523 pass / 0 fail**, against **2522** on `origin/main`; the delta is exactly this case.
- Public surface, both axes, `main` vs PR: **526 type aliases and 893 exported consts on each side, and the key-and-value diff is empty**. No `export type` added, so §6.2's `_` question does not arise. Broken-link sets identical to main.
- No CHANGELOG entry, which is correct here — the change is proof-only and adds no public surface. Consistent with the merged precedent for coverage-only PRs: #1509, #1515 and #1466 all landed without one.

### One non-blocking note

The assertion pins the message but not the branch that produced it. A second mutation — `case 'eof': return { ...state, state: '' }`, i.e. treating `eof` in the newline-required state as if it were `nl` — leaves the new case **passing**, because the parser then falls to the terminal `default` at line 506 and reports the same `'unexpected end'` string (I confirmed this by tagging: the message comes from `@14`, not `@0`, under that mutation).

That does not make the case vacuous — the deletion mutation above proves it covers the branch — but it does mean the test would not notice `eof` handling migrating from `parseNewLineRequiredOp` to the terminal fallback. Asserting `obj[1].metadata` as well would pin the position and close that gap, matching what `errorMetadata` already does elsewhere in this proof. Entirely optional.

### Not verified

- Rust gates and the generator round-trip — `nanvm-lib/` and `fjs/nanvm/` untouched.

PR #1519, opened 2026-08-13
Add `okThen`, the pure `Result` bind

sergey-shandar:
Fixes `fjs/types/result/todo/ok-then.md` (deleted here).

`fjs/types/result` exported the functor map (`mapOk`) but not the monad bind, so every pure `Result` chain re-derived the same guard by hand. `fjs/effects` already shipped the effectful twin, `okStep`.

```js
export const okThen = f => r => r[0] === 'error' ? r : f(r[1])
```

Its `@type` is in `fjs/types/result/module.f.mjs` — quoted here in words because this description strips angle brackets: `okThen` takes `f`, a function from a value to a `Result`, and returns a function from a `Result` to a `Result`. The returned result's ok type is `f`'s ok type; its error type is the **union** of `f`'s error type and the incoming result's error type.

### Two changes to the design in the issue

**The error types are unioned, not unified** — the issue proposed a single shared `E`. Every real site widens its error as it goes: a JSON parse failure (`string`) followed by an rtti schema failure (`ValidationError`). That is precisely why those sites rebuilt the incoming error with `error(v)` — not to handle it, only to retag it into the wider type. With one shared error type each site would need a cast to widen one side before the call; the union states the widening in the type, so nothing is pre-widened and a passed-through `error` stays the very tuple it arrived as. The incoming result's error type binds on the second arrow, matching `mapOk`, so one `okThen(f)` applies to results carrying any error type.

**`addRevision`'s five guards do not become "one composition".** Each step there reads values resolved by earlier ones (`parents` and `subject` are both used three steps later), so a straight `okThen` chain needs a nested closure per link to keep them in scope — which AGENTS.md §6.4 rules out for the same reason it rules out nested `step`. Instead its store-free prefix moves into a new pure `buildRevision`, whose guards stay a flat sequence, and `okThen` is used once at the effect boundary:

```js
const revisionResult = okThen(buildRevision(input))(parentsResult)
```

`addRevision` is left holding only the encode, the write and the cache fold. Behavior and step order are unchanged.

The issue's "roughly ten sites" also overcounts. Most of the surviving `=== 'error'` checks in the tree are either `okStep`-shaped (they return an `Effect`) or `mapOk`-shaped (the step cannot fail); the true pure-bind sites are the four converted below.

### Converted sites

- `fjs/dev/package_json` — `validatePackageJsonText`
- `fjs/media/revision` — `validate` and `decodeText`
- `fjs/cas/evo` — `addRevision`, via `buildRevision`

### Also

`fjs/types/result`'s `@example` still imported from the pre-migration `./module.f.ts` and from a `Result` that now lives in `types.ts`; refreshed to the post-migration form and extended to show `okThen`.

### Checks

- `npx tsc` clean
- `npm test` — 2523 pass, 0 fail (2522 before; `okThenTest` is the new one)
- `npm run cov` — all four changed modules at 100% line / branch / function
- `node fjs/module.mjs ci` and the nanvm update generator produce no diff
- CI green on all 19 checks


o2alexanderfedin (approved):
Verified at `de2122b1` against `origin/main` (`7ae03f04`), Node v23.11.0. Clean — approving.

### What I ran

- `npm test` — **2523 pass / 0 fail**, against **2522** on `origin/main`. The delta is exactly `okThenTest`, matching the description.
- `npm run cov` — `tests 2459, pass 2459`, `all files 99.94`. All four changed modules at **100.00 / 100.00 / 100.00** (line / branch / function): `fjs/types/result`, `fjs/dev/package_json`, `fjs/media/revision`, `fjs/cas/evo`. So the new branches are genuinely exercised, not merely present.

### Proof coverage (§3.2) — mutation-tested, not taken on faith

`okThenTest` covers all three paths (`ok` → `ok`, `ok` → `f`'s own `error`, incoming `error` passed through). I broke each arm separately and confirmed each mutation is killed:

| mutation | result |
|---|---|
| `f => r => f(r[1])` (error-arm removed, never short-circuits) | **FAIL** — `error,odd`: `f` was invoked on the error payload `7` |
| `f => r => r` (ok-arm removed, `f` never called) | **FAIL** — `ok,42` |
| baseline | PASS |

The first mutation is also the direct evidence that `f` is *not* invoked on an `error` value: with the guard removed, `half(7)` runs and produces `'odd'` where the passed-through `error(7)` is expected. The classic gap here — a bind with no error short-circuit test — is not present.

### Type signature

The emitted surface is

```
/fjs/types/result::okThen = <T, R, E>(f: (value: T) => Result<R, E>) => <F>(r: Result<T, F>) => Result<R, E | F>
```

`F` binding on the second arrow mirrors `mapOk`, so one `okThen(f)` applies to results carrying any error type. The proof exercises the union concretely: an incoming `Result<number, number>` through `half: (n: number) => Result<string, string>` type-checks and yields `Result<string, string | number>`. The `E | F` widening is a deliberate departure from the todo's single-`E` sketch and the PR body explains it; it is the strictly more usable signature, and it removes the `error(v)` re-wraps at the converted sites rather than hiding them.

### Not a duplicate

`fjs/types/result` exports `ok`, `error`, `unwrap`, `invert`, `mapOk`; `module.mjs` adds `tryCatch` / `asyncTryCatch`. No existing pure bind. `fjs/effects`' `okStep` is the effectful twin (returns `Effect`), not an overlap. Nothing for the #1510 DRY set to object to here — this addition removes duplication rather than adding it.

I also re-derived the "the todo overcounts, only four true sites" claim rather than accepting it: `resolveParent` / `resolveParents` short-circuit inside `.step` and are `Effect`-shaped, and `decodeReadRevision` rewrites the error message instead of passing it through, so none of the three is an `okThen` site. The claim holds.

### `addRevision` / `buildRevision`

The split is behaviour-preserving: `okThen(buildRevision(input))(parentsResult)` followed by the existing `error` guard is the same short-circuit the old inline `if (parentsResult[0] === 'error')` performed, and step order, the `checkReferences`-before-canonicalize ordering, and the canonicalization itself are unchanged. Declining to force the remaining five guards into a chain (they read values resolved by earlier steps) is the right call and is documented in place.

### Public surface — dual axis

`npm run prepack` in both trees, then `extract.mjs` / `consts.mjs`:

- **Exported type aliases: byte-identical** (46,202 bytes both sides, key-set diff empty). No new `export type`, so no §6.2 `_`-prefix question arises.
- **Exported consts: 893 → 894, one addition**, `okThen` above. Nothing removed, nothing changed, nothing widened to `any` — notably `validatePackageJsonText`, `validate` and `decodeText` keep their exact prior signatures despite being rewritten.

### Other checks

- **Links** — `linkcheck.mjs` broken-link *sets* are identical to main (162 lines each, `diff` empty). Deleting `fjs/types/result/todo/ok-then.md` strands no reference; the new `../../types/result/module.f.mjs` link from `fjs/cas/evo` resolves.
- **CHANGELOG (§8.3)** — present under `Unreleased`, links only `/pull/1519`. Correct that it carries no `**BREAKING CHANGES:**` prefix: this is a pure addition, and the surface diff confirms nothing existing moved.
- **§4 module header** — `@module` survives into `fjs/types/result/module.f.d.mts`.
- **`@example`** — I executed the refreshed block verbatim as a script; it runs clean. Worth noting it is documentation only, not wired to the proof's `example` function, so it is not guarded against future drift — no action needed, just not a covered path.

### Not verified

- Rust gates — `nanvm-lib/` untouched.
- Generator round-trip — `fjs/nanvm/` untouched. I did not re-run `ci-update`.

PR #1523, opened 2026-08-13
Add proof test for toolsCall with absent arguments field

sergey-shandar:
## Summary
Added a new proof test to verify that the `fromRegistry` function correctly defaults a missing `arguments` field to an empty object (`{}`) before dispatching to a tool entry's handler.

## Key Changes
- Added import for `Ts` type from the RTTI module
- Added imports for `runPure`, `fromRegistry`, `toolEntry`, and `okResult` utilities
- Added new `registry` test suite with `toolsCallAbsentArgumentsDefaultsToEmptyObject` test case

## Implementation Details
The test exercises the `fromRegistry` function's behavior when a `toolsCall` request omits the `arguments` field. It verifies that:
- The missing `arguments` field is defaulted to `{}` before reaching the tool entry's handler
- This default is applied at the registry level (bypassing `mcpStep`) so validation against the tool's schema succeeds
- The handler receives the properly defaulted arguments and can execute successfully

This ensures that callers don't need to explicitly provide an empty object for tools that accept no arguments.

https://claude.ai/code/session_01J16odivnRNmLVaqJRxQmmG


o2alexanderfedin (approved):
Approving. The case earns its place, and — unlike #1518 — it pins the branch rather than a message.

Reviewed the real diff (`git diff $(git merge-base origin/main HEAD) HEAD`); the branch is behind `main` by #1518 and #1519, so a plain two-dot diff misleadingly shows it removing `okThen` and the `eof-without-newline` case. The actual change is one new proof case in `fjs/protocol/mcp/proof.f.mjs`.

## Gates

- `npx tsc --noEmit` → 0
- `npm run prepack` from a clean tree → 0
- `npm test` → 2523 pass / 0 fail. Merge base `7ae03f04` is 2522 and `origin/main` is 2524, so the arithmetic is exactly `+1` for this case and `+2` on main for the two commits the branch is missing — nothing lost.
- No CHANGELOG entry, which matches §8.3 and the precedent of #1509 / #1515 / #1518: proof-only changes don't get one.

## Mutation test

Target is `fjs/protocol/mcp/module.f.mjs:215`, `entry.handle(args === undefined ? {} : args)`.

The obvious mutation — deleting the default outright — turns out not to be a fair control: `tsc` rejects it on both sides with `TS2345 … Type 'undefined' is not assignable to type 'Unknown'`, so `npm test` fails at the type-check step before any proof runs. I used mutations that type-check cleanly instead:

| mutation | `tsc` | new case |
|---|---|---|
| `args === undefined ? null : args` | clean | **fails** |
| `args === undefined ? [] : args` | clean | **fails** |
| `args === undefined ? '' : args` | clean | **fails** |

**Negative control:** the first of those, applied to `origin/main`, passes the entire suite — 2524 pass / 0 fail. So the path really was uncovered and this case is what covers it.

**On the #1518 weakness specifically:** that one pinned `obj[1].message`, so a mutation that changed the branch but preserved the message slipped through. This case doesn't have that shape — it asserts on the *successful* result (`result.content[0].text === 'ok'`), and success is only reachable when the default is an object the entry's schema accepts. Every mutation above changes the branch's outcome and every one is observed. I also checked that the assertion isn't load-bearing on incidental structure: replacing the whole default with a constant `entry.handle({})` is killed by 31 other cases across the suite, so the case isn't standing in for coverage that already exists elsewhere.

## One residual gap, not a blocker

The case pins "the default is *an object that validates against the entry's schema*", not "the default is `{}`". `echoArgs` is the empty struct, so anything validates against it — `args === undefined ? { extra: 1 } : args` type-checks and passes the full 2523-case suite untouched. If you want the stronger statement, have the handler capture what it receives and assert it equals `{}`:

```js
let received = /** @type {unknown} */ (undefined)
const entry = toolEntry('echo', 'echoes', echoArgs, a => { received = a; return pure(okResult('ok')) })
…
assertEq(stringify(received), '{}')
```

That would kill the `{ extra: 1 }` mutation too. As written the case still covers the branch that matters, which is why this is a suggestion rather than a finding.

Worth noting the branch will need `main` merged in before it lands (`mergeable_state: blocked`, two commits behind); nothing in the new case conflicts with either missing commit.

PR #1525, opened 2026-08-13
Add defensive branch coverage for endObject non-object guard

sergey-shandar:
## Summary
Added a new test case to the proof object that covers a defensive, unreachable code branch in the `endObject` function. This improves test coverage for edge cases that cannot be reached through normal parsing flow.

## Key Changes
- Added `endObject.nonObjectTop` test case to the proof object
- Tests the defensive guard that handles non-object values at `state.top`
- Creates a synthetic parser state with `top: null` to trigger the unreachable branch
- Verifies the function correctly transitions to the `'result'` state

## Implementation Details
- The test acknowledges that `endObject` is only invoked when `state.top` is an object during normal parsing (guaranteed by value-states like `'{'`, `'{k'`, `'{:'`, `'{v'`, `'{,'`)
- The non-object guard exists as a defensive measure but is unreachable through `parseFromTokens`
- Direct invocation with a crafted state object allows coverage of this defensive branch

https://claude.ai/code/session_01SEkx5h5qGhKB7C6wAmsu2j

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ history.
 
 ## Unreleased
 
+- `fjs/djs/parser`'s `proof` covers `endObject`'s defensive non-object-top
+  guard directly, the same way it already covers `pushKey` and `endArray`'s
+  equivalent unreachable branches
+  [#1525](https://github.com/functionalscript/functionalscript/pull/1525)
 - `fjs/text/ascii` owns the hex-digit codec: `hexDigitValue`,
   `hexDigitCodePoint`, and the `a-f` / `A-F` ranges. The JSON serializer and
   both tokenizers use it instead of rederiving the offsets; the DJS tokenizer
--- a/fjs/djs/parser/module.f.mjs
+++ b/fjs/djs/parser/module.f.mjs
@@ -544,4 +544,22 @@ export const proof = {
             assertEq(result.state, 'result')
         },
     },
+    endObject: {
+        // `endObject` is only ever invoked while `state.top` is an object
+        // (the `'{'`/`'{k'`/`'{:'`/`'{v'`/`'{,'` value-states guarantee it),
+        // so its non-object guard is a defensive branch unreachable through
+        // `parseFromTokens`. Call it directly to cover that branch.
+        nonObjectTop: () => {
+            /** @type {_ParseValueState} */
+            const state = {
+                state: 'exportValue',
+                module: { refs: null, modules: null, consts: null },
+                valueState: '{,',
+                top: null,
+                stack: null,
+            }
+            const result = endObject(state)
+            assertEq(result.state, 'result')
+        },
+    },
 }


o2alexanderfedin (approved):
Reviewed at `cad1b47` (draft — reviewing it anyway, per the usual arrangement). The
case does what it says: it is a real mutant-killer, the branch really is
unreachable through `parseFromTokens`, and the coverage delta is measurable.
**Approving**, with one observation about assertion strength that applies equally
to the `endArray` sibling already on main.

### Mutation test

Removing the guard entirely — `const obj = fromMap(state.top[1])` — is the
mutation that matters, since with `top: null` the true arm cannot be taken.
(`tsc` rejects that edit, correctly, so I ran the runner directly with
`node ./fjs/module.mjs t` rather than through `npm test`.)

| tree | result |
| --- | --- |
| this PR (`cad1b47`) | `pass: 2531, fail: 1` — the one failure is `proof.endObject.nonObjectTop`, nothing else |
| `origin/main` (`28a2e99`) | `pass: 2531, fail: 0` — kills nothing |

So the new case is the only thing in the suite that constrains that branch, and
it did not previously exist.

### The branch is genuinely unreachable through `parse`

Checked both ways rather than trusting the comment.

Statically: `endObject` has exactly three call sites — `parseObjectStartOp`,
`parseObjectNextOp`, `parseObjectCommaOp` — reached from `foldOp`'s dispatch only
at value-states `'{'`, `'{v'` and `'{,'`. `'{'` is written by `startObject`,
which sets `top = ['object', null, '']`; `'{v'` is written by `pushValue`'s last
branch, which is reached only when `top` is non-null and non-array (the `null`
case returns earlier in both `state` cases); `'{,'` is written by
`parseObjectNextOp`'s `','`, which carries `top` through unchanged. `top` is an
object at all three.

Empirically: the guard-removal mutation on `origin/main` breaks nothing across
all 2531 tests, including the parse-level DJS ones — nothing in the existing
suite reaches it. Calling the internal directly is the right and established
answer here, as with `pushKey.nonObjectTop` and `endArray.nonArrayTop` in this
same file.

### Coverage delta (`npm run cov`, both trees)

| | main | PR |
| --- | --- | --- |
| `fjs/djs/parser/module.f.mjs` branch | 93.61 | **94.05** |
| all files, branch | 98.17 | **98.20** |

Lines and functions stay at 100.00 for that file on both sides. `npm test`:
2531 → 2532.

### One observation: the assertion pins the branch but not its result

Unlike the case on #1518, this one *does* distinguish the guarded path — the
mutation above kills it. But it asserts only `result.state`, and that is coarser
than the branch it covers. Changing the fallback from `: null` to `: undefined`
type-checks cleanly and the case still passes (`pass: 2532, fail: 0`), even
though the value pushed into `module.consts` changed. The branch's whole
observable output is that value.

I want to be clear this is a suggestion, not a defect: the pre-existing
`endArray.nonArrayTop` has exactly the same shape, so the new case is faithful to
the established pattern rather than weakening it, and the value is unobservable
through the public API by construction. If you want the case to pin behaviour and
not just reachability, asserting the pushed const (e.g. that
`toArray(result.module.consts)` is `[null]`) would close it — and the same edit
would apply to `endArray`.

### Nit

The comment lists `'{'`/`'{k'`/`'{:'`/`'{v'`/`'{,'`, but only three of those five
dispatch to `endObject`; `'{k'` and `'{:'` go to `parseObjectKeyOp` /
`parseObjectColonOp`, neither of which calls it. The claim stays true (all five
are object states), but the `endArray` comment right above lists exactly the
three states that reach it, so the two read inconsistently.

### Rest of the battery

`npx tsc --noEmit` 0; `npm run prepack` from a clean tree 0; public surface
(`extract.mjs` + `consts.mjs`, after `prepack` in both trees, 46 KB / 65 KB of
output) byte-identical to main; broken-link sets identical (140 on both);
`@module` header survives into `module.f.d.mts`. CHANGELOG entry is at the top of
`Unreleased`, links only `/pull/1525`, and is within the length guidance.

PR #1526, opened 2026-08-13
Exact emitted declarations: keep every @module header, remove the last elided any

sergey-shandar:
Two declaration-fidelity fixes from `todo/migrate-typescript-to-mjs.md`, both of the invisible-in-repository class: `npx tsc` and `npm test` stay green either way, and only a consumer reading the published `.d.mts` sees the difference.

## 1. Every emitted module declaration keeps its `@module` header

Before this change, 28 of the 127 emitted `module.f.d.mts` / `module.d.mts` files had no `@module` documentation header.

The TODO records the mechanism: declaration emit rewrites the import list, and when the header comment is attached to the first `import` statement (no blank line after the leading JSDoc block), the header is dropped with that statement. The TODO named four modules to fix — but that count was as of #1505 and stale. Measured at `main` today:

- **24 modules** had an `@module` header in source that emit dropped — the four known ones (`fjs/types/btree/remove`, `btree/set`, `list`, `nullable`) plus 20 more (`sorted_set`, `range_set`, `rtti` ×5, `byte_set`, `fjs/module.f.mjs`, `effects/node`/`mock`/`eff`/`list`, `media/json/serializer`, `media/html`, `media/nix`, `crypto/vdf`, `path`, `mcp`, `fsc`). Each gets the load-bearing blank line after its header block.
- **4 modules** had no `@module` header at all (`fjs/types/result/module.mjs`, `fjs/types/function/module.f.mjs`, `fjs/effects/module.mjs`, and the `fjs/module.mjs` CLI entry) — a §4-convention gap rather than an emit loss. Each gets a leading JSDoc block with `@module` and a one-sentence description, folding its pre-existing `@import` tags into the block per convention.

**Verified:** `grep -L '@module'` over every emitted module declaration after declaration emit — **0 missing of 127**, up from 99 of 127. The diff for this part is comments and blank lines only.

## 2. `fjs/media/json/schema` emits an exact declaration

`unknownThunk` was typed only through a `@type {const}` cast — the exact middle-row failure of the AGENTS.md §6.2 table: it type-checks, but the emitter has no name for the recursive position. The emitted declaration collapsed `not` to `/*elided*/ any` and four sibling fields (`anyOf`, `items`, `prefixItems`, `properties`, `additionalProperties`) to bare `any` — this was the last `elided` anywhere in the published type surface.

The fix is §6.2's own prescription: an explicit `@type {() => readonly ['const', typeof unknownConst]}` closes the recursion cycle by name, and the emitter prints `typeof unknownConst` instead of inlining and giving up.

No exact `Equal` round-trip assert accompanies it because the schema's Phantom `$out` (`_UnknownConst`) intentionally differs from the derived type in field optionality — the TODO's §6.2 note records this.

**Verified:** the emitted `module.f.d.mts` contains 0 `elided` and 0 bare `any` (was 1 + 5); repo-wide `elided` count across all emitted declarations is **0**.

## Battery

- `npx tsc` clean; `tsc --noEmit false --emitDeclarationOnly && tsc` (the prepack round-trip) clean.
- `npm test` **2532 / 0**.
- CHANGELOG entries added for both fixes; the migration TODO records the corrected header measurement and the schema outcome.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -318,10 +318,12 @@ and the test entry is authored `all.test.mjs`.
 The runtime-import grouping applies to repository-owned relative imports, not to
 external or built-in modules: a FunctionalScript module may depend at runtime on
 external modules and on repository `.mjs`, and there is no relative runtime `.ts`
-import group at all. In a `module.*` file, the blank line after the leading JSDoc
-block is required even when the module has no `@import` tags; it keeps the
-`@module` header detached from the first import/declaration and preserves it
-through declaration emit.
+import group at all. The blank line after the leading JSDoc block is required
+in every file that has one — `module.*` (even with no `@import` tags),
+`types.ts`, and `proof.*` alike; it keeps the header detached from the first
+import/declaration and preserves it through declaration emit. `types.ts` is
+the most easily missed case, since its emitted `types.d.ts` is the published
+documentation for the whole type-level API.
 
 Where each kind of documentation belongs:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ history.
 
 ## Unreleased
 
+- Emitted declarations keep their `@module` documentation header: 24 modules
+  lost it to declaration emit and 4 never had one; all 127 module
+  declarations now carry it
+  [#1526](https://github.com/functionalscript/functionalscript/pull/1526)
+
 - **BREAKING CHANGES:** the npm package ships no `.js` files: the empty
   `types.js` stubs and compiled test files are gone. Import types only with
   fully erased `import type`, and use `fjs/emergent_testing/all.test.mjs` as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ history.
   lost it to declaration emit and 4 never had one; all 127 module
   declarations now carry it
   [#1526](https://github.com/functionalscript/functionalscript/pull/1526)
+- `fjs/media/json/schema`'s emitted declaration is exact: an explicit `@type`
+  naming the recursive schema by `typeof` replaces the `@type {const}` cast
+  that collapsed `not` to `/*elided*/ any` and four sibling fields to `any`
+  [#1526](https://github.com/functionalscript/functionalscript/pull/1526)
 
 - **BREAKING CHANGES:** the npm package ships no `.js` files: the empty
   `types.js` stubs and compiled test files are gone. Import types only with
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,15 +20,15 @@ history.
 
 ## Unreleased
 
-- Emitted declarations keep their `@module` documentation header: 24 modules
-  lost it to declaration emit and 4 never had one; all 127 module
-  declarations now carry it
+- Emitted declarations keep their documentation headers: 24 modules lost
+  `@module` to declaration emit and 4 never had one; all 127 module
+  declarations now carry it, and every `types.ts` / `proof.*` prose header
+  now leads its emitted declaration
   [#1526](https://github.com/functionalscript/functionalscript/pull/1526)
 - `fjs/media/json/schema`'s emitted declaration is exact: an explicit `@type`
   naming the recursive schema by `typeof` replaces the `@type {const}` cast
-  that collapsed `not` to `/*elided*/ any` and four sibling fields to `any`
+  that collapsed `not` to `/*elided*/ any` and five sibling fields to `any`
   [#1526](https://github.com/functionalscript/functionalscript/pull/1526)
-
 - **BREAKING CHANGES:** the npm package ships no `.js` files: the empty
   `types.js` stubs and compiled test files are gone. Import types only with
   fully erased `import type`, and use `fjs/emergent_testing/all.test.mjs` as
--- a/fjs/ci/common/module.f.mjs
+++ b/fjs/ci/common/module.f.mjs
@@ -9,6 +9,7 @@
  *
  * @import { Step, Job, MetaStep, StepType } from './types.ts'
  */
+
 import { actions, images } from '../config/module.f.mjs'
 import { option, array, record, string } from '../../types/rtti/module.f.mjs'
 import { parse as rttiParse } from '../../types/rtti/parse/module.f.mjs'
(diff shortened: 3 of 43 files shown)


o2alexanderfedin (approved):
Approving. I re-derived both headline claims from scratch rather than taking them from the title or the CHANGELOG, and both hold exactly.

## Gates

Both trees cleaned (`git clean -xdfq -e node_modules`) before every measurement.

| | this PR (`a8cbd7c5`) | `origin/main` (`283db73a`) |
|---|---|---|
| `npx tsc --noEmit` | 0 | 0 |
| `npm run prepack` (clean tree) | 0 | 0 |
| `npm test` | 2532 pass / 0 fail | 2532 pass / 0 fail |
| `bin/linkcheck.mjs` broken-link set | identical to main (162 lines, `diff` empty) | — |

## The `@module` survival property — verified, not assumed

Counted authored `module.f.mjs` / `module.mjs` files against their emitted
`module.f.d.mts` / `module.d.mts`, in both trees, each from a clean `prepack`:

| | authored with `@module` | emitted decls with `@module` |
|---|---|---|
| `main` | 123 / 127 | **99 / 127** |
| this PR | **127 / 127** | **127 / 127** |

So 28 emitted module declarations lacked the header on `main`: 4 because the
source never had one (`fjs/effects/module.mjs`, `fjs/module.mjs`,
`fjs/types/function/module.f.mjs`, `fjs/types/result/module.mjs`) and 24 because
emit swallowed it. That is exactly the CHANGELOG's "24 lost it, 4 never had one,
all 127 now carry it" — the numbers are right.

Negative control on the mechanism itself, at this head: deleting
`fjs/types/list/module.f.d.mts`, removing the blank line after the header in
`fjs/types/list/module.f.mjs`, and re-running the emit pass takes
`grep -c '@module'` on the emitted declaration from 1 to 0; restoring it
returns 1. (Deleting the stale `.d.mts` first is required — with it present,
nodenext resolves `./module.f.mjs` through the declaration and the mutated
source never enters the program, so the probe silently passes. Worth knowing for
anyone repeating this.)

## The last elided `any` — verified before/after

Across all 335 emitted `.d.mts` / `.d.ts`, counting `elided` literally and
counting `\bany\b` only after stripping block and line comments (so JSDoc prose
does not inflate it):

| | `elided` | type-level `any` |
|---|---|---|
| `main` | 1 (`fjs/media/json/schema/module.f.d.mts`) | 7 — 6 in the schema decl, 1 pre-existing in `fjs/djs/tokenizer/module.f.d.mts` |
| this PR | **0** | **1** (the pre-existing tokenizer one) |

The full declaration-set diff between the two trees is 29 files: 28 that gain
nothing but the JSDoc header, and `fjs/media/json/schema/module.f.d.mts`, where
`unknownThunk`'s inlined structural expansion is replaced by
`() => readonly ['const', typeof unknownConst]` and the six recursive positions
become exact. No other signature moves anywhere in the tree.

## Public surface

Dual-axis surface diff over the emitted declarations
(`bin/extract.mjs` 46,202 bytes; `bin/consts.mjs` 64,198 bytes — non-empty on
both sides) is **byte-identical** between `main` and this PR. No new
unprefixed public `export type`, no signature widened to `any`; `unknownThunk`
stays a non-exported `declare const`.

Not re-run: the §6.2 `prepack` counter-example (mutate the worked example so the
recursive positions lose their `typeof` names, confirm `tsc --noEmit` passes
while `prepack` fails `TS2344`). This PR touches neither the `prepack` script
nor `tsconfig.json` nor the emit pipeline, so the gate is unchanged by
construction — `prepack` is still `tsc --noEmit false --emitDeclarationOnly && tsc`
and still exits 0 here from a clean tree.

## Nits, none blocking

1. **"four sibling fields" is five.** Both the CHANGELOG entry and the
   `todo/migrate-typescript-to-mjs.md` addition say the cast collapsed `not` to
   `/*elided*/ any` "and four sibling fields to `any`". On `main` the affected
   siblings are `anyOf`, `items`, `prefixItems`, `properties` **and**
   `additionalProperties` — five, which is why the file measures 6 `any` and not
   5.

2. **Three authored files still lose their header**, all pre-existing on `main`
   and all outside the `module.*` scope the CHANGELOG claims:
   `fjs/dev/update/proof.f.mjs`, `fjs/types/rtti/ts/types.ts`,
   `fjs/types/rtti/types.ts` — each has its header flush against the first
   import. Repo-wide the count goes 27 → 3, so the PR's actual scope is right;
   it is only the title's "every `@module` header" that overreaches slightly.
   The two `types.ts` cases are worth a follow-up given that §4 calls `types.ts`
   the most dangerous case, and the rule as written ("in a `module.*` file")
   does not currently reach them.

3. The new CHANGELOG entries are separated from the rest of the `Unreleased`
   list by a blank line; every other entry in that section is contiguous.

Entries link only the PR, sit under `Unreleased`, and no released section was
touched — §8.3 satisfied for a code change.

sergey-shandar:
All three nits addressed in `a7f981f`, and nit 2 grew into the proper fix:

1. **"four sibling fields" → five** in the CHANGELOG entry and the migration TODO — your count was right, which is exactly why the file measured 6 `any` and not 5.
2. **The three out-of-scope files are fixed** — and `fjs/dev/update/proof.f.mjs` turned out to carry an `@module` tag §4 reserves for `module.*` files, now removed along with getting its blank line. Then, re-measuring with your ruler (prose header flush against the first import) instead of mine (`@module` present in the emitted declaration) surfaced **eight more**: `fjs/ci/common/module.f.mjs` + its `types.ts`, `btree/find`, `btree/types`, `function/compare`, and three `rtti` `types.ts` files. Those got lucky rather than being correct — their first import happens to survive into the declaration, so the header rides along *below* it instead of leading the file. All eight get the blank line; measured after: all 127 module declarations carry `@module` **and** every prose header repo-wide now leads its emitted declaration. AGENTS.md §4's blank-line rule is extended from `module.*` to every file with a leading doc block, naming `types.ts` as the easily missed case per your suggestion.
3. **The stray blank line** splitting the `Unreleased` list is removed.

Also noted for anyone repeating your negative control: your observation about deleting the stale `.d.mts` first (nodenext resolving through the declaration so the mutated source never enters the program) matches what I hit while measuring — it's the same resolution-priority mechanism that makes the prepack round-trip check work at all.

`npm test` 2532/0 at this head.

---

PR #1527, opened 2026-08-13
bigfloat: keep `decToBin`'s mantissa at 53 bits when rounding carries

sergey-shandar:
Fixes the `round53-overflow` issue in `fjs/types/bigfloat/todo/` (deleted here).

## Problem

`round53` reduces the mantissa to 54 bits with `decreaseMantissa(...)(twoPow54)`, rounds to 53 by adding the dropped bit back, and returns without re-normalizing. When the 53-bit mantissa is all ones the round-up carries into a 54th bit, so `decToBin` leaves with a mantissa of exactly `2^53` — one bit wider than the function is named for:

```js
decToBin([0b11_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111n, 0]) // 54 ones
// before: [0b10_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000n, 1] // 2^53 — 54 bits
// after:  [0b1_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000n, 2]  // 2^52 — 53 bits
```

The numeric value was already correct (`2^53 * 2^1 === 2^52 * 2^2 === 2^54`); only the representation was out of range. `decToBin` exists to produce the IEEE-754 binary64 significand, so a consumer assuming `abs(m) < 2^53` — to emit the significand field, to compare two `BigFloat`s by mantissa, or to round-trip through a `number` — was wrong on exactly these inputs. Nothing consumes `decToBin` today (`bigfloat`'s importers take only `multiply` and the type), which is why the issue was P3 rather than higher.

## Change

Both return paths go through one `renormalize53` helper: a magnitude of exactly `2^53` is shifted back one bit with the exponent incremented. The bit shifted out is always 0 at that point, so the value is unchanged and no second rounding decision is needed. The tie-to-even and plain round-up paths differed only in the addend, so they collapse into a single `return` with the choice expressed as a conditional.

The alternative — declaring a 54-bit mantissa acceptable and documenting the postcondition as "value-correct, not normalized" — would push normalization onto every future consumer of a function that already normalizes on the way in.

`decToBin`'s JSDoc now states the postcondition: zero returns `[0n, 0]`, and every other input returns exactly 53 significant bits, `2^52 <= abs(m) < 2^53`. It also records what this does *not* cover — see below.

## Proofs

A new `roundingCarry` group covers the carry through both paths in both signs, with the operands written as binary literals like the rest of the file:

- 54 ones and its negation — an exact tie whose round-to-even goes up because `2^53 - 1` is odd.
- 55 ones and its negation — not a tie (the dropped bits are `11`), so the carry comes from the plain round-up path.
- A neighbour that ties *down* and so doesn't carry, keeping the two outcomes side by side.

Each asserts the width with `bitLength` and re-derives `m * 2^e`, so the value is checked to be unchanged and a negative mantissa is checked as strictly as a positive one — `m < 2^53` would have been vacuous for the negative case and would have passed against the old, broken result.

## Not in scope: binary64's exponent range

This fixes the significand *width*; the exponent stays unbounded, so `decToBin` still knows nothing about subnormals or overflow. That gap is now recorded in `decToBin`'s JSDoc and filed as `fjs/types/bigfloat/todo/binary64-exponent-range.md`, with a reproduction: a decimal strictly below the midpoint between the subnormals `3 * 2^-1074` and `4 * 2^-1074` comes back as *exactly* that midpoint, so a consumer rounding onto the subnormal grid ties-to-even upward and lands an ulp above the correctly-rounded `double`. No wrapper can fix it — reaching the subnormal range needs a single rounding to a precision chosen from the target exponent, which is a change inside `round53`, not on top of it.

## Verification

- `npx tsc` — clean.
- `npm test` — 2537 pass, 0 fail. The new proofs fail against the pre-fix implementation (`bitLength` 54 ≠ 53).
- `npm run cov` — `fjs/types/bigfloat/module.f.mjs` stays at 100% line / branch / function.

https://claude.ai/code/session_01XCaYyBsweMU27CRKrjHKsJ


sergey-shandar:
1. Use bin or hex, not hex
2. What about subnormals (when `e === 0x000`)?

sergey-shandar:
Both addressed in 88182ba.

**1. Literals.** The `roundingCarry` operands are now binary literals in the file's existing style, instead of the computed `1n << 54n` constants and decimal values I had. The PR description above is updated the same way.

**2. Subnormals — you're right, and it's worse than a gap in the docs.** `decToBin` fixes the significand *width* and leaves the exponent unbounded, so it knows nothing about either end of binary64's range: no overflow to infinity (`decToBin([1n, 400])` happily returns `[7686445155841023n, 1276]`), and no reduction to the fewer-than-53 bits a subnormal actually carries.

The subnormal end is a correctness bug for any future consumer, not just a missing feature, because **the second rounding cannot be done by the caller**. Take the value one part in `2^126` below the midpoint between the subnormals `3 * 2^-1074` and `4 * 2^-1074`, written exactly as a decimal:

```js
const mantissa = (7n * 2n ** 125n - 1n) * 5n ** 1200n   // dyadic → exact in decimal
decToBin([mantissa, -1200])
// => [7881299347898368n, -1125]  === 7 * 2^-1075 === exactly 3.5 * 2^-1074
```

The input is strictly *below* the midpoint, so the correctly-rounded double is `3 * 2^-1074` — which is what `Number(...)` gives for the same decimal. `decToBin` returns the midpoint itself. A consumer rounding that onto the `2^-1074` grid then hits a tie that never existed in the input, and ties-to-even sends it up to `4 * 2^-1074`: one ulp wrong, with the evidence already discarded. Getting it right needs a single rounding to a precision chosen from the target exponent, i.e. a change inside `round53` where the remainder is still in hand.

That's more than this PR should carry, so it's `fjs/types/bigfloat/todo/binary64-exponent-range.md` — the reproduction above, plus a proposal to parameterize `round53`/`renormalize53` by a target `Format` (`precision`, `minExp`, `maxExp`), derive the available precision as `min(precision, exponent - minExp + precision)`, and signal overflow as a `try*`-shaped `Nullable<BigFloat>` since `BigFloat` can't encode an infinity. `decToBin`'s JSDoc now records the limitation so nobody reads "53-bit significand" as "ready to become a `number`". Happy to fold it into this PR instead if you'd rather have it in one go.

The `Workers Builds` red above was a build that started and finished in the same second with empty output; `npm run website` succeeds on 88182ba locally, and Cloudflare has already re-queued the same commit. Every GitHub Actions job is green.

---

o2alexanderfedin (approved):
Approving. Draft status is not a blocker. I started at `978c29b2`, the head moved to
`88182ba0` mid-review, and everything below was re-run at `88182ba0`. Between the two
heads `module.f.mjs` changes in comments only (executable lines identical) and the proof
cases were re-expressed as binary literals — I checked the literals rather than trusting
them: the 54-ones literal is exactly `2n**54n-1n`, the 55-ones one `2n**55n-1n`, and the
non-carrying neighbour `2n**54n-3n` with target `2n**54n-4n`.

## Gates at `88182ba0`

`npx tsc --noEmit` 0 · `npm run prepack` 0 from a clean tree · `npm test` **2537 pass /
0 fail** against **2532** on `origin/main` (`283db73a`), i.e. exactly the five new cases
and nothing else moved. Broken-link set identical to main (`diff` empty), which also
confirms deleting `todo/round53-overflow.md` stranded nothing — no reference to it
survives anywhere in the repo. Dual-axis surface diff over the emitted declarations
(`bin/extract.mjs` 46,202 bytes, `bin/consts.mjs` 64,198 bytes, non-empty on both sides)
is byte-identical to main; the only declaration change in the tree is the JSDoc block on
`decToBin`.

## I found the carry boundary myself before reading the proof

`decreaseMantissa(…)(twoPow54)` leaves `m54 ∈ [2^53, 2^54)`, so `m53 = m54 >> 1n ∈
[2^52, 2^53)` and a round-up can only ever produce exactly `2^53` — never more. Both
rounding paths reach it:

| input | `origin/main` | this PR | same value? |
|---|---|---|---|
| `2^54-1` (54 ones, exact tie, `2^53-1` odd → up) | `9007199254740992@1`, **54 bits** | `4503599627370496@2`, 53 bits | yes |
| `2^55-1` (dropped bits `11`, plain round-up) | `9007199254740992@2`, **54 bits** | `4503599627370496@3`, 53 bits | yes |
| `-(2^54-1)` | `-9007199254740992@1`, **54 bits** | `-4503599627370496@2`, 53 bits | yes |
| `2^54-3` (tie, `2^53-2` even → down) | 53 bits | 53 bits | yes |
| `2^54-2` (just below) | 53 bits | 53 bits | yes |

Same set the proof picked, arrived at independently.

## Systematic sweep against a trusted oracle

**10,196,014** non-zero inputs — neighbourhoods `2^k ± 300` for `k = 40..70`, all-ones
and near-all-ones patterns `2^k - 1 ± 4` for `k = 1..200`, ~190k pseudorandom mantissas
of widths 1..120 bits, and decimal all-nines `9…9` up to 60 digits — each with a range of
decimal exponents and each in both signs. Oracle: `Number(\`${m}e${e}\`)`, which the spec
requires to be correctly rounded, decomposed to its exact 53-bit `[mantissa, exp2]` via
`BigUint64Array`. All 10,196,014 landed in the normal binary64 range, so all of them were
oracle-compared (none skipped).

| | `origin/main` | this PR |
|---|---|---|
| results whose mantissa is not exactly 53 bits | **7,324** | **0** |
| results disagreeing with native `Number` in value | 0 | 0 |

Exact differential over the same 10,196,014 inputs: **7,324 representation changes, 0
value changes** — the changed set is precisely the width failures, so the rewrite of
`round53`'s two return paths into the single `renormalize53([m53 + (tie ? m53 & 1n : o54), e53])`
expression is behaviour-preserving everywhere except the carry it is meant to fix.

## Mutation testing the new cases

- Revert the fix (`renormalize53` → identity) at this head: `roundingCarry[0..3]` fail on
  the width assertion (`54` vs `53`); 34 other cases still pass. Restoring gives 38/38.
- The value assertion is not decoration: mutate to `m === twoPow53 ? [m >> 1n, e]` (shift
  without compensating the exponent) and the same four fail on `assertEq(m << BigInt(e), value)`
  instead. So each new case pins **both** the width and the value — it does not repeat
  #1518's mistake of pinning a message, nor #1525's of pinning one field.
- Drop the guard entirely (`renormalize53` → `[m >> 1n, e + 1]` unconditionally) and
  `roundingCarry[4]`, the non-carrying neighbour, is what fails. That case does not die
  under the primary mutation, and it is not meant to — it fences the other side of the
  boundary.
- **The same mutation kills nothing on `origin/main`.** `origin/main` *is* the unmutated
  buggy state, and its own suite is green there: 2532/2532, and the bigfloat proof alone
  33/33. Transplanting this PR's `proof.f.mjs` onto `origin/main`'s implementation flips
  that to 34 pass / 4 fail with the identical messages — the cases are what does the
  killing, not anything else in the branch.

## Coverage cannot catch this class — it holds again

`npm run cov` gives `fjs/types/bigfloat/module.f.mjs` **100.00 / 100.00 / 100.00**
(line / branch / function) on *both* sides, with the repo total unchanged at
99.94 / 98.20 / 99.73; test count 2468 → 2473. The bug lived entirely inside a fully
covered expression, so no coverage signal was ever available. Worth saying out loud since
it is the second time in this series. (`npm run cov` was not vacuous here — it ran 2473
real tests.)

## The new JSDoc and `todo/binary64-exponent-range.md`

I checked the substantive claims rather than taking them on trust. The doc's worked
example is right: `decToBin([(7n * 2n ** 125n - 1n) * 5n ** 1200n, -1200])` returns
`7881299347898368@-1125`; rounding that onto the `2^-1074` subnormal grid ties-to-even
gives `4 * 2^-1074`, while the single correctly-rounded `Number` is `3 * 2^-1074` — a
genuine double-rounding error of one ulp, exactly as described. The overflow example
reproduces verbatim: `decToBin([1n, 400])` → `7686445155841023@1276`. Recording it as an
open P3 rather than fixing it here is the right split.

`todo/from-decimal.md`'s cross-link was updated to point at the module instead of the
deleted file, and §8.3 is satisfied: one `Unreleased` entry linking only the PR for the
code change, nothing for the `todo/` edits.

sergey-shandar:
Merge conflict resolved in 5602371 — the queue dropped this for `MERGE_CONFLICT` after #1526 landed.

The conflict was `CHANGELOG.md` only: both sides added entries at the top of `## Unreleased`. Both are kept, mine above #1526's two per §8.3 ("new entries always go above existing ones"). Nothing else collided — #1526 touched no file under `fjs/types/bigfloat/`, and the PR's diff is still the same 6 files / +182 / −90 as at `88182ba`, so the merge added nothing beyond the resolution.

I also checked this branch against the convention #1526 tightened — a blank line after the leading JSDoc block in `module.*`, `types.ts`, and `proof.*` alike. `bigfloat`'s `module.f.mjs` and `types.ts` already have it; `proof.f.mjs` has no leading block for the rule to apply to.

Re-verified on the merged tree: `npx tsc` clean, `npm test` **2537 pass / 0 fail**. Coverage was verified at `88182ba` (100% line/branch/function on the module) and nothing in the merge reaches `bigfloat`.

Needs a re-queue when you're ready — I can't do that myself.

---

o2alexanderfedin (approved):
Re-review at `5602371c`, after the previous approval at `88182ba0`. **Still approving.**

## What actually changed since `88182ba0`

`5602371c` is a merge commit whose parents are `88182ba0` and `d986e0e0` (current
`origin/main`, i.e. #1526). There is no new authored work: the PR-only diff against the
merge base is byte-for-byte the same patch I approved, apart from one CHANGELOG context
line where #1526's entry now sits below this one.

- Every bigfloat blob is identical to the approved head:
  `module.f.mjs`, `proof.f.mjs`, `todo/binary64-exponent-range.md`, `todo/from-decimal.md`
  all resolve to the same object ids at `88182ba0` and `5602371c`.
- `origin/main` did not touch `fjs/types/bigfloat/` or its only runtime dependency
  `fjs/types/bigint/` between `283db73a` and `d986e0e0`, so the merge could not have moved
  the ground under the fix either.

Baseline used for every comparison below: **`d986e0e0`** (`origin/main` as of this review) —
my baseline tree had been sitting on `283db73a` and was moved forward first.

## Battery re-run on the merged tree

- `npm run prepack` from a freshly cleaned tree — exit 0, both trees.
- `npm test` — 2537 pass / 0 fail, against 2532 / 0 on `d986e0e0`. The +5 are exactly
  `roundingCarry[0..4]`.
- Public surface, both axes, after `prepack` in both trees: exported type aliases and
  exported const signatures are byte-identical (46,202 and 64,198 bytes respectively, so the
  extractors did produce output). No new `export type`, prefixed or not; nothing widened to
  `any`.
- File-aware tree diff after `prepack`: the entire blast radius is `CHANGELOG.md`, the four
  bigfloat files, and the todo rename. The only declaration deltas are `decToBin`'s new
  doc comment reaching `module.f.d.mts` and `roundingCarry` appearing in `proof.f.d.mts`.
  No `elided` and no `any` in either emitted bigfloat declaration.
- Link check: 162 reported lines on each side, and the broken-link **sets** are identical —
  the `round53-overflow.md` → `binary64-exponent-range.md` rename strands nothing.

## Re-derived behaviour evidence

I re-ran this rather than carrying the earlier numbers over, and in doing so found that a
naive sweep is worthless here: over 21,600,000 pseudo-random inputs (15–20 significant
decimal digits, both signs, `e ∈ [-30, 29]`) main and this PR produce **identical** results —
0 representation differences, 0 value differences — and both match correctly-rounded
`Number` on all of them. The carry needs the exact value to sit within a half-ulp below a
power of two, which a random mantissa hits with probability ~2^-53.

Targeting that region instead — decimals built exactly from `(2^54-1)·2^(k-1)`,
`(2^55-1)·2^(k-2)`, `(2^55-2)·2^(k-2)`, `(2^54-2)·2^(k-1)` and `2^53·2^k` for
`k ∈ [-80, 80]`, padded with up to 3 extra decimal digits and perturbed by ±3 ulp of the
decimal, both signs — 45,080 inputs:

| | main `d986e0e0` | this PR |
|---|---|---|
| mantissa width ≠ 53 | **23,146** | **0** |
| disagrees with `Number` on value | 0 | 0 |

Exact differential between the two implementations on that set: **23,146 representation
changes, 0 value changes**. So the `round53` return-path rewrite is behaviour-preserving and
the guard is the only thing that moves, which is what the CHANGELOG entry claims.

Mutation check on the merged tree, to confirm the new cases are still load-bearing after the
merge: rewriting `renormalize53` to the identity (verified the edit landed before drawing a
conclusion) fails `roundingCarry` 4 times out of the 38 bigfloat proof cases; restoring it
brings them back.

## CHANGELOG (§8.3)

Sits under `## Unreleased`, links only `/pull/1527`, and characterises the change accurately:
it says the *returned mantissa* changes from 54 to 53 bits and says nothing about values,
which matches the differential — no previously-returned value changes, so the absence of a
`**BREAKING CHANGES:**` prefix is right.

## Not verified

- Nothing in `nanvm-lib/` or `fjs/nanvm/` is touched, so the Rust gates and the generator
  round-trip do not apply.
- `npm run cov` reports a vacuous `100.00` over 0 tests in this environment on `main` too, so
  I am not quoting coverage numbers; as noted last time, `bigfloat` was already fully covered
  before this change, which is exactly why coverage could not have caught this class of bug.

PR #1529, opened 2026-08-13
Add io-effect-migration.md: staged plan for Result-based effects

sergey-shandar:
## Summary

This PR introduces a comprehensive design document outlining a staged migration from the current low-level `Effect<O, T>` abstraction toward an explicit error-handling model using `Result` types. The document establishes the rationale, target API, and a 6-stage implementation plan for making effects recoverable without `try`/`catch`.

## Key Changes

- **New document:** `fjs/effects/todo/io-effect-migration.md` (354 lines)
  - Defines the `IoEffect<O, T, E> = Effect<O, Result<T, E>>` abstraction
  - Introduces a `NotImplemented` error type for unsupported operations
  - Specifies three core composition operations: `step`, `catchStep`, and `resultStep`
  - Establishes union-based type rules for error propagation (following `okThen` precedent)

## Notable Design Decisions

- **Result envelope placement:** The `Result` is part of the operation's declared return type, not wrapped by constructors. This ensures `NotImplemented` can flow through normal effect continuations in Stage 6.
- **Error propagation model:** `step` continues on `ok` and propagates errors (like Rust's `?` operator); `catchStep` recovers from errors; `resultStep` observes both branches.
- **Staged approach:** Six independent, mergeable stages from introduction through runner support, with explicit blocking dependencies and migration-only APIs kept separate during transition.
- **No speculative API:** New combinators (e.g., `historyStep`) are expected only when real consumers require them, not added preemptively.

## Related Context

The document references and coordinates with existing work:
- `map-step-combinator.md` (trailing pure projections)
- `effect-list-fold.md` and `fold-stream-combinator.md` (list/stream operations)
- `fjs/effects/module.f.mjs` (raw `step`, `okStep`, `match`)
- `fjs/effects/eff/module.f.mjs` (fluent wrapper whose API design is deferred to Stage 4)

https://claude.ai/code/session_01GNux2972gSfdfebavKQVuZ

### DIFF
--- a/fjs/effects/todo/io-effect-migration.md
+++ b/fjs/effects/todo/io-effect-migration.md
@@ -90,6 +90,10 @@ their own failures extend that channel, for example:
 IoEffect<ReadFile, Vec, NotImplemented | IoError>
 ```
 
+`IoError` does not exist yet — it names the normalized host-error type this
+migration introduces. Today the fallible operations carry bare `unknown`
+(`IoResult`); replacing that is Stage 3 work.
+
 `NotImplemented` means the runner cannot dispatch the operation and has not
 started it. The program receives control back and decides what an
 incompatible runner means for it: recover, choose a fallback operation, or
@@ -128,9 +132,9 @@ inject `NotImplemented`, and everything would be rewired a second time.
 Two consequences follow, and both are Stage 3 work:
 
 - **Runners change in Stage 3, mechanically.** Every runner handler for a
-  currently-infallible operation starts wrapping its output in `ok(...)`.
-  The already-fallible operations (`Fs`, `Fetch`, `Exec`, `Import`) need only
-  their error type refined.
+  currently-infallible operation starts wrapping its output in `ok(...)`;
+  operations that already return `IoResult` need only their error type
+  refined.
 - Until Stage 6 lands, runners remain **total** over their declared operation
   maps: `NotImplemented` exists in the type model but no runner produces it
   yet.
@@ -161,9 +165,15 @@ resultStep: (e: IoEffect<O, T, E>, f: (r: Result<T, E>) => IoEffect<Q, R, F>)
 its mirror — it unions the success channel and replaces the error type;
 `resultStep` consumes both branches and replaces both.
 
-Do not add `finallyStep` initially; it is derivable from `resultStep` and does
-not add expressive power until real consumers demonstrate a repeated policy
-worth naming.
+Expanded through the alias, `resultStep` is raw `step` at the Io
+instantiation — it adds no branch behavior of its own. It still earns its
+name: the three operations are the canonical vocabulary, and from Stage 5,
+when the raw representation goes private, `resultStep` is the public spelling
+of that instantiation. Declining `finallyStep` is the same principle — a
+derivable form earns a name only by being canonical vocabulary, and
+`finallyStep` has not shown it is; it is derivable from `resultStep` and adds
+no expressive power until real consumers demonstrate a repeated policy worth
+naming.
 
 The new `step` conflicts with today's raw-effect `step`. During migration, keep
 the existing raw API intact and expose the IoEffect operations from a separate
@@ -225,18 +235,21 @@ represented as a pair of effects.
 - [ ] Update every runner handler in the same change as its operation:
       infallible handlers wrap their output in `ok(...)`; fallible handlers
       keep their behavior with a refined error type.
-- [ ] **Migrate operation by operation, fallible operations first.** For the
-      already-fallible operations this stage is a pure type refinement
-      (`unknown` → structured error). For currently-infallible operations
-      (`Write`, `Read`, `Now`, `RandomInt`, `Sandbox`, `CreateServer`,
-      `Listen`) it changes the value consumers receive: continuations that use
-      the value break loudly at `tsc`, while continuations that discard it
-      (`() => next`) keep compiling and silently ignore the new error channel —
-      the very hazard motivating this migration. Convert those operations only
-      together with a sweep of their value-discarding call sites, and
-      consciously decide whether the trivially-total operations (`now`,
-      `randomInt`) are worth the churn before Stage 6 gives their
-      `NotImplemented` a consumer.
+- [ ] **Every operation converts; the sweep is total.** There is no list of
+      operations worth converting and no exemption for trivially-total ones —
+      an operation left on the raw contract is a hole in the error channel,
+      and Stage 6's "a runner may omit any handler" applies to all operations
+      equally. "Fallible first" is an ordering, not a scoping: operations
+      already returning `IoResult` need only a type refinement (`unknown` →
+      structured error), while converting a currently-infallible operation
+      changes the value its consumers receive. Degenerate shapes convert on
+      the same principle — `Result<never, NotImplemented>` simply has only
+      its error branch inhabited.
+- [ ] Convert each operation together with a sweep of its value-discarding
+      call sites: continuations that use the value break loudly at `tsc`,
+      while continuations that discard it (`() => next`) keep compiling and
+      silently ignore the new error channel — the very hazard motivating this
+      migration.
 - [ ] Even operations without domain-specific failures include
       `NotImplemented`.
 - [ ] Fallible operations extend the error channel with their own errors, e.g.
--- a/fjs/effects/todo/io-effect-migration.md
+++ b/fjs/effects/todo/io-effect-migration.md
@@ -97,6 +97,15 @@ treat it as fatal and panic itself (`throw`, e.g. via `unwrap`). Escalation
 belongs to the program — a runner never panics over a missing operation on
 the program's behalf.
 
+That said, runners keep their authority over execution itself. A runner may
+still interrupt or terminate a program — one attempting something malicious,
+exceeding a resource budget, or violating host policy. Interruption is a
+different mechanism from `NotImplemented`: `NotImplemented` is recoverable
+data for a capability the runner lacks, while interruption is the runner
+refusing to continue at all, and nothing in the error channel obliges it to
+hand control back. `NotImplemented` is therefore not a promise that the
+program always regains control, and it is not a security boundary.
+
 ### Where the `Result` envelope lives
 
 **Decision: the `Result` is part of the operation's declared return type**, not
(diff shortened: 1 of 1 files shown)


o2alexanderfedin (commented):
Reviewed at head `b70e4d66b2ecf63fb218b8e3a6907f858c3236d5` against `origin/main` = `c51e6dface83e738f41a37857fdcf58b42616ea8` (the branch's merge-base is exactly that commit). I started at `3e3c6037`; the head moved mid-review, so I re-verified at `b70e4d66`. The delta was the two new paragraphs on runner interruption authority plus the picked-up `virtual/proof.f.mjs` change that came from main — the merge-base diff is still one file, `fjs/effects/todo/io-effect-migration.md`, +366 lines, docs-only. Per §8.3 no CHANGELOG entry is owed, and the Rust and generator gates do not apply. No `.ts`/`.mjs` is touched, so there is no public-surface delta to diff.

The plan is well argued and, on the points where it makes checkable assertions about the tree, almost all of them reproduce. What I re-derived and confirmed:

- `Cont<O, T> = (_: Pr<O, O[0]>[1]) => Effect<O, T>` in `fjs/effects/types.ts` — the "a `Do` node's continuation accepts exactly the operation's declared output" argument for putting the `Result` in the operation's declared return type holds, and the "wrap with `mapStep(do_(…), ok)`" alternative really would leave Stage 6 no typed channel.
- `okThen`'s `@type` is `<T, R, E>(f: (value: T) => Result<R, E>) => <F>(r: Result<T, F>) => Result<R, E | F>` and its body is `r[0] === 'error' ? r : f(r[1])` — the errors are genuinely unioned, not unified, and the passed-through branch is the very tuple that arrived. The `step` / `catchStep` / `resultStep` signatures pinned here are consistent with that precedent, and all three are implementable over `Effect<O, Result<T, E>>` as written (`catchStep`'s `T | R` success channel and `F`-only error channel type-check under `pure(r)` passthrough).
- `Write` is `readonly['write', (stream: WriteConsoles, data: Vec) => void]` today, so it really does have to become `Result<void, NotImplemented>`; the `Fs` members, `Fetch`, `Exec` and `Import` all already return `IoResult<…>`, so "refine the error type" is the right characterisation for them.
- `match` in `fjs/effects/module.f.mjs` does `assert(handler !== null, command)`, and its doc really does argue "a runner cannot resume a command it has no handler for, so there is nothing for a recovery branch to do" — Stage 6 quotes it accurately and is right that it inverts it. The module header does say "**Three functions discriminate `Pure` from `Do`** — `step`, `match`, and `runPure`", so "one of the three sanctioned discriminators" is correct.
- `SandboxResult.result` is `Result<T, unknown>` and is genuinely returned data, not execution status — the carve-out is right.
- `createServer`'s listener, `sandbox`'s thunk and `test`'s body all do carry functions in the payload, so the "command name only" constraint on `NotImplemented` is justified.
- `fjs/cas/module.f.mjs` composes via `eff(…).step(okStep(…))` at two sites and does import `history`/`historyStep`, so both the "decide `Eff`'s fate" item and the "an Io-aware `historyStep` is expected" note describe real code.
- All six relative links resolve, and `node linkcheck` produces a byte-identical broken-link set on this head and on `origin/main` (146 lines both sides) — nothing stranded, nothing newly fixed.

Three things I would change.

**1. Stage 3's list of currently-infallible operations covers about half the set, and the omission is load-bearing.** The doc names seven — `Write`, `Read`, `Now`, `RandomInt`, `Sandbox`, `CreateServer`, `Listen` — and reads as exhaustive, since the following sentence scopes concrete work off it: "Convert those operations only together with a sweep of their value-discarding call sites." But `NodeOp` (`fjs/effects/node/types.ts:267`) has these infallible members too:

- `Test` — `(ctx, name, expectFailure, test) => void`
- `Await` — `(p: unknown) => readonly[unknown]`
- `All` — `<T>(...effects) => readonly T[]`
- `Forever` — `() => never`
- `MemOp` — `memCreate`, `memRead`, `memWrite` (`fjs/effects/memory/types.ts`)

`Test` is the one that makes this look like an oversight rather than a deliberate scoping: Stage 1 already cites "`test`'s body" as a function-carrying payload, so the plan knows `Test` is an operation, yet it is absent from the sweep list. `Forever` (`() => never`) and `All` in particular deserve an explicit ruling rather than silence — `Result<never, NotImplemented>` is a strange thing to write, and Stage 6's "allow a runner to omit an operation handler" applies to them as much as to `write`. Either extend the list to the full set or say plainly that it is illustrative and name the criterion for what is out of scope.

**2. `IoError` does not exist in this repo.** It appears twice — `IoEffect<ReadFile, Vec, NotImplemented | IoError>` and "e.g. `NotImplemented | IoError`" — sitting next to `ReadFile`, `Vec` and `NotImplemented`, which reads as though it were an existing type. `grep -rn IoError` over `fjs/` finds nothing; the only other occurrence in the repo is pseudocode in `todo/new-pl.md`. What exists today is `IoResult<T> = Result<T, unknown>` (`fjs/effects/node/types.ts:16`), i.e. the fallible operations carry `unknown`, which is exactly what the Stage 3 bullet "avoid bare `unknown` as the long-term error type" is asking to replace. The plan is coherent, but it should say `IoError` is a name to be introduced, not cite it as a given.

**3. `resultStep` is definitionally today's raw `step`, which weakens the `finallyStep` argument.** Expanding the alias, `resultStep: (e: IoEffect<O,T,E>, f: (r: Result<T,E>) => IoEffect<Q,R,F>) => IoEffect<O|Q,R,F>` is `(e: Effect<O, Result<T,E>>, f: (r: Result<T,E>) => Effect<Q, Result<R,F>>) => Effect<O|Q, Result<R,F>>` — raw `step` at those instantiations, with no branch behaviour of its own. The doc declines `finallyStep` because it "is derivable from `resultStep`"; `resultStep` is not merely derivable from `step`, it is `step`. That is defensible for Stage 5 onward, where raw `Effect` goes private under `RawEffect` and something has to be the public spelling — but through Stages 2–4, while the raw API is deliberately kept intact, it is a pure re-export. Worth one sentence saying so, otherwise it looks like an oversight against the same DRY standard `finallyStep` was held to.

None of these are blocking and none change the plan's shape. I did not run `tsc`, `prepack`, the test suite or the surface-diff battery: no code is touched at this head, so there is nothing for them to measure.

sergey-shandar:
All three review findings addressed in 0769ec1:

1. **Incomplete infallible-ops list** — resolved by the review's second option, per maintainer direction: the enumeration is gone and Stage 3 now states the criterion instead — every operation converts, the sweep is total, with no exemption for trivially-total shapes ("fallible first" is an ordering, not a scoping). Degenerate shapes get their explicit ruling on the same principle: `Result<never, NotImplemented>` simply has only its error branch inhabited.
2. **`IoError`** — now stated to be a name this migration introduces, normalizing today's bare `unknown` in `IoResult`, rather than cited as an existing type.
3. **`resultStep` is raw `step`** — acknowledged in the Target composition section, with its naming justified by the same canonical-vocabulary principle that declined `finallyStep`.

---

o2alexanderfedin (approved):
Approving. Verified at head `0769ec19`, merge-base `c51e6dfa` = current `origin/main`.
Docs-only (one file, `fjs/effects/todo/io-effect-migration.md`), so no CHANGELOG entry is
required under §8.3 and none was added — correct. Broken-link set is **byte-identical to
main**, and every path the `Related` section names resolves: `map-step-combinator.md`,
`effect-list-fold.md`, `fold-stream-combinator.md`, `../node/todo/ornotfound-combinator.md`,
`node-module-layering.md`, `../../../todo/044-error-handling-pattern.md`,
`fjs/effects/module.f.mjs`, `fjs/effects/eff/module.f.mjs`,
`fjs/types/result/module.f.mjs`. The symbols cited alongside them exist too: `step`
(`fjs/effects/module.f.mjs:141`), `okStep` (`:308`), `match` (`:368`) with the
missing-handler `assert`, and `okThen` in `fjs/types/result/module.f.mjs`.

All three findings from my review at `b70e4d66` are addressed. I re-derived the *new* claims
rather than assuming the fixes are right:

**1. The Stage 3 "currently-infallible operations" list.** Previously it named seven
(`Write`, `Read`, `Now`, `RandomInt`, `Sandbox`, `CreateServer`, `Listen`) while omitting
`Test`, `Await`, `All`, `Forever` and the three `MemOp`s — and that list scoped the concrete
work. The fix does not extend the list, it deletes it: Stage 3 now reads "Every operation
converts; the sweep is total… There is no list of operations worth converting and no
exemption for trivially-total ones." That is the stronger repair — there is no longer an
enumeration that can be incomplete, and it removes the earlier invitation to skip `now` /
`randomInt`, which sat awkwardly next to Stage 6 anyway. Cross-checked against Stage 6,
which does say "Allow a runner to omit an operation handler" with no per-operation
qualification, so the appeal to it holds. The new degenerate-shape line checks out against
the tree as well: `Result<never, NotImplemented>` is exactly `Forever`
(`fjs/effects/node/types.ts:165`, `['forever', () => never]`), one of the operations the old
list had missed.

**2. `IoError` cited as if it exists.** Now stated plainly: "`IoError` does not exist yet —
it names the normalized host-error type this migration introduces. Today the fallible
operations carry bare `unknown` (`IoResult`); replacing that is Stage 3 work." Re-derived
and correct on both halves — `IoResult<T> = Result<T, unknown>` at
`fjs/effects/node/types.ts:16`, and `IoError` still appears nowhere under `fjs/` (the only
occurrences in the tree are the two pseudocode lines in `todo/new-pl.md`). The disclaimer
sits directly under the first use, so the forward reference no longer reads as a citation.
The companion bullet dropped its `(Fs, Fetch, Exec, Import)` parenthetical for "operations
that already return `IoResult`", which is both correct and no longer a list that can drift.

**3. `resultStep` expanding to today's raw `step`.** This was the finding I most expected to
be argued away, and it is not — the text now concedes it outright: "Expanded through the
alias, `resultStep` is raw `step` at the Io instantiation — it adds no branch behavior of
its own." The replacement justification is that a derivable form earns a name by being
canonical vocabulary, which `resultStep` is and `finallyStep` has not shown itself to be.
That is a different argument from the one I objected to, and it is internally consistent:
Stage 5 independently commits to keeping the raw representation under a private name
("`RawEffect`") and to making `step` / `catchStep` / `resultStep` "the canonical composition
API", so `resultStep` really would become the only public spelling of that instantiation.
The `Invariants` list matches.

Nothing further from me. Draft status is not a blocker.

## Not verified

- Whether the staging is the right plan — I checked the document's claims against the tree,
  not the design.
- Stage 5's "double-wraps into `Result<IoResult<T>, NotImplemented>` and fails at its use
  sites"; that is a prediction about code that does not exist yet.

PR #1532, opened 2026-08-13
Refactor proof tests to use main entry point directly

sergey-shandar:
## Summary
Updated the proof test module to call the `main` entry point directly instead of manually composing `testAll` with `defaultReporter`. This simplifies the test setup and ensures the proof tests exercise the actual `fjs t` command entry point.

## Key Changes
- Removed `defaultReporter` from imports (no longer needed in this module)
- Added `main` to imports from the test module
- Updated `runMain` function to call `main(opts)` directly instead of `testAll(defaultReporter(opts))(opts)`
- Updated comments to reflect that the function now runs the actual `fjs t` entry point

## Implementation Details
The change consolidates the test runner invocation to use the same entry point that users interact with, improving test fidelity and reducing duplication of the reporter composition logic.

https://claude.ai/code/session_01W1sxqAFpURDxecPvCQrTHE


o2alexanderfedin (approved):
Approving. Verified at head `56a1105c` against baseline `origin/main` = `373c3325` ("LL(1): match with an explicit stack and a shared-input cursor (#1531)").

The change is exactly equivalent by construction: `main` in `fjs/emergent_testing/module.f.mjs` is

```js
export const main =
    options => testAll(defaultReporter(options))(options)
```

which is character-for-character what `runMain` was inlining, so routing the proof through `main` narrows the proof to the real `fjs t` entry point without changing what runs.

Because #1520 recently reworked the test entry point and the packaging consequences in this area are easy to get wrong from the diff alone, I verified from a packed tarball rather than by reading the change.

**Packaging (packed tarball, not the diff)**

- `npm pack` from a freshly cleaned tree exits 0 on both trees; 590 files in each; packed file lists are **identical** (`diff` clean).
- Extracted both tarballs and diffed recursively: the **only** differing file in the whole package is `fjs/emergent_testing/proof.f.mjs` itself. No `.d.mts` differs, so the emitted public surface is byte-identical to main — nothing to check on the type-surface axis.
- Installed the PR tarball into a scratch consumer (`npm install ./functionalscript-0.44.0.tgz`), added a `lib/module.f.mjs` + `lib/proof.f.mjs` pair and the README's `all.test.mjs` re-export (`import 'functionalscript/fjs/emergent_testing/all.test.mjs'`). `node --test all.test.mjs` discovers and registers the consumer proofs:

```
✔ import("./lib/proof.f.mjs").proof.addsNumbers() ...
✔ import("./lib/proof.f.mjs").proof.addsZero() ...
ℹ tests 2   ℹ pass 2   ℹ fail 0      (exit 0)
```

- Negative control on the same install — replaced one proof with `() => { throw 'negative control' }`:

```
✖ import("./lib/proof.f.mjs").proof.deliberatelyBroken() ...
ℹ tests 2   ℹ pass 1   ℹ fail 1      (exit 1)
```

  So the harness really is running the registered proofs, not silently finding nothing.

**In-repo runners**

- Bare `node --test` (no path argument) still discovers the entry by name: `tests 2476 / pass 2476 / fail 0`, exit 0. 2476 distinct `✔ import(...)` lines for 2476 tests, and the explicit-file run (`npm run cov`, which names `fjs/emergent_testing/all.test.mjs`) reports the same 2476 — no double registration.
- `npm test` (`fjs t`): **pass 2540, fail 0, total 2540**. Baseline `origin/main` at `373c3325`: **pass 2540, fail 0, total 2540**. Exact match, as a pure refactor should be. (The 2540 vs 2476 gap between `fjs t` and `node --test` is the same on main; it is not from this PR.)
- `npx tsc --noEmit`: exit 0. `npm run prepack` (via `npm pack`): exit 0.

**Coverage** — real numbers, `npm run cov` on Node v23 (`--experimental-test-coverage`, `--test-coverage-include=**/module.f.mjs`):

| tree | tests | line | branch | funcs |
|---|---|---|---|---|
| main `373c3325` | 2476 | 99.94 | 98.25 | 99.73 |
| PR `56a1105c` | 2476 | 99.94 | 98.25 | 99.78 |

Identical line/branch, function coverage marginally up (going through `main` covers `main` itself). No regression.

**Other gates** — proof-only change, so per AGENTS.md §8.3 no CHANGELOG entry is required, and none was added; correct. No docs touched, so no link check needed. No `nanvm-lib/` or generator inputs touched. No new exported types anywhere in the emitted surface (see the byte-identical `.d.mts` result above), so the unprefixed-`export type` axis is vacuous here.

One non-blocking nit, entirely optional: the JSDoc on `defaultReporter` still says it is "Exported as a factory so the GitHub format path can be exercised directly from tests". After this PR the proof reaches it through `main` rather than directly. The factory export is still what makes the options injection work, so the sentence is not wrong, just one step behind — worth a word if you touch that block again.

Draft status noted and not treated as a blocker.

PR #1534, opened 2026-08-13
Remove Eff wrapper; convert all modules to raw step combinators

sergey-shandar:
This change completes the migration away from the `Eff` fluent wrapper by converting the last four effect-dense modules (`fjs/cas`, `fjs/cas/evo`, `fjs/emergent_testing`, and `fjs/protocol/mcp/proof`) to use raw `step`, `mapStep`, and `historyStep` combinators directly.

## Summary

The `Eff` wrapper (`fjs/effects/eff/module.f.mjs`) provided a fluent API for chaining effects via `.step()` and `.value`. This change removes all remaining uses of `Eff` from the codebase, leaving it only as a documented experiment in the effects module. All effect composition now uses the flat combinator API.

## Key Changes

- **Removed `Eff` imports** from `fjs/cas/module.f.mjs`, `fjs/cas/evo/module.f.mjs`, `fjs/emergent_testing/module.f.mjs`, and `fjs/protocol/mcp/proof.f.mjs`
- **Converted fluent chains** of the form `eff(e).step(f).value` to raw `step(e, f)` calls
- **Introduced `mapStep`** for chains ending in pure projections: `eff(e).step(x => pure(f(x))).value` → `mapStep(e, f)`
- **Added explicit `historyStep`** calls where prior effect results are needed after subsequent operations (e.g., in `gcStage`, `runModuleMap`, and MCP state read-back), replacing implicit history tuples in the fluent form
- **Flattened deeply nested continuations** in `writeImpl` (`fjs/cas`) by introducing intermediate named bindings (`created`, `removed`, `stated`) that serve as documentation
- **Updated documentation** in `fjs/effects/eff/README.md` to reflect that no production code uses `Eff` anymore, and updated related todo items to remove references to fluent conversion hazards

## Implementation Details

- The conversion is mechanical: every `eff(e).step(f).value` has a direct raw equivalent with no redesign needed
- Chains that previously used implicit history tuples now use explicit `historyStep(history(e), f)` when the prior value is needed, making the intent clear at the call site
- Indentation improved in flat chains (no `eff(…)` / `.value` bracketing overhead), though deeply nested continuations remain equally nested
- Three modules required explicit history handling: `gcStage` (carries timestamp forward), `runModuleMap` (carries totals forward), and MCP proof (carries response forward for state read-back)

https://claude.ai/code/session_01X1p7SirShiavEkwB1YkTNx

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ history.
 
 ## Unreleased
 
+- `fjs/cas`, `fjs/cas/evo` and `fjs/emergent_testing` compose through the raw
+  `step` / `mapStep` / `historyStep` combinators instead of the fluent `Eff`
+  wrapper, which keeps its module but now has no consumers. Behavior is
+  unchanged
+  [#1534](https://github.com/functionalscript/functionalscript/pull/1534)
 - **BREAKING CHANGES:** `fjs/bnf/ll1`'s matcher runs as an explicit-stack
   machine over a cursor into the shared input, so long or deeply nested input
   no longer overflows the JS call stack and matching is no longer quadratic.
--- a/fjs/cas/evo/module.f.mjs
+++ b/fjs/cas/evo/module.f.mjs
@@ -53,8 +53,7 @@
  * @import { Hash, Subject, RevisionData, SubjectState, Cache, Evo } from './types.ts'
  */
 
-import { pure, foldStep } from '../../effects/module.f.mjs'
-import { eff } from '../../effects/eff/module.f.mjs'
+import { pure, foldStep, mapStep, step } from '../../effects/module.f.mjs'
 import { create, read, write } from '../../effects/memory/module.f.mjs'
 import { collectRead } from '../module.f.mjs'
 import { cBase32ToVec, vecToCBase32 } from '../../basen/cbase32/module.f.mjs'
@@ -185,9 +184,9 @@ export const decodeRevisionVec = value => {
  * @returns {(hash: Vec) => Effect<O, Revision | null>}
  */
 export const decodeRevisionBlob = cas => hash =>
-    eff(collectRead(cas.read(hash)))
-        .map(([tag, value]) => tag === 'error' ? null : decodeRevisionVec(value))
-        .value
+    mapStep(
+        collectRead(cas.read(hash)),
+        ([tag, value]) => tag === 'error' ? null : decodeRevisionVec(value))
 
 /**
  * Scans every hash in `cas` and builds a fresh {@link Cache} from the
@@ -202,11 +201,10 @@ export const buildCache = cas =>
         cas.list(),
         emptyCache,
         hash => cache =>
-            eff(decodeRevisionBlob(cas)(hash))
-                .step(revision =>
-                    pure(revision === null ? cache : addRevisionToCache(vecToCBase32(hash), revision)(cache))
-                )
-                .value)
+            mapStep(
+                decodeRevisionBlob(cas)(hash),
+                revision =>
+                    revision === null ? cache : addRevisionToCache(vecToCBase32(hash), revision)(cache)))
 
 /**
  * Scans `cas` once and allocates a memory slot holding the resulting {@link Cache}.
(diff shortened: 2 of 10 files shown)


o2alexanderfedin (approved):
Approving. I treated this as "prove equivalence", not "read the diff", since a wholesale conversion is exactly where a behaviour change hides in the noise.

**Baseline:** `origin/main` at `16142109` (*Exercise `main()` directly in `emergent_testing` proof tests (#1532)*). The branch's merge-base is `373c3325`, i.e. it is one commit behind main; nothing in `#1532` collides with the four converted files, and the test counts below confirm it.

### Gates

- `npm test` — **2540 pass / 0 fail**, against **2540 / 0** on `main`. Exact match, as a pure conversion should be.
- `node bin/linkcheck.mjs` — broken-link *sets* diffed against main: byte-identical (159 lines both sides), so the new `[README](../eff/README.md)` link in `io-effect-migration.md` resolves and nothing was stranded.

### Public surface — no breaking change, and no unprefixed type addition

`Eff` is not removed, only its consumers, so there was no export to break. I checked the *emitted* surface rather than the source anyway:

- `bin/extract.mjs` (type aliases) and `bin/consts.mjs` (const signatures) — identical on both sides, byte for byte.
- File-aware `export type` listing across all emitted `.d.mts`: 589 lines on each side, no diff. So no new unprefixed public `export type` (§6.2) and no signature widened to `any`.
- Stronger: **all 335 emitted declaration files are byte-identical between `main` and this head.** Negative-controlled — appending one comment line to `fjs/cas/module.f.d.mts` made the comparison report `DIFFERS`, so it is not a check that cannot fail.

Byte-identical declarations is also why no `**BREAKING CHANGES:**` prefix is owed under §8.4.

### Equivalence

First algebraically. `eff(e).step(f).value` desugars to `mapStep(historyStep(history(e), f), ([t]) => t)`; with `step(pure(v), g) = g(v)` that reduces to `step(e, v => step(f(v), pure))`, and `step(x, pure)` differs from `x` only by rebuilding continuation objects. So `eff(e).step(f).value ≡ step(e, f)` and `eff(e).map(f).value ≡ mapStep(e, f)` observationally, including the eager-`Pure` behaviour — both forms force a `Pure` head and call `f` at composition time. Since `eff(e)` starts with an empty history, every single-link `.step(x => …)` / `.map(x => …)` saw only the current value, so dropping to the raw one-argument combinators loses nothing. The three chains that *did* read a prior value are exactly the three that became explicit `historyStep`s (`gcStage`, `runModuleMap`, and `withState` in the MCP proof), and I checked each destructuring against the fluent form's positional arguments — `([[k, v], t])`, `([, ts])`, `([state, resp])` all bind what the old `(…tp)` lambdas bound.

Then empirically, which is what I actually trust. I built a differential harness that imports each converted module from *both* trees, drives them with the same scripted interpreter over the raw `Effect` structure (recording every `Do` node's `command` + `payload`, with a real in-memory store for `memCreate`/`memRead`/`memWrite`, recursion into `all`, and invocation of `test` callbacks), and compares the full operation trace *and* the final result:

```
cases=48 mismatches=0 nonTrivialTraces=48 totalTraceLines=295
```

Coverage includes `fileCas.write` / `read` / `list`, `collectRead` (including the mid-stream `error` item and the max-length guard), `casUpload`, `casAddFile`, `gcStage` via `writeImpl`, `buildCache`, `initEvo`, `addRevision` (success, bad-parent, and CAS-write-failure), `readRevision` (valid and invalid hash), `evo.list`/`evo.head`, `runModuleMap` (plain and `GITHUB_ACTIONS`), `registerModule`, `defaultTest` (both `throws` polarities), and the reporter's `result`/`summary` — each under a success script and an error script.

**Negative control, both ways:** with one-token mutations applied to the PR side — `ts.fail !== 0 ? 1 : 0` → `? 7 : 0` in `runModuleMap`, and `st[1].size === offset` → `!==` in `publish` — the harness reported 4 mismatches on exactly the affected cases, then 0 again after reverting. A comparison that cannot fail proves nothing, so I checked that this one can.

One incidental observation: an early harness run with a deliberately non-terminating `readdir` mock blew the stack at a *shallower* depth on `main` than here. That is the expected direction — the fluent form wraps each link in `historyStep` + `mapStep`, so the raw form uses strictly less stack per link. Not a behaviour change, just a small win.

### Conventions

- §8.3 CHANGELOG — code change, entry present, links only `/pull/1534`, no released section touched, length in line with this migration series. The MCP proof is not named in it, correctly: proof-only changes do not need one.
- §3.2 — no new exports; `withState` is a local helper inside a proof file.
- §2 / §4 — no renames, no new modules, no header churn.
- The docs match the code I read: `allvoid-combinator.md`'s three quoted `mapStep(all(…), () => undefined)` sites are exactly the three in `emergent_testing` (`registerModule` ×2, `registerModuleMap`), and `map-step-combinator.md`'s remaining unchecked boxes still have real targets — e.g. `step(now(), t => pure(join(…)))` in `writeImpl` and `step(…, sub => pure(mergeState(…)))` in `runModule`, both pre-existing.

The `eff/README.md` rewrite is honest about what the experiment did and did not settle, and `io-effect-migration.md`'s Stage 4 item is now discharged rather than deleted. Draft status is not a blocker.

PR #1537, opened 2026-08-13
Share EOF flush logic between UTF-8 and UTF-16 decoders

sergey-shandar:
## Summary

Extracts the end-of-input flush logic shared between UTF-8 and UTF-16 decoders into a reusable `eofFlush` factory in the `code_point` module. Both decoders implement the same contract — leftover state becomes exactly one error unit and the state resets to `null` — differing only in how they map state to error codes.

## Changes

- **Added `eofFlush` factory** (`fjs/text/code_point/module.f.mjs`): A generic factory that builds the end-of-input step for any decoder, taking a codec-specific `toError` function as its only parameter. Implements the shared contract: `state === null ? empty : [toError(state)], null`.

- **Refactored UTF-8 EOF handler** (`fjs/text/utf8/module.f.mjs`): Replaced hand-written `utf8EofToCodePointOp` with `eofFlush(utf8StateToError)`, eliminating code duplication and making the contract explicit.

- **Refactored UTF-16 EOF handler** (`fjs/text/utf16/module.f.mjs`):
  - Extracted `utf16StateToError` helper function (`state => state | errorMask`)
  - Replaced hand-written `utf16EofToCodePointOp` with `eofFlush(utf16StateToError)`
  - Simplified JSDoc to reference the shared `eofFlush` contract

- **Added proofs** (`fjs/text/code_point/proof.f.mjs`): Test cases verifying `eofFlush` behavior for both empty state (no output) and non-empty state (exactly one error unit).

- **Updated documentation** (`fjs/text/code_point/README.md`): Documented the streaming decoder skeleton, explaining how `decoder` and `eofFlush` together provide the shared streaming framework that both UTF-8 and UTF-16 codecs build upon.

- **Updated issue tracking** (`fjs/text/utf16/todo/word-classifier-dedup.md`): Clarified that `eofFlush` is the precedent for this kind of shared extraction, and updated references to reflect that the end-of-input extraction is now complete.

- **Removed obsolete issue** (`fjs/text/todo/codec-eof-flush.md`): Deleted the tracking issue for this work, as it is now implemented.

## Implementation Details

The `eofFlush` factory follows the same pattern as the existing `decoder` factory — a higher-order function that captures codec-specific behavior (the `toError` mapping) while implementing the universal contract. This keeps the "exactly one error unit" rule stated once, preventing re-derivation across codecs and making the streaming skeleton more maintainable.


o2alexanderfedin (approved):
Approving. Baseline `origin/main` = `67fb2dac992681f00f4fa32788ff51a6bec763aa`; tested at `ef685e68`.

## The DRY claim: both call sites really were equivalent

The risk with a shared-factory extraction is that the two call sites were not
actually the same, and the shared version silently changes one of them. Here they
were identical, and the reason is `empty === null`
(`fjs/types/list/module.f.mjs:310`), which is exactly what the deleted todo
claimed. `utf8EofToCodePointOp` wrote `state === null ? null : […]` and
`utf16EofToCodePointOp` wrote `state === null ? empty : […]`; those are the same
expression.

I did not take that on the source reading alone. I swept both decoders'
observable behaviour, old vs new, over their input domains — **494,809 probes,
byte-identical output on both trees**:

| probe | count |
| --- | --- |
| `utf8EofToCodePointOp` called directly on every `Utf8State` reachable by feeding 0–3 arbitrary bytes | 288,248 |
| `toCodePointList` over all 1- and 2-byte inputs | 65,792 |
| `toCodePointList` over all 3- and 4-byte inputs from a 16-value boundary alphabet (lead/continuation/`0xC0`/`0xC1`/`0xE0`/`0xED`/`0xF0`/`0xF4`/`0xF5`/`0xFF`) | 69,632 |
| utf16 `toCodePointList` over all 65,536 words, plus every lone surrogate alone and after a BMP char | 69,632 |
| utf16 2- and 3-word sequences over a surrogate-boundary alphabet incl. out-of-range `-1`/`0x10000` | 1,452 |
| empty input; each valid sequence at EOF; every valid×valid and valid×truncated concatenation | 51 |

The EOF cases you would worry about are in there and are non-trivial — e.g.
`[0xE0,0xA4]` flushes to `-2147482588`, `[0xF0,0x90]` to `-2147483120` — so the
comparison is over real flushed error units, not over a domain where the flush
never fires.

**Negative controls** (mutate the PR's `eofFlush`, re-run the same sweep):

| mutation | detected? |
| --- | --- |
| return `state` instead of `null` (no reset) | yes — 288,247 differing lines |
| `[empty, null]` (never flush) | yes — 338,814 |
| flush on the empty state too | yes — 155,994 |
| `utf16StateToError = state => state` (drop `errorMask`) | yes — 3,392 |

So the sweep can fail; it just does not.

## Rest of the battery

- `npx tsc --noEmit` → 0. `npm run prepack` from a clean tree → 0, both trees.
- `npm test`: **2550 pass / 0 fail** vs **2548** on main. +2 = the two new
  `eofFlush` proof cases exactly.
- Public surface (dual-axis, after `prepack`): type aliases **byte-identical**
  (46,199 bytes, non-empty). Consts 889 → 890, the only additions being
  `eofFlush: <S, Cp>(toError: (state: S) => Cp) => (state: S | null) => readonly [List<Cp>, S | null]`
  and the new `proof` key. No signature widened to `any`; `utf8EofToCodePointOp`'s
  emitted signature is unchanged despite `@param`/`@returns` becoming `@type`.
  Emitted `.d.mts` for `code_point`/`utf8`/`utf16` each keep their `@module`
  header and contain no `elided` or `: any`.
- `linkcheck`: 159 broken links on both trees, **identical sets** — deleting
  `fjs/text/todo/codec-eof-flush.md` stranded nothing, and the back-reference in
  `word-classifier-dedup.md` was correctly repointed.
- CHANGELOG (§8.3): entry present, links only `/pull/1537`, and it characterises
  the change correctly — it claims a restructuring ("derive their eof ops from
  it instead of each writing the flush out"), not a behaviour change, which is
  what the sweep shows. Nothing is removed or renamed, so no `**BREAKING
  CHANGES:**` prefix is owed.

`eofFlush` is a new *const* export, so §6.2's `_` prefix (a typedef convention)
does not apply, and it is fully typed rather than public-by-accident.

## The new todo, `unreachable-pending-state-fallthrough.md`

Both of its factual claims check out independently:

- The exhaustiveness argument holds. Over all 65,536 words,
  `isBmpCodePoint` matches 63,488, `isHighSurrogate` 1,024, `isLowSurrogate`
  1,024 — summing to exactly 65,536 with **0** words unclassified, so the
  fall-through at `:196` is genuinely dead. (Negative control: excluding a single
  value from the BMP predicate immediately produces 1 unclassified word.)
- `npm run cov` reproduces the quoted figures exactly:
  `fjs/text/utf16/module.f.mjs | 99.71 | 96.55 | 100.00 | 196`. `code_point` is
  at 100/100 with the new proof cases.

No findings.

PR #1539, opened 2026-08-13
Add RTTI serializable data form with canonical operations

sergey-shandar:
## Summary

Implements the serializable data form for RTTI schemas, completing the two-form architecture where thunk form handles construction and data form handles all schema algebra. This adds `toData` conversion, canonical ordering (`cmp`), structural equality (`equal`), subset testing, and coverage collapse.

## Key Changes

- **New module `fjs/types/rtti/data/module.f.mjs`** (931 lines)
  - `toData(schema)` converts thunk-form `Type` to function-free, serializable `Data`
  - Canonical ordering via `cmp` with SameValue semantics for numbers
  - Structural equality via `equal` that works on canonical forms
  - `subset` test for inclusion checking with coinductive cycle handling
  - Coverage collapse that drops subsumed array/object patterns
  - Kind-wise union operations over six disjoint value kinds (unit, number, string, bigint, array, object)
  - Handles recursive schemas with rule naming and deferred merge fixpoint computation

- **New type definitions `fjs/types/rtti/data/types.ts`** (91 lines)
  - `KindSet<T>` — represents a set of values of one kind
  - `ArraySet` — tuple-with-rest pattern for arrays and tuples
  - `ObjectSet` — struct-with-rest pattern for objects and records
  - `UnionSet` — disjoint union over the six kinds
  - `Node` — either a rule reference or inline union
  - `Data` — the complete serializable form: `[RuleSet, Node]`

- **Comprehensive test suite `fjs/types/rtti/data/proof.f.mjs`** (487 lines)
  - Unit bit encoding tests
  - Primitive and container conversions
  - Recursive schema handling (lists, trees, mutual recursion)
  - Or-cycle normalization (self-cycles, mutual cycles)
  - Coverage collapse verification
  - Serialization round-trip validation
  - Data-driven validator tests

- **Design documentation `fjs/types/rtti/data/README.md`** (122 lines)
  - Explains two-form architecture and rationale
  - Documents set-theoretic representation
  - Describes kind-wise operations and pattern semantics
  - Covers incomplete subset cases and design trade-offs

- **Updated documentation and todos**
  - Moved completed `serializable-data.md` todo to implemented state
  - Added new todo `tuple-extra-elements.md` for thunk-direct validator asymmetry
  - Updated cross-references in related modules

- **Canonical form**: All schema algebra (union, subset, equality) reduces to kind-wise set operations on a canonical representation
- **Lazy conversion**: `toData` runs once when needed; schemas built but never consumed pay nothing
- **Cycle handling**: Recursive thunks are named on first reference; deferred merges resolve via fixpoint iteration
- **Pattern subsumption**: Arrays and tuples share one kind; objects and records share one kind, enabling pattern inclusion tests
- **SameValue semantics**: Number comparison uses `Object.is` to distinguish `-0` from `0` and allow `NaN`
- **Coinductive subset**: Reference cycles are assumed included while being checked, following standard coinductive treatment

https://claude.ai/code/session_01T8BR3aPUJDe5zZMDJKme7j


o2alexanderfedin (commented):
Reviewed at `fa9aba4a` (draft) against `origin/main` `e7e1881a`. Reviewed as a normal PR.

This is a lot of well-organised work and the README is the best part of it — the set-theoretic
framing, the "two forms, one job each" split, and the honest section on `subset`'s deliberate
incompleteness all make the design easy to check. I checked the algebraic claims by property
testing rather than by reading, and they hold up. Two things to fix, one of them the recurring
`_`-prefix defect.

## 1. §6.2 — 17 unprefixed exported types in the new proof

`fjs/types/rtti/data/proof.f.d.mts` publishes these into the package's public type surface:

```
List  Tree  Forest  SelfOr  OrA  OrB  Outer  Inner  T2
X  Y  W  TopOr  TopArr  B2  A2  Rec
```

They are `@typedef`s in `proof.f.mjs`, and JSDoc `@typedef` has no non-exported form — which is
precisely the situation §6.2's `_` prefix exists for. The convention is followed without exception
today: across the whole tree on `main`, `proof.f.d.mts` files export **24 types, 0 of them
unprefixed** (`_TestOp`, `_Benchmark`, `_NodeList`, `_Entries`, `_Tests`, `_MemoryState`, `_Event`,
`_AddOp`, …). This PR takes that to 17 unprefixed. `List` and `Tree` are the uncomfortable ones —
`List` is a name the package already means something specific by, in `fjs/types/list`.

Worth noting the contrast within this same PR: `module.f.d.mts` gets it exactly right — `_Assumed`,
`_Ctx`, `_Key`, `_State`, `_Thunk`, all prefixed. And the seven in `types.d.ts` (`KindSet`,
`ArraySet`, `ObjectSet`, `UnionSet`, `Node`, `RuleSet`, `Data`) are correctly *un*prefixed: they are
the module's intended public type API, documented as such. So it is only the proof file.

## 2. A canonicalization gap the README does not list

The README claims schema identity is a property of the data form, with two stated exceptions (rule
names are structural; a reference is never recognised as `unknown`/`never`). There is a third, and
it does not need a union to trigger — a single-member `or` is enough:

```js
const list = () => ['array', list]

toData(list)         // [{ list: { array: [{ prefix: [], rest: 'list' }] } }, 'list']
toData(or(list))     // [{ list: { array: [{ prefix: [], rest: 'list' }] } }, { array: [{ prefix: [], rest: 'list' }] }]

equal(toData(list))(toData(or(list)))   // false
```

Same rule set, same denoted set, different entry node — one keeps the named reference, the other
inlines the body. The cause looks like the "an entry rule nothing else references is inlined" step:
for `list` the rule *is* referenced (by itself, via `rest`), so the entry stays `'list'`; for
`or(list)` the entry is a fresh inline union, so the body is inlined while the self-reference
survives inside it. `toData(or(list, list))` and `toData(or(tree, tree))` fail the same way, so
**`or` is not idempotent on recursive schemas**: over my 44-schema corpus, `equal(toData(or(t, t)))(toData(t))`
held for 42 and failed for `list` and `tree`.

Either canonicalize the entry (always reference a self-referencing rule, or always inline), or add
this to the README's list of accepted limits. It is not covered by either one currently there.

## The algebra, property-tested

Corpus of 44 schemas (primitives, literals including `NaN` and `-0`, arrays, records, tuples,
structs, unions, options, and five recursive shapes: `list`, `tree`/`forest` mutual, `selfOr`,
`or(list, number)`) crossed with 32 values.

| property | checks | failures |
|---|---:|---:|
| `equal(d)(d)`, `cmp(d)(d) === 0`, `subset(d)(d)` | 44 each | 0 |
| `toData` deterministic across calls (rule-name counter is not leaking state) | 44 | 0 |
| `cmp` antisymmetric: `sgn(cmp(a)(b)) === -sgn(cmp(b)(a))` | 1,936 | 0 |
| `equal(a)(b) ⟺ cmp(a)(b) === 0` | 1,936 | 0 |
| mutual `subset` ⟹ `equal` | 1,936 | 0 |
| `equal` ⟹ mutual `subset` | 1,936 | 0 |
| `or` commutative: `equal(toData(or(a,b)))(toData(or(b,a)))` | 1,936 | 0 |
| `or` idempotent: `equal(toData(or(t,t)))(toData(t))` | 44 | **2** (§2 above) |
| **`subset` soundness**: `subset(a)(b)` ⟹ every corpus value accepted by `a` is accepted by `b` | 8,512 | 0 |

`subset` soundness is the one that would have hurt if it were wrong, and it is clean over every
pair the corpus admits.

## Data-form `validate` vs the thunk-direct validator

Cross-checked all 44 schemas against all 32 values, 1,408 comparisons: **42 divergences, all
explained, none unaccounted for.**

- **16** are the tuple-length divergence this PR documents in `todo/tuple-extra-elements.md`:
  `[number]`, `[]`, `[1]` and `or([number], array(number))` accept extra elements in the thunk
  validator and reject them here. Matches the README's "Tuple length" section exactly.
- **26** are `selfOr` (`X = number | X`), where the *thunk* validator throws — it recurses through
  the pure `or` cycle for any value that is not a number. The data form dissolves the cycle to the
  least fixpoint (`toData(selfOr)` is `[{}, { number: true }]`) and answers correctly. That is the
  data form being strictly better than the existing one, and it is worth a proof case of its own,
  since it is a genuine capability the README's "pure `or` cycles dissolve" bullet only asserts.

## Declaration emit — the #1497 hazard is clear

Checked both trees after `npm run prepack`:

- **`elided`: 0 on `main`, 0 here.** The `#1526` state holds.
- **Type-level `any`: 0 on `main`, 0 here.** Total `any` token count is 50 → 51, and the diff of
  the matches is a single line of JSDoc prose (`types.d.ts:23`, "the array may be **any** length
  `>= prefix.length`"), not a type. The mutually recursive positions emit named:
  `never: UnionSet`, `unknown: UnionSet`, `cmp: (a: Data) => (b: Data) => number`,
  `toData: (t: Type) => Data`, `validate: (data: Data) => (value: Unknown) => ResultE`.
- **§4 headers survive**: `@module` present in both `data/module.f.d.mts` and `data/types.d.ts` —
  the latter is the dangerous one, since every import there is `import type`.

## Minor

`module.f.mjs`'s header and the README both call the form "serializable" without naming a
serializer, and plain `JSON.stringify` does not round-trip it: `bigint` literal sets throw
(`toData(1n)` is `[{}, { bigint: [1n] }]`), and `NaN` and `-0` number literals survive
`JSON.parse(JSON.stringify(...))` as `null` and `0`, so `equal` says false. 3 of 44 corpus schemas
affected. DJS is presumably the intended target rather than JSON; one clause saying so would stop
a reader reaching for `JSON.stringify` and getting a `TypeError`.

## Otherwise clean

- `npx tsc --noEmit` exit 0; `npm run prepack` from a cleaned tree exit 0 (both passes).
- `npm test` 2577 pass / 0 fail vs 2551 on `origin/main`.
- `bin/linkcheck.mjs`: broken links **137 on `main` → 132 here**, and the set diff is five
  *removals* — `todo/172.md`, `todo/662.md`, `todo/668-rtti-function-types.md` and the retired
  `serializable-data.md`. Nothing stranded by deleting `serializable-data.md`.
- §8.3: CHANGELOG entry present under Unreleased, links only `/pull/1539`, no released section
  touched. New module, no breaking change, so no `**BREAKING CHANGES:**` prefix.
- Public surface additions are all in the new module; nothing elsewhere widened.

sergey-shandar:
All review items addressed in bb8ce49:

**§1 — proof typedefs**: all 17 are now `_`-prefixed (`_List`, `_Tree`, …), so `proof.f.d.mts` publishes no unprefixed types; verified against declaration emit.

**§2 — canonicalization gap**: fixed rather than documented. `toData` gains a final *interning* step: a union structurally equal to a named rule's body reads back as a reference to that rule (alphabetically first on a tie). So `toData(or(list))`, `toData(or(list, list))` and `toData(or(tree, tree))` are now structurally identical to `toData(list)`/`toData(tree)`, and the step also collapses re-stated fixpoints — `array(list)` normalizes to `list` itself, and the `outer`/`inner` mutual `or`-cycle now normalizes both entries to the same single-rule form. The collapse and intern passes share one generic bottom-up node rewriter. New proof cases cover idempotence at the entry and in nested positions; the README's canonical-form list documents the rule.

**selfOr capability**: added a proof case — the data-driven validator accepts/rejects against `X = number | X` where the thunk-direct walk cannot terminate.

**Minor (serializer)**: the module header and a new README "Serialization" section now name DJS as the target (plain `JSON.stringify` only when no `bigint` literals are involved) and call out that `NaN`/`-0` literal members don't round-trip textually with either serializer today.

Checks: `npx tsc` clean, 2578/2578 tests pass, `fjs/types/rtti/data/module.f.mjs` stays at 100% line/branch/function coverage.

---

o2alexanderfedin (commented):
Re-reviewed at `a8ce6538` (previous review was at `fa9aba4a`). Baseline: `origin/main` = `87c58701`. Draft status is not a blocker.

Both previously open findings are **resolved**. Re-running the property battery against the rewritten `or` turned up one new problem that I should have caught last round — see §3.

## 1. §6.2 unprefixed exports — fixed

Re-ran the surface diff (`npm run prepack` in both trees, then `bin/extract.mjs`). Output is non-empty on both sides: 528 exported type aliases on `main`, 554 on the PR, 26 added.

`fjs/types/rtti/data/proof.f.d.mts` now exports 17 types, **all `_`-prefixed**:

```
_List _Tree _Forest _SelfOr _OrA _OrB _Outer _Inner _T2 _X _Y _W _TopOr _TopArr _B2 _A2 _Rec
```

The only unprefixed additions in the whole PR are `KindSet`, `ArraySet`, `ObjectSet`, `UnionSet`, all declared in `fjs/types/rtti/data/types.ts` and referenced 54 times from `data/module.f.mjs`. Those are the module's public vocabulary in a real `types.ts`, which is exactly what §6.2 permits. No objection.

## 2. `or` idempotence on recursive schemas — fixed, and the README claim holds

The README now asserts it outright ("a union structurally equal to a named rule's body reads back as a reference to that rule … so `or` is idempotent on recursive schemas"). I re-derived it rather than taking the sentence at face value, over a 51-schema corpus (primitives, literals, tuples, records, arrays, options, unions, and every recursive schema in `proof.f.mjs`: `list`, `tree`/`forest`, `selfOr`, `orA`/`orB`, `outer`/`inner`/`t2`, `x`/`y`/`w`, `topOr`/`topArr`, `a2`/`b2`, `rec`):

```
equal(toData(s))(toData(or(s)))    and  equal(toData(s))(toData(or(s, s)))
102 checks, 0 failures
negative control: equal(toData(number))(toData(string)) === false
```

The two cases that failed at `fa9aba4a` now pass; at that head `toData(or(list))` inlined the body instead of reading back as `"list"`, and it now returns the reference. The three specific README claims each check out: `or(list)`, `or(list, list)` and `array(list)` are all `equal` to `list`.

The new `## Serialization` section also closes the minor note from last round about `JSON.stringify` not round-tripping — the `bigint` / `NaN` / `-0` corners are now stated accurately.

## 3. New finding: `or` is unsound across two differently-named recursive rules

The union of two recursive schemas whose rules carry different names loses **all** array values from both operands, rather than unioning them.

```js
const list = () => ['array', list]   // type list = readonly list[]
const b2   = () => ['array', b2]     // the same type, different rule name

toData(list)         // [{ list: { array: [ { prefix: [], rest: 'list' } ] } }, 'list']
toData(or(list, b2)) // [{}, { array: [] }]          <-- no patterns at all
```

`{ array: [] }` admits no array, so the union rejects values both operands accept:

```
validate(toData(list))([])            -> ok
validate(toData(or(list, b2)))([])    -> error { path: [], message: 'no match' }
```

It is not confined to alpha-equivalent pairs, and it is not confined to array-only schemas — the array kind is emptied while the other kinds survive:

```js
toData(or(tree, y))  // [{}, { number: true, array: [] }]  — accepts 1, rejects []
```

Measured over the same corpus, "each operand is a subset of the union" — the defining property of a union — fails for **26 of 2601 pairs**, every one of them a pair of distinct recursive rules:

```
list|b2, list|rec, tree|y, tree|w, forest|y, forest|w, y|tree, y|forest, y|arr-tree,
w|tree, w|forest, w|arr-tree, b2|list, b2|rec, b2|arr-list, b2|or-list-num, rec|list,
rec|b2, rec|arr-list, rec|or-list-num, arr-list|b2, arr-list|rec, or-list-num|b2,
or-list-num|rec, arr-tree|y, arr-tree|w
```

**This is pre-existing, not a regression from this round.** I checked out `fa9aba4a` and reproduced identical output (`or(list, b2)` → `[{}, { array: [] }]`, `or(tree, y)` → `[{}, { number: true, array: [] }]`). I missed it last round because my 44-schema corpus did not contain two recursive schemas with different rule names, so no pair could exercise the merge path. My earlier "or commutativity, 1,936 checks, 0 failures" was correct but too weak a property — the result is symmetrically wrong in both orders, so commutativity does not see it.

The pattern-merge path appears to bail when two array patterns end in `rest` references to different rules, producing an empty pattern list instead of either keeping both alternatives or coinductively identifying the rules. The coinductive machinery already exists in `subset` — `subset(toData(list))(toData(b2))` is `true` in both directions — so the fix is likely to reuse it here.

Related and worth deciding explicitly: `equal(toData(list))(toData(b2))` is `false` while both `subset` directions are `true`, so **mutual-subset does not imply `equal`** (16 of 2601 pairs). Rule names are part of the form by design — the README says as much when it describes collision counters — so this may be intended, but the README's "schema identity is a property of this form" reads as though it were up to renaming. Worth one sentence either way.

## Battery

- `npm test` — 2580 pass / 0 fail, vs 2553 on `origin/main`.
- CHANGELOG entry present, links only `/pull/1539`.
- Re-ran on the new implementation over the 51-schema corpus, 2601 pairs each: `cmp` antisymmetry 0 failures, `equal` ⟺ `cmp === 0` 0 failures, `or` commutativity 0 failures, `validate` well-formedness 0 failures.

## Settled, not re-litigated

Declaration emit is clean (0 `elided`, 0 type-level `any` on either side; the `any` delta was JSDoc prose).

sergey-shandar:
§3 of the re-review is fixed in 18c9ad7, plus the requested sentence on mutual-subset vs `equal`.

**The bug**: `dropSubsumed` assumed no two *distinct* canonical patterns denote the same set, so on mutual subsumption it dropped both members — and references to α-equivalent rules under different names are exactly the counterexample (`{rest: 'list'}` and `{rest: 'b2'}` subsume each other coinductively). The collapse emptied the array kind.

**The fix**: a mutually-subsumed pattern is now dropped only in favor of an *earlier* one, keeping exactly the first member of each mutual group (strict subsumption still drops regardless of position). The kept spelling then interns to its rule, so the union normalizes fully:

```js
toData(or(list, b2))  // [{ b2: { array: [{ prefix: [], rest: 'b2' }] } }, 'b2']
equal(toData(or(list, b2)))(toData(b2))   // true
validate(toData(or(list, b2)))([])         // ['ok', []]
subset(toData(list))(toData(or(list, b2))) // true — both directions, both operands
```

The mixed-kind case normalizes too: `toData(or(tree, y))` is now `equal` to `toData(tree)`. New proof cases cover the α-pair union in both operand orders, operand-⊆-union, validation, and `or(tree, y)`.

**Mutual-subset vs `equal`**: intended, and now stated in the README's limits — `subset` resolves references coinductively while `equal` is structural identity, so an α-equivalent pair is a mutual `subset` without being `equal` (mutual inclusion implies equality of the sets, not of the spellings), and their union normalizes to whichever spelling sorts first.

Checks: `npx tsc` clean, 2581/2581 tests pass, `data/module.f.mjs` still 100% line/branch/function coverage (the new keep-first branches are exercised by the α-pair and strict-subsumption cases).

---

o2alexanderfedin (approved):
Re-reviewed at `18c9ad77` (previous review was at `a8ce6538`). Baseline: `origin/main` = `87c58701`. **The open `or` unsoundness is genuinely fixed**, and the two minor points are now documented. Approving — draft status is not a blocker.

### `or` unsoundness across differently-named recursive rules — fixed

Re-derived over the **same widened 51-schema corpus** as last round, not a narrower one — corpus size is exactly why this was missed the first time. Same probe script, same 2601 pairs:

| property | `a8ce6538` (prev head) | `18c9ad77` (this head) |
|---|---|---|
| operand ⊆ union (`subset(a, or(a,b))`) | **26 / 2601 fail** | **0 / 2601** |
| `or` commutativity | 0 | 0 |
| `cmp` antisymmetry | 0 | 0 |
| `equal` ⟺ `cmp === 0` | 0 | 0 |
| `validate` well-formedness | 0 | 0 |
| mutual `subset` ⟹ `equal` | 16 | 16 (documented, see below) |

I ran the probe at the previous head as a negative control in the same session: it still reports the 26 failures there (`list <: or b2`, `tree <: or w`, `forest <: or y`, …) and 0 here. Same file, same corpus — so this is a real behavioural change, not a probe that stopped looking.

The fix is the right shape rather than a special case. `dropSubsumed` filtered `!list.some((q, j) => j !== i && le(p, q))`, which for a mutually-subsuming pair drops **both** members — the array kind went empty and `toData(or(list, b2))` became `[{}, { array: [] }]`, rejecting `[]`. It now reads:

```js
list.filter((p, i) => !list.some((q, j) => j !== i && le(p, q) && (j < i || !le(q, p))))
```

A mutually-subsumed pattern is dropped only in favour of an *earlier* one, so exactly the first member of each group survives. Because the list is canonically sorted before this runs, the survivor is order-independent — which is why `or` commutativity stays at 0/2601 despite the filter now being index-sensitive. I also re-ran the idempotence probe from last round: **0 failures over the 51-schema corpus**, negative control (`equal(number, string)`) held false, and the three README identities (`array(list) === list`, `or(list) === list`, `or(list, list) === list`) still hold.

### The 16 mutual-subset-without-`equal` pairs

Not fixed — and correctly not fixed. The README now states the condition accurately: `subset` resolves references coinductively while `equal` is structural, rule names are part of the form, so an α-equivalent pair is a mutual subset without being `equal`, and "mutual inclusion implies equality of the *sets*, not of the spellings". That is precisely what the 16 pairs are (`list`/`b2`, `list`/`rec`, `tree`/`w`, `forest`/`y`, …), and it says the union normalizes to whichever spelling sorts first, which matches the fix above. Documented rather than claimed away.

### Minor: `JSON.stringify` round-trip

The point I raised once is now in the README's Serialization section, and it is accurate on all three counts: DJS covers the whole form including `bigint` literal sets, plain `JSON.stringify` only works with no `bigint` literals, and `NaN` → `null` / `-0` losing its sign is called out as a corner shared by both. Nothing left here.

### Battery

- `npx tsc --noEmit` — exit 0. `npm run prepack` from a cleaned tree — exit 0.
- `npm test` — **2581 / 0**, against **2553 / 0** on `main`. +28, matching the added proof cases.
- Broken-link **sets** vs main: PR **132**, main **137**. Nothing only-in-PR; five entries only-in-main, all repairs (the `todo.md` references from `todo/172.md`, `662.md`, `668-rtti-function-types.md`, and the `data/module.f.ts` link in the now-deleted `todo/serializable-data.md`). No stranded reference from the deletion.
- Public surface, both axes. Ten new exported consts under `/fjs/types/rtti/data`, none widening to `any`. Seven new unprefixed public `export type`s in `data/types.ts` — `KindSet`, `ArraySet`, `ObjectSet`, `UnionSet`, `Node`, `RuleSet`, `Data`. I checked these against §6.2 rather than flagging them mechanically: every one is transitively reachable from `toData: (t: Type) => Data` via `Data = readonly [RuleSet, Node]` → `RuleSet = StringMap<UnionSet>` → `ArraySet`/`ObjectSet`/`KindSet`, and `never`/`unknown` are typed `UnionSet` directly. They are genuine API, not implementation-only types leaking. Everything implementation-only stayed `_`-prefixed (`_Ctx`, `_Assumed`, `_Tree`, `_Forest`, `_A2`, `_B2`, …), so last round's §6.2 fix holds.
- §4 header: `types.ts` is all-`import type` and so is the dangerous case — `@module` survives into the emitted `types.d.ts` (1 occurrence), and into `module.f.d.mts`. No `elided` or collapsed `any` in the emitted declarations.
- §8.3 CHANGELOG: entry present under Unreleased, links only `/pull/1539`, no released section touched.

PR #1541, opened 2026-08-14
Remove unreachable overflow check in base64 decode

sergey-shandar:
## Summary

- `fjs/basen/base64/module.f.mjs` was at 95.65% branch coverage: `decode`'s padded branch checked `length(head) + realBits > maxLength` before combining `head` and the trimmed last chunk, but this condition can never be true.
- `head` is built purely from whole 6-bit chunks (no trimming), so its length is always a multiple of 6. The largest such length `stringToVec` can return without itself overflowing is `maxLength - (maxLength % 6)` = `maxLength - 4`. Adding `realBits` (at most 4, when `padChars` is 1) reaches `maxLength` exactly and never exceeds it — the check was dead code.
- Per `AGENTS.md` §3.2, the branch is removed rather than covered with a contrived test (which isn't possible here anyway, since the condition is mathematically unreachable).
- Along the way, corrected `decodePaddedTailRejections`'s comment in `proof.f.mjs`, which claimed its oversized-input case exercised this check — it was always caught earlier, by `head`'s own `stringToVec` overflow guard.
- `fjs/basen/base64/module.f.mjs` now reaches 100% line/branch/function coverage.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `node --test --experimental-test-coverage --test-coverage-include='fjs/basen/base64/module.f.mjs' fjs/emergent_testing/all.test.mjs` → 100.00% line/branch/func
- [x] `node ./fjs/module.mjs t` → 2581 pass, 0 fail

---


o2alexanderfedin (approved):
Approving. I re-derived the unreachability claim rather than taking it on trust, and it holds — including on `main`, which is the part that mattered here.

Baseline: `origin/main` = `699c960e` ("types/object: add structurallySame ... (#1538)").

**Is the removed branch actually unreachable?**

I instrumented the branch **in the `origin/main` tree** (i.e. the version that still has the check) with a hit counter and drove `decode` at the public entry point:

- content-exhaustive over every string of length 0..3 from a 66-symbol set (full alphabet + `=` + one invalid char), plus every length-4 string over an 8-symbol subset;
- length-exhaustive over every body length 1..3000 x padChars {0,1,2}, with the last body character varied over all 64 alphabet characters for L <= 200;
- a dense band at the real limit, every body length in [174000..174800] x 4 last-characters x padChars {0,1,2}.

Result: **369,227 driven sequences, 0 hits.**

Because the predicate depends only on `length(head)` and `realBits`, I also enumerated its *entire* reachable state space rather than sampling it: `head` is the output of `stringToVec` over whole 6-bit chunks, so `length(head)` in {0, 6, ..., 6*floor(maxLength/6)} = {0, ..., 1048572}, and `realBits` in {2, 4} (padChars is 1 or 2 in this branch; 0 returns early and >2 is rejected). That is **349,526 combinations, of which 0 satisfy `length(head) + realBits > maxLength`** — the maximum is 1048572 + 4 = 1048576 = `maxLength` exactly. So the comment's arithmetic is right: `maxLength % 6 === 4`, and 4 is exactly `realBits`' maximum.

Negative control: retightening the threshold to `maxLength - 8n` makes the same harness report **4 driven hits and 3 predicate states**, so the harness can see a hit when there is one.

**Was it also dead on `main`?** Yes — and I checked this explicitly, because the last PR of this shape (#1540) had a *separate* unmentioned tightening elsewhere in the same diff that was what actually killed the branch, with the CHANGELOG documenting only the no-op deletion. Here `git diff origin/main -- fjs/basen/base64/module.f.mjs` is the removal and its import, nothing else, and the numbers above were produced on the `origin/main` tree. Nothing else in this PR narrows the input reaching the branch.

The proof-comment correction is also correct and is itself evidence of the same point: `decode('A'.repeat(174_764) + '=')` gives `head` a 174,763-character body, so `head`'s own `stringToVec` sees 1,048,578 > 1,048,576 and returns `null` — the removed check never got a look at that input even before this PR.

**Behaviour neutrality.** Swept old vs new `decode` outputs over the same domain (minus the widest band): **360,707 inputs, 0 divergences**, with 10,567 non-`null` old results, so the sweep is not vacuously all-`null`. Negative control: a one-token mutation on the PR side (`padChars > 2` -> `padChars > 1`) makes the same comparison report **1,608 divergences**. No divergence to classify — this is a pure deletion of dead code, neither a regression nor a latent bug being fixed.

**Battery**

- `npm test` — 2617 pass / 0 fail, identical to `origin/main` (2617 / 0).
- `node bin/linkcheck.mjs` — broken-link *sets* identical to main (not just the counts).

**Conventions.** §8.3: this is a code change, so it needs an entry, and it has one linking only `/pull/1541`. §8.4: no `**BREAKING CHANGES:**` prefix, which is right on precedent — #1524 was representation-only with zero value changes and correctly carried none, while #1520's specifier-level break did. Here the emitted behaviour is provably identical over the whole reachable domain, so there is nothing to prefix. No new exported types, so §6.2's `_`-prefix rule is not engaged.

PR #1545, opened 2026-08-14
Migration close-out: drop the .js ignore, browser-native browser-testing plan, @import sweep

sergey-shandar:
Three close-out items from `todo/migrate-typescript-to-mjs.md`, per maintainer direction. (A fourth — the broken `serializable-data.md` link — turned out to be already resolved: #1539 implemented `fjs/types/rtti/data/` and deleted that todo.)

## 1. `.gitignore` drops the blanket `**/*.js` rule

No repository command generates `.js` since #1520 (`prepack` is declaration emit plus a no-emit check), so the rule only hid stale artifacts in pre-#1520 working trees. The `**/*.d.ts` / `**/*.d.mts` ignores stay — declarations are still generated. `**/*.js` deliberately remains in `package.json` `files`, so publishes keep coming from clean checkouts as CI does.

## 2. `browser-testing.md` rewritten browser-native (maintainer's choice: rewrite, not retire)

The 442-line plan was built on transpiling `.f.ts` to browser-loadable `.f.js`. Authored source is now `.f.mjs` — JavaScript that browsers load natively — so the rewrite deletes the transpile step, the generated-`.f.js` module graph, the mixed-graph selection machinery, and the TypeScript-rejection server rules, while keeping everything still true: one shared browser test application behind three runners (HTML page, `fjs browser-test` without Playwright, external `playwright/test` adapter), static `proof`-export selection with clear rejection of `node:`/external graphs, the serializable report protocol, and the validation fixture list. 442 lines → ~160.

## 3. The `@import` sweep — the remaining half of the module-header convention item

Every scattered module-level `/** @import ... */` comment moved into its file's leading JSDoc block per §4: **125 `.mjs` files**. Multi-line `@import` blocks joined into one tag each; blank lines between same-group imports collapsed. The ordering half needed no changes — measured **zero** files with an external/built-in import after a relative one.

## Verification

- `npx tsc` and the `prepack` round-trip (`emitDeclarationOnly && tsc`) clean.
- All **129** emitted `module.f.d.mts`/`module.d.mts` carry `@module`, and every prose header repo-wide leads its emitted declaration — the #1526 properties hold through the sweep.
- `npm test` **2646 / 0**.
- Residual scattered-`@import` sweep: **0** files.

A follow-up commit on this branch checks off the corresponding items in `todo/migrate-typescript-to-mjs.md`, re-measures the `.f.ts` survivor inventory (the browser-testing rewrite removes its 10 mentions), and adds the CHANGELOG entry with this PR's number.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GkqgqTaffDpYFQEJydpqHF)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -128,7 +128,6 @@ _*
 .DS_Store
 **/*.d.mts
 **/*.d.ts
-**/*.js
 
 test-results/
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ history.
 
 ## Unreleased
 
+- Every module-level `@import` tag lives in its module's leading JSDoc block
+  (125 files swept), so emitted declaration headers list their type imports
+  in one place
+  [#1545](https://github.com/functionalscript/functionalscript/pull/1545)
+
 - `types/uint8array`: `toVec` attempts the conversion instead of precomputing a
   byte-count bound; behavior and error message unchanged
   [#1543](https://github.com/functionalscript/functionalscript/pull/1543)
--- a/fjs/asn.1/proof.f.mjs
+++ b/fjs/asn.1/proof.f.mjs
@@ -1,6 +1,9 @@
-/** @import { Vec } from '../types/bit_vec/types.ts' */
-import { empty, isVec, length, msb, uint, unpack, vec, vec8 } from '../types/bit_vec/module.f.mjs'
+/**
+ * @import { Vec } from '../types/bit_vec/types.ts'
+ * @import { SupportedRecord, ObjectIdentifier } from './types.ts'
+ */
 
+import { empty, isVec, length, msb, uint, unpack, vec, vec8 } from '../types/bit_vec/module.f.mjs'
 import { asBase } from '../types/nominal/module.f.mjs'
 
 import {
@@ -19,7 +22,6 @@ import {
     encodeObjectIdentifier,
     decodeObjectIdentifier,
 } from './module.f.mjs'
-/** @import { SupportedRecord, ObjectIdentifier } from './types.ts' */
 
 import { assert, assertEq } from '../asserts/module.f.mjs'
 
--- a/fjs/basen/base128/proof.f.mjs
+++ b/fjs/basen/base128/proof.f.mjs
@@ -1,4 +1,6 @@
-/** @import { Vec } from '../../types/bit_vec/types.ts' */
+/**
+ * @import { Vec } from '../../types/bit_vec/types.ts'
+ */
 
 import { empty, vec, vec8 } from '../../types/bit_vec/module.f.mjs'
 import { asBase } from '../../types/nominal/module.f.mjs'
--- a/fjs/basen/base64/proof.f.mjs
+++ b/fjs/basen/base64/proof.f.mjs
@@ -1,4 +1,6 @@
-/** @import { Vec } from '../../types/bit_vec/types.ts' */
+/**
+ * @import { Vec } from '../../types/bit_vec/types.ts'
+ */
 
 import { assertEq } from '../../asserts/module.f.mjs'
 import { empty, vec, repeat, vec8, maxLength } from '../../types/bit_vec/module.f.mjs'
--- a/fjs/basen/cbase32/proof.f.mjs
+++ b/fjs/basen/cbase32/proof.f.mjs
@@ -1,4 +1,6 @@
-/** @import { Vec } from '../../types/bit_vec/types.ts' */
+/**
+ * @import { Vec } from '../../types/bit_vec/types.ts'
+ */
 
 import { empty, vec } from '../../types/bit_vec/module.f.mjs'
 import { cBase32ToVec, cBase32ToVec5x, vec5xToCBase32, vecToCBase32 } from './module.f.mjs'
--- a/fjs/bnf/data/proof.f.mjs
+++ b/fjs/bnf/data/proof.f.mjs
@@ -9,7 +9,6 @@ import { sort } from '../../types/object/module.f.mjs'
 import { oneEncode, option, range, rangeDecode, repeat0Plus, set } from '../module.f.mjs'
 import { emptyTagMap, toData } from './module.f.mjs'
 import { assertEq } from '../../asserts/module.f.mjs'
-
 import { stringify } from '../../media/json/module.f.mjs'
 import { classic, deterministic } from '../testlib.f.mjs'
 
--- a/fjs/cas/cli/proof.f.mjs
+++ b/fjs/cas/cli/proof.f.mjs
@@ -1,4 +1,6 @@
-/** @import { NodeProgramOptions } from '../../effects/node/types.ts' */
+/**
+ * @import { NodeProgramOptions } from '../../effects/node/types.ts'
+ */
 
 import { commands } from './module.f.mjs'
 import { computeSync, sha256 } from '../../crypto/sha2/module.f.mjs'
--- a/fjs/cas/evo/proof.f.mjs
+++ b/fjs/cas/evo/proof.f.mjs
@@ -22,7 +22,6 @@ import { unwrap } from '../../types/nullable/module.f.mjs'
 import { ok, error } from '../../types/result/module.f.mjs'
 import { nonEmpty, empty as elEmpty } from '../../effects/list/module.f.mjs'
 import { tryUtf8 } from '../../text/module.f.mjs'
-
 import { dialect as revisionDialect } from '../../media/revision/module.f.mjs'
 import {
     buildCache, decodeRevisionBlob, initEvo, evo, emptyCache, syncRevision,
--- a/fjs/ci/nix/proof.f.mjs
+++ b/fjs/ci/nix/proof.f.mjs
@@ -2,9 +2,9 @@
  * Proofs for generated CI flakes.
  *
  * @module
+ *
+ * @import { NixJob } from './types.ts'
  */
-
-/** @import { NixJob } from './types.ts' */
 
 import { assert, assertEq } from '../../asserts/module.f.mjs'
 import { step } from '../../effects/module.f.mjs'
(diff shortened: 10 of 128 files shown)


o2alexanderfedin (approved):
Approving. Baseline for every comparison is `origin/main` at `6684bef2`. I checked the three
changes separately, since they are only loosely related.

## 1. Dropping `**/*.js` from `.gitignore`

This is the one with real silent-breakage potential — a blanket ignore going away can start
tracking generated output — so I went looking for it specifically.

- Nothing was hidden: `git ls-files '*.js'` is 0 and `find . -name '*.js' -not -path
  './node_modules/*'` is 0 on a freshly cleaned PR-head tree. Nothing previously ignored becomes
  trackable, because nothing exists to become trackable.
- `npm run prepack` from a clean tree — exit 0 — emits **343 files, all `.d.mts` (255) or
  `.d.ts` (88), zero `.js`**. Both remaining extensions are still covered by the `**/*.d.mts` /
  `**/*.d.ts` rules that this PR leaves in place.
- **CI drift gate run for real**: after `prepack`, `git add -A && git diff --cached
  --exit-code` → exit 0, 0 staged files. Also clean after `npm test`.
- `package.json` `files` still lists `**/*.js`, matching what the migration doc claims is
  deliberate.

So the ignore rule was genuinely dead and the gate stays green.

## 2. The `@import` sweep

Headline numbers, measured on both trees (note: `grep` here is ripgrep-backed and skips
gitignored paths, so all declaration-file measurements below were taken through
`find … -print0 | xargs -0 grep` to avoid silently reading nothing):

| ruler | `origin/main` | PR head |
| --- | --- | --- |
| `.mjs`/`.ts` lines with `/**` and `@import` on one line | 287 | **0** |
| files with any `@import` in a non-leading JSDoc block (multi-line aware) | 88 | **0** |
| total `@import` tags across `.mjs`/`.ts` | 565 | **565** |

The standalone form goes to zero on both a single-line and a block-structure-aware ruler, and
the total tag count is *unchanged* — which is the check that matters, because a tag silently
dropped rather than moved is the failure mode that widens a type to `any` without `tsc`
complaining.

Confirming that nothing stopped resolving, with `prepack` run in both trees:

- Emitted declarations: **343 files on both sides**, 229 carrying `@module` on both sides,
  **0 occurrences of `elided`** on both sides.
- `any` as a whole word in emitted declarations: **51 on main, 51 on the PR head**, and the
  *sets* match file-for-file — the only differences are line numbers shifting by 1–5 as headers
  grew. (Counting `any` unfiltered is worthless here, so this is `grep -w` over the emitted
  `.d.mts`/`.d.ts` only, then a `diff` of the sorted occurrence lists.)
- Dual-axis public surface (`bin/extract.mjs` / `bin/consts.mjs`): **byte-identical** on both
  axes. No new exported type aliases at all, so §6.2's `_` prefix question does not arise; no
  signature widened.
- 122 emitted declaration files differ, and every changed line is either an `@import` tag moving
  into the header comment or a *statement reordering* — in exactly three files
  (`emergent_testing/module.f.d.mts`, `text/utf16/module.f.d.mts`,
  `types/bit_vec/module.f.d.mts`) an `import type` / `import` / `export type` line moves
  position because the interleaved comment blocks that were pinning emit order are gone. No
  content change.
- §4 blank line after the header, checked structurally across all 255 `.mjs` files: main has 66
  files with no leading `/**` block and 13 missing the blank line; the PR head has 41 and 4, and
  the remaining 4 are a strict subset of main's 13. Improved, not regressed.

`npx tsc --noEmit`/`prepack` exit 0; `npm test` **2646/2646 pass** against **2646/2646** on
`origin/main` (both from cleaned trees, so no `.f.js`/`.f.mjs` double-discovery).
`bin/linkcheck.mjs`: broken-link *sets* identical to main (132 broken relative links, 0
label/target extension mismatches, `diff` empty) — the 122-file sweep strands nothing.

## 3. The browser-testing plan

I checked its factual claims against the tree rather than reading it approvingly:

- "authored source is `.f.mjs`" — `find -name '*.f.ts'` is **0**; the only non-declaration `.ts`
  left is 88 `types.ts` files. Claim holds.
- All four outbound links resolve on disk (`f-mjs-test-and-coverage.md`,
  `todo/migrate-typescript-to-mjs.md`, `fjs/ci/todo/f-mjs-package-support.md`,
  `todo/plan/roadmap.md`), and linkcheck confirms no new broken target. The rewrite drops the
  dead `fjs/fsc/README.md#source-files-…` anchor rather than leaving it.
- The removal of the transpile design is consistent with the tree: there is nothing left to
  transpile, so the `.f.ts → .f.js` half of the old plan was describing work that cannot exist.

In `todo/migrate-typescript-to-mjs.md` I re-derived the two load-bearing numbers rather than
taking them:

- "all **129** emitted module declarations keep their header" — there are exactly 129
  `module.f.mjs`/`module.mjs` files and all 129 carry `@module`. ✅
- "All **20** files that still contain the old extension anywhere" — markdown files containing
  `.f.ts`, whole tree, `CHANGELOG.md` excluded (the doc's own scoping): **20**, and the
  enumeration in the paragraph lists exactly those 20. The previous "21" was stale, so this PR
  fixes it. ✅ (There are also 11 non-markdown files containing the literal — `shouldLoad` and
  friends — but the doc explicitly accounts for those as quotations, so that is scoping, not a
  miss.)

## The one thing I would change: the "125 files" count

Both the CHANGELOG entry ("125 files swept") and the migration doc ("125 `.mjs` files") give a
number I cannot reproduce under any ruler I tried:

| ruler | count |
| --- | --- |
| `.mjs` files in the diff whose `@import` lines change | **124** |
| `.mjs` files in the diff (all of them touch `@import`) | **124** |
| files in the diff with a *removed* `@import` line | 123 |
| files on `main` with a single-line standalone `/** @import … */` | 122 |
| total files in the diff | 128 |

124 is the defensible one — every `.mjs` file this PR touches is a sweep target, and there are
124 of them. (The two that had no single-line standalone form,
`fjs/bnf/data/proof.f.mjs` and `fjs/cas/evo/proof.f.mjs`, had multi-line blocks, so they are
genuinely swept.) 125 appears to be off by one.

Worth fixing precisely because the surrounding paragraph is the one that says *"Re-measure with
the same resolve-against-the-tree method, at the final commit, before claiming a number — prose
that enumerates survivors can itself add mentions, which is how a stale count got published the
first time."* Same class of slip, and the same paragraph is simultaneously correcting the last
one (21 → 20). Everything the number describes is correct; only the number is wrong.

Not blocking — the count is prose, the sweep itself is provably lossless (565 → 565 tags,
identical surface, identical `any` set), and the CHANGELOG follows §8.3 otherwise: one entry for
the code-affecting change, PR link only, released sections untouched, and no entry for the
`todo/` rewrites, which is correct.

Minor formatting note while you are in there: the new entry is followed by a blank line, unlike
every other entry in `## Unreleased`. That makes the whole list loose in Markdown, so every item
renders wrapped in `<p>`, not just this one.

o2alexanderfedin (approved):
Re-review at `5867ec21`, baseline `origin/main` = `725129fe`. Previous review was at `a09addb3`.

**The delta is a merge of `main` and nothing else.** `git diff a09addb3 5867ec21` touches three
files — `AGENTS.md`, `CHANGELOG.md`, `fjs/djs/serializer/module.f.mjs` — and all three arrive
from `main` (#1544, already merged). `fjs/djs/serializer/module.f.mjs` at this head is identical
to `origin/main`'s copy, so the merge did not have to reconcile anything with the sweep. The PR's
own diff against `main` is unchanged: 128 files, 124 `.mjs` + `.gitignore`, `CHANGELOG.md`,
`todo/migrate-typescript-to-mjs.md`, `fjs/emergent_testing/todo/browser-testing.md`.

Because the tree content moved, I re-ran the checks rather than carrying the previous results
forward:

- `npx tsc --noEmit` → 0. `npm run prepack` from a freshly cleaned tree → 0.
- **CI drift gate for real:** after `prepack`, `git add -A && git diff --cached --exit-code`
  → exit 0, 0 staged, 0 untracked. 343 emitted declarations on disk, all `.d.mts`/`.d.ts`,
  0 `.js` on disk, 0 `.js` tracked. Nothing previously ignored became trackable.
- `npm test` → **2646 pass / 0 fail**, exactly matching `fs-main` on `725129fe` (2646/0).
- **`any` / `elided` after the merge:** 51 `any`-bearing lines in the emitted declarations on
  both sides, and the sets are identical file-for-file *and* text-for-text (diff empty on
  `path:text`). 0 occurrences of `elided`. The `@import` consolidation still resolves everything
  it did before.
- **Dual-axis surface diff vs `main`:** the type-alias axis is byte-identical; the const axis is
  904 = 904 with zero set differences and zero signature changes.
- **Sweep invariant after the merge:** 0 standalone `/** @import … */` blocks under a single-line
  ruler, and 0 non-leading JSDoc blocks containing `@import` under a multi-line-aware ruler — the
  files merged in from `main` did not reintroduce one.
- **Lossless:** every changed line across the 124 `.mjs` files is inside a JSDoc block. The 57
  lines that a naive "starts with `*`" filter flags are the continuation lines of the multi-line
  `@import` blocks being removed (`_TokenizerState,`, `} from './types.ts' */`, …), not code.
- **Broken-link sets** are identical to `main`: 132 on both, `diff` empty. Not just equal counts —
  the same set.

### The two open notes from `a09addb3`

**2. CHANGELOG trailing blank line — resolved.** The `Unreleased` list is tight again: the only
blank line between `## Unreleased` and the next heading is the one directly under the heading.
The merge placed #1544's entry immediately after the #1545 entry, which consumed the stray line.

**1. "125 files swept" — still does not reproduce.** Re-derived at this head, not carried over:

| ruler | count |
|---|---|
| all changed files | 128 |
| changed `.mjs` files | **124** |
| changed files of any extension whose diff touches an `@import` line | 124 |
| changed `.mjs` where an `@import` line was actually removed (i.e. moved) | 122 |

The four non-`.mjs` changed files are `.gitignore`, `CHANGELOG.md`,
`todo/migrate-typescript-to-mjs.md` and `fjs/emergent_testing/todo/browser-testing.md`, so there
is no "code files only" reading that reaches 125 either. The claim now appears in two places —
`CHANGELOG.md` (`125 files swept`) and `todo/migrate-typescript-to-mjs.md` line 848
(`125 `.mjs` files`) — and 124 fits both phrasings. Still a documentation nit, still non-blocking;
I am flagging it only because that same paragraph asks for the number to be re-measured before it
is claimed.

Everything else from the previous round is unaffected by this delta and stands. Approving.

PR #1554, opened 2026-08-14
fsm: express the two state-free scans as map projections

sergey-shandar:
Implements `fjs/fsm/todo/scan-as-map.md` (deleted here).

## Why

Two of fsm's stream transforms were written as self-referential `Scan` operators although they carry no state — each ignored the scan state and returned itself:

```js
const stringifyOp = ([sortedSet, max]) => [[stringifyIdentity(sortedSet), max], stringifyOp]
const scanStringify = scan(stringifyOp)

const fetchOp = ([item, _]) => [item, fetchOp]
const scanFetch = scan(fetchOp)
```

`scan` exists for *stateful* streaming transforms. An operator whose only "state" is a reference to itself is an element-wise projection — which is what `map` says directly — and writing it as a scan makes the reader hunt for evolving state that isn't there.

## What

Both become `map`:

```js
/** @type {(entry: Entry<SortedSet<string>>) => Entry<string>} */
const stringifyEntry = ([sortedSet, max]) => [stringifyIdentity(sortedSet), max]
const stringifyEntries = map(stringifyEntry)

/** @type {(entry: Entry<SortedSet<string>>) => SortedSet<string>} */
const entryValue = ([value]) => value
const entryValues = map(entryValue)
```

Six lines become four, `scan` and the `Scan` type drop out of the imports, and both projections gain a one-line JSDoc claim.

`Entry<T>` is the **mutable** tuple `[T, number]` (`fjs/types/range_map/types.ts`) and `_Dfa` stores `RangeMapArray<string> = readonly Entry<string>[]`, so the projection is pinned with an explicit `Entry<string>` return annotation rather than `as const` — a readonly tuple would not be assignable where `toArray(stringifyEntries(setMap))` lands in `newDfa`.

## One deviation from the TODO

The TODO's snippet kept the bindings named `scanStringify` / `scanFetch`. That would leave half the confusion in place — they are not scans any more — so they are renamed to `stringifyEntries` / `entryValues`, which also reads better at the call sites:

```js
const stringMap = toArray(stringifyEntries(setMap))
const newStates = entryValues(setMap)
```

Both are module-private, and no reference to the old names remains anywhere in the tree.

## Verification

- `npx tsc` clean.
- `fjs test`: 2676 pass, 0 fail — `fjs/fsm/proof.f.mjs` unchanged, since this is a pure refactor. Both consumers take a `List`, which `map` produces, so laziness and behavior are unchanged.
- `npm run cov`: `fjs/fsm/module.f.mjs` at 100% lines/branches/functions.

---


o2alexanderfedin (approved):
Approving. Baseline `origin/main` = `4fbf0b71` (merge-base of this head). Since this is a DRY refactor, the whole question is whether the two forms are actually equivalent, so I checked it structurally and then swept it.

**Structural argument.** `scan` and `map` share the same driver:

```
apply    = f => input => () => { const n = next(input); return n === null ? null : f(n) }
mapStep  = f  => ({first, tail}) => ({ first: f(first),  tail: map(f)(tail) })
scanStep = op => ne         => { const [first, newOp] = op(ne.first)
                                 return { first, tail: scan(newOp)(ne.tail) } }
```

When `op` is self-referential — `newOp === op` for every element, which both `stringifyOp` and `fetchOp` were — `scan(op)` reduces literally to `map(x => op(x)[0])`: same `apply` wrapper, so identical thunk shape, identical laziness, identical empty-input behaviour (`next → null` before the step function is ever called), identical per-element evaluation point. And the projections match what the old ops returned in their first slot: `stringifyOp` gave `[stringifyIdentity(sortedSet), max]`, which is `stringifyEntry`; `fetchOp` gave `item`, which is `entryValue`. So this is substitution, not re-derivation.

**Empirical sweep, 40,000 probes.** I loaded `dfa` from both trees in one process and compared canonicalised output over identically-seeded random grammars — rule counts 0–8, `''`-weighted state names so the initial state actually fires, and byte sets built from `one`, `range` and `union` including `empty`:

```
probes=40000  nontrivial=23593  divergences=0
```

`nontrivial` counts the probes that produced a real multi-state DFA rather than a bare initial state, so this is not 40,000 empty runs. Explicit boundaries, checked by hand and identical on both sides:

```
empty grammar    -> [["[\"\"]",[]]]
single rule      -> [["[\"\"]",[["[]",64],["[\"x\"]",65]]],["[\"x\"]",[]],["[]",[]]]
no initial match -> [["[\"\"]",[]]]
```

The empty-grammar case is the one that matters for the state-free property: `setMap` is the empty list there, so both `toArray(stringifyEntries(setMap))` and `entryValues(setMap)` have to bottom out through `apply`'s `next(input) === null` path, which they do.

**Negative controls on the sweep,** so the zero above means something. Perturbing each projection separately and re-running the same 40,000 probes:

```
stringifyEntry: max -> max + 1        divergences=23625
entryValue: ([value]) => value[0..0]  divergences= 2230
```

Both projections are observable through `dfa`, and the comparison detects a change in either. No divergence to classify — nothing here is a latent-bug fix, unlike #1533; this is the #1537 shape, where the two forms are the same expression.

**No surface change at all.** `npm run prepack` exit 0 in both trees, then a checksum of all 343 emitted `.d.mts`/`.d.ts` files: exactly one differs, `fjs/fsm/module.f.d.mts`, and its only difference is the `@import { Fold, Scan }` → `@import { Fold }` comment line. No exported type added, so §6.2 does not come up. `scan`/`Scan` are fully gone from `fjs/fsm/module.f.mjs` — no orphaned import left behind, `foldScan` at :107 is the only remaining scan-family use.

**Other gates:** `npm test` 2676, identical to `origin/main`'s 2676 — exactly what a behaviour-preserving refactor should give. `bin/linkcheck.mjs` broken-link *sets* identical to main, and nothing anywhere referenced the deleted `fjs/fsm/todo/scan-as-map.md`, so removing the completed todo strands no link. `changelog/unreleased/1554.md` links only `/pull/1554` and no `todo/` file; strictly the emitted surface is provably identical so §8.3 would not have demanded an entry, but adding one marked "internal only" is the conservative side of that line and reads fine.

PR #1563, opened 2026-08-14
types/bit_vec: derive front/removeFront from unpackSplit

sergey-shandar:
Implements `fjs/types/bit_vec/todo/front-from-unpack-split.md` (deleted here).

## Why

`bo` already derives `unpackPopFront` generically from the injected `unpackSplit`, and `front` / `removeFront` are exactly its two halves — yet `_Base` demanded both from every bit order, and `lsb` and `msb` each re-derived them by hand.

`startsWith` showed the cost from the other side: it needed "the first `n` bits" and reached for `popFront(n)(v)[0]` because `front` wasn't conveniently in scope, so the same extraction existed **four** ways in one module.

## What

- `front` / `removeFront` are defined once in `bo`, as the two projections of `unpackPopFront`. `_Base` shrinks to the genuinely order-specific parts: `norm`, `uintCmp`, `unpackSplit`, `unpackConcatUint`.
- `startsWith` uses the derived `front`, bound **once per prefix** rather than rebuilding `popFront(n)` per vector — `n` comes from `prefix`, a scope up.
- Two call-invariant partial applications hoisted to module scope per §6.3: `map(unpack)` and the u8 `Unpacked` constructor, neither of which depends on a bit order or on the list being mapped.

`removeFront` hands on the unmasked rest because `pack` re-masks, as the TODO noted.

## Public API

Unchanged. `BitOrder` in `types.ts` still carries `front` / `removeFront`, and `lsb` / `msb` still expose them. Diffing emitted `.d.mts` before and after, the **only** change is the two members leaving `_Base` — an underscore-prefixed private typedef, which §6.2 says is not by itself a breaking change:

```diff
 export type _Base = {
-    readonly front: (len: bigint) => (v: Vec) => bigint;
-    readonly removeFront: (len: bigint) => (v: Vec) => Vec;
     readonly norm: _NormOp;
```

## Verification

- `npx tsc` clean.
- `fjs test`: 2708 pass, 0 fail.
- `npm run cov`: `fjs/types/bit_vec/module.f.mjs` at 100% lines/branches/functions.
- **Differential check against the old bodies.** Since this is bit-level code where a wrong shift passes most tests, I re-ran the previous hand-written `lsb`/`msb` implementations side by side with the derived ones over every length 0–32, 40 values each, and every split point `0 ≤ k ≤ L`, for both orders: **89 760 comparisons, 0 mismatches.**

---


o2alexanderfedin (approved):
Approving. Reviewed at `be26cd8`, baseline `origin/main` = `c3af487f` ("Add task to split AGENTS.md into scoped documents (#1560)").

## Gates

- `npx tsc --noEmit` → exit 0; `npm run prepack` from a clean tree → exit 0.
- `npm test` → `pass: 2708, fail: 0` — exactly matching `origin/main` (`2708`), as a behaviour-preserving derivation should.
- `node bin/linkcheck.mjs` — broken-link **sets** compared line-by-line against main: identical. Deleting `fjs/types/bit_vec/todo/front-from-unpack-split.md` strands nothing (`grep -rn front-from-unpack-split` over the tree: no remaining references).
- Emitted declarations read directly (not via `bin/extract.mjs`). `fjs/types/bit_vec/module.f.d.mts` differs from main by exactly the two dropped `_Base` members and nothing else; `@module` header survives (1 occurrence); no `elided` / `: any`.

## Equivalence sweep — `front` / `removeFront`

I transcribed main's four hand-written bodies verbatim and ran them against the derived ones over the full domain, for both orders: widths `0,1,2,3,7,8,9,15,16,17,31,32,33,63,64,65,127,128,129`, values `{0, 1, mask, mask>>1, top bit, 0xA5…, 0x5A…}`, and `len ∈ {0, 1, w-1, w, w+1, w+2, 7, 8, 9}` — so empty vectors, single bits, and both sides of every byte/word boundary, including `len > length` (where `vec`'s `len <= 0n` guard and the negative-shift behaviour of `>>` both matter). Plus the `maxLength` boundary itself: `w = maxLength (0x100000)` with `len ∈ {0, 1, 8, maxLength-1, maxLength}`.

**4272 comparisons, 0 divergences.** Negative control (perturbing old `msb.front` by `+1n`) produced **1068** divergences, so the harness does discriminate.

That matches the algebra: `lsb.unpackSplit` gives `[uint, uint >> len]`, so `unpackPopFront(len)(unpack(v))[0] = uint & mask(len)` — main's `front` — and `pack({length - len, uint >> len}) = vec(length - len)(uint >> len)` — main's `removeFront`. `msb.unpackSplit` gives `[uint >> (length - len), uint]`, reproducing both msb bodies the same way. The comment's note that `pack` re-masks so the unmasked rest is fine is correct: `pack = vec(length)(uint)` and `vec` masks.

## `startsWith`

Also changed (`popFront(n)(v)[0]` → the derived `front(n)`, bound once per `prefix`). Swept old vs new over 13 prefix widths × 5 values × the same 13×5 vector domain, both orders: **7442 comparisons, 0 divergences**; negative control gave **1387**. The `length(v) < n` guard is untouched, and hoisting `front(n)` out of the `v` closure is safe because `n` comes from `prefix` and `mask(n)` is total for `n >= 0`.

## Laziness / short-circuiting (the #1548 shape)

This PR hoists `map(unpack)` and the u8 `Unpacked` constructor to module scope, which sits right on the path that #1548 broke, so I drove all four with a genuinely **unbounded** self-referential list rather than assuming:

| | this head | `origin/main` |
|---|---|---|
| `lsb`/`msb` `.tryListToVec` | `null` in ~65 ms | `null` in ~65 ms |
| `lsb`/`msb` `.listToVec` | throws `assertion failed` in ~60 ms | same |
| `tryU8ListToVec` | `null` in ~41 ms | same |
| `u8ListToVec` | throws `assertion failed` in ~41 ms | same |

All eight terminate, identically to main — the `foldAbsorbing` short-circuit is intact. Hoisting a partial application of `map` does not force the list, as expected.

## Conventions

- `changelog/unreleased/1563.md` is a per-PR file with no link or PR number inside — correct under the current §8.3.
- `_Base` narrowing is emitted (`export type _Base`), but the `_` prefix marks it implementation-only per §6.2 and `bo` is not exported, so nothing outside the module can construct one. `BitOrder` — the actual public type — still carries `front`/`removeFront` unchanged, so "public types are unchanged" in the entry holds.
- The deleted todo's three tasks are all discharged: `_Base` narrowed and both projections derived in `bo`; `startsWith` routed through the derived `front` and bound once per `prefix`; both call-invariant partial applications hoisted to module scope (§6.3).
- No new proof cases, but nothing new is exported and the existing `bit_vec` proofs still exercise `front`/`removeFront`/`startsWith` through both orders — a derivation whose old and new forms I could compare directly is the weaker case for demanding new proofs, and the 11 714-comparison sweep above is stronger evidence than a case would be.

Clean derivation, no divergence found anywhere in the swept domain.

PR #1561, opened 2026-08-14
AGENTS.md: adopt the commit-message standard

sergey-shandar:
`main` takes one squash commit per PR, titled `PR-title (#NNN)` with the PR
description as its body, so the title and description are the commit message.
A changelog generated from Git history could read nothing else, and the option
expires for every PR that merges before the format is fixed.

New §8.5 documents it: the `topic: short description` title within 72
characters including the ` (#NNN)` GitHub appends, the mandatory `Changelog:`
section — the last section of the description before an optional trailer
block, holding the same list items as the PR's entry file in
`changelog/unreleased/`, or `Changelog: none` — and squash-merge-only landing.
§2 step 6 and CONTRIBUTING.md point at the new section.

§8.5 sits after §8.4 rather than in sequence: `#83-changelog` is linked from
AGENTS.md §2, `changelog/README.md`, and several module `todo/` files, so
renumbering would rot all of them for no gain.

The proposal's release-tag step is **not** adopted: the repository has no Git
tags and is not going to get any. §8.4 records that instead of prescribing
tags — release membership is already in `changelog/X.Y.Z/`, one file per PR
that shipped in it, so a tag would be a second copy of that fact maintained by
hand. The todo keeps the tag section as a rejected design with the reasoning,
and `changelog-from-git-history.md`'s design 1, which took "between release
tags" as its range query, now reads the release commit's bare-version title.

`todo/commit-message-standard.md` stays: the repository settings it asks for
(squash-only, default squash message "Pull request title and description",
branch protection) need admin rights and cannot land in a PR, and until they
are set §8.5 is a convention rather than a rule. Its format section is now
reasoning behind AGENTS.md rather than the normative text.
`todo/commit-message-enforcement.md` was waiting on this adoption and no
longer is.

This PR's own title and description follow the standard it adds. Note that
GitHub strips angle-bracket placeholders from a PR body even inside backticks,
so the prose above spells them out instead — worth knowing when writing a
`Changelog:` section that mentions a type or a generic.

Changelog: none

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01WQFBgRCxeiy5gw1is28t2H

PR #1565, opened 2026-08-14
Implement recursive lock maps for scoped dependency resolution

sergey-shandar:
## Summary

This PR completes the implementation of recursive lock maps in the revision format, enabling scoped dependency resolution for incompatible diamond dependencies. The flat lock map from Stage 1 is widened to allow nested maps at any depth, while maintaining backward compatibility and canonical serialization throughout.

## Key Changes

- **Recursive lock schema**: Replaced flat `record(string)` with a self-referential schema `lock = () => ['record', lockValue]` where `lockValue = () => ['or', string, lock]`, allowing nested maps to any depth
- **Type definitions**: Added hand-written `LockMap` type with recursive structure and `LockSchema` annotation type to support self-reference, with assertion that they match the rtti schema
- **Semantic validation**: Extended `checkReferences` to recursively validate every direct hash string at every nesting depth via new `lockError` function that tracks the path to invalid values
- **Evo API**: Updated `RevisionData` type to use the recursive `LockMap`, with `add` and `revision` operations now recursively validating and canonicalizing hashes while preserving nested scope structure
- **Serialization**: Confirmed existing `stringify(sort)` rule already handles nested objects, so canonical serialization automatically applies to all nesting levels
- **Documentation**: Updated README with nested lock map examples (incompatible diamond case), clarified that nesting is optional and scopes have no format-defined semantics, and documented the widening decision to keep the same dialect
- **Proofs**: Added comprehensive tests for nested lock acceptance, deep nesting, malformed nested values, invalid nested hashes with path reporting, nested key sorting, and deduplication of equivalent nested structures with different property orders
- **MCP schema**: Updated `evo_add` documentation to reflect `lock?` parameter and ensured JSON Schema publication correctly expresses the recursion via `$defs` and `$ref`

## Implementation Details

- The recursive schema uses a module-level `lockValue` constant rather than rebuilding the union on each call, because the rtti data form closes reference cycles by identity
- Nested maps are scopes, not references—only direct string values are validated as hashes, not the map structure itself
- The format records nested information but defines no overlay, replacement, or inheritance semantics; resolvers decide how to interpret scoped bindings
- Backward compatibility is maintained: every flat map remains valid and means what it always did, so the dialect tag stays unchanged rather than becoming `vnd.fjs.revision2`
- Path tracking in `lockError` enables precise error messages like `"lock value for B/D is not a valid hash"` for deeply nested failures

https://claude.ai/code/session_018Ga96cJJGEct1zP4hCvWJj


o2alexanderfedin (approved):
Approving. Draft status is not a blocker.

**Head note:** I did the deep work at `0ded0792` against baseline `origin/main` = `c3af487f`; the head moved to `13513a29` mid-review. That commit is a pure merge of `origin/main` (now `4a3b6877`, after #1561/#1562/#1563 landed) into the branch — `git diff 0ded0792 13513a29 -- fjs/media/revision fjs/cas/evo fjs/mcp/evo changelog/unreleased/1565.md fjs/media/todo` is **empty**, so the PR's own content is byte-identical. I re-ran the gates at `13513a29`: `npx tsc --noEmit` → 0, `npm test` → `pass: 2720, fail: 0`. The marker below is the head I am approving.

## Gates

- `npx tsc --noEmit` → 0 at both heads. `npm run prepack` from a clean tree → 0.
- `npm test` at `0ded0792`: `pass: 2719, fail: 0` vs `c3af487f`'s `2708` — exactly the 11 new proof cases (6 in `media/revision`, 3 in `cas/evo`, 2 in `mcp/evo`).
- `bin/linkcheck.mjs` — broken-link **sets** identical to main; deleting the 382-line `fjs/media/todo/revision-lock-map.md` strands nothing (`git grep revision-lock-map` at the head: no hits).
- Emitted declarations read directly. **No `elided` and no `: any`** in `media/revision`, `cas/evo` or `mcp/evo`, and `@module` survives in all three. This is the §6.2 recursive-constant trap and it is handled: `export declare const lock: LockSchema;` keeps the name, and `revisionSchema`'s field emits as `Or<readonly [LockSchema, undefined]>` rather than an inlined structure that gives up at depth. The named `@type {LockSchema}` (instead of `@type {const}`) is doing exactly the job the JSDoc claims for it.

## Termination — the thing I most wanted to break

"Recursive" plus "resolution" is where non-termination hides, and this repo already shipped one hang (#1548), so I drove every walker over the recursive schema rather than reasoning about it:

| probe | result |
|---|---|
| `toData(lock)` → `dataToTs()` | terminates in 89 ms; emits one rule named `lockValue` and refers to it — the identity-based cycle closing the JSDoc describes actually works |
| `toJsonSchema(evoAddArgs)` | terminates; `$defs: ['lockValue']` |
| `toJsonSchema(revisionSchema)` | terminates; `$defs: ['lockValue']` |
| `checkReferences` on a 5000-deep lock | `ok` in 1.8 s |
| `checkReferences` on a **cyclic** lock map | `RangeError: Maximum call stack size exceeded` in ~1 s — bounded, not a hang |

So nothing loops forever. Two notes on the boundaries, both **not defects**:

- Deep input is stack-bounded, and the bound is the pre-existing one. `validate` gives `ok` at depth 1000 and `RangeError` at depth 5000. On `origin/main`, `rtti/validate` against the *existing* recursive `media/json` `unknown` schema does the same thing at the same depths (`ok` at 1000, `RangeError` at 5000, `RangeError` on cyclic input). The new schema is no worse than the recursive schema the repo already ships, and the README's "to any depth" is a statement about the *format*, which is accurate.
- Cyclic input is unreachable through any documented entry: `validate` is fed by `parseJson`, and `evo_add` by JSON over MCP; neither can produce a cycle. It also matches main's behaviour on the JSON schema, so there is nothing here specific to this change.

The README's argument that lock maps cannot introduce a *revision-hash* cycle — entries select content hashes, never revisions carrying further lock maps — holds independently of any of this.

## Proof coverage (§3.2) — mutation-tested, not trusted by name

Every new/changed production path, mutated at `0ded0792` and run against the whole suite:

| mutation | fails |
|---|---|
| `lockError`: return `null` instead of recursing into a nested map | **2** |
| `lockError`: report `${subject}` instead of the `path.join('/')` scope path | **2** |
| `canonicalLock` (`cas/evo`): return the nested map as-is instead of recursing | **1** |
| `lockValue`: drop the nested-map alternative from the schema union | **16** |
| `evoAddArgs`: drop `lock` from the advertised args | **1** |

All five killed. Worth calling out that the fifth kills exactly one test and that is *correct*: rtti struct validation ignores properties the schema does not name, so removing `lock` from `evoAddArgs` does not change what the server accepts — only what `tools/list` advertises — and `evoAddInputSchemaPublishesTheLockRecursion` is the proof that notices. That proof pins the real JSON Schema shape (`$defs.lockValue`, `anyOf` of a string and an object whose `additionalProperties` `$ref`s back), not just that a property exists.

The proofs also avoid the failure mode that has recurred here (#1525/#1528/#1559 — pinning one field and letting a state-clobbering mutant through): `evoAddAndRevisionCarryNestedLocks` pins the whole serialized revision string, `revisionNestedLockRoundTripsAndCanonicalizes` checks the nested value *and* that re-adding the read-back revision lands on the same CAS address, and `equivalentNestedLockOrders…` pins address convergence in both `media/revision` and `cas/evo`. Sparse/empty nested maps, a depth-4 nest, a nested non-string (structural rejection) and a nested non-hash string (semantic rejection, with the path in the message) are all covered.

## Conventions

- `changelog/unreleased/1565.md`: per-PR file, `**BREAKING CHANGES:**` prefix (§8.4), and **no PR link inside** — the first entry in `unreleased/` that gets §8.3's current rule right.
- The breaking claim is honest and, if anything, conservative: `LockMap` genuinely widens to `string | LockMap`, and the `_lock` → `lock` rename is called out even though `_lock` was `_`-prefixed and had no consumers outside this module pair (`git grep '_lock\b'` on `c3af487f`: only `module.f.mjs` and its own `types.ts`).
- `type _LockMap = Assert<Equal<LockMap, Ts<typeof lock>>>` is non-exported and `_`-prefixed, matching `media/json/types.ts`'s `_Unknown` exactly — the precedent the header cites.
- READMEs across all three modules are updated consistently with the code, including the incompatible-diamond motivation and the "why this keeps the `vnd.fjs.revision` tag" argument.

## One observation on §6.2 naming — not blocking

`LockSchema` is a **new ordinary (unprefixed) export** from `media/revision/types.ts`, and its whole purpose is to let the sibling `module.f.mjs` annotate `lock` — a self-referential const cannot infer its own type, so the alias has to be named and exported to be `@import`ed. That is the exact case §6.2 describes for the `_` prefix: "`_`-prefixed exports are for types a sibling `.f.mjs` must `@import`."

The precedent the doc comment points at goes the other way too: `media/json/rtti/module.f.mjs` writes its recursive schema's type **inline** (`@type {() => readonly['or', typeof primitive, typeof object, typeof array]}`) and publishes no schema-side alias — only the data-model types are public. Here `LockSchema` is the shape of an rtti schema thunk, so adding a third alternative to `lockValue` later would be a visible change to an ordinary public name.

I am not treating this as a defect: the types.ts header lists `LockSchema` in the type-level API deliberately, the alias is unavoidable (unlike JSON's, this cycle has only two nodes and one of them must break inference), and it would show up in the emitted `.d.mts` either way. If publication was not intentional, `_LockSchema` is the spelling §6.2 asks for. Your call.

Solid work — the recursion is closed by identity in the one place it needs to be, the emitted declarations survive it, and every walker over it terminates.

PR #1570, opened 2026-08-14
media: file a todo for a shared lock format

sergey-shandar:
Documentation only — one new `todo/` file and a link to it. No code changes, so no CHANGELOG entry (AGENTS.md §8.3).

## Why

[#1565](https://github.com/functionalscript/functionalscript/pull/1565) deleted `fjs/media/todo/revision-lock-map.md` on completion and preserved its "Future shared lock files" section as an out-of-scope bullet in the revision README. That captured the design decision, per AGENTS.md §7, but left the follow-up work tracked nowhere. This writes it up.

## What the issue proposes

Two halves, separable — and deliberately not presented as equally safe:

**1. `vnd.fjs.lock`, a history-free lock blob.** A separate dialect in `fjs/media/lock/`, `{ dialect, lock }`, importing `fjs/media/revision`'s `lock` schema rather than restating it — one recursive schema, one semantic hash check, one canonical serialization, so a map means the same thing inline and shared. No `subject`/`parents`/`generation`: it is a value, not a step. History for lock content, when wanted, is an ordinary revision whose `snapshot` is such a blob, so no second history mechanism appears.

**2. A revision field referencing shared lock content.** This is the half needing a decision first, because it is *fail-open* where the nested-map widening was fail-closed. Widening `lock` to nested maps kept the dialect precisely because an older reader rejects what it cannot represent. A reference field inverts that: a reader that does not know it still validates the blob and sees `lock`'s documented "no bindings were recorded", so it resolves through mutable heads believing the result is reproducible. That is the silent misread the versioning rule exists to prevent — the same trap that keeps incremental diffs out of the format — so the reference form likely needs `vnd.fjs.revision2` rather than the additive path.

Four open design points are recorded rather than decided: the spelling (`lockRef` vs. widening `lock` to `or(hash, lock)` at the top level), whether an inline map and a reference may combine, whether lock blobs compose (they safely can — entries and references alike select content hashes, and a hash-consistent store cannot cycle among them), and the validation boundary (the media layer can check cbase32 and nothing more, the contract `snapshot` already has).

## Motivation recorded in the issue

An inline map is re-serialized into a new blob on every revision whose bindings did not change — the ordinary case, where the source moved and the dependencies did not — and CAS deduplicates whole blobs, not fragments. Separately, two subjects resolving the same dependency set cannot point at one agreed resolution. A lock-only change correctly produces a new revision; the issue is explicit that this is not about that case.

## Changes

- `fjs/media/todo/shared-lock-format.md` — new, P4/open
- `fjs/media/revision/README.md` — the out-of-scope bullet now links to it

---
_Generated by [Claude Code](https://claude.ai/code/session_018Ga96cJJGEct1zP4hCvWJj)_


o2alexanderfedin (approved):
Reviewed against `origin/main` @ `60d04cb7ea37df06a42def61ceb18ef8ffc4eb50` (the merge-base is exactly that commit). Docs-only, so I checked what a docs-only change can actually get wrong here: every path, count, and existential claim in the new todo, re-derived against the tree rather than taken on trust.

**Paths and links.** `node bin/linkcheck.mjs` over both trees: the broken-link *set* is byte-identical to main — the new file strands nothing and adds nothing broken. I resolved each relative link by hand as well, since the new file sits one directory deeper than the doc it links back to: `../revision/README.md`, `../revision/module.f.mjs`, `./change-content-format.md`, `../../cas/evo/README.md`, `../../todo/group-fs-subdirectories-by-concern.md` all exist. The two section anchors resolve too — `## Lock maps` (line 49) → `#lock-maps` and ``### Widening `lock``` (line 177) → `#widening-lock`.

**Claims about other modules.** These are the ones I expected to find loose, and they hold:

- ``lock`` really is exported from `fjs/media/revision/module.f.mjs` (line 75, `() => ['record', lockValue]`), so "imported, not restated — one recursive schema" is accurate rather than aspirational.
- The fields the doc says a lock dialect would *not* carry are exactly `revisionSchema`'s: `dialect`, `subject`, `parents`, `snapshot`, `generation`, `archived: option(true)`. The `option(true)` idiom the README edit sits next to is real.
- "pure format with no store access" matches the module header, and the validation boundary the doc proposes ("a cbase32 hash and nothing more") is the contract `snapshot` already has via `checkReferences` / `isHash`.
- `dialectEntry` exists in `fjs/media/module.f.mjs` and `revisionDialect` is a real `DialectEntry` export.
- `fjs/cas/evo/README.md` does define `RevisionData` (line 45), and `evo_add` / `evo_revision` are real tools (`fjs/cas/README.md:11`).
- The cited cycle rule is genuinely there: `fjs/todo/group-fs-subdirectories-by-concern.md:130` gives the `cas` ↔ detector cycle as the reason the revision *format* lives under `fjs/media/`, and the same file carries the `vnd.fjs.*` dialect-naming convention the doc appeals to. So "by the same cycle rule" is a correct citation, not a plausible-sounding one.
- `change-content-format.md` really does turn on the same fail-open versioning trap the new doc invokes.

**Conventions.** Heading, `**Priority:** P4`, `**Status:** open`, and the `### Problem` / `### Proposal` / `### Tasks` / `### Related` structure match the template in `todo/README.md` and the sibling `fjs/media/todo/` files. No changelog entry, which is right for a `todo/`-plus-README change under §8.3.

**The README edit.** It appends a pointer to the existing out-of-scope bullet without altering the bullet's claim, and the bullet already named `vnd.fjs.lock`, so the todo and the spec agree on the name.

One thing I'd note rather than object to: the doc is unusually careful about the part that is easy to get wrong — it explicitly separates the safe half (a new dialect, no versioning consequence) from the fail-open half (a reference field an old reader would silently read as "no bindings recorded"), and makes deciding that a task rather than assuming the additive path. That is the correct reading of the format's own versioning rule.

Nothing to fix. Approving.

PR #1571, opened 2026-08-14
Bring emergent_testing to 100% line/branch/function coverage: 97.76% → 100% branch

sergey-shandar:
## Summary

`fjs/emergent_testing/module.f.mjs` was at 97.76% branch coverage (3 uncovered branches). This brings it to 100% line/branch/function coverage:

- **`register`** (the `NodeProgram` entry point used by `fjs/emergent_testing/all.test.mjs`) had no direct proof coverage at all — its `inlineTestContext` ternary (line 384), `engine === 'bun'` ternary (line 385), and `registerModuleMap`'s `modules.length === 0` short-circuit (line 232) were only ever exercised indirectly, and never with the branches that matter.
- Added `registerEmptyModuleMap` / `registerEmptyModuleMapInlineBun`, which run `register` through the `virtual()` effect harness (`effects/node/virtual`) against an **empty** virtual root. With no modules to register, `registerModuleMap` short-circuits before ever invoking the selected `TestContext`'s `test` op — so this reaches `register`'s own branches without ever hitting the virtual harness's `test: todo` stub (which would otherwise throw "not implemented").
- Covers Node's default context (non-inline) in one case, and Bun's inline context (`inlineTestContext: true, engine: 'bun'`) in the other, exercising both sides of each ternary plus the empty-module-map branch.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `node --test --experimental-test-coverage --test-coverage-include='fjs/emergent_testing/module.f.mjs' fjs/emergent_testing/all.test.mjs` → 100.00% line/branch/func
- [x] `node ./fjs/module.mjs t` → 2727 pass, 0 fail

---


o2alexanderfedin (commented):
Reviewed against `origin/main` @ `60d04cb7ea37df06a42def61ceb18ef8ffc4eb50` (merge-base is exactly that commit).

## The headline number is real

`npm run cov` on both trees, `fjs/emergent_testing/module.f.mjs`:

| | line | branch | func |
|---|---|---|---|
| main | 100.00 | **97.76** | 100.00 |
| PR head | 100.00 | **100.00** | 100.00 |

So the title's `97.76% → 100%` checks out exactly. Repo-wide branch also moves 99.63 → 99.69. `npx tsc --noEmit` exits 0; `npm test` goes 2724 → 2726, i.e. exactly the two new cases and nothing else disturbed. The changelog entry matches the current per-entry format.

## But the two new cases don't prove the branches they name

Both cases assert only `assertEq(exitCode, 0)` and discard the returned state (`const [, exitCode] = ...`). With `root: {}`, `registerModuleMap` short-circuits at `modules.length === 0` — the comment in the PR says so itself — so neither `star` nor `ctx` is ever *used*. They are computed, which is enough for the branch counter, and then thrown away.

I mutated `register` and re-ran the two new cases directly (`registerEmptyModuleMap`, `registerEmptyModuleMapInlineBun`):

| mutation to `register` | new cases |
|---|---|
| `() => 0` → `() => 1` (negative control) | **FAIL** ✔ probe is live |
| `o.inlineTestContext ? ' ...' : ''` → `? '' : ' ...'` (inverted) | pass |
| `o.engine === 'bun' ? o.bunTestContext : o.testContext` → arms swapped | pass |
| `const star = ''` (branch deleted) | pass |
| `const ctx = o.testContext` (branch deleted) | pass |
| `step(loadModuleMap(o.env), registerModuleMap(ctx, star))` → `pure(undefined)` (discovery removed entirely) | pass |

The negative control fails as it should, so this isn't a probe that never runs. And the two ternary inversions also survive the **whole** suite: with both arms swapped, `npm test` at this head is still `pass: 2726, fail: 0`.

The last row is the one I'd weigh most: `register` can be gutted to a constant that returns `0` and both new cases stay green.

This is the same shape as #1564 — a case that raises the branch counter but whose assertion cannot distinguish the branch it names. `registerSuffixes` already proves what `registerModule` *does* with `' ...'`; what is still unproven after this PR is `register`'s *derivation* of `star` from `o.inlineTestContext` and of `ctx` from `o.engine`, which is precisely what these two cases are named for.

## Suggestion

Give the cases a non-empty `root` so `registerModuleMap` doesn't short-circuit, and make the selection observable:

- pass distinguishable `testContext` / `bunTestContext` mocks (e.g. each records a different marker) and assert which one received the registrations — that kills the `engine` swap;
- assert on the registered test *names* so the `' ...'` suffix is visible — that kills the `inlineTestContext` swap, the same way `registerSuffixes` does one level down;
- assert on the returned state, not just `exitCode`, so a mutant that returns the right code while doing nothing (or the wrong thing) is caught.

Then the branch counter and the proof would be saying the same thing. As it stands the 100% is accurate but hollow for these two branches, so I'm commenting rather than approving.

sergey-shandar:
Good catch, and confirmed — same #1564 shape. Fixed in 0eae811: `registerSelectsContextAndStar` replaces the two hollow cases. It interprets `register`'s effect through a synthetic runner (`readdir`/`import` faked for discovery, since the virtual harness's own `test` op is `todo` and would throw on a non-empty root) that records which `TestContext` object and which registered name each `test` call actually received, for both the Node/non-inline and Bun/inline combinations. `registerEmptyModuleMap` stays, narrowed to just the `modules.length === 0` short-circuit it actually proves.

Re-ran your mutation table against the new case: the two ternary-arm swaps and the "discovery removed entirely" mutation now all fail as expected.

---

o2alexanderfedin (approved):
Re-reviewed at `0eae811` (previous review was at `492838b`). The open finding from that
review is **resolved** — verified by rebuilding the mutants from scratch rather than by
reading the new test.

## The previous finding, re-tested

Previously `registerEmptyModuleMap` / `registerEmptyModuleMapInlineBun` both ran with
`root: {}`, so `registerModuleMap` short-circuited and neither `star` nor `ctx` was ever
used; mutating either ternary left both cases green. `0eae811` replaces the second case
with `registerSelectsContextAndStar`, which interprets `register`'s effect through a
synthetic runner and asserts *which* `TestContext` object and *which* registered name each
`test` call received.

I rebuilt every mutant against `fjs/emergent_testing/module.f.mjs` at this head — invert
each ternary, delete each ternary (both arms), swap the exit code, and replace the whole
discovery step with `pure(undefined)` — verifying each edit actually landed in the file
before running, and ran the two cases. **All eight die:**

| mutant | source after mutation | result |
|---|---|---|
| negative control | `mapStep(registered, () => 1)` | **dies** — `registerEmptyModuleMap`: `1,0` |
| invert `star` | `o.inlineTestContext ? '' : ' ...'` | **dies** — got `…ok() ...`, wanted `…ok()` |
| delete `star`, keep `''` | `const star = ''` | **dies** — bun case got `…ok()`, wanted `…ok() ...` |
| delete `star`, keep `' ...'` | `const star = ' ...'` | **dies** — node case got `…ok() ...`, wanted `…ok()` |
| invert `ctx` | `o.engine === 'bun' ? o.testContext : o.bunTestContext` | **dies** — identity assertion fails |
| delete `ctx`, keep node | `const ctx = o.testContext` | **dies** — identity assertion fails |
| delete `ctx`, keep bun | `const ctx = o.bunTestContext` | **dies** — identity assertion fails |
| replace discovery | `const registered = pure(undefined)` | **dies** — `0` calls, wanted `1` |
| both ternaries inverted | (combination) | **dies** |

The negative control is live and the surviving-mutant class from #1564 is gone. Both
new cases earn their place: only `registerEmptyModuleMap` kills the `() => 0` exit-code
mutant (`registerSelectsContextAndStar` passes under that one), and only
`registerSelectsContextAndStar` kills the `star`/`ctx`/discovery mutants.

Because `emergent_testing` is the harness itself, I sanity-checked that this is not an
artefact of my standalone runner: with the `ctx`-inversion mutant applied, the real
`npm test` reports

```
import("./fjs/emergent_testing/proof.f.mjs").proof.registerSelectsContextAndStar(): error
Number of tests: pass: 2725, fail: 1, total: 2726
```

so the failure is visible through the repo's own harness, not just through my probe.
The tree was `git clean`ed before every mutant run (no emitted `.f.js`/`.d.mts` present).

## Battery

- `npm run prepack` from a freshly `git clean`ed tree — exit 0 (both passes).
- `npm test`: **2726 pass / 0 fail** here vs **2724 pass / 0 fail** on `fs-main` at
  `origin/main` (`b354477`). +2 = exactly the two added proof exports. Both trees cleaned
  before testing.
- Link check: broken-link **sets** are byte-identical to `origin/main` (144 lines each,
  `diff` of the sorted sets is empty).
- Coverage: unlike the usual vacuous run, `npm run cov` produced real numbers here
  (2662 tests pass at this head, 2660 on main), so the changelog claim is measurable and
  holds *at this head*, not only at the previous one:
  - `origin/main`: `fjs/emergent_testing/module.f.mjs` — line 100.00 / **branch 97.76** / func 100.00
  - `0eae811`: `fjs/emergent_testing/module.f.mjs` — line 100.00 / **branch 100.00** / func 100.00

  So swapping `registerEmptyModuleMapInlineBun` for `registerSelectsContextAndStar` kept
  the coverage result while making the branch actually observable.

## Other checks

- `changelog/unreleased/1571.md` matches the current per-entry convention: no heading, no
  link, no PR number in the body. A proof-only change would not have needed an entry at
  all; having one is harmless.
- No new exported types; `_RegisterMockOps` / `_RegisterMockState` are pre-existing, and
  the added `@import`s (`Import`, `Readdir`) are type-only. No public surface change beyond
  the two proof exports, so I did not run the full dual-axis surface diff.
- Nothing under `nanvm-lib/` or `fjs/nanvm/`, so the Rust gates and the generator
  round-trip do not apply.

## Non-blocking note

`registerSelectsContextAndStar`'s runner re-implements the `all` / `await` bodies of
`makeRegisterRunner` (the comment acknowledges this) and accumulates results into a
reassigned `let calls` rather than threading them through the runner's state channel the
way `_RegisterMockState` does for the other three register cases. `let runner` is forced by
the recursion; `let calls` is not — a state type of
`readonly (readonly [TestContext, string])[]` would have kept the established shape.
Purely a consistency point inside a proof; it does not affect what the test proves, and I
am not asking for a change.

Approving.

o2alexanderfedin (approved):
Re-review at `2c12de6b5d960b9dff74739c0dacc406bcf244d4`. Previously approved at `0eae8111`.

**Merge-only.** `git diff 0eae8111 2c12de6b` is exactly the two Markdown files that came in with `b3544773` (#1570):

```
M	fjs/media/revision/README.md
A	fjs/media/todo/shared-lock-format.md
```

Neither the proof nor the new runner moved, and no test file anywhere changed, so I did not rebuild the mutants — the result from the previous head stands:

- All 8 mutants (invert and delete each ternary, `pure(undefined)` discovery, both-inverted) die, with a live negative control.
- Branch coverage 97.76 → 100.00; test count 2726 vs 2724 on `main` — and `main` gained no tests in this merge, so the delta is unchanged.

The earlier non-blocking style note still applies and is still non-blocking: the new runner duplicates `makeRegisterRunner`'s `all`/`await` and threads a reassigned `let`. Worth folding in if you touch it again; not a reason to hold the PR.

Baseline for this round was `b35447738e756288e6e1da08404b407ec5a9ab7d`, this PR's merge-base. `origin/main` has since moved to `6bd36947` (#1573 merged), which is not in this branch and does not touch `fjs/emergent_testing/`.

Approving.

PR #1573, opened 2026-08-15
media/json: extended bigint codec over a lexeme-first tokenizer

sergey-shandar:
Adds the extended JSON codec — JSON's bare integer syntax as `bigint` — and the
lossless path it needs: a lexeme-first tokenizer and one structural parser that
takes a numeric policy per codec.

## Lexeme-first tokenization

`NumberToken` is now `{ kind: 'number', value: string }`. Scanning a number
accumulates text only: the eager coefficient bigint (`m * 10n + digit`, quadratic
in the digit count) and the `number` exponent buffer are gone, along with `bf`,
the derived `BigFloat` that no consumer in the repository ever read. A
syntactically valid literal therefore always tokenizes, whatever its magnitude,
and negation joins the lexeme instead of multiplying a `BigFloat`.

The dropped `bf` was not merely redundant: its exponent was accumulated as an
ordinary `number`, so past 2^53 it silently lost digits —
`1e999999999999999999999` tokenized with an exponent of `1.0000000000000001e21`.
Nothing read the field, so this was latent rather than observed, but a derived
value that must be right before the token exists is the wrong shape either way.

## One structural parse, a numeric policy per codec

`parser.parse` takes a `NumberPolicy<P>` and hands it each number token; the
policy materializes that token into its own domain or fails with a `Result`.
Standard JSON (`parseFloat`) and extended JSON (`number | bigint`) are two
policies over one tokenizer and one container state machine, so standard parsing
never builds an intermediate `bigint` and extended parsing never goes through a
rounded `number`.

The design in the issue was an intermediate tree holding `NumberToken` values at
its numeric leaves. That representation is not sound in a JSON-shaped tree:
`{"kind":"number","value":"5"}` is ordinary JSON data, and a materializer walking
such a tree cannot tell it from a retained token. Making it sound needs a fully
tagged second tree, built and walked for no gain since materialization runs
immediately after the parse. Parameterizing the one parser keeps the exact lexeme
available to every policy with no representation that can be confused with data.
`fjs/media/json/README.md` records this.

## Numeric policy, settled

The blocking numeric-edge investigation left two extended questions open; they
are answered here and recorded in that issue and in the README:

- a token outside the finite `number` range (`1e400`) is a **parse error**, not
  `Infinity` — the extended domain is exact where it claims to be, so it rejects
  rather than rounds. Ordinary rounding *within* range (`1e-400` to `0`) is
  inherent to `number` and is not an error;
- programmatic `NaN` / `±Infinity` serialize as `null`, which is valid JSON and
  what `JSON.stringify` does with them.

Serialization keeps the runtime types apart, so the round trip is one-to-one:
`0n`/`0`, `0`/`0.0`, `-0`/`-0`, and a `bigint` is always full base-10 digits —
exponent syntax would parse back as a `number`.

One limit is documented rather than contained: a bare integer beyond the
runtime's bigint limit (V8: ~3.2e8 digits, a several-hundred-megabyte document)
throws inside `BigInt`. FunctionalScript has no `try`/`catch`, so it cannot
become a `Result`, and predicting it from a digit count is the size-estimating
preflight DESIGN.md §6 rules out. Nothing before that point narrows, so the
standard parser still reads such a document.

## Also

- `fjs/media/json/number` — bounded lexical helpers (`numberLexeme`,
  `isBareInteger`, `isIntegral`). Exponents are compared as digit text against
  small counts, so `1e-99999999999999999999` is decided from a sign and a length:
  no exponent conversion, no `10 ** exponent`.
- `treeSerialize` in `serializer/module.f.mjs` is the one recursive walk over
  JSON containers, over a leaf-parameterized `Tree<P>` in `types.ts` whose object
  index signature is optional; both codecs supply only a leaf spelling.
- `definedEntries` infers its value type from the record so a generic-valued
  record can be passed.
- `todo/bigint-parse-serialize.md` is deleted; every issue that linked to it was
  rewritten, and several are now unblocked.

`npx tsc` clean, 2787/2787 proofs pass, and every new or changed `module.f.mjs`
is at 100% line and branch coverage. `fjs/media/json/tokenizer`'s 97.06% branch
coverage is pre-existing — verified against a baseline worktree.

Changelog:
- **BREAKING CHANGES:** `media/json`: new `extended` codec — bare integer
  syntax parses to `bigint` and serializes back as ordinary JSON, `-0` stays a
  `number`, and out-of-range tokens are parse errors. `parser.parse` now takes
  a numeric policy, so standard and extended parsing share one state machine
- **BREAKING CHANGES:** `js/tokenizer`: a number token carries only its exact
  lexeme; the derived `bf` field is gone. It accumulated its exponent as a
  `number`, so a long one silently lost digits, and its coefficient had to be
  built as a bigint before the token existed
- `media/json`: new `number` module with bounded lexical helpers
  (`numberLexeme`, `isBareInteger`, `isIntegral`) that classify a token in its
  own length, without narrowing it or evaluating `10 ** exponent`
- `types/object`: `definedEntries` infers its value type from the record it is
  given, so a record whose value type is generic can be passed

https://claude.ai/code/session_01TGZuMdgxFjMMs5ykoH8MVx


o2alexanderfedin (approved):
Reviewed at `4c7c22ce`, baseline `origin/main` = `60d04cb7` (`media/json/parser: hoist the repeated unexpected-token error (#1569)`), which is also this PR's merge-base. Draft status is not a blocker.

**Approving.** I could not find a defect. Below is what I ran, because a codec plus a tokenizer change deserves more than reading it.

### Gates

| check | result |
|---|---|
| `npm run prepack` (clean tree, both passes) | exit 0 |
| `npm test` | 2787 pass / 0 fail (main: 2724 / 0) |
| `bin/linkcheck.mjs`, broken-link **sets** vs main | identical, no entry added or removed |
| CHANGELOG | `changelog/unreleased/1573.md`, per-entry bullets, no links or PR numbers, no headings, `**BREAKING CHANGES:**` on the two breaking bullets — matches `1569.md`'s shape |

### Round-trip: exhaustive where the domain allows, boundary-targeted where it does not

`parse(stringify(v)) === v` (`Object.is`, so `-0` is not `0`) plus text fixpoint `stringify(parse(stringify(v))) === stringify(v)`:

- **bigint, exhaustive**: every integer in `[-200000, 200000]` — 400,001 values.
- **bigint, digit-count boundaries**: `10^d`, `10^d ± 1` and their negations for `d = 1..400`; plus `2^32`, `2^53 ± 3`, `2^63`, `2^64`, `MAX_SAFE_INTEGER ± 3`, both signs.
- **number, boundaries**: every power of two `2^-1074 … 2^1023` with its `±1 ulp` neighbours, every power of ten `1e-323 … 1e308` with neighbours, the subnormal floor `5e-324`, `MAX_VALUE`, `EPSILON`, the `1e21` spelling switch, the `1e-7` spelling switch, `±0`.
- **number, random**: 3,000,000 raw 64-bit patterns, 2,998,853 finite.
- **number, whole-valued**: every integer in `[-200000, 200000]` as a double, which is where the `.0` rule lives.
- **token corpus**: 784 lexemes over sign × integer × fraction × exponent, 639 accepted, all fixpoint after one round.

**0 failures out of ~3.8M round trips.** Range discipline checked separately and matched: `1e400`, `-1e400`, `1e309`, `1.8e308` are `error`; `1.7976931348623157e308`, `1e-400`, `1e-323` are `ok`; `1` + 400 zeros is `ok` as a `bigint` while the same digits with `.0` is `error` — i.e. the bigint branch is deliberately unbounded and the number branch is not. `-0`/`0` discipline: `0 -> 0n`, `-0 -> -0` (number), `0.0 -> 0`, `-0.0 -> -0`, `-0e0 -> -0`; `str(0n) = "0"`, `str(0) = "0.0"`, `str(-0) = "-0"`.

The harness is negative-controlled — four mutations of the codec, each verified to have actually landed in the file before running:

| mutation | detected as |
|---|---|
| drop the `-0` case in `numberSerialize` | 1 round-trip failure |
| bare integer via `Number` instead of `BigInt` | 60,301 round-trip failures |
| drop the `.0` suffix rule | 60,146 round-trip failures |
| accept out-of-range instead of `error` | 2 range failures |

### §3.2 — the new proofs are mutation-tested, not taken on their names

Same seven mutations run against the **shipped** proofs (`extended`, `number`, `parser`, `tokenizer`, `serializer`, `json`, `js/tokenizer`, `djs/tokenizer` — 575 proof invocations, 0 baseline failures):

| mutation | proofs that die |
|---|---|
| drop `-0` serialization | `extended.number.negativeZero`, `containers.array`, `serialize.oneToOne` |
| `Number` instead of `BigInt` | 8, incl. `bigint.beyondSafeInteger`, `bigint.oversized` |
| drop the `.0` rule | 7, incl. `number.wholeFraction`, `number.exponentSpelling` |
| accept out-of-range | all three `overflow.*` |
| `isBareInteger` ignores `exp` | 10 |
| `isIntegral` ignores trailing zeros | `isIntegral.plain.1`, `isIntegral.exponent.3` |
| `exponentAtLeast` negative branch | `isIntegral.exponent.5`, `isIntegral.unbounded.2` |

A deliberate no-op mutation (`BigInt(value) + 0n`) killed nothing, which is the control in the other direction.

### `number` module against an exact bigint oracle

`isIntegral` compared against `coeff % 10^(fracLen - exp) === 0` computed in `BigInt`, over **5,511,056 exhaustive tokens** — sign × `int ∈ {0,1,9,10,100,1230,705,9000}` × every fraction of length 0–4 over all ten digits × 28 exponent forms. Same sweep asserted that `numberLexeme` reconstructs its input token byte-for-byte and that `isBareInteger` matches the lexical predicate. **0 mismatches on all three.** Unbounded cases (400- and 2000-digit coefficients and exponents) terminate and agree with the oracle — including `1.0e-999…9`, where I first wrote down the wrong expected answer and the oracle was right: `10 × 10^(-999…9 - 1)` is not an integer.

### Old vs new tokenizer, token-sequence differential

Full token streams compared main vs PR with `bf` stripped and `bigint`s normalized, over **1,680 inputs**: every `.mjs`/`.ts`/`.js`/`.json` under `fjs/` and `todo/` in the tree — real sources containing `'/'`, `'//'` and `'/*'` string literals, so no regex normalizer was involved anywhere — plus an adversarial corpus of 60-digit coefficients, 40-digit exponents, `123n`, `1.2.3`, `01`, `1e+`, `.5`, `1.`, and bare punctuation.

```
js/tokenizer:         inputs 1680, divergences 0
media/json/tokenizer: inputs 1680, divergences 0
djs/tokenizer:        inputs 1680, divergences 0
```

Negative control (perturb the first PR token): 1680/1680 detected.

**The dropped `bf` was carrying a latent bug, not behaviour.** Nothing outside the three tokenizers ever read `.bf` (`git grep '\.bf\b'`), and on `main` its exponent was accumulated as `b.e * 10 + digit` in a `number`, so it silently lost precision past 2^53:

```
main: 1e999999999999999999999            -> bf = [1n, 1.0000000000000001e+21]   (want 999999999999999999999)
main: 1.0…01e99999999999999999999        -> bf = [10…01n, 100000000000000020000] (want 99999999999999999999)
```

So removing it is a latent-bug removal. One wording nit, non-blocking: the changelog says "no valid number fails to tokenize because a coefficient or exponent exceeds a runtime numeric type" — on `main` no such number *failed* to tokenize either, it tokenized to a quietly wrong `bf`. Silent corruption rather than failure is the stronger thing to have fixed.

### The standard codec is behaviour-identical

`treeSerialize` replaced a hand-written walker and `parse` now routes through a policy, so I diffed main vs PR directly: `json.parse` over 1,754 texts (numeric grammar cross-product plus malformed inputs — `01`, `1.`, `.1`, `1e`, `+1`, `[1,]`, `{"a":}`, unterminated string, empty), and `json.stringify` over 400,264 value/sort-mode pairs (both `identity` and `sort`). **0 divergences**, negative control 500/500.

The only differences are outside `Unknown`, where TypeScript already rejects the input: `undefined` used to throw `Cannot convert undefined or null to object` and now serializes as `null`; a function and a `bigint` used to serialize as `{}` and now serialize as `null`. Garbage in either way, and `null` is the less surprising garbage out.

### Conventions

- **§4 module header** — `@module` survives declaration emit in both new modules (`extended/module.f.d.mts`, `number/module.f.d.mts`, 1 each); the new `types.ts` files have the load-bearing blank line after `*/`. The two new `proof.f.mjs` carry no `@module`, matching the repo (10 of 116 proofs have one).
- **§2 pairing** — every import in both new proofs is already `.f.mjs`.
- **§6.2 `_` prefix** — the added unprefixed exports are the ones that are genuinely public: `Tree`/`TreeObject`/`TreeArray` are the data model `extended/types.ts` instantiates and the README documents, `NumberPolicy` is what a caller of `parser.parse` must now supply, `ParseUnknown` is its result. The implementation-only names stayed prefixed and just gained a parameter (`_JsonObject<P>`, `_StateParse<P>`, `_JsonState<P>`, `_MapEntries`). Not a mechanical violation.
- **§6.2 recursive constants / declaration emit** — read the emitted `.d.mts` directly rather than through `bin/extract.mjs`. No `elided`, no widening to `any`; `parse` emits as `<P>(policy: NumberPolicy<P>) => (tokenList: List<JsonToken>) => Result<ParseUnknown<P>, string>` and `media/json`'s own `parse`/`serialize`/`stringify` signatures are byte-identical to main, so the standard codec's surface is not part of the break.
- `definedEntries` losing `Exclude<T, undefined>` is a genuine, small weakening for a caller that explicitly instantiates `T` with `undefined` in it, but no such caller exists — 28 call sites, `tsc --noEmit` clean — and the JSDoc explains why the inference position had to change.
- The deleted `fjs/media/json/todo/bigint-parse-serialize.md` is the issue this PR closes, and nothing references it.

### Not run

`nanvm-lib/` and `fjs/nanvm/` are untouched, so no Rust gate and no generator round-trip. `npm run cov` reports a vacuous `100.00` over 0 tests in this Node 23 environment on `main` too, so coverage here is unverifiable rather than good or bad.

sergey-shandar:
Thanks — and you're right about the nit, so I took it rather than leaving it.

I reproduced your two numbers against `main`'s accumulator (`e = e * 10 + digit` in a `number`): `999999999999999999999` comes back as `1.0000000000000001e21` and `99999999999999999999` as `100000000000000020000`. So "fails to tokenize" was only ever true of the *coefficient* side, where the bigint had to be built before the token existed; on the exponent side nothing failed, it was quietly wrong. The changelog entry now says that instead, and `fjs/media/json/README.md` records both halves with those examples, since "the derived field was redundant" undersells why it had to go.

Pushed as `3a12932` — changelog and README prose only, no code.

---

o2alexanderfedin (approved):
Re-reviewed at `6c83194a`. Previously approved clean at `4c7c22ce`; this pass only had to establish whether the new head is still clean.

## What moved since `4c7c22ce`

Three commits, and **no code among them**:

```
6c83194a Merge remote-tracking branch 'origin/main' into claude/bigint-parse-serialize-4lcswk
3a129325 changelog: the dropped bf lost exponent digits, it did not fail
b3544773 media: file a todo for a shared lock format (#1570)   <- from main
```

```
 changelog/unreleased/1573.md         |   5 +-
 fjs/media/json/README.md             |  12 +++-
 fjs/media/revision/README.md         |   3 +-   <- from main
 fjs/media/todo/shared-lock-format.md | 125 +++++++++++++++++++++++++++++++++++  <- from main
```

`git diff --name-only 4c7c22ce 6c83194a | grep -v '\.md$'` returns **zero files**. Not one
`.mjs` or `.ts` changed: the codec, all three tokenizers, the parser, the serializer and every
proof are byte-identical to the head that was already approved.

So it is **not strictly merge-only** — there is one authored commit, `3a129325` — but that
commit is prose only, touching the changelog entry and `fjs/media/json/README.md`. The rest of
the delta is main merged in. I checked the merge did not smuggle anything through a conflict
resolution: `git diff b3544773 6c83194a -- fjs/media/todo/ fjs/media/revision/` is empty
(main's two files were taken verbatim), and the merge commit differs from its first parent in
exactly those two paths.

Because no code moved, the evidence from the `4c7c22ce` pass carries forward unchanged and I
did not re-run it: ~3.8M extended round trips at 0 failures (exhaustive bigints ±200k, digit
boundaries to 400 digits, every power of 2 and 10 with ulp neighbours, 3M random doubles);
`isIntegral` matched an exact bigint oracle over 5,511,056 exhaustive tokens; all three
tokenizers at 0 divergences vs main over 1,680 inputs with the negative control firing on all
1,680; the standard codec at 0 divergences over 402k parse/stringify pairs; and the proofs
killing all seven meaningful mutations.

## Verified at this head

- `npm run prepack` from a freshly cleaned tree (`git clean -xdfq -e node_modules`) — exit 0,
  both passes.
- `npm test` — **2787 pass, 0 fail** at `6c83194a`, against **2724 pass, 0 fail** in the
  baseline tree on `origin/main` (`b3544773`). +63, all additions; the tree was cleaned before
  each run so no emitted `.f.js` inflated the count.
- Link check — the broken-link **sets** are identical between the two trees, 144 entries each,
  `diff` empty. The one link the merge added (`fjs/media/revision/README.md` →
  `../todo/shared-lock-format.md`) resolves, since its target came in on the same merge.
- Changelog — `changelog/unreleased/1573.md`, a single per-entry file, no links, no PR numbers,
  no headings, `**BREAKING CHANGES:**` prefix on both breaking bullets. Conforms.

I did re-verify the one thing `3a129325` newly asserts, since a factual claim about main's old
behaviour is worth checking rather than taking on trust. Driving `origin/main`'s `js/tokenizer`
directly and reading the `bf` field off the number token:

```
1e999999999999999999999 -> bf exponent 1.0000000000000001e+21
1e99999999999999999999  -> bf exponent 100000000000000020000
1e5                     -> bf exponent 5      (exact)
1e21                    -> bf exponent 21     (exact)
```

Both figures quoted in the new README paragraph are exactly right, and the short-exponent
controls come back exact, so the probe distinguishes lossy from lossless rather than always
reporting damage.

## Standing notes from the previous review

1. **Now resolved.** The earlier changelog said the dropped `bf` meant "no valid number fails to
   tokenize", which understated things — nothing failed, the exponent accumulated in a `number`
   and silently lost digits past 2^53. `3a129325` rewrites the bullet to say precisely that, and
   the README paragraph documents it with the two concrete values verified above. That was the
   substantive half of the note and it is closed.

2. **Still standing, still non-blocking.** Out-of-domain inputs to the standard codec serialize
   as `null` rather than throwing. Re-confirmed at this head on a clean tree:

   ```
   standard  undefined -> "null"      function -> "null"      1n -> "null"
   control   42        -> "42"
   extended  undefined -> "null"      1n       -> "1"
   ```

   `bigint` is genuinely in-domain for the extended codec, so `1n -> 1` is correct there; the
   note is about the standard codec quietly absorbing `undefined`, functions and `bigint`
   instead of rejecting them. Pre-existing shape, not introduced here.

## Not verified

- Coverage. `npm run cov` reports a vacuous `100.00` over 0 tests in this environment (Node
  v23), on main as well as on this branch, so there is no coverage number to report either way.
- Rust gates — `nanvm-lib/` is untouched by this PR, so they were not run.

No public signature changed relative to `4c7c22ce`, so there was nothing new to check against
the emitted `.d.mts` declarations at this head.

Verdict: still clean. The head moved for a merge plus a changelog/README correction that makes
the entry more accurate than it was when I approved it. Approving.

PR #1574, opened 2026-08-15
types/array: factor the emptiness guard out of head and tail

sergey-shandar:
Implements `fjs/types/array/todo/head-tail-null-guard.md` (deleted here).

`head` and `tail` each wrote `a.length === 0 ? null : …` by hand. Both now share one guard:

```js
const onNonEmpty = f => a => a.length === 0 ? null : f(a)

export const tail = onNonEmpty(uncheckTail)
export const head = onNonEmpty(uncheckHead)
```

## Why not the TODO's first proposal

The TODO led with deriving them from the split accessors — `tail = map(([, t]) => t)(splitFirst(a))`, `head = map(([h]) => h)(splitLast(a))` — on the premise that "`first(a) === null` / `last(a) === null` already expresses" the emptiness guard. **That premise is false**, and taking it would have silently changed behavior.

`first` and `last` go through `fromUndefined`, so a *stored* `null` or `undefined` element is indistinguishable from a missing one. Measured on `main`:

| call | today | derived from the split accessors |
| --- | --- | --- |
| `tail([undefined])` | `[]` | `null` |
| `tail([null])` | `[]` | `null` |
| `head([1, undefined])` | `[1]` | `null` |

`T` is unconstrained here, and JSON data, `List<T>` values and `Nullable<T>` element types all reach these accessors, so this is not a theoretical input.

So I took the TODO's own fallback — factor the shared shape — which keeps the length-based guard exactly as it was. The `onNonEmpty` JSDoc records why the guard is `a.length === 0` and not `first(a) === null`, since that is the trap the next reader will otherwise walk into.

## A defect this surfaced, filed rather than fixed

The premise is false because `splitFirst` / `splitLast` genuinely do conflate the two:

```js
splitFirst([null])        // null — should be [null, []]
splitLast([1, undefined]) // null — should be [[1], undefined]
```

Unlike `first`/`last`, they *can* express "present and null" — `[null, []]` is not `null` — so this is a real defect, not an unavoidable limitation of the return type. It is a behavior change in a different pair of functions, so it is filed as `fjs/types/array/todo/split-accessors-nullish-element.md` with the evidence and a proposed fix, rather than folded in here.

It also interacts with the existing `split-last-nullable-map` TODO, which proposes routing `splitLast` through `nullable.map` — the exact shape that causes this. I added a cross-reference marking the new issue as the one to read first, so that cleanup does not tidy the defect into place.

## Verification

- `npx tsc` clean.
- `fjs test`: 2724 pass, 0 fail.
- `npm run cov`: `fjs/types/array/module.f.mjs` at 100% lines/branches/functions.
- Differential check against `main` over 13 array shapes — empty, single, multi, and every nullish arrangement — `head` and `tail` outputs **identical**.

Changelog:
- `types/array`: `head` and `tail` share one emptiness guard; behavior unchanged

---


o2alexanderfedin (approved):
Verified at `c3678fc5` against `main` at `b3544773`, in two clean worktrees.

The extraction is behaviour-preserving, and the reasoning in the description for *not*
taking the TODO's headline proposal is correct — I reproduced it independently below.

## Behavioural equivalence sweep — `head` / `tail`, old vs new

The whole point of this PR is that `onNonEmpty(uncheckTail)` is the same function as
`a => a.length === 0 ? null : uncheckTail(a)`, so I ran that as a real differential rather
than reading it: the same probe file imported `fjs/types/array/module.f.mjs` from each tree
and the outputs were diffed.

**52 probes per tree**, covering the domain: `[]`, `[1]`, `[1,2]`, `[1,2,3]`, a 5-element
array, `[null]`, `[undefined]`, `[null,1]`, `[1,null]`, `[undefined,1]`, `[1,undefined]`,
`[undefined,undefined]`, nested arrays, `new Array(3)` (all holes), `[1, , 3]` (middle hole),
strings, a frozen array, plus the off-type inputs that reach `.length` anyway — `null`,
`undefined`, `'abc'`, `{length:0}`, `{length:2,0:'x',1:'y'}`, and a no-argument call.
Each probe recorded the **return value** (serialised so that `null`, `undefined` and a
sparse *hole* are three distinguishable things, with `length` and `Object.isFrozen`), and
**whether it threw** (constructor name + message), and whether the input was mutated.

```
diff out-main.txt out-pr.txt   ->   no differences (52/52 identical, both functions)
```

Also run as a **same-process** differential — both modules imported into one Node process
and compared call-by-call over 12 shapes × 2 functions — to rule out any per-process
environment effect: `probes=24 divergences=0`.

Sample of the identical rows, including the ones that matter:

```
head(empty) = ok null                     tail(empty) = ok null
head(single-null) = ok [](len=0)          tail(single-null) = ok [](len=0)
head(trailing-undefined) = ok [1](len=1)  tail(trailing-undefined) = ok [undefined](len=1)
head(sparse-3-holes) = ok [<hole>,<hole>] tail(sparse-middle-hole) = ok [<hole>,3]
head(not-an-array-null) = throw TypeError: Cannot read properties of null (reading 'length')
head(not-an-array-string) = ok "ab"       tail(not-an-array-string) = ok "bc"
head(not-an-array-arraylike-2) = throw TypeError: a.slice is not a function
```

### Negative control (both directions)

A "0 divergences" result is worthless if the harness cannot see a divergence, so I broke
each side in turn and confirmed the sweep reported it. Each mutation was applied with
`node -e` and grepped back out of the file to confirm it actually landed, then reverted with
`git checkout --`.

1. PR side, `onNonEmpty`'s guard `a.length === 0` → `a.length <= 1`:
   **6 divergent rows** (`head`/`tail` of every single-element case flipped to `null`).
2. `main` side, `tail`'s guard `a.length === 0` → `a.length === -1`:
   **2 divergent rows** (`tail([])` → `[]`, and `tail({length:0})` started throwing).

Then re-ran the unmutated sweep and got the clean result back. The harness discriminates.

### The one divergence found, and its classification

`Function.prototype.name`:

| | `main` | PR |
| --- | --- | --- |
| `head.name` | `"head"` | `""` |
| `tail.name` | `"tail"` | `""` |

Point-free definition means the arrow is created inside `onNonEmpty` and never gets a name
inferred from the `const`. Classification: **(a) equivalent** in every sense the module
contracts for. It is observable only through `.name` and stack-trace labels, not through
any return value or thrown error; `head`/`tail` have no in-repo consumers outside
`fjs/types/array/proof.f.mjs`, nothing in the repo reads `.name` off them, and the test
runner labels cases from object keys, not function names. The same module already ships
`first = at(0)` with `first.name === ""` on `main`, so this is the module's existing style
rather than something new. Noting it only because a point-free rewrite is exactly where it
shows up; not asking for a change.

Nothing else moved: arity stays 1 for both, `head !== tail`, inputs are never mutated,
results are freshly allocated per call in both trees, and `head([]) === tail([])` is `true`
in both (both `null`).

## The description's claim about the TODO's first proposal

Confirmed on `main`, not taken on trust — `tail([null])` is `[]` and `head([1, undefined])`
is `[1]` under the length guard, whereas `first([null])` is `null`, so deriving the guard
from `first`/`last` would have changed those. Taking the TODO's fallback instead was the
right call, and the `onNonEmpty` JSDoc that records why is worth having.

I also checked every claim in the newly filed
`todo/split-accessors-nullish-element.md` rather than assuming:

```
splitFirst([null])       = null    splitFirst([undefined])  = null
splitLast([1,undefined]) = null    tail([null])             = []
head([1,undefined])      = [1]
```

All five match the file. `noUncheckedIndexedAccess` is indeed off (`tsconfig.json:96`,
commented out), so the proposed `a[0]` typing note is accurate too. Filing that as a
separate P3 rather than folding a second behaviour change into a DRY extraction is the
right split, and the cross-reference added to `split-last-nullable-map.md` ("read first")
stops the other cleanup from tidying the defect into place.

## Battery

- **`npx tsc --noEmit`** — exit 0 on the PR; exit 0 on `main`.
- **`npm run prepack` from a freshly cleaned tree** (`git clean -xdfq -e node_modules`
  first) — exit 0 on the PR; exit 0 on `main`.
- **`npm test`**, each from a clean tree *before* any `prepack` ran, to avoid the
  emitted-`.f.js` double-count:
  - PR: `pass: 2724, fail: 0, total: 2724`
  - main: `pass: 2724, fail: 0, total: 2724` — exact match.
- **Link check** — `node bin/linkcheck.mjs` on both trees: `BROKEN relative markdown links: 122`
  and `LABEL/TARGET extension mismatches: 0` on each, and the full sorted outputs are
  **identical**, so the broken-link *sets* are unchanged, not merely the counts. The deleted
  `todo/head-tail-null-guard.md` leaves no dangling reference (repo-wide grep for the slug:
  no hits), and the new `split-accessors-nullish-element.md` link from
  `split-last-nullable-map.md` resolves.

## Public surface

Read from the emitted declarations directly (`bin/extract.mjs` is name-keyed and has masked
real changes before, so it was not used).

`fjs/types/array/module.f.d.mts` is **byte-identical** between the two trees:

```
diff fs-main/fjs/types/array/module.f.d.mts fs/fjs/types/array/module.f.d.mts   ->   identical
export declare const tail: <T>(a: readonly T[]) => readonly T[] | null;
export declare const head: <T>(a: readonly T[]) => readonly T[] | null;
```

Genericity survives the point-free rewrite — no collapse to `unknown`, no `any`, no
`/*elided*/`. Repo-wide, a recursive diff of the two prepacked trees shows **no `.d.mts` or
`.d.ts` differences at all** (257 `.d.mts` files each). So this is not a §8.4 breaking
change and correctly carries no `**BREAKING CHANGES:**` prefix.

## Conventions

- **Changelog** — `changelog/unreleased/1574.md`, named by the PR number, two lines, no
  link, no PR number in the body, no heading. Matches the current per-entry convention.
- **§6.2 `_` prefix** — nothing to flag: `onNonEmpty` is a module-private *constant*, not a
  type, and crosses no file boundary. No types were added.
- **§3.2 proof coverage** — the new shared helper is private and both of its arms are
  exercised, twice each: `proof.f.mjs` covers `head([1,20,300])` / `head([])` and
  `tail([1,20,300])` / `tail([])`. No new export needs a new proof.
- **§4 module header** — no new module file; `module.f.mjs`'s header is untouched and
  `@module` still survives into the emitted `.d.mts`.
- **§2 file pairing** — unaffected, no renames.

## Not verified

- **`npm run cov`.** The description reports 100% lines/branches/functions for this module.
  I did not confirm it: `npm run cov` reports a vacuous `100.00` over 0 tests in this Node
  environment, on `main` too, so the number is not measurable here either way. The proof
  cases listed above do cover both arms of the new helper, which is the substantive point.
- **CI itself** — I ran CI's gates locally, not the hosted run.
- **Rust gates** — not run; `nanvm-lib/` is untouched.

A clean DRY extraction that measurably preserves behaviour, keeps the public surface
byte-identical, and leaves the TODO tree in better shape than it found it. Approving.

o2alexanderfedin (approved):
Re-review at `10e332a3876501a7c2312b18f476dda30d28be13`. Previously approved at `c3678fc5`.

**Merge-only.** `git diff c3678fc5 10e332a3` is exactly the two files that came in with `b3544773` (media: file a todo for a shared lock format, #1570):

```
M	fjs/media/revision/README.md
A	fjs/media/todo/shared-lock-format.md
```

Both are Markdown. Nothing under `fjs/types/array/` moved, and the merge dropped nothing from the PR side, so the verification from the previous head carries forward unchanged:

- 52-probe (`head`) and 24-probe (`tail`) differential main-vs-PR, 0 divergences, negative-controlled in both directions.
- The only behavioural difference is `head.name`/`tail.name` becoming `""` from going point-free — no consumers, and `first = at(0)` already does the same thing on `main`.
- No `.d.mts` diff repo-wide, so not a §8.4 breaking change and no changelog entry needed.

Baseline for this round was `b35447738e756288e6e1da08404b407ec5a9ab7d`, which is this PR's merge-base. `origin/main` has since advanced to `6bd36947` (#1573 merged); that is not in this branch and does not touch `fjs/types/array/`.

Approving.

PR #1576, opened 2026-08-15
Bring cas (top-level) to 100% line/branch/function coverage: 96.59% → 100% branch

sergey-shandar:
## Summary

`fjs/cas/module.f.mjs` was at 96.59% branch coverage (3 uncovered branches). This brings it to 100% line/branch/function coverage:

- **`writeImpl`'s three op-failure branches** — a `writeBytes` call failing mid-stream, the per-chunk lease-renewal `rename` failing, and `publish`'s final `stat` reporting a size other than what was written — were all uncovered. Each is real only under a race (a failing disk, a concurrent writer) between two of `write`'s own steps, which the virtual harness's own in-memory filesystem semantics have no way to produce.
- Adds a `drive` helper (alongside the file's existing `casCommand` command-labeling helper) that walks `write`'s effect through synthetic op responses instead of a real filesystem: every call succeeds by default, except the one call under test, which returns an injected error. This reaches each branch directly without needing a real race. `drive` also returns the ordered command log, so a case can assert cleanup actually ran (e.g. the partial staging file's `rm`), not just that the right error tag came back.
- Four proof cases: `casWriteBytesErrorAborts`, `casWriteLeaseRenewalRenameErrorAborts`, `casWritePublishSizeMismatchErrors`, `casWritePublishStatErrorErrorsEvenWithMatchingSize` — the last two together pin both conjuncts of `publish`'s `stat`-based success check.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `node --test --experimental-test-coverage --test-coverage-include='fjs/cas/module.f.mjs' fjs/emergent_testing/all.test.mjs` → 100.00% line/branch/func
- [x] `node ./fjs/module.mjs t` → 2793 pass, 0 fail


o2alexanderfedin (approved):
Approving. I measured the title's claim and mutation-tested each of the three new cases against `origin/main` = `104bfa34` (`types/array: factor the emptiness guard out of head and tail (#1574)`).

### The claim holds

`node --test --experimental-test-coverage --test-coverage-include='fjs/cas/module.f.mjs' fjs/emergent_testing/all.test.mjs`:

| | line % | branch % | funcs % |
|---|---|---|---|
| `104bfa34` (main) | 100.00 | **96.59** | 100.00 |
| `461c1aa9` (this PR) | 100.00 | **100.00** | 100.00 |

96.59 → 100 branch, exactly as the title says, and line/funcs were already at 100. Suite: 2725 pass / 0 fail on main, 2728 pass / 0 fail here — +3, the three new cases, nothing else moved. `tsc --noEmit` and `npm run prepack` (from a clean tree) both exit 0.

`fjs/cas/module.f.mjs` is untouched, so none of this comes from deleting branches.

### The new cases are not vacuous

For each new case I built the mutant that collapses the branch it names, ran the `cas` proofs, and checked the mutant is killed here and killed *nothing* on main:

| mutant | on `461c1aa9` | on `104bfa34` |
|---|---|---|
| `if (wb[0] === 'error')` → `if (false)` | **dies** — `casWriteBytesErrorAborts` | survives |
| `rt === 'error' ? fail : loop` → always `loop` | **dies** — `casWriteLeaseRenewalRenameErrorAborts` | survives |
| `st[0] === 'ok' && st[1].size === offset` → `true` | **dies** — `casWritePublishSizeMismatchErrors` | survives |
| *(control)* publish drops its `mkdir` | dies — 9 pre-existing cases | dies — 9 cases |

The control confirms the harness can see a real behaviour change on both sides, so the three "survives on main" results are meaningful rather than a dead runner. The three pinned messages are distinct (`disk full` / `rename failed` / `publish size mismatch`), so no case is riding on a message another arm also emits.

### Two notes, neither blocking

**1. The third case doesn't test the mismatch it's named after.** `casWritePublishSizeMismatchErrors` drives `stat` to `error('boom')`, which takes the `st[0] === 'ok'` conjunct false. The `stat`-succeeds-with-the-wrong-size case — which is what "publish size mismatch" and the changelog's "a mismatched final `stat`" describe — is never produced. Both weakenings survive the whole suite:

- `st[0] === 'ok' && st[1].size === offset` → `st[0] === 'ok'` (drop the size check): **19/19 pass**
- `st[0] === 'ok' && st[1].size === offset` → `st[1].size === offset` (drop the tag check): **19/19 pass**

So the conjunction is pinned but neither conjunct is. Branch coverage is satisfied by short-circuit, which is why this doesn't show up in the number. A second override — `stat: [ok({ size: 999 })]` — would close it and would match what the test name and changelog already claim.

**2. "the partial staging file is deleted" is asserted in prose only.** Both failure cases document that `write` fails closed by `rm`-ing the partial file, but `drive` discards ops without recording them, so only the returned `[tag, value]` is checked. Gutting the cleanup —

```js
const fail = (curPath, e) => pure(error(e))   // was mapStep(rm(curPath), () => error(e))
```

— leaves 19/19 passing. Same for `publish`'s own `rm(curPath)`. A `drive` that also returned the issued command log would let these cases assert the half of their own comment that currently isn't checked.

Both are refinements to cases that already do real work, not defects in what's here. The coverage claim is accurate and the branches are genuinely pinned, so this is good to merge as is.

Changelog: `changelog/unreleased/1576.md` is a single per-entry file, no headings, no links or PR numbers — matches `1571.md`, the closest precedent (also a proof-only coverage PR, which likewise chose to carry an entry). Only its "a mismatched final `stat`" phrasing runs ahead of what note 1 describes.

sergey-shandar:
Thanks — both notes were real gaps, fixed in f31b028:

1. `casWritePublishSizeMismatchErrors` now drives `stat` to `ok({ size: 999 })` instead of an outright error, so it actually exercises the "ok but wrong size" arm. Added `casWritePublishStatErrorErrorsEvenWithMatchingSize` (a failing `stat` whose error payload's `.size` coincidentally equals the expected offset) to pin the other conjunct. Verified both `st[0] === 'ok'`-only and `st[1].size === offset`-only mutants now die.
2. `drive` now also returns the ordered command log; `casWriteBytesErrorAborts` / `casWriteLeaseRenewalRenameErrorAborts` assert the cleanup `rm` actually ran, not just the returned error tag. Verified the `fail = (curPath, e) => pure(error(e))` (cleanup-skipped) mutant now dies on both.

---

o2alexanderfedin (approved):
Re-reviewed at `f31b028a` (previously approved at `461c1aa9`). The delta is one commit — *"Pin both conjuncts of publish's size check, and cleanup rm"* — touching `fjs/cas/proof.f.mjs` only. It addresses both non-blocking notes from the last round, and I rebuilt the mutants rather than take the renamed/expanded cases at face value.

Baseline for everything below: `origin/main` at `104bfa34` (`types/array: factor the emptiness guard out of head and tail (#1574)`), which is also this PR's merge base — the branch is current.

**Both open notes are now genuinely closed.** `drive` returns `[result, log]`, and the size check is pinned from both sides: `casWritePublishSizeMismatchErrors` now drives `stat` → `ok({ size: 999 })` (a real size mismatch, matching its name at last), and the new `casWritePublishStatErrorErrorsEvenWithMatchingSize` drives `error({ size: 1 })` so a check that dropped the `st[0] === 'ok'` tag and compared only `.size` cannot slip through. I built all three mutants by hand against `fjs/cas/module.f.mjs` at this head and ran the full suite on each:

| mutant | at `f31b028a` | at `104bfa34` |
|---|---|---|
| `st[0] === 'ok' && st[1].size === offset` → `st[1].size === offset` (drop tag conjunct) | **killed** — 1 fail / 2793 | survives — 0 fail / 2789 |
| ... → `st[0] === 'ok'` (drop size conjunct) | **killed** — 1 fail / 2793 | survives — 0 fail / 2789 |
| `fail`: `mapStep(rm(curPath), …)` → `mapStep(pure(undefined), …)` (drop cleanup `rm`) | **killed** — 2 fail / 2793 | survives — 0 fail / 2789 |

Each mutant survives on `main` and dies here, so the kills come from the new cases and not from something already in the suite. The `rm` mutant takes down both `casWriteBytesErrorAborts` and `casWriteLeaseRenewalRenameErrorAborts`, which is what the added `assertEq(log[log.length - 1], 'rm', …)` is for — asserting cleanup *ran*, not merely that the right error tag came back. The mutation script errors out if the substitution does not land, so none of these are silent no-ops.

Rest of the battery at this head:

- `node ./fjs/module.mjs t` — 2793 pass, 0 fail, against 2789 on `main`: exactly the four new proof cases, no collateral.
- Coverage: `node --test --experimental-test-coverage --test-coverage-include='fjs/cas/module.f.mjs' fjs/emergent_testing/all.test.mjs` → 100.00 line / 100.00 branch / 100.00 func. The title's 96.59 → 100.00 branch claim measured true last round and still holds.
- Public surface: none — the diff is `fjs/cas/proof.f.mjs` plus `changelog/unreleased/1576.md`, no implementation change.
- `changelog/unreleased/1576.md` is a well-formed per-entry file, no headings, no links. Proof-only work does not require one, so this is a bonus rather than an obligation.

One cosmetic nit, not worth a push on its own: the PR description's test plan still says 2792, from before the fourth case was added; it is 2793 now.

Approving.

PR #1578, opened 2026-08-15
Document I/O design with effects and update related sections

sergey-shandar:
## Summary

This PR rewrites the I/O specification (§5) to document the effects-based design that is already implemented and in production use, replacing the earlier undecided alternatives. It also updates related sections to reflect this decision and removes the now-completed task from the todo list.

## Key Changes

- **Rewrote §5 (I/O)** with a comprehensive effects-based design:
  - Introduced the `Effect<O, T>` type as plain data (`Pure` thunk or `Do` node with command, payload, and continuation)
  - Documented `Operation` as a name-signature pair defining the host's vocabulary
  - Explained composition via `step` and related combinators
  - Detailed the runner architecture (asyncRun, node, mock, virtual, memory) as interpreters that dispatch commands
  - Clarified consequences for the VM: no external functions, no promises in the language, no mutable I/O state

- **Recorded the decision history** by moving the three earlier alternatives (Isolated I/O, Isolated Asynchronous I/O, State Machine) to §5.6 as superseded approaches

- **Updated §3.2** (Ownership of Mutable Objects) to clarify it is wanted for local mutability, not I/O

- **Updated §3.3** (Priority 3 features) to note that promises are needed for JavaScript interop only, not for I/O

- **Updated §8.3** (Async in Tests) to explain how effects eliminate the testing problems that section originally addressed

- **Updated `spec/todo/3380-promise.md`** to clarify that promises are not required for I/O and are wanted only for JavaScript interop and `async`/`await` sugar

- **Removed `todo/rewrite-spec-io-with-effects.md`** as the task is now complete

## Notable Details

- Added authoritative source references (types.ts, module.f.mjs, README files) that define the effect system behavior
- Included a table of runners showing their world-specific steps and use cases
- Emphasized that `Pure` is a discriminator (not a suspension) and that forcing it observes nothing due to its purity contract
- Clarified that asynchrony is a property of the runner, not the program or language

https://claude.ai/code/session_01CxF1r8nFUoK9RTWJ6WKHwA


o2alexanderfedin (approved):

Docs-only (`spec/README.md`, `spec/todo/3380-promise.md`, and the deleted `todo/rewrite-spec-io-with-effects.md`) — no code touched, so correctly no changelog entry. I re-derived every path, type and existential claim in the new §5 against the sources rather than taking them on trust.

### Every referenced path resolves

All 14 targets exist: `fjs/module.mjs`, `fjs/effects/{types.ts,module.f.mjs,module.mjs}`, `fjs/effects/eff/README.md`, `fjs/effects/node/{types.ts,module.mjs}`, `fjs/effects/node/virtual/README.md`, `fjs/effects/node/memory/module.mjs`, `fjs/effects/mock/module.f.mjs`, `fjs/effects/memory/types.ts`, `fjs/effects/memory/`, `fjs/effects/todo/io-effect-migration.md`, `spec/todo/3380-promise.md`. Broken-link sets are identical to main's (diff empty), and nothing anywhere referenced the deleted todo. The five intra-document anchors (`#5-io`, `#51-effect--the-value`, `#54-runners`, `#55-consequences-for-the-vm`, `#56-earlier-alternatives-superseded`) each match GitHub's slug for their heading, double hyphen in `#51-effect--the-value` included.

### Every type reproduced is verbatim or honestly labelled

| claim in §5 | `fjs/effects/types.ts` |
|---|---|
| `Operation = readonly[string, (..._: readonly never[]) => unknown]` | identical |
| `Effect<O, T> = Pure<T> \| Do<O, T>` | identical |
| `Pure<T> = () => T` | identical |
| `Do` has `command` / `payload` / `continuation` | identical field names |
| `Pr<O, K>` projects into `readonly[parameters, return]` | `O extends readonly[K, (...args: infer P) => infer R] ? readonly[P, R] : never` |
| `Effect` covariant in `O` | the `out O` annotations on `Cont` and `Do` |

The section labels its own simplification ("`Payload` and `Output` are written above for readability"), so the one deviation is declared rather than smuggled.

The behavioural claims hold at the implementation, not just the docs:

- "**the thunk is a discriminator, not a suspension**" — `step` and `match` both branch on `typeof e === 'function'`; there is no tag field.
- "`step` is **not lazy** … a `Pure` head is forced and `f` is called where the composition is written" — `step = (e, f) => typeof e === 'function' ? f(e()) : { ...e, continuation: x => step(e.continuation(x), f) }`. Exactly as described, including that only the `Do` case defers.
- `mapStep`, `historyStep`, `foldStep`, `forEachStep`, `okStep` all exist and are what §5.3 says they are. `okStep` is the `Result` short-circuit (`r[0] === 'error' ? pure(r) : f(r[1])`); the section wisely names its role rather than quoting a signature, since it is not a literal transcription of `okThen`'s `<T,R,E>(f) => <F>(r: Result<T,F>) => Result<R, E|F>` — `okStep` carries a single `E` with no `E|F` widening.
- `export const readFile = do_('readFile')` carries `/** @type {Func<ReadFile>} */`, matching the `// Func<ReadFile>` in the example.

### The runner table is accurate row by row

`asyncRun` is in `fjs/effects/module.mjs` and is the base of both other async runners (`node/module.mjs:211` `runNodeEffect = asyncRun({…})`, `node/memory/module.mjs:60`); `node/virtual` really is `mock` (`import { run } from '../../mock/module.f.mjs'`); and `mock`'s stated `state => effect => [state, result]` is `RunInstance` verbatim: `(state: S) => <O1, T>(effect: Effect<O1, T>) => readonly[S, T]`.

The vocabulary claim survives too, and I checked it because "subprocess" looked like an overreach — it is not: `Fs` includes `Exec = readonly['exec', (command, stdin?) => IoResult<ExecResult>]`, and the Node runner imports `node:child_process`. Filesystem (`Fs`, `Access`), network (`Fetch`, `Http`), console (`Read`, `Write`), clock (`Now`), randomness (`RandomInt`) all check out, as does `MemOp = MemCreate | MemRead | MemWrite` for the key-value slots. The `eff` characterisation is near-quotation of its own README ("under active investigation and its design is unstable", "expressible entirely in terms of `step`").

Supporting claims outside `fjs/effects`: §3.4 does list `async`/`await` as sugar; `fjs/AGENTS.md:113` does treat `throw` as a panic, and no `try`/`catch` appears anywhere in `spec/`; §8 is "Mutable Objects and Ownership Tracking" and §9 does make a function's canonical representation an FJS value, so §5.5's items 3 and 4 point where they say. The three superseded alternatives in §5.6 are faithful to the text this PR replaced.

### One imprecision

> Failures travel in the operation's return type — Node operations return `IoResult<T> = Result<T, unknown>`

`IoResult<T> = Result<T, unknown>` is exact, but "Node operations return" it is broader than `NodeOp` actually is: `Now` (`number`), `RandomInt` (`number`), `Read` (`number | null`), `Write` (`void`), `All` (`readonly T[]`), `Await` (`readonly[unknown]`), `Forever` (`never`) and `Test` (`void`) all return bare values. "Node operations that can fail return `IoResult<T>`" would keep the point — the error channel is the return type, not `throw` — without claiming a uniformity the op-set does not have.

### The todo is fully consumed

All four tasks in the deleted `todo/rewrite-spec-io-with-effects.md` are delivered: §5 rewritten around `Effect`/`Operation`/`step`/runners; §5.1–§5.3 folded in with dependency injection surviving as "the runner is the injected dependency" and §5.2's promise requirement dropped; the authoritative sources linked; and all three consistency targets it named — §3.2's ownership item, §8.3, and `3380-promise.md` — updated. Deleting it is right.

PR #1579, opened 2026-08-15
Remove custom serialization infrastructure (Serializable trait)

sergey-shandar:
## Summary

Remove the custom binary serialization infrastructure that was superseded by the decision to use the AST as the canonical representation of functions and CBOR for generic value serialization. This includes the `Serializable` trait, all its implementations, the little-endian helper trait `Le`, and related serialization/deserialization methods.

## Changes

- **Remove serialization traits and implementations:**
  - Delete `common/serializable.rs` (trait definition and primitive impls for `u8`, `u16`, `u32`, `u64`, `f64`, `bool`, `()`, `Sign`, and tuples)
  - Delete `common/le.rs` (little-endian conversion trait used only by serializers)
  - Delete per-type serializable impls: `vm/{array,bigint,function,object,string}/serializable.rs`
  - Delete `vm/impls/serializable.rs` (Unpacked serialization with tag constants)

- **Update trait bounds:**
  - Remove `Serializable` bounds from `IContainer::Header` and `IContainer::Item` in `vm/internal/icontainer.rs`
  - Remove `serialize()` and `deserialize()` default methods from `IContainer` trait
  - Remove `Serializable` bound from `naive/container.rs`

- **Update documentation:**
  - Remove `Tag` columns from spec tables in `spec/README.md` (§1, §2, §3)
  - Remove "using the tag tables above" wording from §9
  - Update `todo/remove-tags-and-rust-serializers.md` (completed task record)
  - Update `todo/159.md` to reflect removal of `Serializable` from macro dedup work
  - Update `todo/65y-nanvm-conversion-macros.md` to remove references to serialization
  - Update `todo/ast-spec.md` to clarify RTTI schema is sole specification
  - Update `todo/mvp-roadmap.md` to remove tag/shape references
  - Update `tests/README.md` to remove serialization from test scope

- **Remove test coverage:**
  - Delete `serialization()` test function from `tests/test/main.rs`
  - Remove unused imports (`Serializable`, `Iter`, `Property`, `ToArray`, `ToObject`)
  - Remove helper functions `eq_container()` and `eq_value()` used only by serialization tests

- **Update module exports:**
  - Remove `le` and `serializable` from `common/mod.rs`
  - Remove `serializable` submodule declarations from `vm/{array,bigint,function,object,string,impls}/mod.rs`

## Rationale

The custom serialization infrastructure was designed for an earlier plan to serialize values in a tagged binary format. This plan has been superseded:
- The internal AST is the stable, canonical representation of functions (as FJS values)
- Generic value serialization will use CBOR (post-MVP)
- The exact AST shape is specified by an RTTI schema, not by tag tables

The infrastructure is no longer needed and its removal simplifies the codebase.

https://claude.ai/code/session_011br3XAWGh6fCQbV3WsHf19


o2alexanderfedin (commented):

### Nothing still needs what was removed

Whole-crate greps at this head come back empty for `Serializable`/`serializable`, for `common::le`/`to_le`, and for the `TAG_` constants. The only surviving `Serializable` hits in the repo are unrelated prose — `fjs/bnf/README.md:75` ("Serializable Data Representation") and `todo/new-pl.md:247` — neither of which is the Rust trait. Nothing references the deleted `todo/remove-tags-and-rust-serializers.md`.

The bound relaxations are consistent: `IContainer::Header`/`Item` drop `Serializable` and `naive::Container`'s impl drops it from both parameters, which is a widening for implementors rather than a break. `common::sized_index::SizedIndex` is still earned by `IContainer::Items`.

### Rust gates, CI's exact invocations

With `RUSTC_WRAPPER= RUSTC=$(rustup which rustc)`, in `nanvm-lib/`:

| gate | result |
|---|---|
| `cargo fmt -- --check` | 0 |
| `cargo test` | 110 + 1 + 5 passed, 0 failed |
| `cargo clippy -- -D warnings` | 0 |
| `cargo clippy --release -- -D warnings` | 0 |

Identical test counts to `104bfa34` (110/1/5 there too), and `#[test]` count is 111 in both trees — so no test disappeared. The 61 lines out of `tests/test/main.rs` are assertions *inside* the existing `gen_test`, and `eq_container`/`eq_value` went with them because `serialization` was their only caller. That checks out.

(I did not run `cargo clippy --all-targets`; it fails on `src/vm/mod.rs:40` on main too and CI does not pass it.)

### One leftover worth a follow-up

`Iter::eq_by_` (`nanvm-lib/src/common/iter.rs:44`) is now orphaned. Its only caller in the crate was `eq_container` in the removed test block; after this PR the sole occurrence anywhere is its own definition, and it has no test. `Iter` itself stays alive — `vm/join.rs:3` still uses the trait — so this is only the one method. It survives `-D warnings` because a `pub` trait method in a lib crate isn't dead code as far as rustc is concerned, which is exactly why it is easy to miss. Same category as the rest of the PR: infrastructure whose last consumer just left. Not a blocker; deleting it (or noting it in a todo) would finish the sweep.

### Docs

The changelog's second bullet — "the language spec drops the `Tag` columns from its JSON, DJS, and FJS tables" — is exact: three tables, all three Tag columns gone, and no `Tag` string remains anywhere under `spec/`. §9's prose was updated with them ("using the tag tables above" removed) rather than left dangling.

Broken-link sets are identical to main's (diff empty), so the todo rewrites and the two deletions stranded no references. A two-bullet entry file has precedent (`1565.md`, `1573.md`). No `**BREAKING CHANGES:**` prefix looks right here: `nanvm-lib` is a `0.1.0` crate that isn't in the npm package's `files`, and no released entry has ever prefixed a `nanvm-lib` change.

o2alexanderfedin (approved):
Correcting my own slip: the review above says "Approving" but I submitted it as a comment rather than an approval. The verdict stands — all four Rust gates pass with CI's exact invocations, nothing still references the removed trait, and the only note (`Iter::eq_by_` now orphaned) is a follow-up, not a blocker. Approving at the same head.

PR #1581, opened 2026-08-15
Add vnd.fjs.lock dialect for shared lock maps

sergey-shandar:
## Summary

Introduces `vnd.fjs.lock`, a new dialect for storing lock maps as standalone content-addressed blobs that multiple revisions can reference, eliminating redundant serialization when lock bindings remain unchanged across revisions.

## Key Changes

- **New `fjs/media/lock/` module**: Implements the `vnd.fjs.lock` dialect with:
  - `lockSchema` defining the blob structure (dialect tag + lock map)
  - `validate()` and `decodeText()` for structural and semantic validation
  - `encodeText()` for canonical serialization (lexicographically sorted objects)
  - `lockDialect` registry entry for media type detection
  - Comprehensive `proof.f.mjs` test suite verifying round-tripping and canonical serialization

- **Extended `fjs/media/revision/` lock field**: Widened `lock` from inline-only maps to accept either:
  - An inline map (existing behavior, unchanged semantics)
  - A cbase32 hash referencing a `vnd.fjs.lock` blob (new)
  - Introduced `lockField` schema and `LockField` type to distinguish this position from nested hash values
  - Added validation for shared-lock references as cbase32 hashes

- **Schema reuse**: Lock map validation (`lockMapError`) and canonical serialization are imported from `fjs/media/revision`, ensuring identical serialization whether inline or shared—verified by direct byte comparison in proofs

- **Integration points**:
  - `fjs/media` detection now recognizes both dialects and reports correct media types
  - `fjs/mcp/cas` registers `lockDialect` alongside `revisionDialect` for `cas_get` reporting
  - `fjs/cas/evo` handles both inline maps and shared-lock references in `RevisionData`
  - Updated `fjs/mcp/evo` schema to reflect the widened `lock` field

- **Fail-safe versioning**: Widening the existing `lock` field rather than adding a sibling `lockRef` ensures older readers reject unknown hash values outright, preventing the silent misread of "no bindings recorded"
- **No composition**: Lock blobs reference only content hashes, never other lock blobs, keeping every position unambiguous and terminating references in one step
- **Validation boundary**: Like `snapshot`, shared-lock references are validated as cbase32 hashes only; existence and content type checking are deferred to resolvers
- **Canonical convergence**: Two revisions differing only in lock serialization (property order, whitespace) now converge to the same CAS address when using shared references

https://claude.ai/code/session_01BYxxgPByNbnsrzP8eAP5LT


o2alexanderfedin (approved):
Focused on the two things that would actually hurt here — canonicalization and read compatibility.

**Canonicalization.** `encodeText = stringify(sort)` reuses the same `sort` the revision dialect already uses — an `OrderedMap` keyed by `cmp` in UTF-16 code-unit order: total, deterministic, locale-free. Exercised it directly on numeric-looking keys (`"1"/"2"/"10"`), empty-string, accented, astral (😀), lone-surrogate and `__proto__` keys, plus `{}`: all encode stably, shuffled-key inputs converge on one string, and `decodeText(encodeText(x))` re-encodes byte-identically. `proof.f.mjs:180` pins that the map's bytes are identical inline and shared.

**Compat.** No read path was removed — `lock` went from `option(lock)` to `option(or(hash, lock))`, a pure widening. Confirmed against the built module that previously written revisions (inline map, nested map, `{}`, omitted `lock`) all still validate. `checkReferences`'s rewrite is behaviour-preserving for the absent case, and `canonicalLockField` splits on the same `typeof` boundary `lockFieldError` validates on. Registration matches `revisionDialect` exactly. Most of the 189 deletions are the resolved `shared-lock-format.md` todo.

`npx tsc` clean; `fjs t` → 2825 pass / 0 fail; `npm run cov` → 99.96/99.77/99.86 overall, with `media/lock`, `media/revision`, `cas/evo`, `mcp/evo` each at 100/100/100.

PR #1583, opened 2026-08-15
Bring mcp/cas to 100% line/branch/function coverage: 85.71% → 100% branch

sergey-shandar:
## Summary

`fjs/mcp/cas/module.f.mjs` was at 85.71% branch coverage (5 uncovered branches, no `proof.f.mjs` existed yet). This brings it to 100% line/branch/function coverage:

- **Two provably unreachable defensive guards, restructured away.** `cas_get`'s `content: true` path re-derives `type`/`mimeType` via the dialect-aware detector, then checked `fromVec`/`base64Encode` for `null` before encoding the payload. `media/type`'s own `finish()` only ever classifies a blob as `'text'` when its bit length is byte-aligned — the same condition `fromVec` checks — and every byte ever written through `cas_add`/the CAS store is whole chunks, so `base64Encode`'s only failure mode (non-byte-aligned input) can't occur either. Both null-checks are replaced with `assertNotNullish`, mirroring the identical pattern `media/module.f.mjs`'s own `detect()` already uses for the same invariant (down to the comment).
- **Three genuinely-reachable race branches**, covered by a new `proof.f.mjs`: a `writeBytes` failure during `cas_add`, and `cas_get`'s "hash vanished between reads" race on both the metadata-refine path (falls back to the streaming verdict) and the content-fetch path (returns an error — no fallback once inline content was promised). Each is real only under an actual race (a failing disk, a concurrent writer, a GC sweep) between two of the tool's own steps. `proof.f.mjs` reaches them the same way `fjs/cas/proof.f.mjs` does (from PR #1576): a synthetic effect driver that succeeds by default except for the one call under test.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `node --test --experimental-test-coverage --test-coverage-include='fjs/mcp/cas/module.f.mjs' fjs/emergent_testing/all.test.mjs` → 100.00% line/branch/func
- [x] `node ./fjs/module.mjs t` → 2800 pass, 0 fail

---

### DIFF
--- a/changelog/unreleased/1583.md
+++ b/changelog/unreleased/1583.md
@@ -1,0 +1,4 @@
+- `mcp/cas`: `cas_get`'s unreachable `fromVec`/`base64Encode` null-checks
+  become `assertNotNullish`; adds proof coverage for `cas_add`'s
+  `writeBytes`-failure race and `cas_get`'s hash-vanished-between-reads
+  races, bringing the module to 100% line/branch/function coverage
--- a/fjs/mcp/cas/module.f.mjs
+++ b/fjs/mcp/cas/module.f.mjs
@@ -134,6 +134,7 @@ import { identity } from '../../types/function/module.f.mjs'
 import { sha256 } from '../../crypto/sha2/module.f.mjs'
 import { nonEmpty, empty as elEmpty } from '../../effects/list/module.f.mjs'
 import { syncRevision } from '../../cas/evo/module.f.mjs'
+import { assertNotNullish } from '../../asserts/module.f.mjs'
 
 // ── Argument schemas (declared once, used for both inputSchema and validate) ─────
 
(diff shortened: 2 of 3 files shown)


o2alexanderfedin (approved):
Checked both guard removals independently rather than taking the comments' word for it — the unreachability arguments hold.

**`fromVec`** (`module.f.mjs:280`): the guard sits behind `refined.type === 'text'`, and the dialect layer in `fjs/media` only ever rewrites `mime_type`, so `type` can only come from `detectVec`. `finish` emits `'text'` only under `utf8Text(s.utf8) && (s.length & 0b111n) === 0n`, and `push` short-circuits only once the verdict is already `base64` — so a `'text'` verdict implies a complete byte fold, which is exactly what `fromVec` re-checks with the same decoder. No dialect branch or other caller reaches the encode path with an unclassified blob.

**`base64Encode`** (`module.f.mjs:289`): `value` is only ever `collectRead(c.read(key))`, built from `Buffer` chunks in the Node handler, so bit length is always a multiple of 8 and `encode`'s sole null condition (`len % 8n !== 0n`) can't fire. `cas_add` likewise only stores byte-aligned Vecs.

Measured, not assumed: `npx tsc` clean, `fjs test` 2800/2800, and a scoped coverage run over `fjs/mcp/cas/module.f.mjs` reports 100.00 line / 100.00 branch / 100.00 funcs — matching the claim. New `proof.f.mjs` follows the co-located convention and reuses the `drive` technique from `fjs/cas/proof.f.mjs`.

PR #1584, opened 2026-08-15
Add proof object with tagToToken branch coverage test

sergey-shandar:
## Summary
Added a `proof` export to the tokenizer module containing a test case that exercises an unreachable code path in the `tagToToken` function under normal grammar operations.

## Key Changes
- Imported `assertEq` from the asserts module for test assertions
- Added a `proof` object with a `tagToTokenTrueTag` test function that directly calls `tagToToken(true)` to verify it returns `'true'`
- Included detailed comments explaining why this branch is unreachable through normal grammar paths: the `true` tag is reserved by `bnf/data`'s `emptyTagOf` for anonymous nullable sequence rules, which don't exist in the current grammar

## Implementation Details
The test directly invokes `tagToToken` with a literal `true` value to cover a branch that cannot be reached through `tokenizeJs` with the current grammar definition. This ensures comprehensive branch coverage while documenting the architectural constraint that prevents this code path from being exercised during normal tokenization.

https://claude.ai/code/session_01CRmjcUK8N2g2XDm9oxdmfE


o2alexanderfedin (commented):
The proof itself is good — matches the whitebox `export const proof` pattern from `fjs/js/tokenizer/module.f.mjs`, uses `assertEq`, and does hit the branch. Full suite green (2798 pass / 0 fail).

One thing: the justification comment is factually wrong. `emptyTagOf` gives `true` to *any* nullable sequence rule — anonymity isn't the criterion — and `jsGrammar` does contain them. `emptyTagMap(toData(jsGrammar())[0])` yields `{"9": true, "14": true}` (rule `9` is `[]`, `bnf`'s shared `none` from `option`/`repeat0Plus`). So "no rule in `buildToken`/`jsGrammar` is both anonymous and nullable" doesn't hold.

The conclusion survives on a stronger reason, though: `fjs/bnf/descent/module.f.mjs:225` reads `emptyTags[name]` only in the variant arm, and `emptyTagOf` returns a *string* key for variants — every other tag is a variant key or `undefined`. A `true` tag therefore can't enter an AST via `descentParser` from *any* grammar, not just this one. Worth rewording to that, since it's the claim that actually holds.

sergey-shandar:
Re [the review](https://github.com/functionalscript/functionalscript/pull/1584#pullrequestreview-4944271090): good catch — verified independently and you're right on both counts: `emptyTagOf` keys off nullability alone (any nullable sequence, named or not), and `jsGrammar` does contain such rules via `none`/`option`. Reworded the justification to the reason that actually holds: `descentParser` only ever reads `emptyTags[name]` in the variant arm, wrapped in `mrFail` (success:false), and a failed variant only becomes the propagated `result` when nothing in it matches — so it never survives into a successful AST. That holds for any grammar via `descentParser`, not just `jsGrammar`. Pushed in db736a4.

---

o2alexanderfedin (approved):
The rewritten justification holds up claim by claim. `emptyTagOf` (`fjs/bnf/data/module.f.mjs:28-43`) is the only producer of a literal-`true` tag, and `descentParser` reads `emptyTags` in exactly one place — the variant arm at `fjs/bnf/descent/module.f.mjs:225-226` — always wrapped in `mrFail`. Traced the propagation: a failing child is discarded by the seq frame (`:252-253` rebuilds `mrFail(frame.tag, [], startPos)`) and by the enclosing variant frame (`:268-276` keeps its own `emptyResult`), so a `true` tag can only reach the top-level return with `success: false`, and `tokenizeJs` returns the error token before `getTokensFromAstRule`. The earlier "anonymous and nullable" claim is gone.

Proof body unchanged and still hits the branch. `node ./fjs/module.mjs t` → 2829 pass / 0 fail, `proof.tagToTokenTrueTag(): ok`.

PR #1586, opened 2026-08-15
Audit: inline `/** @type {T} */ (v)` casts

sergey-shandar:
This PR documents a comprehensive audit of inline JSDoc type casts (`/** @type {T} */ (expr)`) across the `fjs/` codebase, as called for in AGENTS.md.

## Summary

A mechanical enumeration and analysis of all 357 inline `@type` casts (excluding 221 `@type {const}` casts which are out of scope) was performed against TypeScript 7.0.2 to determine which casts are:
- **Redundant** and can be deleted outright
- **Checking casts** that should use `@satisfies` or annotated declarations instead
- **Assertable** via runtime checks (`assertNotNullish`, `assert(typeof ...)`, etc.)
- **Load-bearing** and must remain due to generic erasure, nominal branding, or genuine type/API mismatches

## Key Findings

| Category | Count | Action |
|----------|-------|--------|
| **Remove** | 181 | Redundant casts; delete with no replacement |
| **@satisfies** | 21 | Expression is assignable; use `@satisfies` to check instead of override |
| **Declare** | 6 | Hoist to annotated `const` declarations |
| **Assert** | 72 | Replace with runtime checks (`assertNotNullish`, `assert(...)`, etc.) |
| **Keep** | 77 | Load-bearing: `any` bridges, generic erasure, nominal branding, TS2589 depth limits, or genuine API mismatches |

**Result:** 208 of 357 casts (58%) need no cast and no replacement machinery—181 simply delete, 27 become checks instead of overrides. Another 72 have real runtime checks available. Only 77 are genuinely necessary.

## Notable Sub-findings

- **`fjs/js/tokenizer/module.f.mjs`**: 37 of 39 casts are redundant; the `_CreateToToken<…>` casts are already contextually typed through `create(def)(…)`.
- **`Array.isArray` narrowing issue**: Eight `Dir` casts sit after guards using `Array.isArray`, which never narrows `readonly T[]` out of a union. Swapping to `instanceof Array` makes the existing checks narrow correctly and deletes all eight casts.
- **MCP/CAS proof `unknown` values**: 40 of the 72 `assert` candidates are unchecked claims about JSON-RPC response shapes. A small set of `assert`-based accessors or rtti `validate` would replace all of them.
- **Literal-range casts**: `fjs/asn.1/module.f.mjs:68` and `fjs/types/function/compare/module.f.mjs:12,17` narrow arithmetic results to literal unions; `assert(i === 0 || i === 1 || i === 2)` narrows identically and checks.

## Proposed Implementation Order

1. Delete the 172 redundant casts outside the two rtti visitors; then take the remaining 9 one at a time (to avoid TS2589 depth limit).
2. Swap `Array.isArray` → `instanceof Array` at the eight `Dir` sites and delete those casts.
3. Convert the 21 + 6 checking casts to `@satisfies` / annotated declarations.
4. Introduce `assert`-based accessors for the MCP/CAS proofs and convert the 72 `assert` candidates.
5. For the remaining 77, open follow-up issues per type/API problem rather than rewriting the cast.

## Deliverable

This PR adds `todo/inline-type-casts.md`, a detailed audit document with:
- Problem statement and methodology
- Summary findings table
- Sub-findings worth acting on first
- Explanation of what genuinely must stay
- Full table of all 357 sites with verdict and replacement guidance per site

https://claude.ai/code/session_01TQcZKuWSt2rEZrCj1jVZih


o2alexanderfedin (approved):
Re-derived the audit mechanically rather than sampling it, and it holds.

- **Table integrity**: a balanced-brace scan of `fjs/` finds exactly 357 non-`const` inline `@type` casts and exactly 221 `@type {const}` casts — both cited figures. All 357 rows match a real cast by file, line *and* type text (no unmatched rows, no orphans). Verdict tallies 181/21/6/72/77 sum to 357 and agree with the summary table and the 208/58% claim.
- **"remove" bucket proven**: deleting all 172 redundant casts outside the two rtti visitors leaves `npx tsc` (7.0.2) clean; the 9 rtti sites are each individually redundant and the group removal trips exactly `TS2589`, as documented.
- **Sub-findings proven**: all 37 tokenizer casts delete clean; switching the 8 `Dir` guards to `instanceof Array` lets all 8 casts delete with `tsc` green; 22+10+8 matches the claimed 40-of-72 MCP/CAS split; exactly one pre-existing `@satisfies`.
- **Spot-checks**: ~20 sites probed both ways — `@satisfies` candidates pass as `@satisfies` but fail deletion; `assert`/`keep` sites fail both with the documented codes (TS7022 at `descent:199`, TS2589 at `protocol/mcp:167`, TS2322 for the literal-range sites).
- **`tsconfig-strict-flags.md`**: every measured number reproduces exactly (0/0/0/0, 8, 31, 202, 209 = 130 TS6196 + 79 TS6133). `@template` count 169 exact; template-literal counts exact; `deno.json` fmt settings quoted correctly.

`npx tsc` clean; `npm test` → 2797 pass / 0 fail. All relative links in the 9 changed docs resolve, and the `spec/README.md` renumbering is consistent with the `NNNN-slug.md` convention.

PR #1587, opened 2026-08-15
Bring sul/module.f.mjs to 100% line/branch/function coverage: 93.75% → 100% branch

sergey-shandar:
## Summary

`fjs/sul/module.f.mjs` was at 93.75% branch coverage (1 uncovered branch). This brings it to 100% line/branch/function coverage:

- **`cascade`'s `for (let index = 0; ; index++)` loop** carried a phantom "loop falls through" branch: V8's coverage instrumentation allocates a branch for the loop's (nonexistent) condition, but the loop only ever exits through one of its two internal `return`s — never by falling off the end — so that branch can never be marked taken by any input, no matter how the loop's logic is exercised.
- Rewrites `cascade` as a recursive `cascadeFrom`, which has no such branch to begin with — same logic, same two exit points, now expressed as two `return` statements in a plain function rather than a loop.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `node --test --experimental-test-coverage --test-coverage-include='fjs/sul/module.f.mjs' fjs/emergent_testing/all.test.mjs` → 100.00% line/branch/func
- [x] `node ./fjs/module.mjs t` → 2828 pass, 0 fail

---

### DIFF
--- a/changelog/unreleased/1587.md
+++ b/changelog/unreleased/1587.md
@@ -1,0 +1,4 @@
+- `sul`: rewrites `cascade`'s `for(;;)` loop as a recursive `cascadeFrom`,
+  removing a phantom loop-exit branch V8's coverage instrumentation could
+  never mark taken, bringing the module to 100% line/branch/function
+  coverage
--- a/fjs/sul/module.f.mjs
+++ b/fjs/sul/module.f.mjs
@@ -32,21 +32,26 @@ export const encode =
 
         /** @typedef {readonly [Id | undefined, S, readonly _HashState[]]} _CascadeResult */
 
-        /** @type {(id0: Id, storage0: S, stacks0: readonly _HashState[]) => _CascadeResult} */
-        const cascade = (id0, storage0, stacks0) => {
-            let id = id0, storage = storage0, stacks = stacks0
-            for (let index = 0; ; index++) {
-                if (index >= stacks.length) {
-                    const [, [newStorage, newStack]] = step(id, [storage, []])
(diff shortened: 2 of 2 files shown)


o2alexanderfedin (approved):
Equivalence and stack depth both check out.

**Equivalence** — two differential harnesses against the pre-PR implementation, zero mismatches in each:
- Isolated `cascade` vs `cascadeFrom` with a mock `step` (full control over `out`/storage/stack): 5,728 cases, exhaustive over `stacks.length` 0–12 × exit-after-1–14 steps × 4 start ids, plus 5,000 random cases up to length 60. Both exit points, empty and single-element `stacks` all agree.
- Full encoder, old module vs new, comparing final id *and* the complete `add` merge log: every bit string of length 0–14 (32,767 cases) plus random and all-zero/all-one inputs up to 100,000 bits.

`stacks.length` is invariant under the splice-replace, so re-reading it per iteration vs per call can't diverge.

**Depth** — logarithmic, not a concern. Depth is `stacks.length + 1`, and each hash level consumes ≥2 symbols per emitted symbol: 1e3 bits → 7 levels, 1e4 → 11, 1e5 → 14, 1e6 → 17, 4e6 → 19. That's +1 per doubling, so nothing physically realizable gets near the ~1,000-frame limit I measured. Tail position, but no TCO needed.

**Coverage** — confirmed both sides: `origin/main` 100.00 line / 93.75 branch / 100.00 funcs; PR head 100.00 / 100.00 / 100.00.

`npx tsc` clean; `node ./fjs/module.mjs t` → 2828 pass / 0 fail.

PR #1588, opened 2026-08-15
fsm: key DFA states without the JSON serializer

sergey-shandar:
Implements `fjs/fsm/todo/sorted-set-key.md` (deleted here).

## Why

The subset construction needed a canonical key for a `SortedSet<string>` and reached for a media-format serializer:

```js
import { stringify } from '../media/json/module.f.mjs'
const stringifyIdentity = stringify(identity)
```

`fjs/fsm` is generic automaton tooling; `fjs/media/json` pulls in its tokenizer, which pulls in the 795-line JavaScript lexer. Building a DFA loaded the whole JS lexer and paid for full JSON string escaping on every state-set key.

## What

`toKey` is added to `fjs/types/sorted_set` — where the data lives — and `fjs/fsm` now imports nothing from `fjs/media`. The `s in dfa` lookup the TODO also flagged is now `at(s)(dfa) !== null`, which is an own-property read rather than a prototype-chain walk.

## The TODO's proposed encoding is not canonical

It says to implement the key "as a `join` over a separator". **That silently breaks the subset construction**, and I checked before writing it:

```
['a,b']    -> "a,b"
['a','b']  -> "a,b"     collision: true
```

State names in `_Rule` are arbitrary strings, so two distinct state sets would share one DFA entry — a wrong automaton, produced silently. JSON escaping is exactly what made the old key injective, so dropping to a bare separator would trade a layering problem for a correctness one.

`toKey` is length-prefixed instead — `3:abc` per element, concatenated:

```js
export const toKey = set => set.map(lengthPrefixed).join('')
```

That is readable back (digits, `:`, then exactly that many code units), so it is injective with no escaping table, and it skips the escaping pass the TODO objects to paying for. The proof pins the property directly, including the comma, colon and newline cases a separator join would collapse, plus `['ab']` vs `['a']` and `['']` vs `[]`.

## Behaviour change

DFA keys — which `run` returns — change from `'["id"]'` to `'2:id'`. I checked the blast radius: **`fjs/fsm` has no runtime importer anywhere in the repo**; the only reference is a doc comment in `types/function/operator/types.ts`. Still a public change, so the changelog entry is prefixed.

The fsm proof no longer hard-codes either encoding: expected states are written as the sets they are and keyed through `toKey`, so it asserts the automaton rather than a string format. That also let it drop `stringify`/`sort`/`fromEntries` in favour of `assertStructurallySame`.

## Verification

- `npx tsc` clean; `fjs test` 2837 pass, 0 fail.
- `npm run cov`: `fjs/fsm/module.f.mjs` and `fjs/types/sorted_set/module.f.mjs` both 100% lines/branches/functions.
- **Encoding-independent differential against `main`.** A key change can hide an automaton change, so I fingerprinted the DFA by BFS from the initial state, numbering states by discovery order — state count, full transition graph, and `run` traces for `a1`, `0.1`, `//`, `::`, `x9_`, `1.2.3`. **Identical**, 195 lines of output.

Changelog:
- **BREAKING CHANGES:** `fsm`: DFA state keys are `sorted_set`'s canonical `toKey` rather than JSON text, so `run` returns different strings for the same automaton. `fjs/fsm` no longer depends on `fjs/media/json`
- `types/sorted_set`: new `toKey` builds a canonical, collision-free key for a set of strings

---


o2alexanderfedin (approved):
Checked the two things that could silently corrupt a DFA here.

**Key injectivity.** `toKey` is a netstring encoding (`${s.length}:${s}` joined), which is uniquely decodable — from any position the first `:` terminates the digit run and exactly that many UTF-16 code units are the element. Brute-forced over an adversarial alphabet (`''`, `':'`, `'0:'`, `'3:abc'`, `'12:x'`, `':::'`, lone surrogates, `'😀'`, `'__proto__'`): 1562 sorted sets and 9724 arbitrary sequences up to length 3, zero collisions. The obvious traps separate cleanly — `['3:abc']` → `5:3:abc` vs `['abc']` → `3:abc`; `['']` → `0:` vs `[]` → `''`. `.length` and slicing agree on code units, so astral elements are fine.

**`in` → `at`.** Safe, and the old check was the riskier one: `'__proto__' in {}`, `'constructor' in {}` and `'toString' in {}` are all `true`, while `at` (`getOwnPropertyDescriptor` + `fromUndefined`) returns `null` for each. `_Dfa` values are always `toArray(...)` arrays, never null, so the `!== null` test can't misfire. A DFA with states literally named `__proto__`/`constructor`/`toString` builds all 5 states and runs correctly.

The `fjs/media` import is gone from `fjs/fsm` entirely, and the deleted todo has no remaining references. `npx tsc` clean; `npm test` → 2837 pass / 0 fail.

PR #1590, opened 2026-08-15
Per-API issues for the casts the cleanup deliberately left

sergey-shandar:
Stacked on #1589 — its last commit is the diff to review here.

`todo/inline-type-casts.md` ends with 36 sites where the cast overrides a real type mismatch, and says each needs an issue against the API it papers over rather than a different cast syntax. This opens those issues for the 20 that share a cause, and records that the remaining one-offs do not.

| New issue | Sites | What it is |
| --- | --: | --- |
| `fjs/effects/todo/do-generic-operation-signatures.md` | 3 | `Func<O>` has nowhere to put a type parameter, so `memCreate`, `memRead` and `all` are cast while their non-generic neighbours in the same file are annotated declarations |
| `fjs/effects/todo/step-continuation-operation-union.md` | 6 | `step`/`okStep` continuations cast their way to the operation union the caller wants, which hides any operation the runner cannot interpret |
| `fjs/effects/node/todo/async-operation-map-assignability.md` | 2 (+2 nearby) | `memoryOperationMap()` is not assignable to `ToAsyncOperationMap<O>` |
| `fjs/js/todo/token-kind-narrowing.md` | 4 | keyword/operator collections typed over `string`, so `has` cannot narrow to a `JsToken` kind |
| `fjs/types/btree/todo/find-path-item-typing.md` | 3 | every tuple `btree/find` builds or indexes |
| `fjs/emergent_testing/todo/mockrun-parameters-inference.md` | 2 | `Parameters<typeof mockRun<…>>[0]` as an argument type |

Two of these are worth reading past the summary. The `ToAsyncOperationMap` cast is the exact hazard `fjs/AGENTS.md` describes — a cast around a value handed to a generic parameter stops each handler being checked against `O`, so a drifted shape is absorbed rather than reported. The `mockRun` one has the same shape for the same reason.

The `token-kind-narrowing` issue also argues against the obvious fix: a type predicate would remove those four casts and AGENTS.md permits one where the alternative is a cast — but only where the predicate body *is* the structural check defining membership. A `Set` lookup asserting a union is not that, so the issue proposes typing the collection instead.

The audit doc gains a table linking all six, and states plainly that the remaining one-offs — a deliberately wrong value in a negative test, a nominal brand, `Function` → `NodeProgram` — have no shared cause and are listed in the site table rather than given an issue each.

https://claude.ai/code/session_01TQcZKuWSt2rEZrCj1jVZih

---
_Generated by [Claude Code](https://claude.ai/code/session_01TQcZKuWSt2rEZrCj1jVZih)_


o2alexanderfedin (commented):
Reviewed just `7338ce8` (the rest is #1589). Sites and arithmetic hold: 19 of the 20 cited casts exist verbatim at the cited `file:line`, the six per-issue counts sum to exactly 20, the audit table has exactly 36 "cast overrides" rows against 74 total matching the reason summary, nothing double-counted, all relative links resolve. Causes check out against source — `Func<O>` genuinely has no type-parameter slot and its neighbours really are annotated declarations; `step`'s signature matches the quote verbatim; the `ToAsyncOperationMap<O>` passage says what's claimed. `npm test` → 2828 pass / 0 fail.

One inaccuracy, in `fjs/js/todo/token-kind-narrowing.md`: the stated cause — the collections holding those strings are typed over `string` — holds for the two `djs` sites (`keywordSet` is `ReadonlySet<string>`), but not for `fjs/js/tokenizer/module.f.mjs:262`. There `keywords` is `/** @type {const} */`, so `kind` already carries the literal union; the cast is needed because `_KeywordToken` excludes `true`/`false`/`null`/`undefined` as separate variants, so `{ kind }` won't distribute over `JsToken`. Retyping the collection won't remove that cast, so "and the four casts go" overreaches.

Minor: `fjs/cas/evo/module.f.mjs:456` is really 466 (inherited from #1589's table, not introduced here).

o2alexanderfedin (commented):
The `fjs/cas/evo/module.f.mjs` 456 → 466 reference is fixed (came in via the base branch).

The `token-kind-narrowing.md` point is still open — the file is byte-identical to 7338ce8; none of the new commits touch it. Re-proved it empirically this time: deleting the cast at `fjs/js/tokenizer/module.f.mjs:262` gives `TS2322` whose message shows `kind` is already the 49-member literal union, not `string`, and fails to distribute over `_KeywordToken | _TrueToken | _FalseToken | _NullToken | _UndefinedToken`. So retyping the collection won't remove that cast, and "the four casts go" still overreaches for that site.

Everything else in this PR's own content is unchanged and still correct: six new todos intact, links resolve, per-issue table still sums to 20. `npm test` → tsc clean, 2838 pass / 0 fail.

PR #1593, opened 2026-08-15
text/utf8: take the BMP and plane bounds from code_point

sergey-shandar:
Implements `fjs/text/utf8/todo/encoder-code-point-bounds.md` (deleted here).

## Why

`fjs/text/code_point` states its own contract — "every predicate below is derived from these constants so the surrogate bounds and the maximum appear exactly once" — and the UTF-16 encoder honours it by dispatching through `isBmpCodePoint` / `isSupplementaryPlane`. The UTF-8 encoder imported nothing but `errorMask` and re-spelled both boundaries:

```js
if (input >= 0x0800 && input <= 0xffff) { … }
if (input >= 0x10000 && input <= 0x10ffff) { … }
```

## What

```js
if (input >= 0x0800 && input <= bmpMax) { … }
if (isSupplementaryPlane(input)) { … }
```

The 4-byte guard needed no new export at all: `isSupplementaryPlane` is already `contains(bmpMax + 1, maxCodePoint)`, which is exactly `>= 0x10000 && <= 0x10ffff`. The 3-byte guard's upper bound is `bmpMax`, so that one constant becomes exported.

The UTF-8-specific thresholds `0x7f` and `0x7ff` stay inline, as the TODO says — those are encoding-length boundaries, not code-point-domain ones.

## One deviation

The TODO asked to export **both** `bmpMax` and `maxCodePoint`. `maxCodePoint` turned out to have no caller once the 4-byte guard goes through `isSupplementaryPlane`, so exporting it would add public API with nothing behind it. It stays private, and `bmpMax`'s JSDoc records why the asymmetry is deliberate: the encoder needs the BMP *boundary* itself to split its 3-byte form, whereas the predicates already answer every question anyone has had about the maximum.

## Verification

- `npx tsc` clean; `fjs test` 2848 pass, 0 fail.
- `npm run cov`: `fjs/text/code_point/module.f.mjs` and `fjs/text/utf8/module.f.mjs` both 100% lines/branches/functions.
- New proof pins the boundary against the predicates derived from it — `isBmpCodePoint(bmpMax)`, `isSupplementaryPlane(bmpMax)`, and both at `bmpMax + 1` — so the split the encoder relies on cannot drift.
- **Exhaustive differential against `main`.** Guard rewrites are exactly where an off-by-one hides, so I encoded **every** input in the function's domain and hashed the output: all 1 114 369 code points from `0` through `0x110100` (past the maximum, to include the falls-through region), all 131 072 error-tagged values sweeping the low bits the later branches test, and the negatives, huge, fractional and non-finite inputs that reach the final `[errorMask]`. **1 245 448 encodings, identical hash.**

Changelog:
- `text/code_point`: `bmpMax` is exported, and `text/utf8`'s encoder takes its BMP and supplementary-plane guards from `code_point` instead of re-spelling the boundaries

---


o2alexanderfedin (approved):
The equivalence is exact, not just close. `fjs/types/range/module.f.mjs:10` defines `contains = (b, e) => i => b <= i && i <= e` — inclusive both ends — and `code_point/module.f.mjs` gives `bmpMax = 0xffff`, `maxCodePoint = 0x10_ffff`, `isSupplementaryPlane = contains(bmpMax + 1, maxCodePoint)`, so both rewritten guards are textually the same predicate as the inline ones.

Confirmed empirically rather than by reading: a differential harness ran `fromCodePointList` from this head and from `origin/main` over 1,376,539 inputs — every value `0..0x110100`, all 131,072 error-tagged values in signed and unsigned form, negatives, `0x110000`, `0x7fffffff`, fractional (`0xffff + 0.5`), `NaN`, `±Infinity`, and the whole surrogate block. Zero byte differences; out-of-range, surrogate and non-integer input still fall through to the same error branches.

The deviation is sound too — `maxCodePoint` has no caller outside `code_point` itself, so the maximum still appears exactly once as a value and the module's "spelled out exactly once" contract is intact rather than weakened.

Deleted todo is the one implemented, nothing references it. `npx tsc` clean; `npm test` → 2848 pass / 0 fail. Changelog entry conforms.

PR #1594, opened 2026-08-15
Improve code coverage

sergey-shandar:


o2alexanderfedin (commented):
Since the description is empty, for the record this deletes two `throw 'unreachable'` guards in `nodeSet`'s `i===2` and `i===3` arms — the only uncovered lines in the file — replacing the `i===3` one with `assert(xL === 2 || xL === 5)` so TSGO still narrows. `package.json` drops the hardcoded test path from `cov`; `deno.json` adds a `cov-html` task. No version bump, no coverage exclusions, no check weakened.

The assert is provably safe: `First<T>` (`fjs/types/btree/find/types.ts:11-19`) permits `i===3` only from `_FirstLeaf2` (len 2) or `_FirstBranch5` (len 5), and `find/module.f.mjs:37-49` agrees — `done(3)` is reachable only at `node.length` 2 or 5. No reachable input throws.

Behavior verified by differential harness against `origin/main`: 51,657 tree comparisons (empty and 1..12 ascending+descending, 300 random 1–300-op mixes over a 60-key space with heavy duplicate-insert/replace, 1000-key monotone both directions, triple re-insert, string keys) — 0 mismatches. Coverage of `fjs/types/btree/set/module.f.mjs`: main 93.97 line / 94.87 branch → head 100/100/100. Old vs new `cov` command: identical 2793 tests and totals. `npx tsc` clean; suite 2857/2857.

One convention item: AGENTS.md §5 requires `changelog/unreleased/1594.md` for a code change and it's absent — every peer coverage PR (#1596, #1597, #1587, #1583, #1576) has one. The empty description also omits the `Changelog:` section, and `ok`/`xL` aren't really commit subjects.

Minor: `deno.json:6` anchors `module\.f\.mjs$` while `cov` on line 5 doesn't — harmless, but they now differ.

sergey-shandar:
> One convention item: AGENTS.md §5 requires `changelog/unreleased/1594.md` for a code change and it's absent — every peer coverage PR (#1596, #1597, #1587, #1583, #1576) has one. The empty description also omits the `Changelog:` section, and `ok`/`xL` aren't really commit subjects.

The behavior is the same!!! Changelog is not required!

PR #1595, opened 2026-08-15
bnf: a repeat rule in the data form, flat AST in descent

sergey-shandar:
Implements `fjs/bnf/todo/667-bnf-repeat-flatten.md`, which is deleted here.

Unbounded repetition had no primitive in the data form: `repeat0Plus(x)` expands
to a right-recursive variant, so nothing in a `RuleSet` said a rule was a *list*,
and a parser building an AST from one emitted the cons-shaped chain that spelling
implies rather than a sequence of items.

## `bnf/data`: deriving the rule

`toData` now recognizes the shape and records it:

```text
R = { some: S, none: E }   S = [item, R]   E = []      ⟶   R = item
```

A `Repeat` is the repeated rule's **name**. That keeps the four rule kinds
disjoint by JavaScript type alone — number → `TerminalRange`, array →
`Sequence`, object → `Variant`, string → `Repeat` — so telling one kind from
another never probes a shape. A first draft wrapped it as `{ repeat: [name] }`,
which shared shape space with `Variant` and needed a field probe to separate
them; a string is free to take here because nothing else in a data `RuleSet` is
one. (The *functional* `DataRule` does have a string case — a Unicode literal —
but `toData` expands it to terminals long before the data form exists, and
`unicode-rules.md` removes it there.)

Recognition is structural, not a `repeat0Plus` special case: `classic()`'s
hand-written `characters` folds identically, which the proofs check. It is also
deliberately narrow — only the unambiguous 0-or-more case:

- exactly two branches, one the empty rule and the other `[item, R]`;
- `item` must not lead back to `R`, so the tail is `R`'s only self-reference;
- `item` must not match empty, since a body that can consume nothing gives the
  same input infinitely many parses.

Everything looser keeps its variant. A 1-or-more chain, a separated list
(`[item, separator, R]`) and an operator-style tree all reach `R` again, but
which of their items are the list's elements is not something the shape says —
naming that needs the schema/action mechanism of #207, not a structural fold.
There is no `min` parameter for the same reason: `repeat` means 0-or-more,
exactly the case the shape settles on its own. Each rejection has its own proof.

Folding a rule leaves its recursive branch — and usually its empty one —
unreachable, so `toData` prunes rules the entry can no longer reach, keeping the
invariant that a generated set holds no dead grammar.

## `bnf/descent`: matching it flat

A `Repeat` is matched by iterating its item, so a repetition is one AST node
holding a flat sequence of the items it matched:

```text
[{tag: ' ', …}, {tag: ' ', …}]
```

rather than the nested `some`/`none` chain. Per-item tags survive — each item
keeps the node it always had — and what disappears is the encoding's scaffolding
plus the trailing `none` node. A round that fails ends the repetition instead of
failing it, rewinding to where that round began.

Rounds accumulate as a list and become an array once, at the end. Appending to an
array per round is the obvious spelling and is quadratic in the items matched;
tokenizing 90k chars of DJS measured 2222 ms against 2375 ms on `main`, so the
flat shape costs nothing.

The matcher also gained `descentParserRuleSet`, mirroring `ll1`'s
`parserRuleSet`. `descentParser(fr)` is unchanged — it is now a one-line wrapper
over it — and the new export is what lets a hand-written rule set exercise the
zero-progress guard that `toData` can never produce.

## `bnf/ll1`: still nested, and why

`dispatchMap` compiles a `Repeat` back into the right-recursive chain it was
folded from, so this backend's AST keeps its shape. That is a dispatch-model
limit, not a missing branch: a dispatch entry *consumes* the symbol it dispatched
on and continues with a chain of rule names, so a nullable item's first set is
inlined into whatever encloses it. A repetition leading a sequence —
`[ws, value, ws]`, the shape of every grammar in `testlib.f.mjs` — has its first
set merged into that sequence's entries, and a looping frame is unreachable from
there.

Making only the non-inlined references iterative would be worse than leaving it:
the same grammar would produce a flat node in one position and a chain in
another. `fjs/bnf/todo/ll1-repeat-flatten.md` tracks the dispatch-model change,
pointing at `new-parser.md` where that decision belongs.

## Knock-on

Pruning removes a rule that three callers were matching by its *generated* name
(`''`, the internal `[item, R]` sequence of a `repeat0Plus`) rather than by the
entry name `toData` returns. `djs/tokenizer` now exports `jsMatcher`, which
returns the matcher and its entry name from one conversion; two proofs use the
entry name.

## Checks

`npx tsc` clean, `fjs test` 2871/2871, and `bnf/data`, `bnf/descent`, `bnf/ll1`
each at 100% line/branch/function coverage.

Changelog:
- **BREAKING CHANGES:** `bnf/data`: `Rule` gains a `Repeat` kind — the bare name
  of the rule to repeat, which keeps the four kinds disjoint by JavaScript type
  alone — and `toData` folds the unambiguous 0-or-more right-recursive shape into
  it, dropping the rules that fold orphans. Serialized rule sets change, and a
  repetition's intermediate rule names no longer exist, so match by the entry
  name `toData` returns. New exports `isRepeat` and `detectRepeat`
- **BREAKING CHANGES:** `bnf/descent`: a repetition now matches iteratively and
  produces one AST node holding a flat sequence of its items instead of a
  right-recursive chain of `some`/`none` nodes; per-item tags are unchanged. New
  export `descentParserRuleSet`, the matcher over an already materialized
  `RuleSet`
- `bnf/ll1` keeps the nested repetition shape — its dispatch model inlines a
  nullable item's first set and has nowhere to put a looping frame — so a
  `Repeat` compiles back to right-recursion there; a repetition's node tags are
  now the item's rather than `some`/`none`
- `djs/tokenizer`: new export `jsMatcher`, the whole-file matcher together with
  its entry rule name, replacing `descentParser(jsGrammar())` plus a hard-coded
  rule name at each call site

https://claude.ai/code/session_01RawVa1TjyiJ7dNRJtNkAyz


o2alexanderfedin (commented):
The parser core holds up under adversarial testing — one doc line to fix.

**Repeat recognition, no false positives.** Ran 15 adversarial grammars through `toData`. Correctly rejected: 1-or-more chain, separated list `[item,sep,L]`, 3-branch variant, nullable item (`option('a')`), item reaching back (`S=['(',R,')']`), mutual recursion A/B, left recursion, and an empty-but-not-`[]` `none` (conservative miss). Correctly folded: `repeat0Plus`, hand-written `{some:[item,R],none}` in either branch order, and the case where the step sequence `S` is also referenced elsewhere. `repeat0Plus(repeat0Plus(a))` folds only the inner one. The `stepRule[1] !== name` identity check is what kills different-rule-same-shape. The only thing a fold loses is the two branch tags — the declared breaking change.

**Flat AST ≡ cons chain**, verified differentially against `origin/main`: 1353 cases across `classic`, `deterministic` and `jsGrammar` (curated plus 400 seeded-random JSON/DJS inputs including corrupted ones), 0 mismatches, 4037 chains flattened and aligned. `tokenizeString`/`tokenizeJs` byte-identical over 65 inputs; `ll1` over 42 inputs identical in success/remainder/shape, differing only in tags.

**String discrimination** is clean — a `Variant`'s keys and values are strings but the rule itself is an object, and `data`/`descent`/`ll1` all import the single `isRepeat` and check in the same order, no residual shape probing.

`npx tsc` clean; `fjs t` → 2872 pass / 0 fail; `bnf/data`, `bnf/descent`, `bnf/ll1` each at 100/100/100.

**The one finding — `fjs/bnf/todo/unicode-rules.md:183`:**

> `[LL(1) repeat flattening] — unaffected by this split; the shipped Repeat encoding never used a bare string.`

It is a bare string. This contradicts line 88 of the same file, `data/types.ts` (`export type Repeat = string`), and changelog line 2. The deleted `667-bnf-repeat-flatten.md` explicitly forbade this encoding and the PR deliberately reverses that — every other touched todo records the reversal honestly, so this line reads as leftover pre-reversal wording that would mislead whoever picks up `unicode-rules.md`.

Minor: changelog entries 1 and 2 run ~380 and ~330 chars against the ~250-char guidance in `changelog/README.md`.

Re #1600 (same file): non-adjacent hunks — this PR at `tokenizer/module.f.mjs:485,523`, #1600 at `:529` — textual conflict unlikely, and #1600's unreachability argument still holds under the iterative matcher.

o2alexanderfedin (approved):
The `unicode-rules.md:183` finding is resolved, and I checked the new wording rather than taking it on faith: it now says the shipped `Repeat` encoding is a data-layer string, not a functional one. `fjs/bnf/data/types.ts` has `export type Repeat = string` (data layer), while `fjs/bnf/types.ts:49` has `DataRule = Variant | Sequence | TerminalRange | string` — the functional string case the unicode split actually removes. So "unaffected by this split" is now true for the stated reason, and it agrees with line ~86 and the changelog.

The new work stands on its own:

- **`ll1-ast-divergence.md`** — the 6-row table is factually correct, every row asserted by a passing proof, and byte-identical to the one in `fjs/bnf/ll1/README.md`. The headline claim (difference 5, grouping destroyed) is real: `[[A,B],C]` and `[A,B,C]` both yield `("A" ("B") ("C"))`. "Only the variant agrees" holds, the cited `[ws, value, ws]` shape exists in `testlib.f.mjs`, links resolve, and nothing dangles to the deleted `ll1-repeat-flatten.md`.
- **`proof.f.mjs` `descentDivergence`** — all 6 run and `assertEq` on both backends. Discrimination proved by mutation: breaking the repeat re-entry failed exactly `repetition`; dropping dispatch tags failed exactly `variant` + `takenOption`.
- **`6799421`** reverts nothing — purely +9 lines adding a `jsGrammar` proof that the entry rule is a `Repeat`. `jsMatcher` untouched.

No regression in what I verified last round: `npx tsc` exit 0, suite 2877 pass / 0 fail, and `bnf/data`, `bnf/descent`, `bnf/ll1` all still 100/100/100.

Changelog length is still over the ~250-char guidance (entry 3 is ~556) — non-blocking, as before.

PR #1596, opened 2026-08-15
Bring asn.1 to 100% line/branch/function coverage: 90.91% → 100% branch

sergey-shandar:
## Summary

`fjs/asn.1/module.f.mjs` was at 90.91% branch coverage (6 uncovered branches). This brings it to 100% line/branch/function coverage:

- **`parsedTagDecode`'s 8-way `||` assert** narrows a decoded tag's class + primitive/constructed bits (top 3 bits of the first byte) to one of exactly eight values. Every existing test in this file only ever decoded a universal-primitive tag (top bits `000`), so the other seven `===` comparisons in the chain never had a case where they were the one making the assert true — they were always short-circuited past.
- Adds `tagClass`, one `check` round trip per class×primitive/constructed combination (universal/application/context/private × primitive/constructed), covering all eight.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `node --test --experimental-test-coverage --test-coverage-include='fjs/asn.1/module.f.mjs' fjs/emergent_testing/all.test.mjs` → 100.00% line/branch/func
- [x] `node ./fjs/module.mjs t` → 2856 pass, 0 fail

---

### DIFF
--- a/changelog/unreleased/1596.md
+++ b/changelog/unreleased/1596.md
@@ -1,0 +1,3 @@
+- `asn.1`: adds proof coverage for `parsedTagDecode`'s tag class ×
+  primitive/constructed combinations, bringing the module to 100%
+  line/branch/function coverage
--- a/fjs/asn.1/proof.f.mjs
+++ b/fjs/asn.1/proof.f.mjs
@@ -296,6 +296,22 @@ export const proof = {
         // Re-encoding an UnsupportedRecord returns it unchanged (it is already the TLV bytes)
         assertEq(encode(decoded), raw, 'encode should round-trip UnsupportedRecord')
     },
+    // parsedTagDecode narrows the top three bits of the tag's first byte
+    // (class + primitive/constructed) to one of exactly eight values via an
+    // 8-way `||` assert. Every other test in this file only ever exercises
+    // the universal-primitive tag (top bits 000), so each of the other seven
+    // comparisons has never had a case where it's the one that turns the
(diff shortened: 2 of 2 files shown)


o2alexanderfedin (approved):
Measured both ends rather than taking the numbers from the body: `origin/main` → line 100.00 / branch 90.91 / funcs 100.00; this head → 100.00 / 100.00 / 100.00.

The stated cause is accurate — `fjs/asn.1/module.f.mjs:72-75` is exactly the 8-way `||` assert over `firstByte & 0b111_00000n`, which came in with #1589 replacing a compile-time cast with a runtime narrowing assert. All eight tag constants are right per X.690 (`0x02/0x22` universal P/C, `0x42/0x62` application, `0x82/0xA2` context, `0xC2/0xE2` private), and the low 5 bits `00010` stay under `tagNumberMask` so no long-form path is taken by accident.

The important part for a coverage PR: these assert rather than merely execute. `check` compares the decoded tag against the caller's literal, not against implementation output. Confirmed by mutation — changing `classPcMask` to `0b011_00000n` made exactly the four bit-8-set cases fail (2852 pass / 4 fail), so the new cases can actually fail.

Co-located `proof.f.mjs` under `export const proof`, reusing the file's existing `check` helper, all eight as individually named tests. `npx tsc` exit 0; `fjs test` → 2856 pass / 0 fail. Changelog conforms, with #1571/#1576 as precedent for shipping an entry on a proof-only PR.

PR #1597, opened 2026-08-15
Bring dev to 100% line/branch/function coverage: 97.50% → 100% branch

sergey-shandar:
## Summary

`fjs/dev/module.f.mjs` was at 97.50% branch coverage (1 uncovered branch). This brings it to 100% line/branch/function coverage:

- **`loadModuleMap`'s `s === '.' ? '' : s` prefix computation** only ever saw the `s === '.'` side (`INIT_CWD` unset). The normal case — `INIT_CWD` set to a real subdirectory, as it is under `npm run` — was never exercised, so the `false` side of the ternary had no test.
- Adds `loadModuleMapStripsInitCwdPrefix`, which sets `INIT_CWD` to a real subdirectory and verifies discovered keys are still relativized to `./`-prefixed paths (not left carrying the `INIT_CWD` prefix).

## Test plan

- [x] `npx tsc --noEmit`
- [x] `node --test --experimental-test-coverage --test-coverage-include='fjs/dev/module.f.mjs' fjs/emergent_testing/all.test.mjs` → 100.00% line/branch/func
- [x] `node ./fjs/module.mjs t` → 2857 pass, 0 fail

---

### DIFF
--- a/changelog/unreleased/1597.md
+++ b/changelog/unreleased/1597.md
@@ -1,0 +1,2 @@
+- `dev`: adds proof coverage for `loadModuleMap`'s `INIT_CWD`-prefix
+  stripping, bringing the module to 100% line/branch/function coverage
--- a/fjs/dev/module.f.mjs
+++ b/fjs/dev/module.f.mjs
@@ -168,4 +168,13 @@ export const proof = {
         const [, result] = virtual({ ...emptyState, root })(loadModuleMap({}))
         assertEq(Object.keys(result).join(','), './a.f.ts')
     },
+    loadModuleMapStripsInitCwdPrefix: () => {
+        // With `INIT_CWD` set (the normal `npm run` case), discovery starts
+        // from that subdirectory and each key is relativized against it, so
+        // keys are still `./`-relative to `INIT_CWD` rather than carrying it.
+        /** @type {Dir} */
+        const root = { 'sub': { 'a.f.ts': () => ({}) } }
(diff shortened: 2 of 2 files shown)


o2alexanderfedin (approved):
Measured both ends: `origin/main` → 100.00 line / 97.50 branch / 100.00 funcs; this head → 100.00 / 100.00 / 100.00, matching the claim. `npx tsc` clean; suite 2857 pass / 0 fail.

The test is meaningful, proved by mutation: changing `module.f.mjs:110` from `s === '.' ? '' : s` to `const prefix = ''` makes only `loadModuleMapStripsInitCwdPrefix` fail, with `[ 'sub/a.f.ts', './a.f.ts' ]` — so it genuinely pins that the `INIT_CWD` prefix is stripped.

No env-hygiene concern, which was my main worry with a test like this: `loadModuleMap` takes `env` as a parameter (`loadModuleMap({ INIT_CWD: 'sub' })`) and nothing under `fjs/dev/` touches `process.env`, so there's no global state to leak across tests. Placement is right too — `proof.f.mjs` covers the exported `shouldLoad`, while module-private internals live in the inline `export const proof` alongside the sibling `loadModuleMapDefaultsToCwdWhenInitCwdUnset`. Changelog conforms.

One pre-existing observation, not for this PR: the opposite mutation (`const prefix = s`) survives both tests, since `relativize('.', './a.f.ts')` and `relativize('', './a.f.ts')` agree — the `''` branch is a no-op at this call site. That's a property of the existing code and the older test, not something introduced here.

PR #1585, opened 2026-08-15
Add 30 todo items documenting refactoring opportunities

sergey-shandar:
This commit adds 30 new todo documents across the codebase, each identifying a specific refactoring opportunity where code is duplicated, inconsistently named, or could be better organized.

## Summary

These todo items represent a systematic audit of code duplication and organizational issues across the fjs and nanvm-lib projects. Each document follows the established format with Problem, Proposal, and Tasks sections, prioritized from P3 (high) to P5 (low).

## Key Categories of Issues Identified

**Abstraction Extraction (11 items)**
- `kindset-eliminator`: Name the `KindSet` absent/whole/members trichotomy once
- `shared-helper-reuse`: Consolidate tiny helpers (`dedup`, `assoc`, comparators)
- `export-node-accessors`: Export private `Node` accessor functions from rtti/data
- `mapped-list-to-vec`: Parameterize the `try*ListToVec` + `mapUnwrap` pattern
- `strip-undefined-set-op`: Extract `withoutUnits` set operation to rtti/data owner
- `continuation-bit-layout`: Name the base128 varint layout constants once
- `byte-length-fields`: Compute SHA-2 byte lengths once, not per-consumer
- `sync-interpreter-owner`: Export shared synchronous memory interpreter
- `corpus-eliminators`: Export operator-corpus format rules as eliminators
- `proof-fixture`: Share Evo proof fixture between cas/evo and mcp/evo
- `node-program-options-factory`: Export the `NodeProgramOptions` factory

**Derivation Instead of Duplication (8 items)**
- `ivm-from-unpacked`: Collapse `IVm`'s eight `From` bounds to `From<Unpacked<Self>>`
- `tokenize-string-derive`: Derive metadata-free tokenizers from metadata-aware ones
- `rfc6979-vector-harness`: Parameterize RFC 6979 test-vector walker
- `trivia-merge-rule`: State ws/nl coalescing contract once in js/tokenizer
- `detect-via-push`: Derive `detect` from the `push` state machine
- `proof-drain-collectread`: Use exported `collectRead` instead of hand-rolled `drain`
- `bigint-eq-cmp-owner`: Unify `BigInt` equality and ordering decision paths
- `codepoint-type-owner`: Create shared `CodePoint` type owner in code_point module

**Type/Module Organization (5 items)**
- `codepoint-type-owner`: Establish code_point as the owner of the CodePoint type
- `numeric-coercion-module`: Create numeric_coercion.rs module for ToNumeric
[...]

PR #1598, opened 2026-08-15
types/bit_vec: share the mapped list-to-vec concatenation

sergey-shandar:
Implements `fjs/types/bit_vec/todo/mapped-list-to-vec.md` (deleted here).

## Why

"Concatenate a mapped list into one vector" was written twice — once for `Vec` elements inside `bo`, once for `U8` elements as top-level exports — with the same two steps and the same `mapUnwrap` lift, differing only in the element-to-`Unpacked` map. The module already had that abstraction for the opposite direction: `mappedChunkList` parameterizes chunking by an in-map and an out-map, with `chunkList` / `unpackedChunkList` as its instantiations.

## What

```js
const mappedListToVec = g => ({ unpackConcat }) => {
    const f = unpackListToVec(unpackConcat)
    const mg = map(g)
    return list => f(mg(list))
}
```

`tryListToVec` inside `bo` becomes `mappedListToVec(unpack)({ unpackConcat })`, and `tryU8ListToVec` becomes `mappedListToVec(u8ToUnpacked)` — point-free, since the bit order is now the only remaining parameter. Both unwrapping variants stay `mapUnwrap(...)` one-liners.

`map(g)` is bound once outside the returned function, matching `mappedChunkList`'s `mf` and §3.3. The two module-scope `map(...)` bindings I added in [#1563](https://github.com/functionalscript/functionalscript/pull/1563) are gone: mapping is the combinator's job now, so only the element maps themselves remain.

The `bo` parameter is typed structurally (`{ readonly unpackConcat: _UnpackConcat }`) rather than as `BitOrder`, because `bo`'s own body applies this before the record it returns exists. The public `tryU8ListToVec` keeps its declared `(_: BitOrder) => …` contract.

## Verification

- `npx tsc` clean; `fjs test` 2856 pass, 0 fail.
- `npm run cov`: `fjs/types/bit_vec/module.f.mjs` at 100% lines/branches/functions.
- **Declaration emit byte-identical** to `main` — this touches two exported functions and one record member, so the published contract is checked rather than assumed.
- **Differential check** across both bit orders and all four entry points: `Vec` lists that are empty, single, ragged, and interleaved with zero-length members, a 64-element mixed-length list, `U8` lists from empty to 300 bytes, and the over-`maxLength` case that must answer `null` rather than build. **Identical** for `lsb` and `msb`.

Changelog:
- `types/bit_vec`: the `Vec` and `U8` list concatenations share one combinator, mirroring `mappedChunkList` in the chunking direction; behavior unchanged

---


o2alexanderfedin (approved):
Ran a differential harness with the pre-PR module beside the new one, both bit orders, all four entry points (`tryListToVec`, `listToVec`, `tryU8ListToVec`, `u8ListToVec`): **2118 comparisons, 0 differences**. Covered empty, single, zero-length-interleaved, ragged mixed-bit-length lists, a 64-element mixed list, 300 random `Vec` lists and 200 random `U8` lists, thunk and generator lists, repeated reuse of the partially-applied function, and the `maxLength`/`maxLength+1` boundary. Null and throw paths matched exactly too (6 `null` results, 12 throws from the `mapUnwrap` variants).

Binding `map(g)` once is safe — it's hoisted into the same scope `unpackListToVec(unpackConcat)` already occupied, and `map` and both element maps are pure, so there's no per-call state to lose. Declaration emit is unchanged: `fjs/types/bit_vec/module.f.d.mts` is byte-identical to `main`.

`npx tsc` exit 0; `fjs t` → 2856 pass / 0 fail. The deleted todo is the one implemented, and nothing references it, `mapUnpack`, or `mapU8ToUnpacked` afterward. Changelog conforms.

Nit, PR body only — nothing in the diff repeats it: `mappedChunkList`'s instantiations are `chunkList` and `uintChunkList` (`module.f.mjs:447`, `:456`); `unpackedChunkList` doesn't exist. Same wrong symbol as in the todo I flagged on #1585, now retired with that file.

PR #1599, opened 2026-08-15
effects/io: introduce IoEffect and its composition API

sergey-shandar:
Stages 1 and 2 of the IoEffect migration
(`fjs/effects/todo/io-effect-migration.md`): the fallible effect abstraction and
the operations that compose it. Additive — no operation, runner, or consumer
produces an `IoEffect` yet, which is stage 3 and stage 4.

`IoEffect<O, T, E>` is `Effect<O, Result<T, E>>` — the raw effect with its
failure made part of the type. `NotImplemented` is
`readonly['notImplemented', string]`, the error a runner answers with when it
cannot dispatch an operation. The composition API is `step`, `catchStep`,
`resultStep`, the `pureOk` / `pureError` lifts, and `mapStep`.

## What the layer buys

Raw `step` knows nothing of `Result`, so a fallible chain silently continues
past a failure:

```js
const a = writeFile(...)
const b = step(a, () => console('written'))
```

`'written'` is printed whether or not the write succeeded, and it is precisely
the *value-discarding* continuation that hides it — one that read the value
would not have compiled. With Io `step` the same line means what it looks like:
the continuation runs only on `ok`, and an error propagates on its own. That is
the structured replacement for exception-style propagation, in a language whose
`throw` is a panic and which has no `try`/`catch`.

## The union rules

The subtle part, so it is pinned. Following `okThen`, error types are
**unioned, not unified**: neither side is pre-widened, and a branch that is
passed through stays the very tuple it arrived as rather than being rebuilt to
retag it. `step` unions the errors and replaces the success type; `catchStep`
mirrors it, unioning the success channel and replacing the error type;
`resultStep` consumes both branches and replaces both. `types.ts` asserts each
signature at a concrete instantiation, so a "simplification" that unified an
error channel fails there rather than at some future call site.

**Raw `okStep` now unions its error types too.** Io `step` is raw `step` over
`okStep`, the adapter that already writes the `ok` / `error` branch — but
`okStep` unified the two errors, the one thing this layer must not do, so its
type now quantifies the incoming error on the second arrow, exactly as
`okThen` (its documented pure sibling) already did. A strict generalization:
every previous instantiation is `F = E`, so the raw consumers it stays
available for are untouched, and `tsc` confirms that over the whole repository.

## Naming decisions

- **`pureOk` / `pureError`, not `ok` / `error`.** Those names are
  `fjs/types/result`'s, and every consumer during the migration both builds
  bare `Result`s and lifts them; `pure` is not free to shadow either. The lifts
  are entry points — nothing else can produce an `IoEffect` until stage 3 — not
  speculative API.
- **`resultStep` is raw `step`** at the Io instantiation and adds no branch
  behavior; it is implemented as that function with a narrower type. It is
  named anyway because the three operations are the canonical vocabulary, and
  from stage 5, when the raw representation goes private, this is the public
  spelling of that instantiation. `finallyStep` is declined on the same
  principle read the other way.
- **`mapStep` is included, an Io `historyStep` is not.** Without `mapStep`,
  every site converted in stage 4 would spell its trailing pure projection as a
  step, which `map-step-combinator.md` establishes is the wrong shape.
  `historyStep` is expected as soon as `fjs/cas` migrates, but no consumer
  exists yet to shape it.

## `NotImplemented`

Ordinary recoverable effect data, not a fatal runner condition: the runner has
not started the operation, the program gets control back and owns the policy —
recover, choose a fallback, or panic itself. It names the operation **by
command name only**, since a payload may hold functions (`createServer`'s
listener, `sandbox`'s thunk, `test`'s body); a type-level assert against the
JSON data model keeps that claim checked, so adding a function-bearing field
fails to compile (verified — TS2344) instead of merely contradicting a comment.
It is **not a security boundary**: a runner may still interrupt or terminate a
program out of band, and nothing in the error channel obliges it to hand
control back.

## Coverage

Type-level, in `types.ts`: the alias is transparent (still the `Pure | Do`
union a runner discriminates), both channels widen — one way only, so an
unhandled error type is a compile error — `O` covariance survives the alias,
and the four signatures resolve exactly as pinned. Runtime, in
`io/proof.f.mjs`: propagation (the continuation is `todo`, which throws if it
runs), recovery, joined operation sets, mixed error types, pass-through by
reference identity, and each operation over a `Do` node — 100%
line/branch/function coverage on the new module, with `fjs/effects` still at
100%.

`npx tsc` is clean and `fjs t` passes (2879 tests); `npm run ci-update`
regenerates nothing.

Both stages are here because they share a branch. They are independently
mergeable in principle — stage 1 is the first commit — so say the word if you
would rather review them as two pull requests.

Changelog:
- `effects/io`: new `IoEffect<O, T, E>` (`Effect<O, Result<T, E>>`) and
  `NotImplemented` types — the fallible effect abstraction with an explicit
  error channel — plus its composition API: `step` propagates an error,
  `catchStep` recovers, `resultStep` observes both, with `pureOk` / `pureError`
  lifts and `mapStep`. No operation or runner produces one yet
- `effects`: `okStep` unions its two error types instead of unifying them,
  matching `okThen`. A strict generalization — existing call sites are the
  `F = E` instantiation


o2alexanderfedin (approved):
Types-only confirmed the decisive way: `tsc --emitDeclarationOnly` at `origin/main` vs this head diffs to one new file (`fjs/effects/io/types.d.ts`, exporting exactly `NotImplemented` and `IoEffect`, with the `_`-prefixed asserts dropped as claimed) plus JSDoc-comment-only changes in `fjs/effects/types.d.ts` and `module.f.d.mts`. Zero existing signatures changed; the 10 lines in `module.f.mjs` are all inside the module JSDoc block.

The alias composes for real, not just in prose — wrote a scratch `.ts` applying `step`, `mapStep`, `runPure`, `okStep` and `match` to an `IoEffect<ReadOp, string, NotImplemented>`; all compile, so the README's transparency claim holds.

Discrimination checks out as passing type asserts: `NotImplemented extends Result<unknown, unknown>` is `false` and the converse too, `'notImplemented'` can't collide with `'ok'`/`'error'`, nested `Result<NotImplemented, NotImplemented>` stays discriminable by the outer tag, and `Extract` narrows correctly against a future `['ioError', string]`. The serializability assert is load-bearing — adding a `() => void` element to `NotImplemented` fails with `TS2344` at `types.ts:113`, exactly as the body says.

`npx tsc` exit 0; `fjs t` → 2857 pass / 0 fail (body says 2856; the merge from main added one). Links resolve, the todo checks Stage 1 and leaves 2–6 open, types-only-in-`types.ts` matches `fjs/AGENTS.md` §3.5 and the `types/option` precedent, changelog conforms.

PR #1600, opened 2026-08-15
Bring djs/tokenizer back to 100% line/branch/function coverage: 98.86% → 100% branch

sergey-shandar:
## Summary

`fjs/djs/tokenizer/module.f.mjs` was at 98.86% branch coverage (2 uncovered branches) — a regression since PR #1553 previously brought it to 100%.

- **`stringDecodeScan`'s `default:` arm in the `'escape'` state.** The grammar's own `escape` rule (`buildToken`'s `string`) only ever accepts `"`, `\`, `/`, `b`, `f`, `n`, `r`, `t`, or `u` right after a backslash — any other character fails to parse before a token reaches this scan at all, so the fallback was dead. Replaced with an `assert`-based narrowing, the same technique `asn.1`'s `parsedTagDecode` already uses and #1594 just applied to `types/btree/set`, so the switch stays exhaustive for the type checker without a defensive fallback that can never run.
- **`tokenizeJs`'s `len < cpm.length ? cpm[len][1] : finalMetadata` ternary.** I verified empirically (probing dozens of malformed inputs with temporary debug logging) that entering the enclosing `if` always means `len < cp.length`: the `ok === false` case only ever happens with `len === 0` (nothing matched at position 0), and a `len !== cp.length` failure with `ok === true` is by definition a partial match. `repeat0Plus` never fails after successfully consuming the whole input, so there's no way to reach this branch with `len === cp.length` — the `finalMetadata` fallback was unreachable. Removed it.

`js/tokenizer`'s remaining 97.99% func score (also flagged by a repo-wide sweep) is unrelated and **not** a regression — the three uncalled closures are `unionConflict`'s own test fixtures, deliberately never invoked, as already documented in that file. No change needed there.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `node --test --experimental-test-coverage --test-coverage-include='fjs/djs/tokenizer/module.f.mjs' fjs/emergent_testing/all.test.mjs` → 100.00% line/branch/func
- [x] `node ./fjs/module.mjs t` → 2857 pass, 0 fail

## Changelog

- `djs/tokenizer`: restores 100% branch coverage — narrows `stringDecodeScan`'s escape-character switch via `assert` instead of a dead default arm, and removes an unreachable fallback in `tokenizeJs`'s error-position lookup


o2alexanderfedin (approved):
Both removals check out — the crux being that the new asserts can't fire on reachable input.

**`stringDecodeScan` escape arm.** The function is module-private; its only caller is `decodeJsonString` (`:384`), whose only caller is `toJsToken`/`toJsTokenWithMetadata`'s `case 'string'` — which runs after `ok && len === cp.length` and after the `unterminated`/`numError` structural-tag check, so the code points always come from a fully-parsed `string` rule. That rule (`:74-93`) admits only `set('"\\bfnrt')`, `solidus: '/'`, and `u: ['u', repeat(4)(hex)]` after a backslash — exactly the nine asserted, and no error-recovery path yields a `string` tag. The function already leans on this same narrowing for the `'unicode'` state, so the technique isn't new here. Empirical sweep on this head: 835 inputs (every code point 0x00–0xFF plus surrogate/astral edges, in `"\X"`, `"a\Xb"`, `"\X`, plus truncated and malformed `\u` forms) — zero throws.

**`tokenizeJs` ternary.** In `bnf/descent` a top-level failure always unwinds to the root's `startPos` = 0, so `!ok ⇒ len === 0`; and `ok && len !== cp.length ⇒ len < cp.length` because `physicalIdx` clamps. `cpm.length === cp.length ≥ 1` since empty input returns early. `finalMetadata` is still used by the two later paths, so no dead binding.

Coverage measured both sides: main 100.00 / 98.86 / 100.00 → head 100.00 / 100.00 / 100.00. `npx tsc` clean; `fjs t` → 2857 pass / 0 fail; all 18 CI checks green.

Re #1595 (same file): semantically independent, but `git merge-tree` shows one trivial textual conflict in the import block — #1595 rewrites the `descentParser` import, this one the adjacent `assertEq` import. Both body hunks merge cleanly; whichever lands second needs a one-line rebase.

Nit: `changelog/unreleased/1600.md` is present and well-formed, but the PR description has no `Changelog:` section — CONTRIBUTING asks for the list items duplicated there since the description becomes the squash commit.

PR #1601, opened 2026-08-15
cas: remove the unreferenced casUpload path

sergey-shandar:
Implements `fjs/cas/todo/casupload-dead.md` (deleted here).

## Verifying the TODO's claims first

Every factual claim checked out before anything was deleted:

- **`casUpload` has no consumer.** Its only references outside its own two proofs were prose. It still hard-coded `~/cas_upload/` and built its own `fileCas(sha256)(home)` — the residue of the MCP `type: 'url'` upload flow removed for symlink-escape reasons.
- **`casAddFile`'s doc was false on both halves.** It claimed "Both the CLI `cas add` and the MCP `add` delegate to this; the MCP layer additionally deletes the source file on success." The MCP `cas_add` handler takes inline `text`/`base64` content, calls `c.write` directly, and deletes nothing — it does not import `casAddFile` at all. The CLI is the only caller.

## What

`casUpload` and its two proofs are gone, and `casAddFile`'s doc now names its single consumer and says what the MCP tool actually does instead. The `sha256` import went with it — `casUpload` was its only user in that module.

## Stale references the removal would otherwise have created

Three documents described `casUpload` as existing code with migration requirements, which would have become quietly false:

- `fjs/cas/plan/staging.md` had a "Compatibility with the existing `casUpload` path" section planning a migration into `_stage/`, including a warning that its `rename` preserves the source mtime and would defeat a mtime-based grace period. It now records that the path was deleted and that a Strategy 1 cleaner has one staging location to scan, keeping the mtime hazard visible as history rather than as a live constraint.
- `fjs/cas/plan/staging-lease.md` deferred to that section; it now says the same in two sentences.
- `fjs/cas/todo/write-closed-helpers.md` cited `okStep` applying to "two steps inside `write` (`createExclusive`, and `casUpload` beside it)" — now one step.

`changelog/0.34.0.md` also mentions `casUpload`, and is deliberately untouched: it is published release history.

## Verification

- `npx tsc` clean.
- `fjs test`: **2855 pass, 0 fail, against a measured 2857 on `main`** — exactly the two deleted proofs, `casUploadSuccess` and `casUploadFailureKeepsSource`, and nothing else moved.
- `npm run cov`: `fjs/cas/module.f.mjs` still 100% lines/branches/functions — the deleted code took its own coverage with it and left no gap behind.

Changelog:
- **BREAKING CHANGES:** `cas`: `casUpload` is removed. It had no caller and was the remnant of an upload flow deleted in 0.32.2; content is added through `casAddFile` (CLI) or the MCP `cas_add` tool


o2alexanderfedin (approved):
Checked the deadness the ways a grep alone wouldn't catch. `git grep casUpload` across all tracked files at `origin/main` — including `nanvm-lib/`, `Cargo.toml`, `.github/`, `spec/` — yields only `fjs/cas/module.f.mjs`, its two proofs, and prose in `plan/`, `todo/` and published changelogs. No barrel re-export (`fjs/mcp/module.f.mjs:44` and `fjs/mcp/cas/module.f.mjs:127` import only `fileCas`/`collectRead`), no CLI command-table entry, no `package.json`/`deno.json` task, no MCP tool-name map hit. `~/cas_upload/` survives only as history in `fjs/mcp/README.md:233` and `changelog/0.32.2.md`.

The doc correction is right on all four sub-claims: `casAddArgs` is `{content, type: 'text'|'base64'}`, the handler calls `c.write(nonEmpty(...))` at `:199`, there's no `rm`/`unlink` anywhere under `fjs/mcp/`, and `casAddFile` isn't imported there at all. Its sole production caller is `fjs/cas/cli/module.f.mjs:30`.

`sha256` has zero remaining occurrences in the file. `npx tsc` clean; `fjs t` → 2855 pass / 0 fail against 2857 on main, exactly the two deleted proofs. Coverage of `fjs/cas/module.f.mjs` is 100/100/100 on **both** sides and the all-files totals are unchanged, so those proofs weren't covering anything else — which was the main thing worth checking before deleting them.

Deleted todo matches the work, nothing references it, changelog conforms and the body carries the matching `Changelog:` section.

PR #1602, opened 2026-08-16
media/type: make detect a projection of the streaming fold

sergey-shandar:
Implements `fjs/media/type/todo/detect-via-push.md` (deleted here).

The module doc claimed both detectors read the signature table "through the same eliminator" — true of the step, but the driving fold was written twice. `detect` is now:

```js
export const detect = bytes => magicMime(push(detectInit)(bytes).magic)
```

## Checking the equivalence the rewrite rests on

The TODO said to verify the fixed-point claim before relying on it. It holds, and more simply than the TODO suggested — `magicStep` opens with `if (m.tag !== 'scan') { return m }`, so **every** non-`scan` state is an unconditional fixed point, returning the identical object. The three cases then line up exactly:

| final magic | old `detect` | `push` |
| --- | --- | --- |
| `matched` | breaks there | `isSettled` is `true` on `matched`, breaks there |
| `dead` | breaks there | may keep folding — `isSettled` is `!utf8.valid \|\| !utf8.text` — but `dead` cannot move again |
| `scan` | runs every byte | `isSettled` never fires, so it runs every byte too |

Both start from `magicInit` (`detectInit.magic`) and iterate the same `iterable(u8List(msb)(…))`, so the ragged-`Vec` zero-padding is unchanged as well.

The middle row is the only behavioural difference in the fold, and it is the one now pinned by proof: a `dead` magic gets fed many more bytes than the old loop ever gave it.

## Cost

That extra folding is not free on text input, and the JSDoc now says so. `push` stops once the *whole* verdict is settled, not once the magic state is, so a signature match or non-text bytes still exit early but valid text is read to the end — 100 KB of ASCII measured **1.2 ms before, 82 ms after**. It is the same work `detectVec` already does on the same bytes (~87 ms there), and `Vec` is capped at `maxLength`, so the cost is bounded. Recorded on the export rather than traded away, since `detect` has no production caller and the alternative is keeping the duplicate fold.

## Kept, not deleted

The TODO offered deleting `detect` outright, since its only callers are its own proofs — `fjs/mcp/cas` uses `detectStream` and `fjs/media` uses `detectVec`. (The `detect` in `fjs/media/proof.f.mjs` is a different function, the dialect registry's.) I kept it: it is exported public API with a documented contract, it is the natural pure entry point, and rewriting it removes the duplicate fold without a breaking change. Deleting it would be a separate decision about the package surface, not a deduplication.

## Verification

- `npx tsc` clean; `fjs test` 2860 pass, 0 fail.
- `npm run cov`: `fjs/media/type/module.f.mjs` at 100% lines/branches/functions.
- New proof cases pin both absorbing states as fixed points from the outside, where `magicStep` is private: a `matched` signature followed by non-extending bytes and by a 64-byte tail, a `dead` magic likewise, and the same dead-magic input still reading as `text` through `detectVec` — which is what confirms the fold really did continue past the settled magic.
- **Differential against `main`, 542 inputs**: every prefix length of all eight signatures, each with a trailing tail, every single-byte mutation of every signature byte, text and JSON samples, all ragged `Vec` bit-lengths from 0 to 40, and 400 pseudo-random byte strings of lengths 1–12. **Identical.**

Changelog:
- `media/type`: `detect` reads the magic state off the streaming detector's fold instead of driving a second copy of it; same results, but it now reads a whole text `Vec` rather than stopping at the first non-signature byte


o2alexanderfedin (approved):
The fixed-point argument holds at source: `magicStep` (`module.f.mjs:87`) opens with `if (m.tag !== 'scan') { return m }` — a bare identity return, no reconstruction or mutation, and every other return is inside the `scan` branch. So `matched` and `dead` are absorbing, and since `detect` reads only `magicMime(state.magic)`, the extra folding `push` does past a `dead` magic (while `isSettled` waits on `!utf8.valid || !utf8.text`) can't change the answer. `utf8Step` is total over 0–255 and can't throw, so nothing that previously short-circuited can now fail late — that was my main worry with the concession in the body.

Differential test against `origin/main`, 4221 inputs: all signature prefixes with ASCII/NUL/0xFF/multibyte tails, every single-byte mutation of every signature byte, all 256 single bytes, empty, ragged `Vec`s at every bit length 0–48, text/JSON, six invalid-UTF-8 shapes, die-early and die-late blobs, 2500 pseudo-random strings, and 100 KB blobs — **0 differences in `detect`, 0 in `detectVec`**.

`npx tsc` clean; `fjs t` → 2860 pass / 0 fail; all 18 CI checks green. Deleted todo is the one implemented, no dangling references, changelog conforms and the body carries `Changelog:`.

Non-blocking note: on text input the fold no longer exits at the first dead byte — 100 KB ASCII measured 0.40 ms before vs 78 ms after. It's the cost `detectVec` already pays on the same bytes, `detect` has no production caller, and `Vec` caps at 128 KB, so it's bounded. Worth knowing rather than changing.

sergey-shandar:
@o2alexanderfedin thanks — the `utf8Step` totality check is the part I hadn't thought to make explicit, and it's the right question: my equivalence argument only covers *which* magic state you end up in, not whether the extra folding can fail. Good catch that it closes.

I took the performance note as "worth knowing", not "worth changing", and left the behaviour alone. But knowing it shouldn't live only in a review thread, so I recorded it where a caller will meet it — `detect`'s JSDoc now carries the cost characteristic, and the PR body and changelog entry say it too rather than claiming a pure no-op.

I reproduced your measurement before writing it down: 100 KB of ASCII, **1.2 ms before, 82 ms after**, with `detectVec` at ~87 ms on the same bytes. Same order as your 0.40 → 78 ms. The JSDoc states the shape of it — early exit on a match or on non-text bytes, whole-`Vec` read on valid text, bounded by `maxLength`, and the same work `detectVec` already does — so someone scanning many large text blobs for signatures alone knows they are paying for the UTF-8 fold too.

Two things I deliberately did **not** do:

- **Make `push` stop on a settled magic.** That would fix `detect`'s cost by changing `detectVec`'s semantics — the UTF-8 factor has to keep running for `finish` to classify text at all. Wrong trade.
- **Give `detect` its own stop predicate.** That reintroduces the second fold this PR removes, which is the whole point of the change.

If the cost ever does matter, the honest fix is a `push` variant parameterized by its settling predicate, and that wants its own issue rather than a hedge here.

Changes since your review are `7bada23` (JSDoc) and `8018510` (changelog wording); no behaviour touched, `fjs test` still 2860 / 0.

---

o2alexanderfedin (approved):
Re-approving at the new head. The delta is documentation-only — filtering the diff for non-comment lines returns nothing, and `detect`'s body is byte-identical to 26b19bd, so the 4221-case differential from my earlier review still covers this head.

Measured the documented cost rather than taking the numbers on faith (100 KB, avg of 20–50 reps after warmup):

| input | old `detect` | new `detect` |
| --- | --- | --- |
| ASCII text | 0.22 ms | 55.5 ms (`detectVec` 56.7 ms) |
| PNG, match at byte 8 | 0.13 ms | 0.17 ms |
| non-text binary | 0.18 ms | 0.23 ms |

All three JSDoc claims hold: the early exit survives both on a signature match and on non-text bytes, only valid text pays the full fold, and the cost matches `detectVec`'s on the same bytes. `isSettled` settles `dead` on `!valid || !text`, so control bytes count as non-text exactly as written. Absolute numbers differ from the ones in the commit in the usual machine-variance direction; the characterization is right.

Changelog conforms — three wrapped lines, `Topic: description` style, no links, and the measurements live in JSDoc where `changelog/README.md` wants them. `npx tsc` clean; `fjs test` → 2860 pass / 0 fail.

PR #1603, opened 2026-08-16
tokenizer: bring back to 100% coverage

sergey-shandar:
## Summary
- Restore full line/branch coverage for `fjs/js/tokenizer/module.f.mjs` by exercising the `union` throw path directly via a type-erased `unionX` helper instead of an unreachable-in-practice callback construction
- Drop unused `RangeMapArray`/`StringToken`/`NumberToken`/`BigIntToken`/`ErrorToken`/`WhitespaceToken`/`NewLineToken`/`IdToken`/`CommentToken`/`EofToken` imports and add the now-used `Reduce` import


o2alexanderfedin (approved):
Coverage claim is exact, measured with `npm run cov`:

| | line | branch | funcs |
| --- | --- | --- | --- |
| `fjs/js/tokenizer/module.f.mjs` on `origin/main` | 100.00 | 100.00 | **97.99** |
| same file at `c9442a6` | 100.00 | 100.00 | **100.00** |
| all files, main → head | 100/100/**99.86** → 100/100/**100.00** |

So this closes precisely the global 99.86% funcs gap. The cause is confirmed by #1600's own body — the old fixture *declared* three arrows it never invoked, which is what cost the function score; `unionX(0)(1)(2)` allocates none.

I checked the type-erasure angle since it runs opposite to #1594/#1600 removing unreachable branches, and it holds up: the `throw` proof key is a first-class convention here (`fjs/emergent_testing/module.f.mjs:65,91`), and this throw is a real merge guard rather than a dead defensive branch. `unionX` is module-local, not exported, and `const union = unionX` re-narrows to `_CreateToToken<T>`, so every production call site (`module.f.mjs:184`) keeps its old type — no safety loss and nothing reachable from production.

Every dropped import is genuinely unreferenced; `Reduce` is used at lines 165 and 174. `npx tsc` exit 0; suite 2855 pass / 0 fail; CI green.

One nit: the description has no `Changelog:` section, which CONTRIBUTING.md:201 makes mandatory either way. Proof-only with no behavior change, so no `changelog/unreleased/1603.md` file is needed — just `Changelog: none` at the end of the body.

PR #1604, opened 2026-08-16
effects/io: add the IoEffect composition API

sergey-shandar:
Stage 2 of the IoEffect migration
(`fjs/effects/todo/io-effect-migration.md`): the operations that compose an
`IoEffect`. Stage 1 — the `IoEffect` and `NotImplemented` types — shipped in
#1599. Still additive: no operation, runner, or consumer produces an
`IoEffect` yet, which is stage 3 and stage 4.

The API is `step`, `catchStep`, `resultStep`, the `pureOk` / `pureError` lifts,
and `mapStep`, in `fjs/effects/io/module.f.mjs`.

## What the layer buys

Raw `step` knows nothing of `Result`, so a fallible chain silently continues
past a failure:

```js
const a = writeFile(...)
const b = step(a, () => console('written'))
```

`'written'` is printed whether or not the write succeeded, and it is precisely
the *value-discarding* continuation that hides it — one that read the value
would not have compiled. With Io `step` the same line means what it looks like:
the continuation runs only on `ok`, and an error propagates on its own. That is
the structured replacement for exception-style propagation, in a language whose
`throw` is a panic and which has no `try`/`catch`.

## The union rules

The subtle part, so it is pinned. Following `okThen`, error types are
**unioned, not unified**: neither side is pre-widened, and a branch that is
passed through stays the very tuple it arrived as rather than being rebuilt to
retag it. `step` unions the errors and replaces the success type; `catchStep`
mirrors it, unioning the success channel and replacing the error type;
`resultStep` consumes both branches and replaces both. `types.ts` asserts each
signature at a concrete instantiation, so a "simplification" that unified an
error channel fails there rather than at some future call site.

**Raw `okStep` now unions its error types too.** Io `step` is raw `step` over
`okStep`, the adapter that already writes the `ok` / `error` branch — but
`okStep` unified the two errors, the one thing this layer must not do, so its
type now quantifies the incoming error on the second arrow, exactly as
`okThen` (its documented pure sibling) already did. A strict generalization:
every previous instantiation is `F = E`, so the raw consumers it stays
available for are untouched, and `tsc` confirms that over the whole repository.

## Naming decisions

- **`pureOk` / `pureError`, not `ok` / `error`.** Those names are
  `fjs/types/result`'s, and every consumer during the migration both builds
  bare `Result`s and lifts them; `pure` is not free to shadow either. The lifts
  are entry points — nothing else can produce an `IoEffect` until stage 3 — not
  speculative API.
- **`resultStep` is raw `step`** at the Io instantiation and adds no branch
  behavior; it is implemented as that function with a narrower type. It is
  named anyway because the three operations are the canonical vocabulary, and
  from stage 5, when the raw representation goes private, this is the public
  spelling of that instantiation. `finallyStep` is declined on the same
  principle read the other way.
- **`mapStep` is included, an Io `historyStep` is not.** Without `mapStep`,
  every site converted in stage 4 would spell its trailing pure projection as a
  step, which `map-step-combinator.md` establishes is the wrong shape.
  `historyStep` is expected as soon as `fjs/cas` migrates, but no consumer
  exists yet to shape it.

## Coverage

Type-level, in `types.ts`: the four signatures resolve exactly as pinned, next
to stage 1's asserts for the alias's transparency and one-directional widening.
Runtime, in `io/proof.f.mjs`: propagation (the continuation is `todo`, which
throws if it runs), recovery, joined operation sets, mixed error types,
pass-through by reference identity, and each operation over a `Do` node — 100%
line/branch/function coverage on the new module, with `fjs/effects` still at
100%.

`npx tsc` is clean and `fjs t` passes (2877 tests); `npm run ci-update`
regenerates nothing.

Changelog:
- `effects/io`: the `IoEffect` composition API — `step` propagates an error,
  `catchStep` recovers, `resultStep` observes both — with the `pureOk` /
  `pureError` lifts and `mapStep`. No operation or runner produces an
  `IoEffect` yet
- `effects`: `okStep` unions its two error types instead of unifying them,
  matching `okThen`. A strict generalization — existing call sites are the
  `F = E` instantiation

---


o2alexanderfedin (approved):
Verified the short-circuit by execution rather than by reading, since it's the whole point of the layer. A 27-check harness drove real `Do`-node effects through a `match`-based runner, asserting on a side-effect log: on `ok` each continuation runs and receives the unwrapped value; on `error` it never runs and the error propagates. In a 4-link chain the first error wins and the three later continuations produce no log entries. Since `assertEq` is `===`, pass-through is reference identity rather than a re-wrap — confirmed for `step`'s error, `catchStep`'s `ok`, `resultStep`'s tuple and `mapStep`'s error.

`catchStep` touches only the error channel and *replaces* the error type with the recovery's `F` alone rather than widening — which is what Stage 3's `NotImplemented | IoError` union needs. `resultStep` hands `f` the intact `Result`.

Still additive, with one exception worth naming: diffing exported declarations at the merge base vs this head, the only pre-existing signature change is `okStep`, `(r: Result<T,E>) => …` → `<F>(r: Result<T,F>) => Effect<O, Result<R, E|F>>`. Runtime body byte-identical, `F = E` recovers the old type, and its three call sites in `fjs/cas` and `fjs/effects/node` typecheck unchanged. Everything else is pure addition.

The proofs discriminate — four mutations each killed a named proof: continuation-on-error → `step.propagates` + `step.overFailedDo`; recovery-on-ok → `catchStep.preservesOk`; `f` over the error payload → `mapStep.error`; value-equal re-wrap → `step.errorTupleIsUnchanged`. Unifying `_StepSig`'s error channel fails `tsc` at `types.ts:118`.

`npx tsc` clean; `fjs t` → 2904 pass / 0 fail; `fjs/effects/io/module.f.mjs` at 100/100/100 with all of `fjs/effects` at 100 (relevant now that #1605 gates on it); `ci-update` leaves the tree clean. Changelog and `Changelog:` section both conform; the todo marks Stage 2 done with Stages 3–6 remaining.

PR #1608, opened 2026-08-16
bnf: share the matcher cursor, AST, and result constructors

sergey-shandar:
Implements `fjs/bnf/todo/669-bnf-matcher-shared-core.md`, which is deleted here.

`bnf/ll1` and `bnf/descent` are two machines over one contract.
[`fjs/bnf/README.md`](https://github.com/functionalscript/functionalscript/blob/main/fjs/bnf/README.md#logical-eof-in-parser-input)
stated that contract normatively — callers supply physical symbols only, each
backend synthesizes exactly one logical EOF after them, public positions stay
physical, and internally a backend tracks the complete cursor
`(idx, eofConsumed)` because consuming EOF is progress even though `idx` does not
move. Nothing owned it in code. Each backend derived it separately, and since
#1531 both derived it *identically*.

Three things were duplicated: the cursor and its accessors, the AST type family
under two spellings, and the `mrSuccess` / `mrFail` pair. So was the prose — the
same explanation of why `cp.length + 1` is a position appeared in both modules
and in the README, and drifted whenever one copy was corrected.

## `fjs/bnf/matcher`

One module owning `Cursor`, the `Ast<L>` / `AstSequence<L>` / `AstTag` /
`AstResult<L, P>` family, and `leafAt` / `symbolAt` / `physicalIdx` /
`mrSuccess` / `mrFail`. One rather than three because the three are
interdependent: `leafAt` produces AST leaves, and `mrSuccess` builds AST nodes
positioned by a cursor, so splitting them would put one concept behind three
imports.

## What deliberately stays put

The machines. `descent` backtracks — two frame kinds, a per-frame rewind to the
sequence's start, and a furthest-failure high-water mark that outlives the
rewinds. `ll1` is predictive — one frame kind, and a cursor that never moves
backwards. Their public results are different types on purpose, and one matcher
covering both would be worse than two.

The leaf. A backend picks `L` for what it keeps of a consumed symbol, and
`CodePointMeta<T>` is the whole reason `descent` exists, so it stays there.
`symbolAt` takes the leaf reader as a parameter rather than branching on the
shape — the one place the two differ.

## `symbolOf`, and one deviation from the issue

The issue specified `symbolAt(identity)` in `ll1`. That does not type-check:
`symbolAt`'s own `L` cannot be inferred from a fully generic
`<T>(value: T) => T`, since TS tries to unify `T => T` with `L => number`.
Rather than an inline cast — which
[fjs/AGENTS.md §3.2](https://github.com/functionalscript/functionalscript/blob/main/fjs/AGENTS.md#avoid-as-type-assertions)
discourages — the type is pinned on a declared const, the form that document
prefers:

```js
/** @type {(leaf: CodePoint) => number} */
const symbolOf = identity

const symbolAtCp = symbolAt(symbolOf)
```

`descent` binds `symbolAt(([symbol]) => symbol)` the same way. Each partial
application sits at module scope, per
[fjs/AGENTS.md §3.3](https://github.com/functionalscript/functionalscript/blob/main/fjs/AGENTS.md#place-curried-partial-applications-at-their-dependencys-scope):
the binding's scope says it depends on the leaf shape and nothing else.

## Old names are deleted, not aliased

`descent`'s `AstRuleMeta<T>` / `AstSequenceMeta<T>` and `ll1`'s `_AstRule` /
`AstSequence` / `AstTag` are gone rather than aliased to the new names, per
[DESIGN.md §2](https://github.com/functionalscript/functionalscript/blob/main/DESIGN.md#2-the-api-is-the-most-important-part-of-quality).
There was exactly one external importer, `djs/tokenizer`, updated here.

Beyond the issue's task list: `showAst` in `bnf/testlib.f.mjs` carried its own
hand-rolled copy of the AST node shape, which is itself evidence the family is
shared. It now uses `Ast<_Leaf>`.

## Prose

`fjs/bnf/README.md` keeps the normative statement and points at where it is
implemented; the mechanics live on `Cursor` in the shared JSDoc; each backend's
README keeps only what is true of it — `ll1` never rewinds, `descent` rewinds and
tracks a furthest failure.

## Checks

`npx tsc` clean, `fjs test` 2917/2917, and `npm run cov` reports 100.00 across
all files — which #1605 now enforces.

Changelog:
- **BREAKING CHANGES:** new `bnf/matcher`, one owner for what every BNF matcher
  backend shares: the `Cursor` a match runs on, the `Ast<L>` / `AstSequence<L>` /
  `AstTag` family it builds, the `AstResult<L, P>` it returns, and the
  `leafAt` / `symbolAt` / `physicalIdx` / `mrSuccess` / `mrFail` functions over
  them. `bnf/ll1` and `bnf/descent` derived all of it separately and identically
- **BREAKING CHANGES:** `bnf/descent`: `AstRuleMeta<T>` and `AstSequenceMeta<T>`
  are now `Ast<CodePointMeta<T>>` and `AstSequence<CodePointMeta<T>>` from
  `bnf/matcher`, and `AstTag` moves there too. `CodePointMeta<T>` stays — a leaf
  that carries metadata is this backend's own concept
- **BREAKING CHANGES:** `bnf/ll1`: `_AstRule`, `AstSequence` and `AstTag` are
  gone; `MatchResult`'s first element is `Ast<CodePoint>`
- `djs/tokenizer` takes the AST types from `bnf/matcher` rather than
  `bnf/descent`

https://claude.ai/code/session_01RawVa1TjyiJ7dNRJtNkAyz


o2alexanderfedin (approved):
The extraction is behavior-preserving. `leafAt`, `symbolAt`, `physicalIdx`, `mrSuccess`/`mrFail` are byte-equivalent to the copies they replace — same expressions, same property order in the result record — with only the leaf reader parameterized; `identity` for ll1 and `([symbol]) => symbol` for descent reproduce `cp[pos]` / `cp[pos][0]` exactly. EOF synthesis stays in the backends: the `pos <= cp.length` guards and `pos + 1` advancement are untouched, so each machine still synthesizes exactly one logical EOF.

Public positions stay physical — descent still clamps via `physicalIdx(cp.length)` at the single place it builds `DescentMatchResult`, and ll1's `remainderAt` now calls that same clamp instead of an inline `Math.min` (`ll1/module.f.mjs:240`). No `eofConsumed` component and no EOF-counting index reaches a caller.

Executed, main vs head:
- **1113 backend cases** (21 grammars × 53 inputs) — `eof`, `[eof,eof]`, `repeat0Plus(eof)`, `{a,e:eof}`, `[eof,range]`-first variants, nullable-then-`eof`, `classic()`/`deterministic()` with and without a trailing `eof`; inputs covering empty, valid, first-symbol failure, mid-input failure and exactly-at-EOF. Compared success, descent `idx` plus the full `failure` record, ll1 `remainder`, and raw AST JSON: byte-identical.
- **70 tokenizer cases** through `tokenizeString`, `tokenizeJs`, `tokenize`: identical.
- `repeat0Plus(option(eof))` OOMs — but identically on both sides, so pre-existing rather than introduced here.

`npx tsc` clean; `fjs t` → 2921 pass / 0 fail; `npm run cov` → 100/100/100.

**Merge-order warning:** this conflicts non-trivially with #1611. `git merge-tree` gives 8 conflict hunks — `ll1/module.f.mjs` (5), `ll1/types.ts` (2), `ll1/README.md` (1) — plus a delete/modify conflict on `fjs/bnf/todo/669-bnf-matcher-shared-core.md`, deleted here and modified there. Whichever lands second needs a real rebase of the ll1 module, not a mechanical merge.

Passing note: `fjs/bnf/todo/207.md:274` and `todo/new-parser.md:106,191` still name `AstRuleMeta`, which no longer exists.

PR #1609, opened 2026-08-16
js/tokenizer: give the trivia coalescing rule one owner

sergey-shandar:
Implements `fjs/js/todo/trivia-merge-rule.md` (deleted here).

## Why

"A maximal run of whitespace/newline trivia collapses to a single token, and a run containing any newline is an `nl`" is a contract of `JsToken`, decided independently in two modules — as two range-map states in `fjs/js/tokenizer`, and as a four-way tag table in `fjs/djs/tokenizer`'s `scanFunc`. The DJS copy exists only to reproduce the JS tokenizer's output, so a change to the absorption rule on one side leaves the proofs as the only thing catching the drift.

## What

`fjs/js/tokenizer` — which defines `JsToken` — owns it:

```js
/** @type {(a: TriviaKind, b: TriviaKind) => TriviaKind} */
export const mergeTrivia = (a, b) => a === 'nl' || b === 'nl' ? 'nl' : 'ws'
```

with `TriviaKind = 'ws' | 'nl'` added to `types.ts`.

Both trivia state ops now read the rule instead of encoding it, through a pair of shared state constants so a run of trivia allocates nothing per character — the old ws→nl transition built a fresh `{ kind: 'nl' }` each time.

DJS maps its grammar tags to the same kinds and keeps only the bookkeeping. Its four branches become one:

```js
const inputKind = triviaKind(input)
const stateKind = triviaKind(stateTag)
if (inputKind !== null && stateKind !== null) {
    return [null, mergeTrivia(stateKind, inputKind) === stateKind ? state : [input, null, []]]
}
```

The restart is needed exactly when the merged kind is not the one the pending token already has — ws followed by nl — which reproduces all four original rows.

## Verification

- `npx tsc` clean; `fjs test` 2914 pass, 0 fail; `npm run cov` 100% aggregate.
- Proof covers all four combinations of `mergeTrivia`, since not re-deriving them on each side is the point.
- **Differential against `main`, 1705 cases**: every trivia run up to length 4 over all four trivia characters (341 runs), each in five contexts — alone, between identifiers, leading, trailing, and between numbers — comparing **both** tokenizers' full token streams. Identical, and the two agree with each other on every trivia token in all 1705.

## A note on the harness

My first two attempts at that differential were broken and **both reported "identical" while comparing two equally-broken outputs** — first calling `djsTokenize` with the wrong arity, then reading a `.kind` that does not exist on `JsTokenWithMetadata`. The guards I had written (row count, non-empty) did not catch it because the rows were populated with the string `"undefined"`.

The harness now also fails loudly on any row containing `undefined` or `THREW`, and separately asserts that the two tokenizers agree with each other — so a broken harness cannot pass as a clean diff. The numbers above are from the corrected run. Worth stating because a differential that cannot fail is worse than none.

Changelog:
- `js/tokenizer`: new `mergeTrivia` states the whitespace/newline coalescing rule once; `js` and `djs` tokenizers both read it instead of each encoding the four cases

---


o2alexanderfedin (approved):
`mergeTrivia` reproduces both originals exactly. JS side: `parseWhitespaceStateOp` was ws→`state``{kind:'ws'}`, nl→`{kind:'nl'}`; `parseNewLineStateOp` was ws→`state`, nl→`state` — and `triviaState[mergeTrivia(kind, x)]` yields ws/nl/nl/nl across those four. Sharing the state constants is safe: `_ParseWhitespaceState`/`_ParseNewLineState` are exactly `{readonly kind}` (`types.ts:167,170`), nothing dropped. DJS side: all four rows match including the only restart (`state ws` + `input nl` → `[input, null, []]`), and the `inputKind !== null && stateKind !== null` guard has the same domain as the old `isNlTag`/`isWsTag` pair, so non-trivia still falls through to the general branch.

No fold-order concern — each op is a per-character state transition and both call sites pass `(pendingKind, incomingKind)` in that order; `mergeTrivia` is an `nl`-wins disjunction, so commutative and associative regardless.

Differential main vs head over **8899 inputs** — runs to length 4 over ` \t\n\r`, CRLF/CR variants, newline first/middle/last, start/end/empty, 19 contexts, plus NBSP, U+2028, U+2029, VT, FF, BOM and ideographic space — **0 differences** in either tokenizer, comparing full streams with kind, value, line and column. `TriviaKind` loses nothing: `rangeSetNewLine`/`rangeSetWhiteSpace` are `lf,cr`/`ht,space` only, and U+2028/2029 already tokenize as `error` on both revisions.

`npx tsc` exit 0; `fjs test` → 2914 pass / 0 fail; `npm run cov` exit 0 under the 100% gate. Deleted todo is the one implemented, no dangling refs, changelog and `Changelog:` both conform.

Non-blocking: `fjs/djs/tokenizer/todo/vocabulary-single-source.md` still names `isNlTag`/`isWsTag` at lines 39, 63, 66, 82 and 244-245 — now `triviaKind`.

PR #1610, opened 2026-08-16
media/note: add the vnd.fjs.note dialect for text items

sergey-shandar:
`vnd.fjs.note` is a new dialect-tagged JSON format for human-authored text items — notes, todos, issues, calendar events — stored as CAS blobs. It follows the `vnd.fjs.lock` template and starts deliberately minimal: the dialect tag, the text, and optionally the subjects the item depends on (referenceable from the text by index) and a priority.

```json
{ "dialect": "vnd.fjs.note", "text": "ship it once [0] lands", "dependencies": ["write the spec"], "priority": "P2" }
```

Design:

- **Minimal on purpose, because that is the extension strategy.** rtti structs are open, so under the versioning rule every future capability — a title, tags, event dates, a status — lands as an additive optional field that keeps the tag. The candidates are collected in `fjs/media/note/todo/extend-note-format.md`, each to be measured against the versioning rule and the interpretable-in-isolation rule (including the warning that `status` is the one candidate that may force a new dialect).
- **A value, not a step.** No timestamps, author, or history fields: an evolving note is an ordinary `vnd.fjs.revision` chain whose `snapshot` is a note blob, so editing, merging concurrent edits, and archiving a finished todo come from the existing history mechanism, and `revision` stays the only one.
- **`text` is required and may be `''`** — an absent text and an empty one would otherwise be two spellings of one blob, the same argument that makes `lock` required in `vnd.fjs.lock`.
- **`dependencies` entries are subject identity strings, never content hashes.** A blocked-by relation tracks the live item as it evolves; pinning a dependency to immutable content is the revision layer's `lock`, so the two fields compose without overlapping — the note says *what* it depends on, the revision's lock says *which content* those subjects resolved to. The field is optional because its absent value is the constant "depends on nothing" (the rule that keeps `archived` optional in the revision dialect).
- **The text references dependencies by index: `[0]` names `dependencies[0]`.** A bracketed decimal integer is a reference only when it indexes an existing entry; anything else is ordinary text, so there is no such thing as a structurally valid note with a *broken* reference and the convention adds no validation stage. Entry order is therefore significant — reordering renumbers references, as reordering `parents` changes a revision's meaning. An escape for a literal in-range `[N]`, and a reader-side reference-extraction helper, are tracked as additive follow-ups in the module's todo.
- **`priority`, not severity, and literals, not numbers.** Severity states a fact about a problem's impact; priority states the author's decision about what to work on first — an authored-items format carries the decision, and a separate `severity` stays available as its own additive field if defect tracking ever needs one. The scale is the closed literal union `'P1'`–`'P5'`, `todo/README.md`'s own issue vocabulary, so one scale ranks the repository's issues and a note blob alike; a numeric rank would pose questions (is `1` high or low? is `2.5` allowed?) only a semantic check could answer. Absent means unprioritized — a constant default, deliberately distinct from every rank including `P5`. Widening the scale later (`P0`) is the same fail-closed path as widening `lock`.
- **No semantic refinement stage.** A note has no hash fields — dependencies are identities, unconstrained like `revision`'s `subject`, and the priority union is enforced structurally — so structural validation is the whole check and `noteDialect` registers with `detect` with no refinement.

`fjs/media/proof.f.mjs` already used a hypothetical `vnd.fjs.note` schema as a test fixture; this PR promotes that exact shape into the real dialect and the proof now imports it. `fjs/mcp` registers the dialect alongside `revisionDialect` and `lockDialect`, so `cas_get` reports a note under its own media type.

Verified: `npx tsc` clean, `fjs test` green (2,935 tests), and the new module at 100% line/branch/function proof coverage.

Changelog:
- `media/note`: new module — the `vnd.fjs.note` dialect, served as
  `application/vnd.fjs.note+json`: a human-authored text item (a note, todo,
  issue, or calendar event) as a history-free blob; edits, merges, and
  archiving come from `vnd.fjs.revision`. The shape is deliberately minimal —
  `dialect`, `text`, an optional `dependencies` list naming the subjects an
  item depends on, referenceable from the text by index as `[0]`, and an
  optional `priority` on the `P1`–`P5` scale of `todo/README.md` — so every
  future capability lands as an additive optional field under the same tag
- `mcp`: `cas_get` detects `vnd.fjs.note` blobs and reports them under their
  own media type

https://claude.ai/code/session_01UMivFEnpSE7JP8X3gP7zV8


o2alexanderfedin (approved):
The dependency-order risk I was most worried about — `vnd.fjs.lock` sorts, and here the array is index-referenced from the text — doesn't materialize. `encodeText = stringify(sort)` reuses lock's serializer, but `sort` is applied only inside `objectSerialize` (`fjs/media/json/serializer/module.f.mjs:174`); arrays go through `arraySerialize` untouched. Confirmed empirically: `{deps:['a','b']}` and `{deps:['b','a']}` encode to distinct byte strings with order preserved, so `[N]` references stay valid.

Property order is canonicalized (`dependencies` < `dialect` < `priority` < `text`), `undefined` optionals are dropped by `definedEntries`, and encode is idempotent across every adversarial input tried: astral chars, lone high/low surrogates (escaped `\ud83d`/`\ude00`), combining vs precomposed, `"`/`\`/tab/LF/CR/U+0001/U+007F, U+2028/2029, empty text, `[]`, and dependency strings containing quotes and lone surrogates. `decodeText(encodeText(x))` reproduced `x` in all 19 cases.

`validate()` rejects wrong dialect tag, missing/null/non-string `text`, non-array or non-string-element `dependencies`, and `'P0'`/`'p1'`/`''`/`null`/numeric `priority`; accepts empty text, `[]`, absent deps, and unknown fields (open rtti structs, deliberate). An out-of-range index like `text:'[7]'` with one dependency is silently allowed — by design, and the README says so.

`npx tsc` clean; `fjs test` → 2935 pass / 0 fail; `npm run cov` exit 0 with `fjs/media/note/module.f.mjs` at 100/100/100. The −29 lines are just `fjs/media/proof.f.mjs` dropping its hypothetical inline `noteSchema` fixture for the real import, assertions retained.

Two non-blocking notes: `dependencies: option(array(string))` admits two spellings of "depends on nothing" (`[]` vs absent) that hash to different CAS blobs, unlike `revision`'s `archived: option(true)` — the README owns that trade-off explicitly, so just noting it. And `fjs/media/note/todo/extend-note-format.md:8` still says "v1 is deliberately minimal — `dialect` + `text`", now stale since `dependencies` and `priority` shipped.

PR #1611, opened 2026-08-16
bnf/ll1: converge AST with descent backend

sergey-shandar:
## Summary

The LL(1) parser backend now produces ASTs that match the descent backend's shape. Rules are entered *before* their first symbol is consumed, so every rule invocation owns a node. This makes the AST a shared contract across backends rather than backend-specific.

## Key Changes

- **Dispatch model redesign**: Rules are now entered before consuming their first symbol. The dispatch map selects a variant's branch or repetition's next round by lookahead, but the rule itself builds a node.

- **Sequence handling**: Every item in a sequence, including the leading item, now owns its own node. Previously the leading item's node was absorbed into the enclosing dispatch.

- **Nullable items**: Items that match empty now produce empty nodes, making it visible in the AST that they were considered. Previously they disappeared entirely.

- **Variant tags**: A taken variant's tag now lands on the branch's own node, not the enclosing sequence's node.

- **Repetition**: `Repeat` rules are now matched iteratively into one flat node holding all matched items, instead of being compiled into right-recursive chains. This matches the descent backend's behavior.

- **Left recursion detection**: The builder now throws `left recursion …` when a rule is reachable from itself in first position, preventing infinite loops. The one valid self-reference is a `Repeat` of itself, which matches zero rounds when no lookahead is in the item's first set.

- **Proof updates**: The `descentEquivalence` proof group in `proof.f.mjs` now pins the shared AST shapes across both backends, replacing the old `descentDivergence` group that documented the differences.

- **Type refinements**: `_DispatchRule` now uses `empty: _DispatchBranch | undefined` instead of `emptyTag`, and `_DispatchBranch` replaces the old `_DispatchRuleCollection` to better reflect the new semantics.

- **Documentation**: Updated `README.md` files to reflect that the AST is now a shared contract. Removed the `ll1-ast-divergence.md` todo documenting the divergence, as it is now resolved.

## Implementation Details

The matcher remains an explicit-stack machine but now uses frame types that distinguish between sequence matching (`_SeqFrame`) and repetition matching (`_RepeatFrame`). Task types similarly distinguish between rule invocation (`_RuleTask`) and repetition rounds (`_RepeatTask`).

The dispatch builder detects left recursion by tracking which rules are currently being computed in `current` and throwing when a rule references itself in first position. Repetitions of themselves are allowed and produce empty dispatch maps.

`types/btree/set` gains one proof case (replace the first value of a two-value leaf): the old ll1 dispatch builder was the only code in the suite reaching that branch, so the branch moves to the module's own proof to keep aggregate coverage at 100%.

## Verification

- `npx tsc` clean; `fjs test` 2964 pass, 0 fail; `npm run cov` 100/100/100, exit 0.
- The `descentEquivalence` proof group pins one AST per grammar and asserts it against both backends; the `parser` group asserts the whole JSON-like test grammar produces the same AST as `bnf/descent`, input by input.

Changelog:
- **BREAKING CHANGES:** `bnf/ll1`: the AST now matches `bnf/descent`'s — a node per rule invocation, entered before its first symbol is consumed, and a flat node per `Repeat` — so the AST is one contract across backends. `dispatchMap` stores per-rule first sets instead of consumed-symbol rule chains and throws `left recursion …` at build time for a left-recursive grammar

https://claude.ai/code/session_016RdPMsNDdTA2p8xzmDQ9Zh


o2alexanderfedin (approved):
All six divergences from `ll1-ast-divergence.md` converged — verified by execution against the deleted table row by row, including row 5: `[[A,B],C]` → `((("A") ("B")) ("C"))` vs `[A,B,C]` → `(("A") ("B") ("C"))`, now distinguishable. Broader fuzz over 17 grammars × 3058 inputs: 1883 accepted pairs, **0 AST differences** against descent.

The accepted language is unchanged — and where it moved, it moved toward correct. Same corpus, ll1@head vs ll1@main: 55,062 checks, 9 accept flips, every one a case where main was wrong and head now matches descent (notably `fsc/json` rejected `{"a":1,"b":[1,2,{"c":null}]}` on main and accepts it now). Zero accept/reject disagreement with descent anywhere.

Termination holds. `repeat0Plus(option('a'))`, `repeat0Plus('')` and repetition-of-repetition are now rejected at build with `left recursion …` — these are genuinely left-recursive (descent hangs on them on both revisions), so this turns a silent zero-round match into a diagnostic. All-nullable variants and nullable sequence prefixes terminate; djs, fsc and testlib all still build, so no false positives.

Proofs discriminate: `bothBackends` pins a string and asserts against both backends, and mutating the node-building back to the old shape fails exactly the right ones — spliced leading item breaks `leadingItem`/`nesting`/`skippedOption`/`takenOption`/`repetitionInNullablePrefix`; dropped branch tag breaks `variant`/`takenOption`; flattened repetition breaks `repetition`. `btree/set/proof.f.mjs` does belong here — reverting it drops that module to 97.37% branch and fails `npm run cov`.

**Merge with #1608 is clean.** Since it landed, I re-checked: ll1 still imports `leafAt`, `mrFail`, `mrSuccess`, `physicalIdx`, `symbolAt` from `../matcher/module.f.mjs` with no local reimplementation, the 669 todo is properly gone, `descent` is unchanged apart from README prose, and 30 EOF-grammar cases give ll1 the same success flag and remainder as `origin/main`. Results are byte-identical between the pre- and post-merge heads.

`npx tsc` clean; suite 2964 pass / 0 fail; coverage 100/100/100.

Two minor: the body has no `Changelog:` section (AGENTS.md §5), and `1611.md` embeds a PR link that `changelog/README.md` says to omit.

o2alexanderfedin (approved):
Re-approving at the new head. The delta is `changelog/unreleased/1611.md` alone (+5/−15), no code — `npx tsc` exit 0, `fjs test` 2964 pass / 0 fail.

Both my notes are addressed: the embedded PR link is gone and the body now carries a `Changelog:` section matching the file verbatim.

Nothing load-bearing was lost in the trim — the `**BREAKING CHANGES:**` marker survives, and so do the two things a user actually needs: the AST-shape change and the build-time `left recursion` throw, which is what produced the 9 accept-flips I measured. Only illustrative detail went. At 388 chars it's over the README's ~250 hint but shorter than 1608 (908) and 1607 (623), so it's in line with practice.

PR #1614, opened 2026-08-16
text/utf16: out-of-range guard emits errorMask, not 0xffffffff

sergey-shandar:
Implements `fjs/text/utf16/todo/decoder-oob-sentinel.md` (deleted here).

## The problem

The UTF-8 and UTF-16 decoders share the `code_point` `errorMask` contract, but their *input-domain guards* — the first check each per-unit step runs on a raw code unit — disagreed on the sentinel:

```js
// utf8ByteToCodePointOp — shared constant
if (byte < 0x00 || byte > 0xff) { return [[errorMask], state] }

// utf16ByteToCodePointOp — bare magic literal
if (!u16(word)) { return [[0xffffffff], state] }
```

Verified on `main`:

```
utf8  toCodePointList([-1, 256])    -> [2147483648, 2147483648]
utf16 toCodePointList([-1, 65536])  -> [4294967295, 4294967295]
```

`0xffffffff` is a second, un-named "invalid" value that only *happens* to have bit 31 set, so downstream `(cp & errorMask) !== 0` still fired. The JSDoc documented the literal as the contract ("an error code `0xffffffff` is returned"), cementing an accidental value. Every other error path in the same function goes through `… | errorMask`.

## The change

- The guard emits `[[errorMask], state]`, matching UTF-8.
- JSDoc rewritten to state the shared shape rather than the literal, with a comment on why the guard emits the *bare* mask: the word is out of domain, so there is no payload to tag — unlike every other error path below it.
- Three proof expectations pinning `4294967295` move to `2147483648`.

Marked `**BREAKING CHANGES:**` in the changelog. The type signature is unchanged and `& errorMask` detection is unaffected, but `0xffffffff` was named in the JSDoc as the returned error code, so a consumer comparing against it observably breaks.

**No `outOfRange` helper.** The TODO left this open ("decide once both guards are identical"). Decided against: only the *returns* match — the conditions differ (`byte < 0x00 || byte > 0xff` vs `!u16(word)`) — and the return is a single expression whose entire content is `errorMask`, which is already the shared constant. A helper would put a name in front of it, not share anything new.

## Something the TODO got wrong

The TODO asserted the two guards are "structurally identical — raw unit out of range → emit one error unit, pass `state` through unchanged". True of the **return**, not of the **condition**: UTF-8's is strictly weaker. `U8` is just `number`, and a bare range comparison admits fractions:

```
toCodePointList([1.5])   -> [1.5]          <- a fraction reported as a valid code point
toCodePointList([0.5])   -> [0.5]
toCodePointList([200.5]) -> [-2147483448]  <- tagged, but fractional payload
toCodePointList([255.5]) -> [2147483648]   <- right, but only because 255.5 > 0xff
```

UTF-16 already handles this: `u16` is `Number.isInteger(i) && isInU16Range(i)`, with a doc comment explaining that the BMP/surrogate predicates partition only the *integers*, so a fraction falls between two ranges. UTF-8's `contTag`/`isLeadByte` dispatch has the same property and no integer check; `fjs/text/utf8/proof.f.mjs` has no fractional case, which is why it survived.

Out of scope here — it changes a decoder's accepted domain and wants its own proof cases — so it is filed as `fjs/text/utf8/todo/byte-guard-accepts-non-integers.md` rather than folded in.

## Also in this PR

`fjs/text/utf16/todo/word-classifier-dedup.md` quotes line numbers in the function I edited, so they shifted. While fixing them I found its quoted code was **already stale on `main`**: it shows a three-branch recovery arm, but the third branch was removed by an earlier PR that proved it unreachable. So its premise ("the two blocks differ only in whether the flushed error is prepended") no longer holds literally — the fresh arm is three-way, the recovery arm two-way. Refreshed the quote and added a note that its proposed `restart` extraction is a behavioral claim to prove with a differential, not a textual dedup. Not implemented.

## Verification

- `npx tsc` — clean.
- `node fjs/module.mjs test` — 2957 pass, 0 fail.
- `npm run cov` — 100.00% overall; `text/utf16`, `text/utf8`, `text/code_point` each 100.00% lines/branches/functions.
- Differential against `main`, 529 cases: 426 in-domain (every boundary of every range the decoder discriminates on and its neighbours, all pairs, plus all 216 three-unit combinations over the surrogate-relevant subset), 90 out-of-domain (negative, fractional, above range, `NaN`, each alone and interleaved with a pending surrogate), and 13 encoder cases. Result: the diff is exactly the 90 out-of-domain rows, and `sed 's/4294967295/2147483648/'` over `main`'s output makes it byte-identical to the branch — so the sentinel value is the only behavior change.
- The harness asserts 344 of the in-domain rows carry a tagged error (`-2147…`), so "identical" is a statement about exercised error paths, not just clean decodes. It also fails on any empty/`THREW`/`undefined` cell, so a crash on both sides cannot masquerade as a match.

## Coordination

`fjs/text/todo/666-utf16-encode-errormask.md` (the TODO flagged as adjacent) is about `codePointToUtf16`, the **encoder** — untouched here, and covered unchanged by the 13 encoder rows in the differential. No ordering constraint either way.

## Changelog

- **BREAKING CHANGES:** `text/utf16`: the decoder's out-of-range guard emits the
  shared `errorMask` instead of a magic `0xffffffff`, matching the UTF-8 decoder.
  A code unit outside `[0x0000, 0xFFFF]` or non-integer now decodes to
  `2147483648`, not `4294967295`.


o2alexanderfedin (approved):
No consumer can observe the sentinel change, which was my main question given this is an output change rather than a refactor.

The utf16 out-of-range guard is unreachable from every in-repo path: the only caller of utf16's `toCodePointList` is `stringToCodePointList` (`module.f.mjs:290`), fed by `stringToList`, whose `charCodeAt` can only yield integers in `[0, 0xFFFF]`. Repo-wide, nothing compares against `0xffffffff`/`4294967295` outside docs and the historical `changelog/0.45.0/1540.md`, and the sole downstream inspector — `fjs/media/json/serializer/module.f.mjs:77` — tests exactly `(codePoint & errorMask) !== 0`.

The amplification I was worried about is real but unreachable, and measured: for a hypothetical raw `List<U16>` caller, `[-1]` now gives `codePointToString` `'\u0000'` instead of `'\uffff'`, and utf8 `fromCodePointList` gives `[errorMask]` instead of a bogus `[247,191,191]` — arguably a fix in itself. Nothing externally visible: no CAS, hash or wire path reaches the guard, since all of those go through strings.

Differential over 40+ cases (negatives, `1.5`, `NaN`, `±Infinity`, `56319.5`, all surrogate boundaries, valid pairs, unpaired in both positions, interleaved, empty): the only differing rows are the out-of-domain ones, `4294967295` → `2147483648`. Everything else byte-identical.

The new `byte-guard-accepts-non-integers.md` claim is true and reproduces verbatim on both branches (`[1.5] -> [1.5]`), so filing rather than fixing leaves utf8/utf16 no more inconsistent than before. Deleted todo matches the implementation, and `word-classifier-dedup.md`'s line refs all check out.

`tsc` clean; 2957 pass / 0 fail; `npm run cov` 100/100/100. Changelog carries `**BREAKING CHANGES:**`, body has the section.

PR #1615, opened 2026-08-16
effects/io: migrate the generator programs and cas to Io composition

sergey-shandar:
Stage 4 of the IoEffect migration
(`fjs/effects/todo/io-effect-migration.md`), first tranche. Stage 4 is
explicitly module-by-module — "migrate consumers module by module… validate
`npx tsc` and `fjs t` after each migration PR" — so this PR takes the modules
that migrate cleanly and adds the combinators the rest will need. **It does not
finish the stage**; see *What is left* below.

Migrated: `fjs/ci`, `fjs/ci/nix`, `fjs/nanvm/update`, `fjs/dev/update`,
`cas list`, and `fjs/cas`.

## Four combinators, each required by a consumer here

Not mirrored from the raw API speculatively — the stage's own rule:

- **`history` / `historyStep`** — `syncMcp` needs the source text after the
  directory has been created. The raw `historyStep` cannot serve: it would
  carry each link's `Result` into the tuple, so every later link would
  destructure results it has no intention of handling. The Io one carries `ok`
  values and short-circuits, which is what keeps a fallible chain flat.
- **`foldStep` / `forEachStep`** — `nixFlakes` writes one flake per job, `cas
  list` prints one line per hash. The raw `forEachStep` runs every item
  whatever each one answered, and its `void` accumulator *hides* that: a
  `Result`-valued effect is accepted silently in a `void` return position.

`items` stays a **raw** effect in both folds. The two shapes that need them — a
list the caller already holds and one an infallible operation produced — have
no error channel, so requiring one would mean lifting at every call site to
express something no consumer has.

## What the migration showed

- **Programs end in `exitStep`.** A chain now runs to its tail, so a failed
  write is the host's message on stderr and exit code `1` rather than a stack
  trace. `fjs/dev/update`'s missing-source proof moved out of its `throw` group
  for exactly that reason — it asserts the message and the code now.
- **`catchStep` earns its keep in `cas`'s `gcStage`**, whose error channel is
  now `never`. The claim that a best-effort sweep cannot fail the upload it is
  piggy-backed on is in the type: it recovers per item (one undeletable file no
  longer stops the sweep) and again around the whole sweep (a fresh store has
  no `_stage/` yet). A new proof pins it — an `rm` that fails on an expired
  staging file leaves the upload succeeding.
- **`resultStep` is what cleanup-on-failure wants.** `writeImpl` deletes its
  partial staging file and *then* reports; propagation alone would skip the
  delete.
- **`okStep` has no consumer left but the Io `step` itself.** Stage 5 can
  inline it and drop the export; this PR leaves it exported, since stage 2
  promised it to raw consumers for the duration of the migration.

Guidance updated for the new semantics: `fjs/AGENTS.md` §3.4 now opens with the
Io layer, and the `fjs/effects` module header says which of its rules a
fallible chain should follow through the Io combinators instead.

## What is left

Still composing through `unwrapStep`, each its own follow-up PR:
`fjs/emergent_testing`, `fjs/cas/evo`, `fjs/mcp`, `fjs/protocol/mcp` (and its
stdio transport), `fjs/dev`, and the proofs that drive them. The todo lists
them, and every remaining site is greppable by that one name.

## Checks

`npx tsc` clean; `fjs t` 2948 pass / 0 fail; `npm run cov` exit 0 at 100%
line/branch/function; `npm run ci-update` regenerates nothing.

Changelog:
- `effects/io`: new `history`, `historyStep`, `foldStep` and `forEachStep` —
  the fallible chain keeps its earlier values, and a fold stops at the first
  error instead of running every item
- `ci`, `dev/update`, `nanvm/update`, `cas list`: a failed write is reported on
  stderr with exit code 1 rather than thrown
- `cas`: the staging upload composes through `fjs/effects/io`. A failed `now`
  or `randomInt` propagates instead of panicking, and a failed sweep can no
  longer fail the upload it runs before

---


o2alexanderfedin (approved):
Ran every migrated program on both trees. `npm run ci-update` and `npm run dev-update` exit 0 and leave the tree clean on both sides, so `.github/workflows/ci.yml`, all `nix/generated/*/flake.nix`, `.vscode/mcp.json` and `generated.rs` match the committed baseline either way. `cas add`/`list`/`get` against a temp `$HOME` store gave identical hashes, identical listing order and identical restored bytes, with `diff -r` over the two `.cas` trees reporting nothing.

Error paths were the thing to get right, and they hold. Forced EACCES on `ci.yml`, on `nix/generated/node24/flake.nix`, on `.vscode/mcp.json` and on `generated.rs`; a missing `.copilot/mcp.json`; `cas add` of a missing file; `cas add` with `.cas` read-only; `cas list` with `.cas` chmod 000; `cas get` of an absent hash — **every one exits 1 on both sides**, nothing silently became exit 0. Main printed an uncaught-exception dump where head prints the host message (`EACCES: permission denied, open '…'`).

One error-path difference, non-blocking and an improvement: with `.cas/_stage` replaced by a regular file, main reported `ENOTDIR … open <stage path>` because raw `step` discarded the `mkdir` result and carried on, while head reports `EEXIST: mkdir '…/_stage'` — the failure surfaces one link earlier. Same exit code, same store state.

All four combinators have real consumers in this PR, per the stage's own rule — `history`/`historyStep` in `dev/update` and `cas` (×2), `foldStep` in `cas` `random256`, `forEachStep` in `cas` `gcStage`, `cas list` and `ci/nix`. Short-circuit verified by mutation: breaking `foldStep`'s fails `proof.foldStep.shortCircuits` and `proof.forEachStep.stopsAtTheFirstError`; breaking `historyStep`'s fails `proof.historyStep.propagates` and `propagatesFromHead`.

`npx tsc` clean; `fjs t` → 2948 pass / 0 fail; `npm run cov` exit 0 at 100% across the board. Todo marks Stage 4 in progress with the remaining modules listed.

sergey-shandar:
On the code-quality finding at `fjs/dev/update/proof.f.mjs:35` — "superfluous argument passed to function main": the argument is required, and I've left the call as it is.

`main` is a `NodeProgram`, which is `(options: NodeProgramOptions) => Effect<O, number>`. The implementation ignores the options and so is written `() => exitStep(syncMcp())`, but the *declared* type takes one, and `tsc` rejects `main()` for that reason. The bot is comparing against the arrow's arity rather than the JSDoc type. The pre-existing `main` proof two lines up (and `fjs/nanvm/update`'s) passes `defaultNodeProgramOptions` for the same reason; the new line just matches them.

Calling it as a `NodeProgram` is also the point of the proof: it checks the exit code and the stderr line a *runner* would see, which is what changed in this PR — a missing source used to be a panic and is now exit 1 with the host's message.

---

PR #1616, opened 2026-08-16
effects: file the remaining stage 4 consumer migrations

sergey-shandar:
Documentation only — a `todo/` file, no code.

Stage 4 of the IoEffect migration is module-by-module by design. The first
tranche ([#1615](https://github.com/functionalscript/functionalscript/pull/1615))
took the modules whose chains migrate without touching a published type. What
is left — `fjs/emergent_testing`, `fjs/dev`, `fjs/cas/evo`,
`fjs/protocol/mcp` (+ stdio) and `fjs/mcp` — each changes one: `Reporter<O>`,
`Evo<O>`, the MCP handler and transport shapes. That is why they are separate
reviews rather than one sweep.

Until now that worklist was a paragraph in the parent plan and a name to grep
for. This turns it into an issue: the order (later modules depend on types the
earlier ones settle), the ripple each one causes, and the tail each chain ends
at — `exitStep` for the test runner, an MCP error response for the tool
handlers, a JSON-RPC error for the transport.

Two things the file is explicit about, since both are easy to get wrong later:

- `Evo.add` already returns a domain `Result<Hash, string>`, which must **not**
  be collapsed into the effect's error channel — it is returned data, like
  `SandboxResult.result`.
- The goal is a *considered* policy at every site, not zero panics.
  `unwrapStep` stays where a module genuinely has no answer to a failure; it
  just has to say so in a comment.

It also records the exit condition — delete the file when the grep returns only
commented sites, and check off Stage 4 in the parent plan — and the fact that
Stage 5 inherits one loose end from this work: `okStep`'s only remaining
consumer is the Io `step` itself.

Changelog: none

---


o2alexanderfedin (commented):
Two of the type signatures the file quotes don't match the declarations.

**`Reporter<O>`** — `fjs/effects/todo/io-effect-consumer-migration.md:38` says its members return `Effect<O, void>`, but that holds for only two of three. `fjs/emergent_testing/types.ts:50`:

```
readonly test: (file: string, path: Path, set: TestEntry) => Effect<O, SandboxResult<unknown>>
```

The "`void` return position swallows a `Result` silently" hazard is real for `result` and `summary`; `test` isn't subject to it.

**`revision`** — `:47` groups it with `list` and `head` as `Effect<MemOp, …>`. `fjs/cas/evo/types.ts:123` declares it `Effect<O | MemOp, Result<RevisionData, string>>`. Two knock-ons: it's served by `readRevision(cas)` and never reads the cache slot, so propagating the slot doesn't make it `IoEffect<MemOp, …, NotImplemented>`; and it already returns a domain `Result` exactly like `add`, so the "must **not** be collapsed into the effect's channel" rule two lines down applies to `revision` too, not just `add`.

Minor: `:37` says `fjs/emergent_testing` (9 sites) — there are 8 `unwrapStep` call sites in `module.f.mjs` (121, 122, 130, 136, 179, 202, 323, 347); the 9th grep hit at 342 is a comment. Every other module's count matches exactly.

Otherwise checked out: the grep command reproduces the module list on `main` as written, the #1615 attribution is accurate, `okStep`'s only non-proof consumer is Io `step`, all referenced paths exist, and `Changelog: none` is right for a todo-only PR. `npx tsc` clean, `fjs t` 2964 pass / 0 fail.

sergey-shandar:
All three corrected in fad67fa — checked each against the declarations, and all three were right.

**`Reporter<O>`** — `test` returns `Effect<O, SandboxResult<unknown>>`, so the `void`-swallows-a-`Result` hazard is `result` and `summary` only. The file now says which two members it applies to and that `test` is exempt.

**`revision`** — two corrections, since the second changes the plan rather than just the prose:

- It is `Effect<O | MemOp, Result<RevisionData, string>>`, and `readRevision(cas)` never reaches the cache slot, so propagating the slot does **not** move it. The file now says it stays put until the memoization its `MemOp` is reserved for arrives.
- Its domain `Result` is under the same do-not-collapse rule as `add`'s. That rule now names both, with `SandboxResult.result` as the precedent.

**Counts** — 8 `unwrapStep` call sites in `fjs/emergent_testing`; the 9th hit was the comment at 342. While checking, `fjs/dev` was off in the other direction: 4 bare `unwrap`s, not 3 + an `all`. It now names the operations they guard (a `readdir`, an `import`, two `all`s), which is more use to whoever picks it up than the number.

Rebased onto main now that #1615 has merged. `npx tsc` clean; `fjs t` 2976 pass / 0 fail.

---

o2alexanderfedin (approved):
All three corrections land — approving.

- `Reporter<O>` now splits the members: `result`/`summary` carry the `void`-position hazard, `test` returns `Effect<O, SandboxResult<unknown>>` and doesn't. Matches `fjs/emergent_testing/types.ts:48-50`.
- `revision` is now separated from `list`/`head` with the right reason: served by `readRevision(cas)`, never reaches the slot, so it doesn't move until the memoization its `MemOp` is reserved for arrives — and it's grouped with `add` under the "must not be collapsed" rule. Matches `fjs/cas/evo/types.ts:110,112,114,123` and the `evo` literal in `module.f.mjs`, where `list`/`head` compose `unwrapStep(read(cacheKey))` and `revision` doesn't. The `SandboxResult.result` analogy holds — it's `Result<T, unknown>` at `fjs/effects/node/types.ts:256`.
- 9 sites → 8. The 9th grep hit is the comment at line 342.

Re-checked the rest of the file against this tree rather than trusting the edit: `cas/evo` 5 (233, 240, 241, 570, 574), `protocol/mcp` + stdio 5 (293, 296, 317, 342 + stdio 83), `mcp` 1 (87), the four proof files exist and contain `unwrapStep`, and the revised `fjs/dev` "(4 bare `unwrap`s: a `readdir`, an `import`, and two `all`s)" matches lines 61, 81, 90, 124, all reachable from `loadModuleMap`. `ToolEntry<O>.handle` is one homogeneous `Effect<O, ToolsCallResult>`, so "every `fjs/mcp/evo` tool handler has to answer for it" is right even though `evo_revision` itself doesn't move.

Both main catch-ups are clean: `git diff 0effe0f d11c9c5` is the one doc file and nothing else. `npx tsc` 0, `fjs t` 2976 pass / 0 fail, `cov` 100/100/100, `ci-update` leaves the tree clean. `Changelog: none` still correct for a todo-only PR.

PR #1617, opened 2026-08-16
tsconfig: enable the four low-cost strictness flags

sergey-shandar:
Implements step 1 of [`todo/tsconfig-strict-flags.md`](https://github.com/functionalscript/functionalscript/blob/main/todo/tsconfig-strict-flags.md): turn on `noImplicitReturns`, `noImplicitOverride`, `isolatedModules` and `noFallthroughCasesInSwitch`, so properties the tree already has stop being unenforced.

The first three were clean, as the issue measured. `noFallthroughCasesInSwitch` was not: it had drifted from 0 to 2 sites since the measurement. Both are switches whose exhaustiveness TypeScript cannot see, so the end of the enclosing `case` was reachable and control would have fallen into the next one — in both places a silent continuation into an unrelated state rather than a failure. That is the argument for the flag, so they are fixed here rather than deferred:

- `djs/tokenizer` — `stringDecodeScan`'s `escape` state switches a code point against the `text/ascii` constants, which are `number` rather than literal types. No such switch can ever be exhaustive to TypeScript, so the `\u` clause becomes a `default`; the assertion above it already establishes that `\u` is the only case left.
- `types/btree/set` — `switch (x.length)` over a tuple union, where TSGO does not narrow the length to a literal union ([typescript-go#4613](https://github.com/microsoft/typescript-go/issues/4613)). Fixed the way the same function already worked around that regression at `case 3`: bind the length, `assert` its two possible values, switch on the binding.

The issue file stays — three of its four steps are still open — but its status moves to `wip` and its measurement table is refreshed against TypeScript 7.0.2 on this branch (`noUnusedParameters` 8, `noPropertyAccessFromIndexSignature` 33, `noUncheckedIndexedAccess` 212, `noUnusedLocals` 218 = 137 `TS6196` + 81 `TS6133`), records the two fallthrough fixes and their causes, and notes that counts drift and should be re-measured before the next step. `todo/strict-static-analysis.md` is updated to match.

`npx tsc`, `npm test` (2976 proofs) and `npm run cov` (100% lines/branches/functions) pass; `npm run ci-update` produces no diff.

Changelog: none

### DIFF
--- a/fjs/djs/tokenizer/module.f.mjs
+++ b/fjs/djs/tokenizer/module.f.mjs
@@ -393,7 +393,12 @@ const stringDecodeScan = (cp, state) => {
                 case latinSmallLetterN: return [[lf],        { kind: 'normal' }]    // \n → line feed (LF)
                 case latinSmallLetterR: return [[cr],        { kind: 'normal' }]    // \r → carriage return (CR)
                 case latinSmallLetterT: return [[ht],        { kind: 'normal' }]    // \t → horizontal tab (HT)
-                case latinSmallLetterU: return [null, { kind: 'unicode', acc: 0, count: 0 }]  // \u → start 4 hex digits
+                // `\u` is the only case the assertion above leaves. It is a
+                // `default` rather than a `case latinSmallLetterU` because the
+                // ASCII constants are plain `number`s, so the switch can never
+                // be exhaustive to TypeScript and the clause would fall through
+                // into `unicode` below.
+                default: return [null, { kind: 'unicode', acc: 0, count: 0 }]  // \u → start 4 hex digits
             }
         }
         case 'unicode': {
--- a/fjs/types/btree/set/module.f.mjs
+++ b/fjs/types/btree/set/module.f.mjs
@@ -80,7 +80,11 @@ const nodeSet = c => g => node => {
             case 2: {
                 // insert
                 const value = g(null)
-                switch (x.length) {
+                // TODO: remove after TSGO fix the regression.
+                const xL = x.length
+                // See https://github.com/microsoft/typescript-go/issues/4613
+                assert(xL === 1 || xL === 2)
+                switch (xL) {
                     case 1: { return [[x[0], value]] }
                     case 2: { return [[x[0]], value, [x[1]]] }
                 }
--- a/todo/strict-static-analysis.md
+++ b/todo/strict-static-analysis.md
@@ -14,7 +14,7 @@ and `cargo test` in both profiles on every one of them.
 The JavaScript half runs `npx tsc` and the test suites (`fjs test`,
 `node --test`, `deno task cov`, `bun test --coverage`) — and nothing else. No
 lint, no formatting check, no unused-code check, no package-correctness check.
-`tsconfig.json` leaves eight checking flags off. There is no equivalent of
+`tsconfig.json` leaves four checking flags off. There is no equivalent of
 `-D warnings` anywhere on this side.
 
 That asymmetry is the gap to close, using well-known tools rather than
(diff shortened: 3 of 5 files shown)


o2alexanderfedin (commented):
Both source hunks check out and the branch is green, but two of the four refreshed counts don't reproduce.

On this head with TypeScript 7.0.2:

```
npx tsc --noUncheckedIndexedAccess  -> 212 errors   (doc: 211)
npx tsc --noUnusedLocals            -> 218 errors   (doc: 214)
                                       137 TS6196 + 81 TS6133   (doc: 133 + 81)
```

`todo/tsconfig-strict-flags.md:37,38,61,78-80`, and the PR body's "refreshed against TypeScript 7.0.2 on today's tree". `noUnusedParameters` 8 and `noPropertyAccessFromIndexSignature` 33 are correct. Same numbers on `origin/main`, so it isn't a rebase artifact — the `TS6196` count is off by four.

Also `todo/tsconfig-strict-flags.md:4` still says `**Status:** open`, but step 1 of its own task list is now checked; `todo/README.md` calls that `wip`.

Verified clean, for the record:

- **`djs/tokenizer:396`** `case latinSmallLetterU:` -> `default:`. The unchanged `assert` above restricts `cp` to nine constants, eight of which the preceding cases take, so `default` is reachable only by `latinSmallLetterU` and returns the identical tuple. Outside the nine, `assert` throws before the switch in both versions.
- **`types/btree/set:80-90`** bind + `assert` + `switch (xL)`. `i === 2` selects `_FirstLeaf1`/`_FirstLeaf2` only (`_FirstBranch3` is pinned to `1`, `_FirstBranch5` to `1 | 3`), and `find` only emits `done(i3(node[0]))` for length-1 and `done(i5(node))` for length-2 — so `x.length ∈ {1,2}` and the new assert can't fire. The 100% branch gate proves both arms are exercised. Case bodies unchanged.
- Restoring `main`'s two files under this tsconfig yields exactly two `TS7029`s at exactly these two sites, so nothing else was silently touched.
- 270 `.d.mts` files emitted from this branch and from `origin/main` are byte-identical — no published API change, which is also why `Changelog: none` is right here.
- `npx tsc` 0, `fjs t` 2976 pass / 0 fail, `npm run cov` 100/100/100, `prepack` 0, `ci-update` leaves the tree clean.

sergey-shandar:
You're right on both counts — fixed in 0830f70.

The cause was mine: I measured the table before rebasing this branch onto `main`, so the numbers were taken on the pre-#1615 tree. Re-measured on this head, TypeScript 7.0.2:

```
noUncheckedIndexedAccess -> 212
noUnusedLocals           -> 218   (137 TS6196 + 81 TS6133)
```

Those are now in `todo/tsconfig-strict-flags.md` and in the PR body, along with the two downstream references to the old figures (`shrinks the 211`, `fix the 133 stale @import entries`). `noUnusedParameters` 8 and `noPropertyAccessFromIndexSignature` 33 were unaffected.

`**Status:**` is now `wip`, per `todo/README.md`.

Thanks for the independent verification of the two source hunks, especially the `find` argument for why `x.length ∈ {1,2}` at `i === 2` — that's the part the type system can't currently show, and it's worth having in the record.


---

o2alexanderfedin (approved):
Counts fixed — approving.

Re-measured on `0830f70`: `--noUncheckedIndexedAccess` 212, `--noUnusedLocals` 218 = 137 `TS6196` + 81 `TS6133` (those two codes are the whole population). Matches the table, the prose, and the task list. The PR body's "refreshed against TypeScript 7.0.2" sentence now agrees too, and a repo-wide grep for surviving `211`/`214`/`133` turns up only unrelated hits. `Status: wip` matches `todo/README.md` with step 1 of four checked.

`git diff 6a30c0b 0830f70 --stat` is the one `.md` file only, so the two source hunks I verified last round are untouched. Branch green: `npx tsc` 0, `fjs t` 2976 pass / 0 fail, `cov` 100/100/100, `ci-update` leaves the tree clean.

PR #1618, opened 2026-08-16
media/json: export colon; djs names the shared MapEntries

sergey-shandar:
Implements `fjs/media/json/todo/serializer-shared-atoms.md` (deleted here).

## The problem

`fjs/djs/serializer` already imports the serializer atoms it shares with JSON (`objectWrap`, `arrayWrap`, `stringSerialize`, `numberSerialize`, `nullSerialize`, `boolSerialize`), but two pieces of the same property-serialization vocabulary were re-declared instead:

1. **`colon`** — `const colon = [':']` in both `fjs/media/json/serializer/module.f.mjs` and `fjs/djs/serializer/module.f.mjs`, used in the same `stringSerialize(k), colon, f(v)` fragment.
2. **The entry-sorter type** — djs spelled `(entries: List<ObjectEntry<Unknown>>) => List<ObjectEntry<Unknown>>` inline.

## The change

- `colon` is exported from `fjs/media/json/serializer`, beside `objectWrap`/`arrayWrap`; djs imports it and drops its copy.
- `fjs/djs/types.ts` gains `export type _MapEntries = TreeMapEntries<Primitive>`; the serializer imports it and drops its local typedef (along with the now-unused `ObjectEntry` import).

**The TODO was half-stale, and its proposal would have added a duplicate.** It cited `fjs/media/json/module.f.mjs:70` for `colon` and `:74-76` for local `Entries`/`MapEntries` types. Neither is there any more — `colon` moved into `serializer/`, and the generic sorter type already exists as `TreeMapEntries<P>` in `fjs/media/json/types.ts`. So instead of adding the new `MapEntries<P>` to `json/serializer` that the TODO proposed, djs instantiates the alias that already exists, in `types.ts` where the type belongs.

That lines it up exactly with `fjs/media/json/extended/types.ts`, which already does `export type _MapEntries = TreeMapEntries<Primitive>` — the existing precedent, now followed by a second family.

## The soundness question, and what it settles

Reusing `TreeMapEntries` at djs's leaf set is only valid if djs's `Unknown` really is `Tree<djs.Primitive>`. djs writes its containers out by hand, so this is a claim, not a definition. It holds — `fjs/djs/types.ts` now pins it:

```js
type _Unknown = Assert<Equal<Unknown, Tree<Primitive>>>
```

Checked that the pin is load-bearing rather than vacuously true: substituting `Tree<JsonPrimitive>` (dropping `bigint | undefined`) makes `tsc` fail with `Type 'false' does not satisfy the constraint 'true'`.

That equality is the central premise of the open `fjs/djs/todo/663-json-djs-tree-type.md`, whose remaining task is "re-express `fjs/djs`'s aliases using the same shared generic tree". Since it now compiles, that task is a **rename, not a redefinition** — it cannot change any type's meaning. #663 is updated to record the pin, and to point at `extended/types.ts` as the worked precedent for the rest. Not implemented here; the aliases are left as they are.

## Verification

- `npx tsc` — clean.
- `node fjs/module.mjs test` — 2964 pass, 0 fail.
- Declaration emit inspected for the new public surface: `export declare const colon: List<string>` and `export type _MapEntries = TreeMapEntries<Primitive>`.
- Differential against `main`, 315 rows: 35 values (atoms, nested containers, duplicate-reference graphs that trigger const hoisting, `undefined`, `bigint`) × `json.stringify` / `extended.stringify` / `djs.stringify` / `stringifyAsTree` / `countRefs` × `sort` and `identity`. Byte-identical.
- Harness guards: every serializer's return is checked to actually be a `string` (its first version conflated a legitimate `undefined` djs output with a broken call — caught by the guard and fixed by recording `typeof` alongside the value); 76 rows are asserted to serialize a property and 6 to hoist a const, so the paths that used the duplicated `colon` are exercised rather than assumed. The colon guard was initially vacuous because the `string:` type prefix I added put a colon in every row — corrected to strip the prefix first.

## Also in this PR

Two other TODOs cross-referenced the deleted file (`fjs/media/json/todo/stringify-sorted-canonical.md`, `fjs/media/nix/todo/serializer-validation-split.md`). Both references rewritten to point at `fjs/media/json/serializer/module.f.mjs` and note the question is settled for these atoms, so nothing dangles.

## Changelog

- `media/json`: the serializer exports `colon`, the `key : value` separator,
  which the DJS serializer now shares instead of declaring its own

---


o2alexanderfedin (approved):
Approving. Serializer output is unchanged — that was the thing worth proving here, since a shared `colon` would move every content hash in the system.

Differential harness across two worktrees (merge-base `c101940` vs this head), running `json.stringify`/`serialize`, `extended.stringify`/`serialize`, `djs.stringify`/`stringifyAsTree`/`serializeWithoutConst`/`countRefs` over a 53-value corpus × `sort`/`identity` = 734 rows per side. Corpus included keys *containing* colons (`"a:b"`, `":"`), the empty key, escape-needing keys, unicode keys, negative zero, `1e21`/`1e-7`, lone surrogates, bigints to `2n**200n`, depth-50 nesting, and duplicate-reference graphs that trigger `const c0 = …` hoisting.

**733/734 identical.** The one delta is the intentional new `colon` export. 187 rows exercise a `"key":value` fragment and 12 exercise const hoisting, so the paths that used the duplicated atom are covered rather than assumed. `serialize` rows were joined with a `U+0001` separator instead of `''`, so chunk boundaries were compared too — the colon still arrives as its own chunk in the same position.

Also checked:

- `colon` was `[':']` in both places at the merge base; the surviving one is the same literal, only gaining an explicit `@type {List<string>}`.
- `MapEntries` — `Assert<Equal<Old, New>>` compiles against the deleted spelling (negative control confirms `Equal` isn't vacuous). No widening, no lost `readonly`. `djs.Object ≡ TreeObject<Primitive>` and `djs.Array ≡ TreeArray<Primitive>` hold as the 663 doc claims.
- Declaration diff vs the merge base is exactly two entries: `+export declare const colon: List<string>` and `_MapEntries` relocating into `fjs/djs/types.ts`.
- Deleting `serializer-shared-atoms.md` isn't an overreach — at the merge base `media/json` had already dropped its local `colon`/`MapEntries`, so djs's was the only live duplication and both halves are now gone. The 663 doc's claim reproduces verbatim: substituting `Tree<JsonPrimitive>` yields `fjs/djs/types.ts(30,24): error TS2344: Type 'false' does not satisfy the constraint 'true'`. No dangling references to the deleted file.
- `npx tsc` 0, `fjs t` 2964 pass / 0 fail, `cov` 100/100/100, `prepack` 0, `ci-update` leaves the tree clean. Changelog entry is 144 chars, conformant, and character-identical to the PR body.

Non-blocking: `fjs/media/nix/todo/serializer-validation-split.md` now has two consecutive `Related` bullets both leading with `fjs/media/json/serializer/module.f.mjs`. Both are accurate, it just reads as a repeat.

PR #1620, opened 2026-08-16
todo/ast-spec: design the stage 1 function AST

sergey-shandar:
Working document `todo/ast-spec-stage1-discussion.md` for designing the stage 1 function AST, developed subject by subject; once every subject is decided, the result is distilled into a concrete design in `todo/ast-spec.md` and the working document is deleted.

Current state of the design:

- A function body is a **single operation node** — the root of a computation DAG whose nodes are connected by natural JS object references; sharing is expressed by referencing the same node and is semantic. There is no index space — the originally proposed `["local", i]` is removed.
- Non-resulting computations (asserts, fail-fast guards) are merged by the **`["comma", result, ...asserts]` operation** — an ordinary, composable operation introduced after stage 1 without changing the body's shape; stage 1 ships without it.
- Operation nodes: constants, object constructors, `["array", ...]`, `["args"]`, `["at", a, b]`, `["call", f, a]`, `["bindCall", o, p, a]`, and the later `["comma", ...]`.
- Assumptions analyzed: **A1** purity (accepted), **A2** engine interrupts with tiered cross-engine agreement (accepted), **A3** always-throws preserved (accepted), **A4** order preservation (**rejected** under the opaque-error contract: errors carry no information, so evaluation order of independent branches is unobservable).
- Decided subjects: 1 (reference DAG, hash-as-written), 2 (`["args"]`), 3 (lazy-operator extension path), 4 (property order is part of the value), 8 (`comma` membership with the as-if rule: proof and cache count as execution). Open: 5 (validation), 6 (command vocabulary), 7 (bare body node vs metadata wrapper). Parked: 9 (canonical graph serialization and hashing).

Changelog: none


o2alexanderfedin (commented):
Four things, the first two being leftovers from the A4 reversal.

**Subject 4 still states the pre-rejection position.** `:332-335`:

> "properties evaluate in the order written, per JS insertion-order semantics, **which subject 8 adopts exactly**"

A4 is rejected (`:143`), and subject 8 now says the opposite — "the array guarantees membership, not order" (`:402-404`), "**Membership is semantic; order is not** (A4 rejected)" (`:490`). History confirms it's a leftover: subject 4 got that wording in `9d8cb20b` when subject 8 was still "Anchored, JS-deterministic evaluation"; `be57f5f2` reversed subject 8 and left subject 4 alone. Subject 1 got an explicit `*Note (A4 rejected):*` patch at `:245-249` for exactly this. The *conclusion* still stands on independent ground — key order is observable in the resulting object — it's the "evaluate in the order written" claim and the subject 8 cross-reference that are stale.

**`:236-239` — a sentence broken by an edit, now saying the opposite of the rule.**

> "a source `const` is a shared interior node — or an assert-branch root when its value is unused **that other nodes reference**."

At `9d8cb20b` this read "an anchor-list entry that other nodes reference"; a later commit spliced in the em-dash clause and left the relative clause stranded. As it parses now it attributes "that other nodes reference" to an assert-branch root, which subject 8's own well-formedness rule forbids: "an assert entry must not be reachable from any other entry" (`:419-422`).

**`:516` — broken link.** `[mvp-roadmap](../../nanvm-lib/todo/mvp-roadmap.md)` from `todo/` resolves outside the repo. Should be `../nanvm-lib/todo/mvp-roadmap.md`, which is the form every other file uses (`todo/ast-spec.md:51,67`, `todo/fjs-nanvm-integration.md:8,107,132`). The only one of the file's 17 links that doesn't resolve.

**Missing `**Priority:**`.** `todo/README.md:63-67` mandates the Priority/Status block; all 36 other files under `todo/` carry one, including `todo/ast-spec.md:3`. `**Status:** open` here is fine, and a project-level discussion document is an accepted shape (`todo/README.md:23-29`).

Minor: `:511` attributes tags 28/29 to RFC 8949 — they're IANA-registry tags from the value-sharing spec, not defined in 8949 itself.

Also, on the PR metadata, since it becomes the squash commit: the title isn't `<topic>: <short description>` (`CONTRIBUTING.md:178`) though the branch commits already use `todo/ast-spec:` correctly; there's no mandatory `Changelog:` section (`AGENTS.md:87-92` — `Changelog: none` is right here, and correctly no entry file); and the description documents a superseded draft — it says "arrays of sequential operations", "reference previous results by index", `["local", -1]`, and "seven open subjects", where the current document has nine subjects, no index space, and `["local", i]` explicitly removed at `:224-232`.

Verified fine: subjects 1-9 and A1-A4 complete, no gaps, duplicates or dangling cross-references. The other three points that flipped during authoring are each stated once and coherently — guard-skipping (`:485-489` only, as non-conforming debug mode), result revelation (`:440-452` plus the as-if proof/cache rule `:454-464`), asserts as contracts/tests (`:466-482`, no surviving "asserts are validation" text), and ordering ownership (`:424-439`). FS principles 1-2 match `spec/README.md:5-6`; the `fjs compile` claim checked by execution (`{"a":{"z":1},"b":2,"c":{"z":1}}` → JSON expands the sharing, `.f.js` emits `const c0 = {"z":1}` and reuses it); the five command names and the `a.indexOf` observable difference match `spec/todo/2330-property-accessor.md`; the CBOR tasks match `nanvm-lib/todo/mvp-roadmap.md:336,362`. Branch green: `tsc` 0, `fjs t` 2972 pass / 0 fail, `cov` 100/100/100, `ci-update` clean.

sergey-shandar:
All review findings addressed in bdac5b8 (pushed as a371ed3 after rebase) plus PR metadata:

- **Subject 4** — re-grounded on the constructed value: JS objects preserve insertion order observably (`Object.keys`, iteration, `JSON.stringify`), so key order is part of the result; the stale "evaluate in the order written / subject 8 adopts exactly" claim and cross-reference are gone. Property-operand *evaluation* order is now explicitly as free as any other (A4 rejected).
- **Subject 1** — the spliced sentence is restored: "a shared interior node *that other nodes reference* — or an assert-branch root when its value is unused."
- **Link** — `../nanvm-lib/todo/mvp-roadmap.md`.
- **Header** — `**Priority:** P2` added (matching `todo/ast-spec.md`).
- **Tags 28/29** — attributed to the IANA value-sharing registration instead of RFC 8949.
- **PR title/description** — retitled to `todo/ast-spec: design the stage 1 function AST`; description rewritten for the current reference-DAG/branch-array model (nine subjects, no index space) and ends with `Changelog: none`.

---

o2alexanderfedin (approved):
All six addressed — approving.

**Subject 4** now rests on the independent ground rather than the stale one: "the written key order is part of the constructed *value*, not of the schedule … This holds with A4 rejected — the evaluation *order* of property operands is as free as any other (subject 8); what is fixed is the result." The `(subject 8)` cross-reference resolves to `:494` "Membership is semantic; order is not (A4 rejected)", which says what's claimed, and the `fjs compile` claim matches `spec/README.md:130` "Object properties are emitted in sorted key order."

**`:237-240`** parses now, with the relative clause back on the shared interior node: "a source `const` is a shared interior node that other nodes reference — or an assert-branch root when its value is unused." Agrees with `:415-419` and with the well-formedness rule at `:423-426` — the branch that is a root is exactly the one nothing references.

Link is `../nanvm-lib/todo/mvp-roadmap.md` and resolves; all 17 link targets in the file check out. `**Priority:** P2` is present above `**Status:**`, matching `todo/README.md`'s block order, and every `todo/*.md` now carries one. CBOR reads "the IANA-registered value-sharing tags 28/29", which is right.

Metadata all corrected: title conforms, `Changelog: none` present with no entry file, and the description now matches the current document — nine subjects with 1/2/3/4/8 decided, 5/6/7 open, 9 parked, matching every `**Status:**` in the file, with `["local", -1]` and the index space gone.

No new inconsistency from the ~32-line fix: guard-skipping appears only at `:488-493` as a non-conforming debug mode, result revelation at `:444`/`:456`, asserts-as-contracts at `:470-486`, ordering ownership at `:427-443`. The one remaining superseded-reasoning site, `:241-242`, carries its own `*Note (A4 rejected):*` patch. Merge dropped nothing, and the branch is green (`tsc` 0, `fjs t` 2972 pass / 0 fail, `cov` 100/100/100, `ci-update` clean).

Non-blocking nuance, not a defect: integer-index keys iterate in ascending numeric order regardless of insertion, so for an all-integer-key object sorting would in fact be unobservable. The rule as written is conservative rather than wrong, and the conclusion is unaffected.

o2alexanderfedin (commented):
The comma design holds up — two small things from the last commit.

**The PR body has the operands reversed.** It says merged by `["comma", result, ...asserts]`, which is the pre-`1f6fcc1f` spelling. The document says result *last* in all four places it appears: `:18`, `:72`, `:443`, and the example at `:97` (`// the result: last, as in JS (a, b) → b`), plus `:470` "the result operand at its fixed last position". `1f6fcc1f` is the commit that moved it, and the body didn't follow. It becomes the squash commit.

**`:32` — "all three of" is followed by four functions** (`f`, `g`, `h`, `k` at `:35-38`). `9c9ff2cc` added `k` and the "last spelling" sentence but left the count. The same sentence calls them "statement spellings", while `:41-42` says `k` is "an expression-bodied arrow, no block, no `return`".

Lower confidence, same edit: `:29-30` says the block-with-`return` form "lowers to a `comma` node verbatim, `toString(f)` prints it back verbatim", but `:41-43` says the expression-bodied spelling is "the natural form for `toString(f)` to print". Since all four share one AST, only one can come back verbatim. Reads as loose wording about the comma pattern surviving the round trip rather than the literal text, but the earlier sentence no longer supports it as written.

Checked, and this was the part that mattered — the comma operation coheres with all four positions the earlier subjects settled:

- "Membership is semantic; order is not" survives at `:538`, and **result-last does not reintroduce an ordering commitment** — it's a canonical *spelling* constraint occupying the same slot the old fixed head did, with `:79-82` still granting "any order, possibly in parallel (A4)".
- The root rule is restated in comma terms at `:463-465`: "an assert operand must not be reachable from any other operand of the same `comma`", echoed in subject 5 at `:389-392`.
- Reveal-after-all-succeed with the as-if rule at `:486-512`, now correctly scoped to the `comma`'s value rather than the function's, with the body-root case called out at `:488-490`.
- Spec-owns-spelling / engines-own-schedule at `:468-485`.

The JS claim is true and I ran it: the comma operator yields the rightmost operand, and all four spellings behave identically in Node (`5` for input `3`, throw for `-1`), including that `return assert(...), a + 2` parses as `return (assert(...), a + 2)`. The framing that a JS engine implements "one legal schedule (left-to-right, eager) of the same semantics" is compatible with A4's rejection rather than in tension with it — and result-last is what makes the JS reading agree with reveal-after-all-succeed.

No superseded text survived the four-commit sequence: grepping the pre-comma vocabulary (`entry`/`entries`, `branch array`, `resulting branch`, `index 0`, `fixed head`) returns nothing, and every `["comma"` occurrence is `...asserts, result`. Subjects 3, 5, 7 and the A4 bullets were all updated in step. The earlier fixes are intact — subject 4's value-based resolution `:367-375`, the whole `const` sentence `:274-278`, the in-repo roadmap link `:564`, `**Priority:** P2`, the IANA tag wording. Nine subjects, no gaps, statuses matching the body's split. All 13 links resolve; `spec/todo/2340-operators.md:33` does mark Comma `not allowed` and `spec/README.md:8` reads "It whitelists complete semantic patterns." Branch green.

PR #1622, opened 2026-08-16
types/rtti: validate rejects over-long tuples

sergey-shandar:
Implements `fjs/types/rtti/todo/tuple-extra-elements.md` (deleted here).

## The problem

Two consumers of the same schema disagreed on the same value. Verified on `main`:

```
validate([42])([42, 'extra'])  ->  ['ok', [42, 'extra']]
data form                      ->  not in the set
```

`Ts<readonly [42]>` is the exact tuple `readonly [42]`; a 2-element array is not assignable to it. The data form already constrained length exactly (`{ prefix }` with no `rest`). `validate` walked the schema's entries and never looked at the value's length.

## The change

`constContainerValidate` takes a `sizeOk` predicate, so the shared walk stays in one place and only the rule that differs lives at the call sites — tuples require `value.length === size`, structs pass `() => true`.

## What breaks

Two classes, both `ok → error`, both correct under `Ts<T>`:

```
validate([42])([42, 'extra'])              base ok  ->  error   extra element
validate([number, option(string)])([42])   base ok  ->  error   missing trailing optional
validate([unknown])([])                    base ok  ->  error
validate([option(number)])([])             base ok  ->  error
```

The second class is the subtle one. A trailing element whose schema admits `undefined` — `option(t)` is `or(t, undefined)`, and `unknown` includes it — makes a *shorter* array pass the element walk, because the walk reads `value[1]` as `undefined` and `undefined` satisfies the schema. `Ts<readonly [number, string | undefined]>` still has length 2, so `[42]` is not assignable to it; only a length check rejects it.

A third, smaller change: for a missing element whose schema *rejects* `undefined`, base already errored via the walk, at the first missing index (`path: ['0']`). The length check fires first, so it now reports at the root (`path: []`), where the length lives.

**`parse` is left alone.** The TODO said to add the length check to "`constContainerValidate`'s tuple instantiation *and its `parse` counterpart*", on the premise that both "accept" extra elements. `parse` does not accept them — it constructs a fresh value of exactly the schema's length, truncating extras and filling a missing optional:

```
parse([42])([42, 'extra'])              ->  ['ok', [42]]
parse([number, option(string)])([42])   ->  ['ok', [42, undefined]]
parse([42])([])                         ->  ['error', …]      nothing to fill with
parse({a:42})({a:42, b:'x'})            ->  ['ok', {a: 42}]
```

Its module header documents this as the design, with a reason: *"forward compatibility with extended serialization formats: a schema-based parser keeps working when newer versions of the format add extra fields or tuple elements."* So `parse` already satisfies the `Ts<T>` consistency the TODO argues for — by construction rather than by rejection, and its result already has the exact type. Rejecting would trade that away for nothing.

The rtti `README.md` now carries the resulting table; every cell in it was verified by execution:

| schema | value | `validate` | `parse` | data form |
| --- | --- | --- | --- | --- |
| `[42]` | `[42, 'extra']` | error | `[42]` | not in the set |
| `[42]` | `[]` | error | error | not in the set |
| `[number, option(string)]` | `[42]` | error | `[42, undefined]` | not in the set |
| `{ a: 42 }` | `{ a: 42, b: 'x' }` | ok | `{ a: 42 }` | in the set |

`data/README.md`'s "Tuple length" section pointed at the deleted TODO; it now points here.

## One more thing

`Object.entries(rtti)` moved to the `rtti =>` scope. It depends on the schema alone, and the length check needs its size; recomputing it per validated value was already wasteful. This is the "place curried partial applications at their dependency's scope" rule.

## Verification

- `npx tsc` — 0 errors.
- `node fjs/module.mjs test` — 2979 pass, 0 fail. Before updating any proof, exactly **one** test failed: `validate/proof.const.tuple.extraItems`, the case the TODO names. Nothing else in the tree depended on the old behavior.
- **Mutation-tested**, after review found the first version of `missingItems` was vacuous (see below). All seven mutants of the length check now fail the suite: `<=` (2 fail), `>=` (2), `!==` (7), `size + 1` (6), `size - 1` (6), `() => true` (4), `() => false` (5).
- Differential against `main`, 850 rows (17 schemas — tuples of arity 0–3, nested, thunk-element, structs, array, record, primitives, const — × 25 values × `validate`/`parse`): **48 rows change, all `validate`, all tuple schemas, every one `ok → error`.** Zero `parse` rows move; zero struct, array, record, or primitive rows move; nothing became more permissive.
- Harness guards: every call is checked to actually return an `['ok'|'error', …]` tuple (the first version asserted a string return, copied from an earlier harness — it failed loudly on 850 rows rather than passing vacuously), and both outcomes are required to occur in quantity so "identical" is not a statement about an all-error corpus.

## Review fixes (commit `48bf601`)

- **`missingItems` did not pin the short side.** Both its schemas reject `undefined`, so both cases already errored on base through the element walk — `value.length <= size` survived the entire suite. Added `missingOptionalItems`, which covers what only the length check can reject, and `missingItems` now pins the root path with `assertErrorPath([])`, which the `<=` mutant also breaks.
- **The changelog understated the break**, presupposing too-short was already an error. Reworded; the README table gained the missing-element and optional-trailing rows.
- **The "Note on `main`" was wrong and is removed.** It claimed 29 `tsc` errors on the base. They came from stale gitignored `.d.mts` artifacts in my working tree — left by an earlier declaration-emit check — shadowing the real sources. After deleting them, `npx tsc` exits 0 on both base and head. My error, not `main`'s.

## Changelog

- **BREAKING CHANGES:** `types/rtti`: `validate` requires an array to match its
  tuple schema's length exactly, matching `Ts<T>` and the data form. An extra
  element is rejected, and so is a missing trailing `option`/`unknown` element,
  which used to pass. Structs are unchanged: extra keys are still accepted.


o2alexanderfedin (commented):
The implementation is right — three things around it.

**The changelog understates the break.** It says "A too-short tuple now reports at the root rather than at the first missing index", which presupposes too-short was already an error. For any tuple whose trailing declared element accepts `undefined`, it flips `ok` → `error`. Against the merge base:

```
validate([number, option(string)])([42])  base ok  -> PR error
validate([unknown])([])                   base ok  -> PR error
validate([option(number)])([])            base ok  -> PR error
```

The new behavior is correct — `Ts<readonly [number, string|undefined]>` has length 2, so `[42]` isn't assignable — but a consumer with an `option`/`unknown` trailing element hits a second break class the entry doesn't mention. The README section shows only the extra-element row too.

**The `missingItems` proof doesn't pin the short side.** Mutating `fjs/types/rtti/validate/module.f.mjs:138` to `value.length <= size` leaves the whole suite green — 2978 pass / 0 fail — because both cases (`[42]` vs `[]`, `[42,'hello']` vs `[42]`) already errored on base through the element walk, so they assert nothing the walk didn't already give. Coverage stays 100% since both `sizeOk` branches are still taken. A case that kills it: `validate([number, option(string)])([42])` — PR `error`, `<=` mutant `ok`. This is also what the body's "New proofs cover both sides of the length check" claims. The other six mutations I tried (`() => true`, `() => false`, `size+1`, `size-1`, `>=`, `!==`) are all caught.

**The "Note on `main`" doesn't reproduce.** It reports 29 `tsc` errors on `9f380ec` in `fjs/cas/` and `fjs/effects/io/`. After `npm ci`, `npx tsc` exits 0 with no output on both the merge base and this head. Worth dropping before the squash.

Verified clean otherwise — the semantic delta is exactly the tuple length constraint:

- Differential harness, 48 schemas × 48 adversarial values × `validate`+`parse` = 4608 rows base-vs-PR: 275 diffs, all in `validate`, none in `parse`, 104 `ok→error` and **0 `error→ok`**. Every diff is on a schema containing a tuple; zero on `struct*`, `array_*`, `record_*`, recursive arrays, primitives, consts, or `or`/`option` over non-tuples.
- Cross-checked against an independent reference implementation written from the spec rather than from this code: 48 schemas × 3000 random values = 144,000 rows, 0 mismatches with the PR; the same reference with only the length line removed matches base on all 144,000. So the delta is the exact-length constraint and nothing else.
- All 9 importers of `rtti/validate` use struct schemas only — every `option(...)` in the tree is an object key, never a tuple element — so no caller semantics changed and no other proof needed updating.
- All six cells of the new README table verified by execution; `TupleTs` really is the exact mapped tuple, so the `Ts<readonly [42]>` argument holds. The deleted todo's three Proposal items are covered, with the deliberate `parse` deviation captured in the README and at `validate/module.f.mjs:100-103`.
- `npx tsc` 0, `fjs t` 2978 pass / 0 fail, `cov` 100/100/100, `prepack` 0, `ci-update` leaves the tree clean. Title and changelog format both conform.

sergey-shandar:
All three confirmed and fixed in `48bf601`. Thank you — the mutation testing in particular caught a proof that was pulling its weight only by coincidence.

**1. `missingItems` didn't pin the short side.** Reproduced: `value.length <= size` passes 2978/0. You're right about why — both schemas I picked (`[42]`, `[42, 'hello']`) reject `undefined`, so the element walk already errored on them and the assertions carried no new information. `assertOk`/`assertError` also hid it, since neither looks at the path.

Two changes. `missingOptionalItems` covers what only the length check can reject:

```js
assertError(validate([number, option(string)])([42]))
assertError(validate([option(number)])([]))
assertError(validate([unknown])([]))
// Present-but-undefined is the declared length, so it stays ok — the
// check is on length, not on element presence.
assertOk(validate([number, option(string)])([42, undefined]))
```

And `missingItems` now uses `assertErrorPath([])` instead of `assertError`, so it pins the root path that the length check produces — the `<=` mutant reports `['0']` there and fails it too. Re-ran your seven mutants against the new suite:

| mutant | failures |
| --- | --- |
| `value.length <= size` | 2 |
| `value.length >= size` | 2 |
| `value.length !== size` | 7 |
| `value.length === size + 1` | 6 |
| `value.length === size - 1` | 6 |
| `() => true` | 4 |
| `() => false` | 5 |

**2. The changelog understated the break.** Correct — "now reports at the root rather than at the first missing index" presupposes too-short was already an error, which is only true when the missing element's schema rejects `undefined`. Confirmed your three cases against `9f380ec`:

```
validate([number, option(string)])([42])   base ['ok',[42]]  ->  error
validate([unknown])([])                    base ['ok',[]]    ->  error
validate([option(number)])([])             base ['ok',[]]    ->  error
```

The entry now leads with the exact-length requirement and names both classes. The README table gained the too-short and optional-trailing rows, and I re-verified all twelve cells by execution — including `parse([number, option(string)])([42])`, which pads to `[42, undefined]` rather than erroring, while `parse([42])([])` does error because there is nothing to fill with.

**3. The "Note on `main`" doesn't reproduce because it was wrong.** Not a stale `node_modules` — `npm ci` didn't change anything. My working tree had 61 gitignored `.d.mts` and 53 `.d.ts` files left over from a declaration-emit check I ran in an earlier session; `fjs/effects/io/module.f.d.mts` was stamped 13:10 against a `module.f.mjs` from 15:08, so TypeScript resolved the `effects/io` imports against a declaration file predating the stage-4 migrations and reported exports that do exist. Deleting them takes `npx tsc` to 0 on both base and head. The note is removed from the PR body. Sorry for the noise — I should have checked what `tsc` was actually resolving before attributing it to `main`.

---

o2alexanderfedin (approved):
Both addressed — approving.

**The short side is pinned.** Re-ran the mutation: `value.length <= size` now fails the suite (2970 pass / 2 fail), killed by `tuple.missingOptionalItems` and `tuple.missingItems`; restoring gives 2972 / 0. The new case is genuinely testing the new constraint rather than the element walk — `validate([number, option(string)])([42])`, `validate([option(number)])([])` and `validate([unknown])([])` are all `ok` on `f8837a43` and `error` with path `[]` here, and `[42, undefined]` still passes. Strengthening `missingItems` with `assertErrorPath([])` pins the path claim too. All seven mutations now caught (`() => true` 4 fails, `() => false` 5, `size+1` 6, `size-1` 6, `>=` 2, `!==` 7).

**The changelog now covers the second break class** — "and so is a missing trailing `option`/`unknown` element, which used to pass" is exactly the `ok` → `error` case that was missing. Conformant, and byte-identical to the body's section. All four README rows verified by execution including the `parse` and data-form columns.

The "Note on `main`" is gone, replaced with the stale-`.d.mts` explanation; `npx tsc` exits 0 on both `f8837a43` and this head, so that resolves it.

Merge dropped nothing: `git log f8837a43 --not refs/pr/1622` is empty and the diff is only the six intended files. `cov` 100/100/100, `prepack` 0, `ci-update` leaves the tree clean.

One non-blocking nit: `fjs/types/rtti/validate/proof.f.mjs:183-184` says the `missingOptionalItems` cases are what pin the short side and "with `length <= size` the whole suite still passes without them" — that's no longer true, since the same commit strengthened `missingItems`. Deleting `missingOptionalItems` and applying the mutant still fails (1 fail, `missingItems`). The PR body has it right; only the comment understates it. Doubly pinned is better than the comment claims.

PR #1623, opened 2026-08-16
types/rtti: todos for deleting validate and adding close

sergey-shandar:
Two todos, replacing the open/closed investigation this PR originally carried. The design question is settled by the maintainer, so the investigation file is gone and these record the decisions instead. **No code touched.**

- `fjs/types/rtti/todo/retire-validate.md`
- `fjs/types/rtti/todo/close-type.md`

Plus cross-links into the existing `fjs/media/json/todo/rtti-parse.md`.

## 1. Delete `validate`

`parse` becomes the way to read a value against a schema. Structs and tuples are open — extras accepted on the way in, absent from what `parse` constructs on the way out.

**Deleting is better than reverting #1622.** The exact-length check lives *only* in `validate`; `parse` never had one. So the module takes the check with it — no revert, and no intermediate state where the two readers disagree. `parse` is already open on both counts, verified against `805aabe`:

```js
parse([42])([42, 'extra'])              // ['ok', [42]]              extra element accepted, dropped
parse({ a: 42 })({ a: 42, b: 'x' })     // ['ok', { a: 42 }]         extra key accepted, dropped
parse([number, option(string)])([42])   // ['ok', [42, undefined]]   short array, trailing optional filled
parse([42])([])                         // ['error', …]              42 excludes undefined, so position 0 is required
```

The last two are the object rule from `data/types.ts:42-44` — an absent member reads as `undefined`, so a member is required exactly when its set excludes `undefined` — applied to both kinds.

### The two-step objection, and its answer

The real cost of deleting `validate` is what it leaves behind: reading JSON against a schema becomes text → extended value → `parse` → `Ts<T>`, two traversals plus a discarded intermediate, where `validate` at least returned its argument.

The answer is not to keep a validator — it's `fjs/media/json/todo/rtti-parse.md`, the RTTI-aware JSON parser, which goes from text to `Ts<T>` in one pass. That todo already exists and already specifies the entry point.

It's **strictly better** than `validate` for this use rather than merely equivalent: it still holds the original number token, so it rejects `1.00000000000000001` against an rtti `bigint` — something no validator over an already-materialized value can do, because by then the token has rounded to `1`. It also already states it reuses `rtti/parse`'s structural behavior ("drop extra struct fields/tuple elements where the current parser does"), so it inherits openness rather than needing its own rule.

So the readers that remain are `parse` for a runtime value and the JSON parser for text, and neither has a reason to return its input unchanged.

### Migration

The todo lists all nine importers with line numbers, and the three ways `parse` is not a drop-in: the result is a fresh value so identity is not preserved; extras are gone from it, which matters at `media/module.f.mjs:123` where `matchWith` hands the value to a caller-supplied `extraValidate`; and it allocates where `validate` returned its argument, worth measuring in `media`'s dialect detection since that runs per candidate dialect.

## 2. `close` type

```ts
() => ['close', Struct | Tuple, Rest?]
```

Declared members match exactly; remaining members must match `Rest`; omitted or `undefined` `Rest` means no remaining members.

**The data form already carries this.** `ArraySet` and `ObjectSet` are both `{ members, rest? }`, and all four corners validate correctly today with no code change — verified:

```js
validate([{}, { object: [{ props: { a: num }, rest: never }] }])({ a: 1, b: 2 })  // error  — closed object
validate([{}, { object: [{ props: { a: num } }] }])({ a: 1, b: 2 })               // ok     — open object
validate([{}, { array: [{ prefix: [num], rest: unknown }] }])([1, 2, 3])          // ok     — open tuple
validate([{}, { array: [{ prefix: [num] }] }])([1, 2, 3])                         // error  — closed tuple
```

So `close` is the schema-form spelling of a `rest` that already exists: `never` when `Rest` is omitted, `R` otherwise. `never` being the empty set is why "no remaining members" falls out rather than needing a special case.

The todo flags one thing to decide rather than assume: `parse` currently *drops* extras, so a closed container has to **error** instead, or `close` and the open form behave identically through `parse` and the tag has no observable effect.

## Things I verified that the one-line descriptions didn't cover

- **Deletion is clean.** All nine importers take only `validate`; nothing imports the four symbols it re-exports from `common`; `parse` imports from `common` directly, not through `validate`; neither `deno.json` nor `package.json` names the path.
- **The data form still describes tuples as closed**, so it disagrees with `parse` independently of `validate` — `toData` maps a `Tuple` to `{ prefix }` with no `rest`, and an absent `rest` means closed on arrays but open on objects. Recorded in `retire-validate.md` as separable follow-on work so it isn't lost, not folded into the deletion.
- **`common` has no `match` function**, despite `rtti/README.md` describing one. `visit` and `orVisit` are the only dispatchers; noted so the `close` implementer doesn't go looking.

Both files state that TypeScript is not the authority here: the open `TupleTs` mapping is commented out because TypeScript could not handle it, and #1622 read that as a design decision. Each also carries a "don't 'fix' this again" section aimed at the specific failure mode — someone reads `Ts<T>`, concludes tuples are exact, adds a length check.

## Verification

- Every line reference checked against `805aabe`; two corrected while checking (the `Type` union is `types.ts:79-94`, not `78-105`).
- Both code blocks run exactly as written and produce the stated results; the bindings they use are named in the files so they're reproducible.
- All three new cross-document relative links resolve, and the one sentence quoted from `rtti-parse.md` is quoted exactly (my first attempt paraphrased it inside quote marks).
- `npx tsc` — 0 errors. `npm run ci-update` leaves the tree clean.

Changelog: none — `todo/` only.


o2alexanderfedin (approved):
Approving. Every factual claim holds — I executed them rather than reading them.

Reproduced exactly as annotated, via a script importing from the branch:

- The four data-form corners (`:62-65`) with `num = { number: true }`: closed object `error`, open object `ok`, open tuple `ok`, closed tuple `error`.
- The three asymmetry-2 cases (`:94-96`) with `numOrUndef = { number: true, unit: 2 }`: `{}` `ok`, `[]` `error`, `[undefined]` `ok`. Bit `2` is the `undefined` bit, since `unitList` is `['null','undefined','false','true']`.
- The #1622 characterization: `validate([42])([42,'extra'])` error, `validate([number, option(string)])([42])` error, `validate({a:42})({a:42,b:'x'})` ok.
- The four §3 `parse` behaviors: truncates to `['ok',[42]]`, fills to `['ok',[42,undefined]]` (index genuinely present), drops extra keys, and still errors on `parse([42])([])`.
- The "`toData` cannot produce two of the four" claim — ran it over the whole schema vocabulary (`Tuple`, `Struct`, `array(t)`, `record(t)`, `record(never)`, `array(never)`); no schema yields a per-position array pattern with a `rest`, nor a `props`-bearing object pattern with `rest: never`.
- The load-bearing claim that the commented-out open-tuple mapping at `../ts/types.ts:78-80` is a *tooling* limit, not a design choice: uncommenting it gives `types.ts(79,13): error TS2589: Type instantiation is excessively deep and possibly infinite`. Correct.

References all resolve, with the quoted snippets byte-exact: `../ts/types.ts:78-80`, `../data/types.ts:22-55,:34-55,:42-44`, `../data/module.f.mjs:296-320` (`arraySet` normalizing `never`→absent at 298, `objectSet` top→absent at 315), `:371-383`, `:414-423`, the two README sections, and `fjs/djs/todo/663-json-djs-tree-type.md`.

No contradiction with the rtti or data READMEs — they state the same absent-`rest` rules this restates. No duplication either: `172.md` is the validate/parse skeleton, and the one prior overlap, `tuple-extra-elements.md`, was deleted by #1622.

`Status: open` matches all six tasks unchecked, `Changelog: none` with no entry file is right for todo-only, and the branch is green (`tsc` 0, `fjs t` 2972 pass / 0 fail, `cov` 100/100/100, `ci-update` clean).

o2alexanderfedin:
Reviewed after merge — one small internal contradiction worth a follow-up, since it would misdirect whoever picks the task up.

`fjs/types/rtti/todo/close-type.md:114` says:

> `- [ ] Add the `Visitor` / `match` case; implement it in `parse`.`

but `:73-74` of the same file says the opposite, and is right:

> "Note there is no `match` function in `common` despite `../README.md` describing one; `visit` and `orVisit` are the only dispatchers."

`common/module.f.mjs` exports `verror, prependPath, primitive0Validate, constPrimitiveValidate, isArray, isObject, eachEntry, orVisit, visit`, and `common/types.ts` has no `match` at all. The task should name `visit` — as written it sends the implementer looking for the very function the body warns doesn't exist.

Also non-blocking: `data/README.md:141` links to `../README.md#tuples-are-closed-structs-are-open`, and `retire-validate.md:149-150` directs replacing that exact section. Neither task list mentions the inbound anchor, so it will dangle when that work is done. Nothing is broken today.

Everything else checked out:

- **Consumer audit.** Exactly 9 importers of `validate`, matching the 9-row table, and all 9 line numbers correct. Each takes only `validate`. `parse`'s visitor covers the identical variant set, so no consumer loses a schema form. The `media/module.f.mjs:123` caveat is real (`matchWith(validate(type))(extraValidate)` passes the validated value into the caller's predicate), and "once per candidate dialect per value" matches `dialects.find(({match}) => match(value))` at `:149`. Nothing imports the four re-exports through `validate` — `parse` takes them from `common` directly — and neither `deno.json` nor `package.json` names the path.
- **Executed, not read:** all four `parse` lines at `retire-validate.md:27-30` (including the fill being a genuine `undefined` at a present index), all four data-form corners at `close-type.md:31-34`, the `parse` vs data-form divergence at `:127-128` with `toData([42])` yielding `{prefix}` and no `rest`, and the post-#1622 `validate` behavior.
- The `TupleTs` "TypeScript could not handle it" claim: uncommenting gives `TS2589` plus `TS2574`. Holds.
- All 15 relative links resolve, every line reference and quoted snippet is byte-exact in both directions across the three files.
- Dropping `open-closed-values.md` carries the substance over — the `TupleTs` argument, the four-corner block verbatim, the `rest`-default asymmetry, the `subset`/`equal`/`cmp` constraints. Two items were dropped rather than migrated (the `[undefined]` vs `[]` invariant, and the ordering link to `663-json-djs-tree-type.md`); `todo/README.md:87-93` allows capture "in another issue", so the replacements satisfy it. Nothing still references the deleted file.
- `tsc` 0, `fjs t` 2972 pass / 0 fail, `cov` 100/100/100, `ci-update` clean.

PR #1624, opened 2026-08-16
types/rtti: delete validate; parse is the reader

sergey-shandar:
Implements `fjs/types/rtti/todo/retire-validate.md` and `fjs/dev/package_json/todo/remove-module.md` (both deleted here).

`parse` is now the only schema-form reader. Structs and tuples are open — extras accepted on the way in, absent from the fresh value `parse` builds — and #1622's exact-length check goes with the module it lived in, so there was nothing to revert.

## Migration

All nine importers moved to `parse`. Two things my first pass got wrong, both caught by `tsc`:

- **`fjs/media/module.f.mjs` already binds `parse`** to the JSON parser, so the rtti reader is imported as `rttiParse`. A blind rename produced `Duplicate identifier 'parse'`.
- **`fjs/protocol/mcp/module.f.mjs` had three more call sites** than my first patterns matched (`initializeParams`, `toolsListParams`, `toolsCallParams`).

## `fjs/dev/package_json` is removed

Eight sites migrated by changing an import. The ninth did not: `fjs/dev/package_json` is documented as *"reading metadata without losing unrelated fields before write-back"*, with a proof pinning `Object.is(result, input)` — the single failure when I first ran the suite. It initially kept the property with a local `mapOk(() => value)` and a cast.

That module has **no consumers anywhere**, and it was the only thing left in the tree needing "check the shape, return the original" — the reader shape this PR retires. So it is deleted, and the exception goes with it: `mapOk(() => value)` and its cast never land.

Done in this PR rather than a follow-up because the todo exists only on this branch and the exception it describes is *introduced* by this branch — a separate PR would have to land the exception first and then remove it.

Confirmed before deleting: no `.mjs`/`.ts`/`.json`/`.yml` outside the module names it, and nothing else in-tree parses `package.json` — `fjs/ci` only checks whether the file exists (two of its three mentions are comments; the third puts a `package.json` into a *virtual* filesystem to test the CI generator).

## Documentation

- `rtti/README.md`: the "Tuples are closed, structs are open" section is replaced by **"Structs and tuples are open"**, with the openness table, the one rule that explains both length rows (*an absent member reads as `undefined`, so a member is required exactly when its set excludes `undefined`*), and a **"This is deliberate; please do not 'fix' it"** subsection naming the exact failure mode.
- `parse/module.f.mjs` and `common/module.f.mjs`: headers rewritten. The former had asserted the *opposite* of this PR's thesis — *"this makes both structs and tuples effectively closed at runtime, matching the TypeScript type produced by `Ts<T>`"* — in the file a reader opens, and shipping in `parse/module.f.d.mts`. Both now describe openness and carry the guard; verified in the emitted declarations via `npm run prepack`.
- `ts/types.ts`: `TupleTs` says the commented-out open mapping is the accurate one and TypeScript's inability to express it is a renderer limitation, not a statement about the value model.
- `parse/proof.f.mjs`: `extraItemsAcceptedAndDropped`, `extraKeysAcceptedAndDropped`, `shortArrayFillsAnOptionalPosition`, each commented with *why*.
- `data/README.md`'s "Tuple length" section rewrote a dead anchor and a now-false claim, and documents the **data-form divergence** with executed examples.
- Broken references repointed in `fjs/mcp`, `fjs/mcp/evo`, `rtti/ts`, `rtti/data`, `fjs/bnf/todo/207.md`, `spec/todo/3360-type-annotations.md`.

**`todo/inline-type-casts.md` I deliberately did not edit.** It has nine rows for the deleted module, but its category counts and totals are derived from that table. Rather than fabricate arithmetic I left the rows and noted that they describe a deleted module and a re-audit is the way to refresh the file.

## Todos

**Resolved by the deletion**, so removed: `172.md` (one container skeleton for `validate` and `parse` — one walker remains). `proof-shared-asserts.md` reduced to the one item that was never about the two-proof split.

**Filed**: `fjs/types/rtti/todo/data-form-open-tuples.md`. The data form still describes tuples as closed, so `parse([42])([42,'extra'])` → `['ok',[42]]` while `validate(toData([42]))([42,'extra'])` → `error`. It carries the concrete fix, plus two things verified while writing it: step 2 (dropping `arraySetValidate`'s minimum-length check) is genuinely required, since with `rest: unknown` alone a *shorter* array is still rejected; and the divergence is arrays-only, because `arraySet` normalizes a `rest` of `never` away while `objectSet` normalizes `unknown` away.

## Verification

- `npx tsc` — 0 errors. `node fjs/module.mjs test` — 2902 pass, 0 fail. `npm run cov` — 100.00% lines / branches / functions.
- Every README example, openness-table cell, and JSDoc example verified by execution.
- **Differential across the migrated consumers**, 25 rows through their public entry points (`decodeRequest`, `revision`/`lock`/`note` validate and decode, and `detect(dialects)(bytes)` over real UTF-8 bytes):

  | | result |
  | --- | --- |
  | changed rows | 10 of 25 |
  | `ok` ↔ `error` flips | **0** |
  | dialect-detection rows changed | **0 of 8** |

  Changes are confined to returned-value shape: key order follows the schema, extras are dropped, absent optionals are materialized as explicit `undefined`. That last makes `'params' in req` true where it was false, but `json_rpc`'s dispatcher destructures and tests `=== undefined`, and there is no `in`/`Object.keys`/`hasOwn` in `json_rpc` or `mcp`.

### Harness bugs found by its own guards

Four, disclosed because each weakened a claim I had already made:

1. A `detectText` export I invented, with an escape hatch in my harness that recorded `no-detectText-export` and passed. Replaced with the real `detect(dialects)(bytes)`.
2. Test payloads missing the required `dialect` literal, so no schema could match.
3. **An invalid cbase32 test hash**, so the revision path was erroring on *both* sides and the `extraValidate` ok-path — the riskiest part of the migration — was never exercised. The guard that allowed this was a generic "some rows matched"; it is now per-dialect, plus a check that both `checkReferences` refinements reach their ok path.
4. **`git stash` stopped producing a `main` baseline** once the migration was committed, so a re-run reported a misleading *0 changed rows*. The numbers above come from a real `origin/main` git worktree.

## Changelog

- **BREAKING CHANGES:** `types/rtti`: `validate` is deleted — `parse` is the way
  to read a value against a schema. Structs and tuples are open: extras are
  accepted and absent from the value `parse` constructs.
- **BREAKING CHANGES:** `dev/package_json` is removed. It had no consumers, and
  its schema is three optional fields any caller can declare and `parse`.


o2alexanderfedin (commented):
The migration itself is correct — **0 ok↔error flips at any consumer entry point**, and the CAS paths are safe. Four documentation problems, one of which ships in a published declaration and contradicts this PR's own thesis.

**`fjs/types/rtti/parse/module.f.mjs:11` asserts the opposite of what this PR establishes.**

> "This makes both structs and tuples **effectively closed** at runtime, matching the TypeScript type produced by `Ts<T>`"

and `:178` — `// tuples are closed: extra elements are dropped`. That is verbatim the inference the new `README.md:58` subsection ("This is deliberate; please do not 'fix' it") and `ts/types.ts:88-89` ("Do not cite the exact mapping as evidence that tuples are closed and add a length check to `../parse/module.f.mjs`; that inference is what produced #1622") exist to prevent — sitting in the very file that reader opens. The same header still describes `validate` as live at `:8`, `:24-25`, `:81`, `:113`. All of it is in the emitted `parse/module.f.d.mts:8,11,24`, and per `fjs/AGENTS.md` §2 the JSDoc on `module.f.*` is the authoritative statement of the invariant.

Same class, minor: `common/module.f.mjs:2,6,20` still has "Shared kernel for RTTI consumers (`validate`, `parse`)", "`validate` keeps the original value", "so `validate`'s pure pass/fail check pays no allocation" — also shipped in `common/module.f.d.mts`.

**`fjs/types/rtti/data/README.md:145` — dangling anchor, and the section is now false.** It links `../README.md#tuples-are-closed-structs-are-open`, but this PR renames that heading to `### Structs and tuples are open` (`README.md:31`), so the fragment resolves to nothing — the only such anchor in the tree. The surrounding text still says the exact length matches "the `Ts<T>` type-level mapping and the thunk-direct `validate`" (deleted here) and "how the **three** runtime consumers line up" (now two). The body says broken references in `rtti/data` were repointed; the top of the file was fixed, this section was missed.

**A decision the deleted todo said should not be lost, is lost.** `retire-validate.md:122` ("One thing deleting `validate` does not fix") plus its final task recorded the concrete fix — map `Tuple` to `{ prefix, rest: unknown }`, drop `arraySetValidate`'s minimum-length check at `data/module.f.mjs:994-998` — and said it could be split into its own issue. It isn't done, and nothing in `fjs/types/rtti/todo/` records it now (`close-type.md`'s `rest: unknown` is an illustrative example, not this task). Removing the `| data form |` column from the README table took out the other place it lived. The divergence is real and now undocumented: `parse([42])([42,'extra'])` → `['ok',[42]]` vs `validate(toData([42]))([42,'extra'])` → `['error',{path:[],message:'unexpected value'}]`. `todo/README.md:87-89` asks for capture before deleting.

Verified clean — this is the part that mattered:

- **Differential harness, 164 rows** through the real entry points (`decodeRequest`, `revision`/`lock`/`note` validate + decode/encode, `detect` over real UTF-8 bytes, `validatePackageJson`, the mcp param schemas, `media/json/rtti`). 39 rows changed value, **0 verdict flips at any consumer**. The only verdict flips anywhere are raw tuple probes — and no consumer schema contains a tuple, which is exactly why #1622's exact-length check is unobservable from every migrated site. `detect` 0 of 24 changed; `package_json` 0 of 18, with `Object.is(result, input)` still true.
- **`media/module.f.mjs`**: `matchWith` yields only a boolean and `detect` never returns the parsed value, so the rewrite reaches `extraValidate` alone; `isValidRevision` → `checkReferences` reads `parents`/`snapshot`/`lock`/`generation`, all present in the rebuilt value, and its ok path is genuinely exercised. No value can match a dialect it previously didn't.
- **CAS is safe.** `encodeText`'s single caller (`cas/evo/module.f.mjs:461`) applies it to `buildRevision` output assembled from typed fields, never a `decodeText` result; `stringify` omits `undefined`-valued keys, so the materialized `archived: undefined` / `lock: undefined` never reach a hash. The only affected round trip is a blob carrying an undeclared field, which no in-repo path performs.
- `'params' in req` does become true, but both `json_rpc:100` and `mcp:281` destructure and test `=== undefined`; no `in`/`Object.keys`/`hasOwn`/spread over a decoded request in either module.
- Proofs not vacuous: reintroducing a tuple length check in `parse` → 3 failures; making `validatePackageJson` return `parseShape(value)` → 1 failure.
- Declaration diff is only the two removed `validate` files plus `decodeRequest` retyping `Validate<…>` → `Parse<…>` and `validatePackageJson`'s signature. `tsc` 0, `fjs t` 2906 pass / 0 fail, `cov` 100/100/100, `prepack` 0, `ci-update` clean. Changelog conforms with the `**BREAKING CHANGES:**` prefix and every importer updated here.

sergey-shandar:
All four fixed in `88175b5`. The first one is the worst kind of miss — I wrote a guard against an inference while leaving that exact inference asserted in the file the reader opens.

**1. `parse/module.f.mjs` asserted the opposite.** Rewritten. The header now leads with *"Structs and tuples are open"*, states the rule that explains both length behaviours (a member is required exactly when its set excludes `undefined`, on both kinds), and carries the same guard:

> **Do not read "the result has the schema's length" as "tuples are closed" and add a length check here.** The set a tuple schema describes includes longer arrays; `Ts<T>` renders the closed approximation only because TypeScript cannot express the open one … and taking that rendering for the model is what produced #1622.

Also fixed the `:178` example comment (`// open: a longer array is accepted, and the extra is not carried over`), the four stale `validate` references in the header, and the two `Mirrors validate's …` factory comments. `constContainerParse`'s doc now says *why* both kinds are open — it iterates the **schema's** entries, so a longer array or undeclared key is never visited.

**2. `common/module.f.mjs`.** Its two consumers are now `parse` and the data form; the header and the four helper comments say that. The `eachEntry` note kept the no-allocation point but attributed to "a caller whose whole question is did every entry succeed" rather than to `validate`.

Both verified in the emitted declarations via `npm run prepack`: `common/module.f.d.mts` has **zero** live `validate` references, and `parse/module.f.d.mts` mentions "closed" only inside the guard sentence.

**3. `data/README.md:145`.** You're right that the top of the file was fixed and this section was missed. The anchor was dead — I confirmed it's the only broken cross-document anchor I introduced by walking every `](file#anchor)` in the tree (the 3 remaining hits are pre-existing directory-style links in `bnf`). Rewritten around the real state, with both pairs verified by execution:

```js
parse([42])([42, 'extra'])             // ['ok', [42]]
validate(toData([42]))([42, 'extra'])  // ['error', { path: [], message: 'unexpected value' }]

parse({ a: 42 })({ a: 42, b: 'x' })             // ['ok', { a: 42 }]
validate(toData({ a: 42 }))({ a: 42, b: 'x' })  // ['ok', { a: 42, b: 'x' }]
```

**4. The decision is recorded again** — `fjs/types/rtti/todo/data-form-open-tuples.md`. It carries the concrete fix the deleted todo held: map `Tuple` to `{ prefix, rest: unknown }`, drop `arraySetValidate`'s minimum-length check at `:994-998`.

Two things I checked rather than copied across:

- **Step 2 is genuinely required, not belt-and-braces.** With `rest: unknown` alone a *shorter* array is still rejected, because `arraySetValidate` requires `value.length >= pn` whenever `rest` is present: `{ prefix: [num, numOrUndef], rest: unknown }` vs `[42]` → error, where `parse([number, option(string)])([42])` → `['ok', [42, undefined]]`.
- **The divergence is arrays-only.** Struct keys already agree, because `arraySet` normalizes a `rest` of `never` away while `objectSet` normalizes a `rest` of `unknown` away — opposite identity elements for the same field.

The todo also flags `arraySetSubset`'s length reasoning as the thing most likely to break (its `p.rest === undefined && qn === pn` arm stops being reachable if every tuple gains a `rest`), and the canonical-form requirement that `{ prefix }` and `{ prefix, rest: unknown }` must not compare unequal once they denote the same set.

`npx tsc` 0, `fjs t` 2906 pass / 0 fail, `cov` 100/100/100, `ci-update` clean.

---

o2alexanderfedin (approved):
All four addressed — approving.

**The `parse` header now inverts the claim and guards it.** `:8-11` is "**Structs and tuples are open.** A value carrying more than the schema declares is accepted … the extras are accepted on the way in and absent on the way out", and `:27-32` adds the guard directly: "**Do not read 'the result has the schema's length' as 'tuples are closed' and add a length check here.** … taking that rendering for the model is what produced #1622." That's the right place for it — it's the file the tempted reader opens. No lowercase `validate` survives in either module's prose, only real identifiers, and both `@example` comments are corrected. Confirmed in the emitted `parse/module.f.d.mts` and `common/module.f.d.mts`: the only occurrences of "closed" that ship are inside the warning.

The new sentence at `:19-21` ("an absent member reads as `undefined`, on both kinds") checks out by execution — `parse([number,option(string)])([42])` → `['ok',[42,undefined]]`, `parse({a:number,b:option(string)})({a:1})` → `['ok',{a:1,b:undefined}]`, required cases still error.

**Anchor fixed** to `#structs-and-tuples-are-open`, matching `README.md:31`; repo-wide grep for `tuples-are-closed` returns nothing, and the stale "thunk-direct `validate`" / "three runtime consumers" text is gone.

**`todo/data-form-open-tuples.md` captures the decision and adds to it.** Against the deleted section it carries over the divergence example, the `arraySet`/`objectSet` opposite-identity explanation and both fix steps, plus a `subset`/canonical-form/serialization checklist the original lacked. Its load-bearing claim verified by hand-building the pattern: with `rest: unknown` alone a *shorter* array is still rejected — `validate([rules,{array:[{prefix,rest:unknown}]}])([42])` → `['error',{path:[],message:'unexpected value'}]` while `parse([number,option(string)])([42])` → `['ok',[42,undefined]]` — so step 2 really is needed as well as step 1. Line citations check out (`data/module.f.mjs:296-320`, the length check inside `:994-998`, `arraySetSubset`'s `p.rest === undefined && qn === pn` at `:378`), and "inherited, not introduced" holds — the check dates to `daf48b68` (#1539), predating #1622.

A 27-schema × 24-value differential of `parse(s)(v)` vs `validate(toData(s))(v)` gives 25 verdict flips, **all tuple-length disagreements**, including one nested inside an object, and no struct-key flip — so "the divergence is arrays-only" is right in both the README and the todo.

**`common`'s consumer list is correct**: importers of `rtti/common/module.f.mjs` are exactly `parse/module.f.mjs` and `data/module.f.mjs`, and the narrowed wording matches — `data/module.f.mjs:28` imports only `eachEntry, isArray, verror`, no `visit`.

The 106 changed lines in the two production modules are comment-only: stripping block and line comments from both at `61e16f11` and at this head gives byte-identical files. Green: `tsc` 0, `fjs t` 2906 pass / 0 fail, `cov` 100/100/100, `prepack` 0, `ci-update` clean.

sergey-shandar:
`fjs/dev/package_json` is now removed in `ad9ed02`, implementing the todo added earlier in this PR — which is deleted along with the module.

**Heads up that this changes an already-reviewed PR.** I did it here rather than in a follow-up because the todo exists only on this branch and the exception it describes is *introduced* by this branch. A separate PR would have to land `mapOk(() => value)` and its cast first, then remove them — so the exception never enters `main` at all this way. Happy to split it back out if you'd rather re-review the migration as approved.

What it removes: the one remaining site needing "check the shape, return the original", which is the reader shape this PR retires. Prerequisites re-checked before deleting — no `.mjs`/`.ts`/`.json`/`.yml` outside the module names it, and nothing else in-tree parses `package.json` (`fjs/ci` only checks the file exists).

`npx tsc` 0, `fjs t` **2902** pass / 0 fail (down 4 with the module's proofs), `cov` 100/100/100, `ci-update` clean. Changelog gains a second `**BREAKING CHANGES:**` line — `files` globs every `.mjs` and there's no `exports` map, so the module was deep-importable.

**One correction to the differential I reported earlier.** Re-running it after the removal gave *0 changed rows*, which is wrong. The cause: `git stash` was my baseline, and that worked only while the migration was uncommitted — once committed, stashing reverts nothing relevant and compares the branch against itself. I redid it against a real `origin/main` git worktree:

| | result |
| --- | --- |
| rows | 25 |
| changed | 10 |
| `ok` ↔ `error` flips | **0** |
| detect changed | **0 of 8** |

Same 10 rows as before across the same consumers, so the earlier conclusion holds — but the earlier run's method would not have caught a regression introduced by a *committed* change, and I should have noticed that when I re-ran it rather than when the result looked impossible.

---

PR #1625, opened 2026-08-16
dev: remove package_json

sergey-shandar:
Implements `fjs/dev/package_json/todo/remove-module.md`, deleted along with the module it describes.

#1624 merged at `88175b5`, before the removal commit on that branch — so this is a fresh change off the new `main`, not a re-push of that work.

## Why

The module has no consumers anywhere, and it is the only place left needing *"check the shape, return the original"* — the reader shape retired when `validate` was deleted. That is exactly why it needed a local exception in #1624, which is now on `main`:

```js
export const validatePackageJson = value =>
    mapOk(() => /** @type {PackageJson} */ (value))(parseShape(value))
```

Every other `validate` call site migrated to `parse` by changing an import. This one could not, because it is documented as *"reading metadata without losing unrelated fields before write-back"* and a proof pinned `Object.is(result, input)`. Removing the module takes the exception with it.

The whole surface was a three-optional-field schema and two functions over it:

```js
{ name: option(string), version: option(string), scripts: option(record(string)) }
```

A caller that needs this declares the schema and calls `parse`.

## Prerequisites, re-checked against `main`

The todo's first task is "confirm no consumer has appeared since this was filed", so these were re-run rather than carried over:

- No `.mjs`, `.ts`, `.json`, `.yml`, or `.yaml` outside `fjs/dev/package_json/` names it.
- Nothing enumerates it in a module list — `fjs/dev`'s `loadModuleMap` discovers by filesystem scan, so it simply stops finding it.
- Nothing else in-tree parses `package.json`. `fjs/ci` is the only other module that touches the file: two mentions are comments, and `ci/proof.f.mjs` puts a `package.json` into a *virtual* filesystem to test the CI generator against a Node-shaped repo, without reading its contents.
- The only remaining reference in the tree is `changelog/0.44.0.md`, which is published history and left alone.

## Verification

- `npx tsc` — 0 errors.
- `node fjs/module.mjs test` — 2902 pass, 0 fail (down 4: the module's own proofs).
- `npm run prepack` — 0, and no declaration left behind for the deleted paths.
- `npm run ci-update` — leaves the tree clean.

**No differential.** Nothing imports the module, so there is no caller whose behavior could change — a differential would compare two identical outputs and prove nothing. The observable delta is the published surface, which is what makes the changelog entry breaking: `package.json` has no `exports` map and `files` is `["**/*.js", "**/*.d.ts", "**/*.mjs", "**/*.d.mts"]`, so `fjs/dev/package_json/module.f.mjs` was deep-importable from the npm package.

## Changelog

- **BREAKING CHANGES:** `dev/package_json` is removed. It had no consumers, and
  its schema is three optional fields any caller can declare and `parse`.

---


o2alexanderfedin (approved):
Approving. Genuinely unused, and the breaking-change handling is right.

Grepped the whole worktree with no include filter — so `.github/`, `nanvm-lib/`, `*.nix`, `*.yml`, `*.toml`, `*.rs`, `*.md`, `*.json`, lockfiles — for every export (`packageJsonSchema`, `validatePackageJson`, `validatePackageJsonText`, and the typedefs `PackageJson`, `JsonTextError`), the path fragment `package_json`, and a case-insensitive `package[_-]?json` sweep. Three hits, all benign: `deno.lock:145` (a lockfile key), `fjs/ci/proof.f.mjs` (a local parameter named `packageJson`, unrelated), and `changelog/0.44.0.md:374-376` (published history, correctly untouched). No dangling `@import` — the module's own pointed outward and nothing pointed in. `fjs/dev/module.f.mjs` discovers modules by filesystem scan, so there's no list to update, and nothing references the deleted todo.

Re-verified the todo's claims against the tree as it is now rather than trusting them, since #1624 landed in between: `package.json` has no `exports` map and `files` is `["**/*.js","**/*.d.ts","**/*.mjs","**/*.d.mts"]`, so the module *was* deep-importable — the `**BREAKING CHANGES:**` prefix is required and present, and "update every importer" is vacuous with none. The `fjs/ci` claims hold: `config/module.f.mjs:25` and `deno/module.f.mjs:7` are comments, and `ci/proof.f.mjs:34` puts `package.json` into a virtual filesystem without reading its contents.

Declaration diff at base vs head is exactly the two removed files (358 → 356), nothing else shipped changed. All four tasks in `remove-module.md` are done, and its design content is already captured in `fjs/types/rtti/README.md` §"Structs and tuples are open" — the todo's own `Related` pointer — so nothing unique dies with it.

`tsc` 0, `fjs t` 2902 pass / 0 fail (down exactly 4 from base, the module's four proofs, matching the body), `cov` 100/100/100 with the changed denominator, `prepack` 0, `ci-update` clean. Entry is ~150 chars and conformant, byte-identical to the body's `Changelog:` section.

PR #1626, opened 2026-08-16
todo/edag: table the operations, rename the AST to EDAG, file the __proto__ bug

sergey-shandar:
Follow-up to #1620, continuing to iterate on the stage 1 design of FunctionalScript's stable function representation, plus two issues found along the way.

**Renamed: AST → EDAG (expression DAG).** *AST* is now reserved for BNF parser output (`fjs/bnf`, `fjs/djs`); the function representation is the EDAG everywhere else. Both halves of the name are load-bearing — the graph has no statement nodes at all (`const` is sharing, `if` is `"?:"`, `return` is the root, an assert is a `","` operand), and sharing and acyclicity are both semantic, so "tree" was never right. The term comes from *Compilers: Principles, Techniques, and Tools*, with the status inverted: there a DAG is derived from a tree by common subexpression elimination and its sharing is an unobservable optimization, here the EDAG is primary and its sharing is authored and observable. Files renamed to `todo/edag-spec.md`, `todo/edag-stage1-discussion.md`, `todo/edag-graph-visualizer.md`; `spec/`, `nanvm-lib/`, `DESIGN.md` and the plan documents follow.

**The core invariant**, now stated explicitly: *any validated EDAG behaves on the VM exactly as the corresponding source behaves on a JavaScript engine* — same value when both complete (A1, with A2's tiered cross-engine guarantee), failure exactly when JS always throws (A3), with evaluation order and error identity outside behavior (A4). It binds every validated graph, not just compiler output, so restrictions must be structural rather than conventional; and since every EDAG has a source form, the printed source is the semantic reference and the invariant is testable by running both.

**Operations tables.** The operations we want, with their stage:

- structural — constants, `{…}`, `["[]", …]`, `["args"]`, `[".", o, p]`, `["()", f, args]`, `[".()", o, p, args]`, `["own", o, k]`, and the later `[",", …]` and `["=>", frame, body]`;
- operators tagged by their **JS symbol** — `["+", a, b]`, `["===", a, b]`, `["?:", c, t, e]` — with arity distinguishing unary from binary as in JS, and the short-circuit operators marked lazy;
- other — `["throw", node]`, `["self"]`, `["frame"]`, `["Number", node]`.

Tags are JS syntax wherever JS has syntax; word tags survive only where it has none (`"args"`, `"frame"`, `"self"`, `"throw"`, `"own"`).

**New design points recorded:**

- **Throwing is the only effect** in a pure language, so the whole membership apparatus exists for it and "may throw" is the single predicate an optimizer needs — which narrows A4's speculation ban: a *provably* non-throwing operand may be speculated.
- **Property keys are restricted** — a permitted string constant, a number constant, or a **unary** `"+"` / `"Number"` node — so a computed string never reaches prototype-chain lookup. The 2330 abuse cases become unrepresentable rather than checked, and computed-string access is the separate `["own", …]` operation, lowered from the recognized `Object.getOwnPropertyDescriptor(o, k)?.value` pattern.
- **Closed-scope model** — `["frame"]` is the captured array as `["args"]` is the argument array, so a function body's only leaves are constants, `args`, `frame` and `self`. Nothing refers outward, which resolves the nesting corners subjects 3 and 9 had left open and makes each function hash independently.
- **`["self"]` keeps recursion acyclic**, so content addressing works for it with no special machinery; mutual recursion remains open.
- **`let`, loops and TCO** (subject 11): a mutable variable cannot be a DAG node, so `let` must lower to explicit state threading.
- **`toString(f)` returns real, evaluable source** (subject 12): sharing must print as `const` bindings, and those same bindings carry frames — the EDAG's explicit frame is JS's implicit lexical capture.
- **Declare-before-use plus self** is what makes frames constructible; the proposed relaxation is an acyclicity rule, sound because textual order is unobservable.

**`spec/todo/2340-operators.md`: the comma operator is now allowed**, having been rejected as "useful only when we want to mutate" — the assert pattern is the counter-example, since a discarded operand in a pure language can only throw.

**New issues filed** (design only, no implementation):

- `todo/proto-property-key.md` — `__proto__` as a property key. The identifier and string spellings assign a prototype while only `["__proto__"]:` creates a property, which is where "JSON is a subset of FunctionalScript" breaks: `JSON.parse` has no prototype special case, so one text means a data property as JSON and a prototype assignment as a module. Proposes rejecting the two prototype-assigning spellings in FS source, always emitting the computed one, and records the dependency on constant-string computed keys.
- `fjs/djs/todo/proto-key-handling.md` — the current compiler mishandles it in both directions: the module emitter writes `export default {"__proto__":3}`, which evaluates to an object without the property, and the parser accepts source that is not valid FunctionalScript. Includes conformance fixtures, and the note that both output formats share one key serializer, so the fix must be per-emitter or it would put computed keys into `.json` output.
- `todo/edag-graph-visualizer.md` — a P4 issue for a pure EDAG-to-Mermaid/DOT renderer, since sharing is semantic and every textual view hides it.

Changelog: none


o2alexanderfedin (commented):
The `not allowed` → `1` flip is sound; the fallout from the tag rename isn't fully propagated. All of the below re-verified at `7cab6af0`.

**`spec/todo/2340-operators.md:41` names a tag that no longer exists.** It says the spellings "lower to the AST's `comma` operation", but `d827c9a1` renamed it to `","` throughout the discussion document — 6 occurrences of `[","`, zero of `["comma"`. The spec line was written one commit earlier and not carried forward.

**Same sentence points at the wrong file.** It cites `todo/ast-spec.md` as "where the exact semantics live"; that file contains zero occurrences of "comma" (it's the RTTI-schema proposal). The semantics are in `todo/ast-spec-stage1-discussion.md`, subject 8. The link resolves, the claim about its contents doesn't.

**`cond` is both retired and still referenced.** The new material says "the hypothetical `["cond", …]` of subject 3 is not needed" (`:211`) and "No separate `cond` node is needed: `?:` is the branch node" (`:500`), but `:57` still says "the control-flow branches of the future `cond` (subject 3)" and `:725` "Future: a control-flow branch operand (`cond`, subject 3) carries its guards as a `","` node inside the branch". `:725` was touched by the rename in this PR, so it was edited and left stale.

**The table's Notes column uses 2330's other names.** `:125` labels `["[]", object, property]` "instance property" and the next row "instance method call", but subject 6 at `:555-556` maps them the other way: "This proposal's `"[]"` = its `at`; `"[]()"` = its `at_call`". `instance_property`/`instance_method_call` are 2330's compile-time-known-name forms, which `:563` and `:590` lean toward keeping out of the AST entirely — so the table names the two things the document says the AST doesn't carry.

**The `#1620` miscount carried over unfixed:** `:32` "all three of" introduces four functions (`f`, `g`, `h`, `k`), still called "The statement spellings" while `:41` describes `k` as "an expression-bodied arrow, no block, no `return`".

**The description is stale against the head** — it says subject 6 is "leaning no" on symbolic structural tags, but `d827c9a1` decided the opposite ("Decided for the structural tags: they are JS syntax too"), and it still enumerates `["at", …]`, `["call", …]`, `["bindCall", …]`, `["comma", …]`, all four superseded. It becomes the squash commit.

Verified fine:

- The `not allowed` → `1` flip matches the file's own convention (`==`/`!=` stay `not allowed`), nothing else in `spec/` still calls comma disallowed, and it doesn't collide with `spec/README.md:8`'s whitelist principle — the old "the general ban stands" paragraph was correctly removed rather than left contradicting.
- Settled positions hold: every spelling is `[",", ...asserts, result]` — result last — and nothing reintroduces an ordering commitment (A4 still rejected at `:261-263`, "Membership is semantic; order is not" at `:655`, the unreachability rule at `:583`). The new "proven non-throwing may be speculated" carve-out appears in both places it belongs. The body doesn't repeat #1620's reversed-operand error.
- Table rows all correspond to something the document defines, nothing in prose is missing from it, no count is asserted. Behavioral claims checked by running node: comma yields the last operand, `&&`/`||`/`??` short-circuit, `?:` evaluates one arm, and `const p = [1,2,3].indexOf; p(1)` throws `TypeError` — the argument for a distinct `"[]()"`. The `"array"`-would-collide-with-`"[]"`-at-arity-2 point is correct, and the `DESIGN.md §8` citation fits. All 12 links and the `#operations` anchors resolve.
- `Changelog: none` with no entry file is right for `todo/` + `spec/todo/`; title conforms; header block intact. Branch green — `tsc` 0, `fjs t` 2902 pass / 0 fail, `cov` 100/100/100, `ci-update` clean.

One scope note: the head moved twice during review. The findings above are all re-checked at `7cab6af0`, but the new material in `ac253348` (throw operation, free-variable gap) and `7cab6af0` (no `if` operation) arrived after the pass and I haven't reviewed it.

o2alexanderfedin (commented):
The tag reshuffle is clean — no collisions, and the rename propagated. Four carried-over items are still open, plus one new inconsistency.

**Still open from the last pass:**

- `spec/todo/2340-operators.md:41` — still "lower to the AST's `comma` operation". The tag is `","` (`:16`, `:129`, `:698`); `"comma"` is no longer a tag anywhere. That file was untouched by all eight new commits.
- Same line still cites `todo/ast-spec.md` as "where the exact semantics live" — `grep -in comma todo/ast-spec.md` is zero hits; it's the RTTI-schema proposal. Subject 8 of the discussion document is the right target.
- `:57` "the control-flow branches of the future `cond` (subject 3)" and `:807` "Future: a control-flow branch operand (`cond`, subject 3)" still contradict `:291-294` ("The hypothetical `["cond", …]` of subject 3 is not needed") and `:581` ("No separate `cond` node is needed: `?:` is the branch node").
- `:32` "all three of" still introduces four functions (`f`, `g`, `h`, `k` at `:35-38`), and `:41` then treats all four as one AST.

**New — `["self"]` and the closed-scope model label the same fact opposite ways.** `:216` says the closed-scope model "**resolves the nesting corner** flagged in subjects 3 and 9: a node cannot be shared across a function boundary … not a rule to enforce so much as a consequence of the model". `:243` says `["self"]` "inherits the same **open corner**: once `["=>", …]` nests, a node using `["self"]` cannot be shared across nesting depths (subject 3, subject 9)". The underlying fact agrees; the status doesn't — and both cite the same two subjects, whose text was rewritten in this PR to say resolved (`:590`, `:831-834`).

Minor: subject 6 is `**Status:** open` and still reads "Leaning toward 2" at `:651`, while `:668-671` records "**Decided**: … option 2 above". No question is left open in its text.

The description is also still stale — it enumerates `["at", …]`, `["call", …]`, `["bindCall", …]`, `["comma", …]`, `["function", …]` (now `"."`, `"()"`, `".()"`, `","`, `"=>"`), still says "leaning no" on symbolic structural tags where subject 6 records three **Decided** rulings the other way, and doesn't mention any of the new operations. It becomes the squash commit.

**Verified — the collision question resolved correctly.** Enumerated every (tag, arity) pair:

| arity | tags |
|---|---|
| 0 | `[]`, `args`, `self`, `frame` |
| 1 | `throw`, `[]`, `+`, `-`, `!`, `~` |
| 2 | `[]`, `.`, `()`, `=>`, `,`, and the binary operators |
| 3 | `.()`, `?:`, `,` |
| n | `[]`, `,` |

No two distinct operations share a pair. The critical case — `["[]", a, b]` as a two-element array — is unique now that access moved to `"."`, and the old objection was *retired* rather than left standing: `:677-679` says "The earlier objection — that `["[]", a, b]` would read as both a two-element array and `a[b]` — disappeared with access moved to `"."`." Repo-wide grep for the superseded tags finds only four deliberate historical mentions, all framed as superseded.

Prior finding #4 is fixed: subject 6 now maps `"."` = `at`, `".()"` = `at_call` (`:640`), the table rows read "property access" / "method call", and `instance_property`/`instance_method_call` stay out of the AST.

The new operations cohere with the settled positions — A4 rejected and "membership is semantic; order is not" (`:798`), result last (`:19-21`), assert-unreachability (`:722-724`), reveal-after-all-succeed with the as-if rule (`:746-772`), spec-owns-spelling (`:728`), throwing-is-the-only-effect (`:320`). `frame`/`self`/closed-scope are function-boundary rules, orthogonal to `","` membership. `["self"]` is argued at `:226-232` as precisely what keeps a recursive function acyclic and hashable, consistent with subject 5's no-cycles rule. The one real pressure point against "no normalization" — frame slot order needing to be canonical for hash-as-written — is flagged open at `:861-863` rather than glossed.

JS claims verified by running node: `arguments.callee` throws in strict mode, `(1,2) → 2`, `{} === {}` false while `const x={}; [x,x]` shares, a named function expression self-references without leaking the name, `null[0]` throws, `const p = a.indexOf; p("bison")` throws while `a.indexOf("bison")` returns 0, and `(a => (a, a+2)).toString()` round-trips verbatim. Spec cross-claims check out against `3110-function.md`, `3111-function-frame.md`, `2330-property-accessor.md` and `mutability.md`. Subjects 1-11 complete, no gaps; 48 links and anchors all resolve. Green: `tsc` 0, `fjs t` 2902 pass / 0 fail, `cov` 100/100/100, `ci-update` clean.

Scope note: the head moved twice more during this pass. Everything above is re-verified at `26036ae2`, but subject 12 (`b00a1c93`), printed consts carrying frames (`392b326d`) and `26036ae2` arrived after and are unreviewed.

o2alexanderfedin (commented):
One new item; the seven from my last comment all still stand at this head (I won't repeat them).

**`todo/ast-graph-visualizer.md:8-9` cites the wrong file — same shape as finding 2.**

> A function's AST is a **DAG whose sharing is semantic** ([ast-spec](./ast-spec.md)): a node referenced twice is evaluated once, and `[x, x]` is a different function from `[{}, {}]`.

`todo/ast-spec.md` has no occurrence of *DAG*, *shar*(ing), or *graph* — the single "shar" hit is `:62` "test vectors **shared** by the FJS and Rust implementations" — and no `[x, x]` / `[{}, {}]` example. Those live in `todo/ast-spec-stage1-discussion.md:72` and `:826`, and the DAG model is subject 1. The `### Related` bullet at `:60` is fine; it's the in-text semantic citation that's unsupported.

`decc56f2` didn't touch finding 5 — its diff is confined to `:848-884`. The new closing paragraph at `:878-884` ("For now the design is built around `["self"]`") is about mutual recursion and redesign freedom, a different question, and it's correct on its own terms: `DESIGN.md §2` does say "If the new version can have a better, simpler API, change it — never hesitate", and the link resolves. The resolved-vs-open label on the nesting corner at `:216` / `:243` is untouched.

Verified clean in the new material:

- The completed example at `:848-860` type-checks against the table: `["()", ["self"], ["[]"]]` matches `["()", object, args]` with `args` "one node yielding an array" (`:127`), and `["[]", ...node]` with zero operands (`:124`; subject 5's validation imposes no minimum arity). `f` takes no parameters, so an empty args array is right, and "needing no frame entry at all" matches `:865-868`. Ran the JS: `f() = 12`, `g() = 6`, self-reference inside the body works with no TDZ error. The `// error — b is not defined before f` comment is the 3140 static rule, not a runtime claim, so the run doesn't contradict it.
- Everything else in the visualizer checks out: result-last and assert operands match `:129`/`:698`; the lazy set `&&`/`||`/`??`/`?:` matches `:177-178`; `["args"]`/`["frame"]`/`["self"]` as leaves match `:187-188`/`:213-215`; `["=>", frame, body]` matches `:130`; "a body's only inbound edges are its frame" matches `:213-220`. The JSON/DJS claims are `spec/README.md:127-128` verbatim. It specifies a pure function, data in / text out, with Mermaid and DOT as output formats rather than libraries — no dependency implied.
- `**Priority:** P4` / `**Status:** open` are both allowed values, and with no task list there's nothing for the status to be out of sync with.
- Nothing regressed: subjects 1-12 each appear once, links and anchors resolve, and the settled positions in subjects 8/A3/A4 are untouched by the diff. Green — `tsc` 0, `fjs t` 2902 pass / 0 fail, `cov` 100/100/100, `ci-update` clean.

Worth noting for the squash: the description now also doesn't mention `todo/ast-graph-visualizer.md`, so it would describe a PR that adds a file it never names.

o2alexanderfedin (commented):
Reviewed at `5f1a5ff4` (the EDAG rename landed mid-pass; it does fix the half-renamed ASDAG I'd otherwise have flagged). My eight earlier findings all still stand — not repeating them, except to note #7 is now further out of date: the description doesn't mention `own`, `Number`, the key restriction, the core invariant, or the rename, and still names `todo/ast-spec-stage1-discussion.md`, a path that no longer exists.

**Subject 5's validation rule admits binary `"+"`, which reopens exactly what the restriction closes.** `:729-730` permits "a permitted string constant, a number constant, or a `"+"` / `"Number"` node" — no arity qualifier, while the Operations statement at `:210` says `"+"` **(unary)**. Binary `"+"` is the same tag and yields a string:

```
typeof ("constr" + "uctor")        // 'string'
({})["constr" + "uctor"] === Object // true
```

So `[".", o, ["+", "constr", "uctor"]]` passes subject 5 as written, against `:213` "a run-time-computed **string** can never reach `"."`" and `:190-193`. Subject 5 is the normative list an implementer codes from, so the qualifier needs to be there too.

**The object constructor's JS equivalence fails for `"__proto__"`**, and the new invariant makes that load-bearing:

```
Object.keys({ __proto__: proto })                    // []      — no own property
Object.getPrototypeOf({ __proto__: proto }) === proto // true    — prototype set
Object.keys({ ["__proto__"]: proto })                // ['__proto__']
```

An EDAG object constructor with key `"__proto__"` means "own property"; printed as `{ "__proto__": v }` a JS engine sets the prototype instead. Subject 5's validation list has no `__proto__` key rule, and `:150` says "anything a hostile graph could express, validation must have already ruled out".

**Subject 4's rationale is false for integer-index keys**, also newly load-bearing. It says "JS objects preserve insertion order … so reordering an object constructor's keys changes the resulting object":

```
Object.keys({ "2":'two', "1":'one', b:'bee', "0":'zero' })  // ['0','1','2','b']
Object.keys({"1":1,"2":2}) vs Object.keys({"2":2,"1":1})     // identical
```

Index keys (canonical `ToString(ToUint32)` in `[0, 2^32-2]`) always sort first, ascending. So `{"2":a,"1":b}` and `{"1":b,"2":a}` are two EDAGs with two hashes but one JS value — the print/run/compare test can't distinguish them. The conclusion (don't sort constructor keys) still stands on the non-index cases; only this rationale over-claims.

**The core invariant cites A2 for a guarantee A2 denies.** `:137` says "the same value whenever both complete (**A1, A2**)", but A2 at `:474-480` says non-CA engine results "**may vary** … object/function identity, identity-based ordering … and `toString(f)` text". Two conforming engines can complete with different values, so at most one matches the JS engine.

**Two lists weren't updated for the new tags.** Subject 6 at `:801-802` still says "Word tags now survive only where JS genuinely has no expression spelling: `"args"`, `"frame"`, `"self"`, `"throw"`" — `own` is missing. And subject 12 at `:1129-1133` still says exactly two operations need runnable workarounds (`throw`, `self`), but `["own", o, k]` prints as `Object.getOwnPropertyDescriptor(o, k)?.value`, which renders the bare built-in `Object` — while `:1140-1143` says "the only leaves to render are constants, `["args"]`, `["frame"]` and `["self"]`". Subject 10 still lists built-in namespaces as open.

Lower confidence: subject 6's new mapping table sends `[".", o, p]` to "`instance_property` + `at`", but 2330:78-85 says `instance_property` covers only *implemented* names and "All other property names generate `own_property` commands".

Verified clean: **no tag/arity collisions** after re-enumerating with `own` and `Number` added — every tag is distinct, so the arity-2 array-vs-access collision hasn't returned. The `Number` claims are all correct, including `Number(1n)` → `1` while `+1n` throws, so "accepts bigints, unlike unary `+`" holds; it complements 2330's `+`-before-an-index rule rather than duplicating it. Number→key stringification can't produce a prohibited name. `own` correctly can't reach the prototype chain (`Object.getOwnPropertyDescriptor({}, 'constructor')` is `undefined`). All 19 links resolve, subjects 1-12 complete, `Changelog: none` right for a docs-only PR. Green at `212cb201`: `tsc` 0, `fjs t` 2902 pass / 0 fail, `cov` 100/100/100, `ci-update` clean.

sergey-shandar:
All seven open findings from the three passes are fixed in 7fe5d99, plus the description. Note the AST → EDAG rename since your last pass moved several of these lines and renamed three files.

- **`2340-operators.md:41` naming a dead tag** — now `` `","` ``.
- **Same line citing the wrong file** — was pointing at the RTTI-schema proposal; now cites `todo/edag-stage1-discussion.md`, subject 8, where the semantics actually live.
- **`cond` retired but still referenced** — both sites fixed: the baseline now says "the control-flow branches of `"?:"`", and subject 8's future note says "an arm of `"?:"`".
- **"all three of" introducing four functions** — now "all four of".
- **`["self"]` labelling the nesting property an *open corner*** while the closed-scope model calls the same fact resolved — the self bullet now states it as the model's consequence: a node cannot be shared across nesting depths, which is what makes each body hash independently.
- **Subject 6 marked `open` and still "leaning"** while recording three decisions — now `**Status:** decided`, with the options/leaning paragraph replaced by the ruling. Its stale consequence went too: "prohibited names must be rejected at *runtime*" was superseded by the syntactic property-key rule, so a prohibited name is now a validation error at construction and a computed string never reaches `"."` at all. Recorded as an explicit correction rather than a silent edit.
- **Visualizer citing `edag-spec` for a claim that file doesn't make** — the in-text citation now points at subject 1; `### Related` lists both, distinguishing the RTTI schema from where the semantics are worked out.
- **Description** — rewritten for the current head: the rename and its rationale, the core invariant, the current tag vocabulary, every new operation, and the visualizer file it previously never named.

On the scope notes: the material that arrived after each pass — `throw`, `["self"]`, `["frame"]` and the closed-scope model, no-`if`, subject 11 (`let`/TCO), subject 12 (`toString`), the property-key restriction, and the rename — is all unreviewed, so a fresh pass would be worth it.

---

o2alexanderfedin (commented):
Reviewed at `9fa4fd98`; re-checked the findings below at `8072798b`. Good progress — 1, 2, 3, 5, 6, 8, 10 are fixed and the new text is correct in each case. The `__proto__` compiler bug reproduces exactly as filed, both halves:

```
$ node ./fjs/module.mjs compile tmpproto/id.f.js tmpproto/id.out.json   (exit 0)
  in:  export default { __proto__: 3, a: 1 }
  out: {"__proto__":3,"a":1}
```

JS reads the input as own keys `["a"]`; the emitted JSON means `["__proto__","a"]`. Silent, exit 0. `parseObjectStartOp` at `fjs/djs/parser/module.f.mjs:396-401` falls through `case 'string': case 'id':` into `pushKey` as documented.

**The fix in `proto-key-compiler-bug.md:43-48` would break the JSON output it says to protect.** It calls `stringify` / `stringifyAsTree` collectively "the `.f.js` module path", but `fjs/djs/module.f.mjs:45-47` splits them the other way:

```js
outputFileName.endsWith('.json') ? stringifyAsTree(sort)(result[1]) : stringify(sort)(result[1])
```

`stringifyAsTree` **is** the `.json` path. And both funnel through the single `propertySerialize` inside `buildSerialize` (`fjs/djs/serializer/module.f.mjs:92-99`, reached from `serializeWithoutConst:128` and `serializeWithConst:131`). So the task "`propertySerialize` in the **DJS** serializer emits `["__proto__"]:`" would put `["__proto__"]:` into `fjs compile x.f.js out.json` output — which `:42` calls "not even valid JSON" and `:45` says fixing there "would break JSON output". The fix has to be per-emitter, not in the shared helper.

**Subject 5 still admits binary `"+"`** — the one I'd most want closed. `:739-741` is untouched: "a permitted string constant, a number constant, or a `"+"` / `"Number"` node", no arity qualifier, while Operations `:210` says `"+"` **(unary)**. `typeof ("constr"+"uctor")` is `'string'` and `({})["constr"+"uctor"] === Object` is `true`, so `[".", o, ["+", "constr", "uctor"]]` still passes validation as subject 5 words it.

**`proto-property-key.md:46-47` cites the wrong file.** "FunctionalScript has no prototypes at run time ([function-frame](../spec/todo/3111-function-frame.md))" — that file has zero occurrences of "prototype"; it's entirely about NaNVM frames and captured consts. The claim is supported by `2330-property-accessor.md:27` ("In FS we plan to not have prototype chains") and `design-principles.md:32,36`. The discussion doc's three other citations of 3111 all use it correctly.

**`7fe5d994` stranded a cross-reference.** `:798-799` still says "the static/numeric split that remains is a bytecode specialization, **as option 2 said**", but that commit deleted subject 6's numbered `Options:` list. There is no longer an option 2 in subject 6. (The one at `:1112` is subject 11's own list, fine.)

**`proto-key-compiler-bug.md` is in the wrong directory.** Its content is entirely `fjs/djs/serializer/` and `fjs/djs/parser/`, and `todo/README.md` states as a bolded rule: "a bug or task scoped to `fjs/foo/bar/` goes in `fjs/foo/bar/todo/{slug-kebab}.md`, not in a parent directory and not here." `fjs/djs/todo/` exists with 10 such files. (`proto-property-key.md` in root `todo/` is right — it's a language-design question.)

Still open from earlier, unchanged: subject 4's "reordering keys changes the resulting object" rationale is false for integer-index keys (`Object.keys({"2":a,"1":b})` and `Object.keys({"1":b,"2":a})` are both `["1","2"]`); `:137` cites A2 for a guarantee A2 denies; subject 6's word-tag list at `:807-808` omits `"own"`; `:792` maps `[".", o, p]` to `instance_property` where 2330:85 sends non-implemented names to `own_property`. Finding 4 is half-fixed — "all four of" now, but `:33` still calls them "statement spellings" while `:43` says the last is expression-bodied.

Two smaller ones on the bug file: the header says `**Status:** open` with `**Blocked by:** computed property keys`, but `:80` says "The two rejections are not blocked and can land first" — `todo/README.md` has `blocked` as its own status; and `:63` writes bare `#2410`, which `todo/README.md` reserves for GitHub PR/issue numbers (`proto-property-key.md:98` does it correctly).

The description is now current on everything else, but doesn't mention either new todo file, and the title doesn't cover them.

Everything else verified: all paths and symbols in both new files exist, the quoted `spec/README.md:48` and `:103` lines are byte-exact, the JS behaviors all reproduce, subjects 1-12 and A1-A4 complete, every link and anchor resolves. Green: `tsc` 0, `fjs t` 2902 pass / 0 fail, `cov` 100/100/100, `ci-update` clean.

sergey-shandar:
All findings from the latest two passes are addressed in 288fc71, plus the reverse-direction test. I verified each claim by running the code rather than taking it on trust — every one held, including the two that reverse things I had written.

**The serializer prescription was wrong, and worse than stale.** Confirmed `fjs/djs/module.f.mjs:45-47`: `stringifyAsTree` is the `.json` path, `stringify` is the module path — the opposite of what the issue said — and both reach the single `propertySerialize` in `buildSerialize`. So the fix as filed would have emitted `["__proto__"]:` into `.json` output, exactly what the same file called "not even valid JSON". The task is now "parameterize key serialization per output format", not "patch the shared helper".

This also invalidated my reproduction, which used `stringifyAsTree` — i.e. it was showing *correct JSON behavior*. Re-ran on both paths:

```
stringify       (.f.js path): "export default {\"__proto__\":3}"   ← the bug
stringifyAsTree (.json path): "{\"__proto__\":3}"                  ← correct
```

The bug is real, on the module path; the issue now reproduces it there.

**Binary `"+"` — confirmed and closed.** `typeof ("constr" + "uctor")` is `'string'` and `({})["constr" + "uctor"] === Object` is `true`, so subject 5 as worded admitted a run-time-rebuilt prohibited name. It now says **unary** `"+"` / `"Number"`, with that attack spelled out as the reason the qualifier is load-bearing.

**Integer-index keys — confirmed.** `Object.keys({"2":a,"1":b})` and `Object.keys({"1":b,"2":a})` are both `["1","2"]`. Subject 4 now states the exception explicitly, including that two such EDAGs are one JS value with two hashes that print-run-compare cannot separate. The conclusion (don't sort constructor keys) still stands on the non-index cases, and is marked as resting there.

**The 3111 citation — confirmed wrong.** `grep -c prototype spec/todo/3111-function-frame.md` is 0. Now cites `2330-property-accessor.md`.

Also fixed: the core invariant no longer cites A2 for an unconditional agreement A2 tiers; subject 6 adds `"own"` to the word-tag list, drops the reference to the options list `7fe5d99` deleted, and maps `"."` per 2330's actual routing (non-implemented names go to `own_property`); subject 12 counts `["own", …]` among the operations needing a printed form and notes it renders the built-in `Object`, which is a fourth thing the printer emits and why built-in namespaces stay open in subject 10; subject 5 gains a `__proto__` key rule; "statement spellings" is now "other spellings", since one of the four is expression-bodied.

**Moved** `proto-key-compiler-bug.md` → `fjs/djs/todo/proto-key-handling.md` per `todo/README.md`'s scoping rule, with paths rebased and the bare `#2410` replaced by an explicit link. `Blocked by` is dropped from the header since only the accept-computed-keys task is blocked; that's now noted on the task itself.

**Added**, per a request outside the review: `fjs compile protoGood.js out.json` must emit plain JSON with no `["__proto__"]:` artifact — pinning the deliberate divergence between the emitters and closing the loop `proto.json → .js → .json`.

Title and description now cover all three new files.

---

o2alexanderfedin (commented):
All twelve are fixed and the replacements check out — including the binary `"+"` hole, which is genuinely closed: I enumerated every form subject 5 now permits (permitted string constant, number constant, unary `"+"`, `"Number"`) and none can yield a string at run time. `Number(1n)` is a number while `+1n` throws, `Number(Symbol())` throws, and number→key stringification only produces `"0"`, `"NaN"`, `"Infinity"`, `"1e+21"`… — none prohibited.

The correction landed in `fjs/djs/todo/proto-key-handling.md` but not in its sibling, so the two files now contradict each other.

**`todo/proto-property-key.md:28-35` still uses the JSON emitter as the example of the bug.** It says "**The DJS serializer emits the unsafe form today**, so it does not round-trip its own output" and demonstrates with `stringifyAsTree(sort)(o) // '{"__proto__":3}'`. But `stringifyAsTree` is the `.json` path (`fjs/djs/module.f.mjs:46`), and `proto-key-handling.md:53-55` — which this file links to — now says of exactly that output: "**The JSON output is already correct and must stay.** `JSON.parse('{"__proto__":3}')` yields an own property, so JSON round-trips today." `288fc71a` switched the bug file's example from `stringifyAsTree` to `stringify`; this one kept the old one. Confirmed: `stringify(sort)(o)` → `'export default {"__proto__":3}'`, `stringifyAsTree(sort)(o)` → `'{"__proto__":3}'`.

**`:52` and `:93-94` instruct the fix the bug file says would break JSON.** `:52` reads "Emit `["__proto__"]:` for that key, **always**", and the task says "Serializer: emit `["__proto__"]:` for that key (`fjs/djs/serializer`)". That's the single shared module, and `proto-key-handling.md:49-51` now says patching it "would emit `["__proto__"]:` into `.json` files, which no JSON parser accepts", with `:152-157` requiring the JSON emitter not to.

Smaller, all from the same edit pass:

- `edag-stage1-discussion.md:753` and `:757` duplicate the sentence — "anything else is a validation error." then "Anything else is a validation error, which is what keeps computed-string prototype access unrepresentable…". Splicing artifact.
- `:719`, `:757`, `:818` are left un-rewrapped in a file otherwise wrapped at ~72 columns.
- `:726` keeps "FS has no run-time prototypes" in the present tense, which is the phrasing you corrected to "plans to have no prototype chains at run time" in `proto-property-key.md:46`.

Verified on the rest: the moved file is in the right place, nothing references the old path, and all six of its relative links resolve. The corrected serializer description is accurate and its proposed per-emitter fix is implementable — `buildSerialize` is module-private with `stringify`/`stringifyAsTree` as the two distinguishable entry points. `ee8db8db`'s reverse-direction claim holds under the real CLI: compiling a `__proto__`-keyed `.json` gives byte-exactly `{"__proto__":{"a":42}}` with no computed-key artifact, while the `.js` direction reproduces the bug end-to-end. Subjects 1-12 complete, links and anchors resolve, `Changelog: none` present with no entry file, title conforms. Green: `tsc` 0, `fjs t` 2902 pass / 0 fail, `cov` 100/100/100, `ci-update` clean.

sergey-shandar:
All five fixed in 4c1eef7. Every one was collateral from my own previous edit pass — I corrected the bug file and left its sibling behind.

**The two contradictions in `todo/proto-property-key.md`:**

- The demonstration used `stringifyAsTree`, the JSON path, whose output the bug file now (correctly) calls fine. It now shows `stringify(sort)(o)` → `'export default {"__proto__":3}'`, and states explicitly that the JSON emitter is correct as it stands, with a pointer to the per-emitter fix.
- "Emit `["__proto__"]:` for that key, **always**" and the task naming `fjs/djs/serializer` both prescribed exactly what the bug file says would break JSON. The proposal now splits by output format — computed key in FunctionalScript output, plain key in JSON, diverging on purpose — and the task notes the two formats share one key serializer today, so the change is per-emitter.

**The three smaller ones in `edag-stage1-discussion.md`:** the duplicated "anything else is a validation error" sentence is gone (the bullet is rewritten rather than patched, so the reasoning reads in one pass), the spliced lines are re-wrapped to the file's ~72 columns, and subject 4's "FS has no run-time prototypes" now matches the phrasing you corrected in the sibling — "plans to have no prototype chains at run time".

Thanks for catching the serializer inversion two passes ago in particular; that one had propagated into a prescribed fix that would have broken JSON output, and it isn't visible without reading the dispatch in `fjs/djs/module.f.mjs` against the shared `buildSerialize`.

---

o2alexanderfedin (approved):
Approving. The three documents now agree, and I verified the serializer facts by execution rather than reading:

```
own keys: ["__proto__"]
stringify(sort)(o)        = "export default {\"__proto__\":3}"
stringifyAsTree(sort)(o)  = "{\"__proto__\":3}"
JSON.parse own __proto__? true ["__proto__"]
```

Matching `fjs/djs/module.f.mjs:45-47`. Every claim the three files make about which emitter does what is correct against that.

The prescription is now consistent: `proto-property-key.md:58-63` "In **FunctionalScript output**, emit `["__proto__"]:` … In **JSON output** the plain `"__proto__":` key stays … The two emitters diverge here on purpose", its task at `:102-106` noting "both formats share one key serializer today, so the change is per-emitter", and `proto-key-handling.md:57-59` "Parameterize key serialization per output format … rather than changing the shared helper." No "always" left anywhere. The duplicated sentence is gone, and `:727-728` now reads "FS plans to have no prototype chains at run time", matching the corrected sibling and `2330:27`.

Two small things, neither blocking:

**`proto-property-key.md:44` lost its antecedent.** The new JSON paragraph was inserted between the `stringify` code block and the sentence that follows it, so "Evaluating that text gives an object with **no** properties" now sits directly after "which `JSON.parse` reads back as an own property" — as read, the two adjacent sentences contradict. The intended referent is the module output at `:35`. Moving `:44-47` up against the code block, or naming the referent ("Evaluating the module output…"), fixes it.

**`edag-stage1-discussion.md:820` is still 120 columns** in a file otherwise wrapped at ~72 — the third of the three I mentioned; the other two were rewrapped. `proto-property-key.md:52` and `proto-key-handling.md:74` are the same shape if you want to sweep them together. No rule in `AGENTS.md` covers line width, so purely cosmetic.

Verified otherwise: subjects 1-12 complete with no gaps; all links and anchors in the changed files resolve; no stranded cross-references remain — `([Operations](#operations))` at `:757`, `proto-property-key.md:117-118`'s pointer to subject 12, and `edag:769`'s all resolve to text saying what the referrer claims. `proto-key-handling.md:31-51`'s implementation claims check out (`propertySerialize` inside `buildSerialize` at `serializer/module.f.mjs:92-98`, `stringifyAsTree = serializeWithoutConst` at `:196`, `stringify` via `serializeWithConst` at `:191`), and the three-spelling behaviors reproduce in node. Title conforms, `Changelog: none` with no entry file is right, and the body names every file the PR adds. Green: `tsc` 0, `fjs t` 2902 pass / 0 fail, `cov` 100/100/100, `ci-update` clean.

PR #1627, opened 2026-08-17
Add self-describing BNF grammar module

sergey-shandar:
## Summary
This PR adds a self-describing BNF grammar module that implements the canonical Backus-Naur Form metasyntax using this package's own functional grammar representation. This allows the BNF grammar to parse BNF grammars themselves.

## Key Changes

- **New module `fjs/bnf/self/module.f.mjs`**: Implements the complete BNF metasyntax (`<rule-name> ::= <expression>`) as a functional grammar. Transcribed from the canonical Wikipedia BNF-of-BNF example with three key adaptations:
  - Rewrites left-recursive `<rule-name>` rule to right-recursive form (`<letter> <rule-char>*`) for LL(1) compatibility
  - Reorders alternatives in `<expression>` and `<list>` rules to try longer matches first (greedy ordered-choice semantics)
  - Removes quote characters from `<symbol>` to prevent greedy matching from consuming closing quotes as content; quotes are added back only in `character1`/`character2` rules

- **New documentation `fjs/bnf/self/README.md`**: Comprehensive guide explaining:
  - The canonical BNF grammar structure
  - Why three specific adaptations were necessary for recursive descent parsing
  - How to use the grammar with the package's parser backends
  - Reference to proof examples

- **New test/proof file `fjs/bnf/self/proof.f.mjs`**: Demonstrates the grammar works correctly with test cases covering:
  - Single and multiple rules
  - Rule references and literals
  - Whitespace handling
  - Single and double-quoted strings
  - Rejection of invalid syntax (missing operators, brackets, line endings, empty rule names)

- **Updated `fjs/bnf/README.md`**: Added reference to the new self-describing BNF module

## Implementation Details

The grammar entry point is the `syntax` rule, which matches one or more BNF production rules. All three adaptations from the canonical form preserve language equivalence while enabling LL(1) recursive descent parsing. The module exports only the `syntax` rule; internal rules are private constants.

https://claude.ai/code/session_01DKKwQ1J1Fjsk1gEsXCXk3R


o2alexanderfedin (commented):
The self-describing claim holds, and I checked it rather than took it: I wrote out the canonical BNF-of-BNF as 18 rules (ellipses expanded) and fed it to `descentParser(syntax)` — accepted, `idx === len === 1353`, a fully-consuming parse. All 18 rules also parse individually, and 7 malformed variants are rejected. The README's `## Trying it` snippet runs verbatim. Gates are green: `tsc` clean, 3043/3043, `cov` exit 0 with `fjs/bnf/self/module.f.mjs` itself at 100/100/100, `ci-update` no diff, 18/18 checks.

The proof does prove — 13 of 17 grammar mutations redden it, and the three README-documented adaptations are each independently load-bearing.

Three things (all separate from the Codex bot's points, which I verified and agree with — the LL(1) throw, the missing `~`, and the missing changelog file all reproduce):

1. **`fjs/bnf/self/README.md:71` is false.** "All three rewrites accept exactly the same language as the rule they replace" doesn't hold for the third — and `module.f.mjs:19` already says so correctly, calling it "not a left-recursion fix but a **correction**". Removing both quotes from `symbol` strictly narrows the language: canonical `<character1> ::= <character> | "'"` accepts `"` via `<symbol>`, so canonical accepts `<a> ::= "x"y"` and `<a> ::= 'x'y'` while this grammar rejects both. Verified through `descentParser`.

2. **Eight grammar branches are unpinned** — deleting or narrowing each leaves the proof fully green. The two that matter most are `module.f.mjs:132` (`character1.quote`) and `:139` (`character2.quote`): those are exactly the branches the deviation comment at `:19-25` and `README.md:62-69` exist to justify, and nothing asserts them. The rest: `eol.crlf` (`:64`), `letter.upper` (`:98`), `letter.lower`'s `z` bound (`:99`), `ruleChar.digit` (`:159`), `ruleChar.dash` (`:160`), `text1`'s empty case (`:144`). One leaf each for `<a> ::= "it's"`, `<a> ::= '"'`, a CRLF line end, `<Abc>`, `<a1>`, `<opt-whitespace>` and `<a> ::= ""` closes all eight. Coverage can't catch these — the module is declarative data, so it reads 100% whether a branch is ever matched or not.

3. **Title has no `<topic>: ` prefix** (`CONTRIBUTING.md:178`). The last 32 consecutive merged non-release PRs (#1614–#1651) are 32/32 on this. Something like `bnf/self: add the BNF metasyntax described by itself` fits.

Adjacent to the bot's changelog point: beyond the missing `changelog/unreleased/1627.md`, the description also has no `Changelog:` section, which `AGENTS.md:88` requires independently of the file.

PR #1628, opened 2026-08-17
djs: computed property keys, and __proto__ only in that form

sergey-shandar:
Fixes `todo/proto-property-key.md` and `fjs/djs/todo/proto-key-handling.md`, both deleted here.

JavaScript gives the three spellings of a `__proto__` key two meanings: `{__proto__: v}` and `{"__proto__": v}` replace the prototype and add no property, while `{["__proto__"]: v}` is an ordinary property. FunctionalScript plans no prototype chains at run time, so the first two have no meaning to give — they are now compilation errors — and the module emitter writes the computed form, the only spelling whose evaluation reproduces the value. It previously wrote `"__proto__":`, so it did not round-trip its own output.

The computed key is not a `__proto__` special case: `{ ["a"]: 1 }` is available for any constant string key in any DJS file, as a third spelling next to the identifier and the string literal. The general form — a key computed from an expression — waits on expressions and stays in `spec/todo`.

The two emitters diverge on purpose. Key spelling is a parameter of `buildSerialize` rather than a property of the shared JSON helper: `stringify` (module output) writes `["__proto__"]:`, `stringifyAsTree` (JSON output) keeps `"__proto__":`, which `JSON.parse` already round-trips and which is the only form a JSON parser accepts. Tests pin both, so a later refactor cannot unify the paths and reintroduce the bug.

Proofs, at 100% line/branch coverage: computed-key acceptance and its five error shapes; all four `__proto__` rejections; both emitter spellings including the hoisted-`const` case; and a serialize → evaluate → structurally-equal round trip over an 18-value corpus (every leaf type, both containers, shared values, `__proto__` as a primitive and as an object value), with an end-to-end `compile` group that also checks the semantics behind the text — an own property, prototype untouched.

**One conformance test from the todo could not land as written.** It expected `fjs compile proto.json out.js` to succeed on a JSON document with a `__proto__` key, on the premise that JSON input goes through a different reader. `transpile` in fact reads every input with the DJS parser; only the output branches on `.json`. So that document is rejected as input — exactly the subset exception the todo documents, now pinned by a test and stated in `spec/2480`. Whether `fjs compile` should pick the reader by extension the way it picks the writer is filed as `fjs/djs/todo/json-input-reader.md`; it would also make today's bigint/`undefined`/comments-in-`.json` inputs an error, so it is a separate decision.

Changelog:
- **BREAKING CHANGES:** `djs`: `{ __proto__: v }` and `{ "__proto__": v }`
  are compilation errors; the key is written `{ ["__proto__"]: v }`, which
  is also what the module emitter now produces
- `djs`: computed property keys `{ ["a"]: 1 }` with a constant string key

https://claude.ai/code/session_0152yFio8m9UwqeNwSENqQoL


o2alexanderfedin (approved):
Approving. This closes the `__proto__` line of work correctly, and I checked the behavior by running it rather than reading it.

Real CLI, real files: `{["__proto__"]:{"a":42},"b":[1,2]}` → `.f.js` is byte-identical and recompiles as the identity; → `.json` is `{"__proto__":{"a":42},"b":[1,2]}` with no bracket artifact. Importing the emitted `.f.js` in node gives `Object.keys` = `["__proto__","b"]` with `getPrototypeOf === Object.prototype` and a plain data descriptor — the own key, not the prototype assignment. The shared-reference case keeps identity too. A `.json` input carrying `"__proto__"` exits 1 with `__proto__ requires the computed key form` and writes nothing, which is the documented subset exception in `spec/2480` and `todo/json-input-reader.md`.

The per-emitter split is real and pinned. Unifying the two emitters (`serializeWithoutConst = buildSerialize(jsKeySerialize)`) — the exact regression the design warned about — fails 3 proofs including `protoKey.jsonOutput`. Reverting the JS emitter fix fails 7. Dropping the parser rejection fails 7. All three mutants died against a 367/0 baseline over the `fjs/djs/**` proof trees.

The computed-key grammar boundary holds: across 32 probes, only a single string token is accepted inside `[...]`. `["constr"+"uctor"]`, `[name]`, `[c]` for a declared `const`, `[1]`, `[["a"]]`, `["a","b"]` and the rest are all `unexpected token`. The `"+"` hole is closed. Escapes are judged on the decoded value — `{"__proto__":1}` rejected, the bracketed form accepted.

One note, wording only. `fjs/djs/proof.f.mjs:106` — `roundTrip`'s comment ("serialize → evaluate → structurally the same"), echoed in the PR body, describes a loop that re-reads the output with `transpile`, i.e. the DJS parser. That is FS-semantics evaluation, but it's the one loop that can't catch a parser/serializer pair agreeing on a spelling node reads differently — the exact bug class this PR fixes. I closed that gap by hand (the node `import()` results above) and every case is correct; the behavior the comment claims is real, only its strength is overstated.

Verified otherwise: 12 key classes (`""`, `a-b`, `"3"`, `é😀`, `constructor`, `a\nb`, `a"b`, `class`, `__proto`/`___proto__`) all evaluate equal under node and parse equal from JSON; only `__proto__` gets brackets and no other key needs them. No dangling reference to either deleted todo — `grep -rn` finds only links to `spec/2480`. Every task in both deleted files is delivered except the `proto.json` conformance fixture, which is carried into `todo/json-input-reader.md` and pinned by `protoKey.jsonDocumentRejected` rather than dropped. `spec/2470`/`2480` numbering is free, sorts right in the index, carries the `Depends on` footer, all links resolve, and every behavioral claim in both was checked by execution. Title conforms, changelog entry matches the body byte-for-byte and correctly declares the breaking change. Green: `tsc` 0, `fjs t` 2942 pass / 0 fail, `cov` 100/100/100, `prepack` 0, `ci-update` clean.

o2alexanderfedin (commented):
Re-reviewed after the root-vs-import change. The code is correct and well-pinned — six of seven mutants died, including a true revert of the import change (`parseModule` using `parserFor(path)`) and seeding `transpile` with `parseFromTokens`. The remaining items are in the PR body and one spec file the delta missed.

Findings below are against `4c60524c`; `5c045f45` only adds `todo/remove-json-from-parser.md`, whose eight relative links all resolve and which closes my earlier concern about a dropped open question — "`.json` inputs stop accepting DJS extensions" is tracked there now.

**The body's `Changelog:` section still doesn't match `changelog/unreleased/1628.md`, and the gap grew.** `CONTRIBUTING.md:197-207` requires them identical. Body item 1 says "are compilation errors;" where the file says "are compilation errors **in a FunctionalScript module**;" — and that qualifier is exactly what this delta is about, since in a `.json` root they aren't errors. The body is also missing the file's third item entirely (the `fjs compile` input-language one). Body has 2 items, file has 3. Separately that third entry is ~430 chars / 6 wrapped lines against `changelog/README.md`'s "about three wrapped lines, ~250 characters"; items 1 and 2 are 225 and 73.

**The body paragraph beginning "One conformance test from the todo could not land as written" is now false in all three of its claims.** It says `fjs compile proto.json out.js` is rejected — it exits 0 and emits `export default {["__proto__"]:5}`. It says "`transpile` in fact reads every input with the DJS parser; only the output branches on `.json`" — `transpile` now seeds `parserFor(path)` at `transpiler/module.f.mjs:128`. And it files the question as `fjs/djs/todo/json-input-reader.md`, which isn't in the tree (grep for it returns zero hits repo-wide). The branch answered the question that paragraph poses; it should go.

**`spec/1000-json.md:18` is now the only file in `spec/` still stating the old rule.** "Such a document is still compiled — `fjs compile` reads a `.json` file as JSON — but it is not also a module." This delta narrowed that everywhere else — `spec/README.md:150-151` "This applies to the file named on the command line", `spec/2480:40-41` "The extension of each file **named on the command line** picks its language" — and 1000-json kept the unqualified form, which the code contradicts directly:

```
$ node fjs/module.mjs c main-proto.f.js out.json   # imports ./proto.json
proto.json:1:2 - error: __proto__ requires the computed key form
exit=1
```

**Minor: nothing pins that a `.json` root's imports are read as FunctionalScript.** Changing `transpileWithImports`'s hard-wired `foldNextModuleOp(parseFromTokens)` to `foldNextModuleOp(parserFor(path))` leaves the suite green at 2950/0. Under that mutant a `.json` root propagates the JSON reader down its whole import tree, contradicting the docstring's "The root is the only file this applies to." The real code is right — a chain `a.json` → `b.json` → `c.json` with `{"__proto__":1}` in the last correctly fails — but both new proofs use a `.f.js` root, so none covers it.

Verified otherwise: `tsc` 0, `fjs t` 2950 pass / 0 fail, `cov` 100/100/100, `prepack` and `ci-update` clean. Root/import asymmetry confirmed both directions, and 2480's `proto.json → a.f.js → out.json` loop is byte-identical under `cmp`. `spec/todo/2140`'s Node table reproduces exactly — all five rows, on v23.11.0 rather than the 22 it cites — as does its "the parser has it exactly backwards" claim: fjs accepts the bare `.json` import and rejects the `with { type: "json" }` spelling with `unexpected token`. The Stage 3 claim for import-text checks out. `spec/todo/README.md`'s new §2.1 fills a real gap (the file jumped `## 2. DJS` → `### 2.2.`) and matches `spec/README.md:60`'s own §2.1 title. Mutants that died: the two above plus `pushStringKey` dropping the language guard, `pushStringKey` → `false`, `jsKeySerialize = stringSerialize`, and the emitter unification.

PR #1629, opened 2026-08-17
emergent_testing: migrate to IoEffect composition

sergey-shandar:
Stage 4 of the io-effect migration, the first of the queued consumer modules (`fjs/effects/todo/io-effect-consumer-migration.md`). Eight `unwrapStep` sites become one: the reporter's writes, `sandbox`, `awaitIfPromise` and the `all` parallelism propagate their failures instead of panicking.

`Reporter`'s members carry their `Result` in the type, which is why this module went first. `result` and `summary` returned `Effect<O, void>`, and TypeScript accepts an effect of any value type where a `void` one is expected — so a reporter whose writes were fallible type-checked while its failures went nowhere. The channel is pinned to the standard node one (`NotImplemented | IoError`) rather than made a type parameter: a reporter is free in which operations it performs, but it fails the way node IO fails, and a parameter would be threaded through every caller unchanged.

`allOk` is new, beside `all` in `fjs/effects/node/module.f.mjs`. `all`'s envelope is the runner's (`OpResult`), so handing it `IoEffect`s nests one `Result` inside another and returns `readonly Result<T, E>[]` — the value-discarding hazard, one level in where it is harder to see. `allOk` collapses that to the first error. Every effect still runs: the short-circuit is in the result, not the execution. `fjs/dev`'s two `all` sites want the same combinator, so it is already where that migration needs it.

Two defects fell out of the sweep. `registerModuleMap` discarded `all`'s result through a `() => undefined` continuation, and `register` answered exit `0` whether or not registration succeeded — it now ends with `exitStep`.

One panic stays, with the reason in a comment: `Test` hands its callback to an external framework (node `--test`, Bun, Deno) whose contract is a raw `Effect<…, void>`, so there is no channel to answer through, and a throw is what those frameworks already report as a failed test.

`testAll` needed a tail that is not `exitStep`. That one answers `0` for every success, and this chain's success value *is* the exit code (`1` when a test failed), so conflating them would report a failing suite as a passing run. The local `exitCodeStep` keeps the computed code and reports only a channel failure; `register`, whose success carries nothing, uses `exitStep` unchanged.

Validation: `npx tsc` clean, `fjs t` 2903 pass / 0 fail, `npm run cov` 100/100/100. The suite count is 2902 on the base and 2903 here — checked explicitly against the base, since this module is the test runner and a silent drop in discovered tests would otherwise read as a pass. The coverage gate first failed at 99.96% on exactly the new error branches, so `reporterWriteFailure` runs `main` against a runner whose `write` always fails and asserts exit `1`.

Changelog:
- **BREAKING CHANGES:** `emergent_testing`: `Reporter`'s members and the
  `registerModule` / `runModuleMap` / `testAll` chain are fallible `IoEffect`s.
  A failed reporter write now ends the run with exit `1`, as does a failed
  `register`, which used to answer `0` either way.
- `effects/node`: new `allOk` — `all` in the `ok` channel, answering with the
  first error instead of a list of results for the caller to collapse.

### DIFF
--- a/changelog/unreleased/1629.md
+++ b/changelog/unreleased/1629.md
@@ -1,0 +1,6 @@
+- **BREAKING CHANGES:** `emergent_testing`: `Reporter`'s members and the
+  `registerModule` / `runModuleMap` / `testAll` chain are fallible `IoEffect`s.
+  A failed reporter write now ends the run with exit `1`, as does a failed
+  `register`, which used to answer `0` either way.
+- `effects/node`: new `allOk` — `all` in the `ok` channel, answering with the
+  first error instead of a list of results for the caller to collapse.
--- a/fjs/effects/node/module.f.mjs
+++ b/fjs/effects/node/module.f.mjs
@@ -11,6 +11,7 @@
  * @module
  *
  * @import { Vec } from '../../types/bit_vec/types.ts'
+ * @import { Result } from '../../types/result/types.ts'
  * @import { Effect, Func, Operation } from '../types.ts'
  * @import { List } from '../list/types.ts'
  * @import { All, Access, Await, Console, CreateExclusive, CreateServer, Dirent, Engine, Env, Exec, ExecResult, Fetch, FileStat, Forever, Fs, Headers, Http, IncomingMessage, Import, IoError, IoErrorInfo, IoResult, Listen, MakeDirectoryOptions, Mkdir, Module, Now, NodeOp, NodeProgramOptions, OpResult, RandomInt, Read, ReadBytes, ReadConsoles, ReadFile, Readdir, ReaddirOptions, RequestListener, Rename, Rm, Sandbox, SandboxResult, Server, ServerResponse, Stat, Test, TestContext, TestFn, Write, WriteBytes, WriteConsoles, WriteFile, _UtfList, _WriteLoop, } from './types.ts'
@@ -22,7 +23,7 @@ import { toCodePointList } from '../../text/utf8/module.f.mjs'
 import { codePointListToString } from '../../text/utf16/module.f.mjs'
 import { reverse } from '../../types/list/module.f.mjs'
 import { length } from '../../types/bit_vec/module.f.mjs'
-import { error as resultError } from '../../types/result/module.f.mjs'
+import { error as resultError, ok as resultOk, unwrap } from '../../types/result/module.f.mjs'
 import { do_, mapStep, pure, step } from '../module.f.mjs'
 import { mapStep as ioMapStep, pureError, pureOk, step as ioStep } from '../io/module.f.mjs'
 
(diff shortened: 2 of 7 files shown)


o2alexanderfedin (approved):
Approving — the migration is correct. I ran a differential harness (base `4a40f320` vs head, 21 scenarios through `main`/`testAll`/`register`/`defaultTest`): 14 identical byte-for-byte on stdout, stderr and exit code, and every one of the 7 differences is a former panic now becoming a propagated failure with exit `1`. No swallowed errors, no reordering, no lost short-circuiting.

Three notes, none blocking.

**The `register` exit-code change has no proof.** `module.f.mjs:427` replaces `mapStep(…, () => 0)` with `exitStep(registered)`, and the changelog calls this out as BREAKING. Reverting it to the old always-zero form leaves the suite green — 31/0. The only proof touching it is `registerEmptyModuleMap` (`proof.f.mjs:445`), which asserts `0` on the *success* path, and the 100% gate is met either way since the line runs regardless of its value. My harness confirms the new code is right (failing `test` op → exit `1`); nothing holds it there. A `register` proof whose `test` op answers `error(['notImplemented', 'test'])`, asserting exit `1`, would close it.

**`allOk` has no proof in its own module, and its first-error rule is unproven.** `fjs/effects/node/module.f.mjs:99,128`. The JSDoc promises "keeping the **first** error in list order"; mutating `okList` to keep the *last* one survives, 31/0. Its only coverage is indirect via `reporterWriteFailure` — which does kill a mutant that drops propagation entirely. Its siblings `exitStep` (both branches), `errorMessage` and `both` all have direct proofs in `fjs/effects/node/proof.f.mjs`; the new one is the odd one out, and the property the doc singles out is the one nothing checks.

**`testAll`'s operation set widened without a changelog mention.** Declaration emit goes `Program<O | All | LoadModuleOperations>` → `… | Write`, so a consumer supplying its own runner must now handle `write` (for `errorExit`). The entry also calls `testAll` a fallible `IoEffect`; it's the tail that leaves the layer and still answers `number`.

Verified otherwise: `tsc` 0, `prepack` clean, `fjs t` 2903 pass / 0 fail against 2902 on base — delta exactly +1, the new `reporterWriteFailure`, no silent test drop. `cov` 100/100/100. `ci-update` leaves the tree clean. Declaration-emit diff vs base touches only `registerModule`, `runModuleMap`, `testAll`, `defaultTest` and the new `allOk` — no unintended type surface; `types.ts` adds no `any` and no assertions. `all` is untouched and `allOk` purely additive, with `fjs/dev:74,120` unaffected. Mutants that died: `exitCodeStep` swallowing the channel error, `exitCodeStep` behaving like `exitStep` (4 proofs — the split is genuinely held), `okList` never propagating. Both todo files match the code, including the surviving `unwrapStep` at `module.f.mjs:130` being documented as the external-framework boundary rather than left as an unexplained holdout.

PR #1630, opened 2026-08-17
dev: migrate module discovery to IoEffect composition

sergey-shandar:
Stage 4 of the io-effect migration, the second of the queued consumer modules, following #1629. Four bare `unwrap`s leave `fjs/dev`: `allFiles`, `loadFile` and `loadModuleMap` propagate their failures instead of panicking, so a directory that cannot be listed or a module that will not import is the caller's decision rather than a stack trace.

Propagating is the right answer for discovery specifically, and the JSDoc says so: it reads the whole source tree, and a map silently missing the file it was asked about is worse than no map at all.

Both halves of the plan's prediction for this module held. It needed **no new vocabulary** — `allOk` landed with #1629 precisely because `fjs/dev`'s two `all` sites wanted the same combinator, which is what the ordering was for. And `loadModuleMap` becoming fallible turned `testAll` and `register`'s raw `step` into the Io one, so the two modules met at the same tail exactly where the plan said they would; that is the whole four-line `fjs/emergent_testing` diff here.

This is a semantic change to a public function, not only a typing one: callers that relied on `loadModuleMap` throwing now receive an `error` result they have to handle. `fjs/emergent_testing` is its only consumer in the repository, and it handles it — `fjs t` reports the reason on `stderr` and exits `1` where it used to unwind.

Validation: `npx tsc` clean, `fjs t` 2952 pass / 0 fail, `npm run cov` 100/100/100, `fjs ci` leaves the tree clean. Coverage stayed at 100% unaided here, since the migration adds no branches — so unlike #1629 the gate did not force a proof, and `loadModuleMapReportsUnreadableRoot` was added anyway to pin the behavior change it does make.

Rebased onto `main` after #1629 squash-merged, so this now contains only the `fjs/dev` commit.

Changelog:
- **BREAKING CHANGES:** `dev`: `loadModuleMap` is a fallible `IoEffect`. A
  directory it cannot list or a module that will not import is now an `error`
  the caller handles, where discovery used to panic.

### DIFF
--- a/changelog/unreleased/1630.md
+++ b/changelog/unreleased/1630.md
@@ -1,0 +1,3 @@
+- **BREAKING CHANGES:** `dev`: `loadModuleMap` is a fallible `IoEffect`. A
+  directory it cannot list or a module that will not import is now an `error`
+  the caller handles, where discovery used to panic.
--- a/fjs/dev/module.f.mjs
+++ b/fjs/dev/module.f.mjs
@@ -178,6 +178,16 @@ export const proof = {
         const [, result] = virtual({ ...emptyState, root: {} })(loadModuleMap({ INIT_CWD: 'missing' }))
         assert(result[0] === 'error', result)
     },
+    loadModuleMapReportsUnimportableModule: () => {
+        // The other half of the same breaking change, and the half `readdir`
+        // does not cover: discovery lists the file, then the import of it
+        // fails. `'a.f.ts': []` is content rather than a module, so the
+        // virtual runner's `import` refuses it. Without this, a `loadFile`
+        // that turned every import failure into a success would still pass
+        // every other proof here.
+        const [, result] = virtual({ ...emptyState, root: { 'a.f.ts': [] } })(loadModuleMap({}))
+        assert(result[0] === 'error', result)
+    },
     loadModuleMapStripsInitCwdPrefix: () => {
         // With `INIT_CWD` set (the normal `npm run` case), discovery starts
         // from that subdirectory and each key is relativized against it, so
--- a/fjs/dev/module.f.mjs
+++ b/fjs/dev/module.f.mjs
@@ -3,16 +3,16 @@
  *
  * @module
  *
- * @import { Access, All, Env, Import, Readdir } from '../effects/node/types.ts'
- * @import { Effect } from '../effects/types.ts'
+ * @import { Access, All, Env, Import, IoError, Readdir } from '../effects/node/types.ts'
+ * @import { IoEffect, NotImplemented } from '../effects/io/types.ts'
  * @import { Dir } from '../effects/node/virtual/types.ts'
  * @import { Module, ModuleMap, LoadModuleOperations } from './types.ts'
  */
 
-import { all, import_, readdir } from '../effects/node/module.f.mjs'
+import { allOk, import_, readdir } from '../effects/node/module.f.mjs'
 import { cmp as strCmp } from '../types/string/module.f.mjs'
 import { unwrap } from '../types/result/module.f.mjs'
-import { pure, step } from '../effects/module.f.mjs'
+import { mapStep, pureOk, step } from '../effects/io/module.f.mjs'
 import { join, relativize, toPosix } from '../path/module.f.mjs'
 import { assert, assertEq } from '../asserts/module.f.mjs'
 import { emptyState, virtual } from '../effects/node/virtual/module.f.mjs'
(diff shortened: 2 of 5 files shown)


o2alexanderfedin (commented):
The migration is correct — a 13-scenario differential harness against `origin/main` shows every success path identical and every failure path changing panic→`error` with a byte-identical payload. But one half of the declared breaking change has no proof behind it.

**A mutant that turns every import failure into a success passes the whole gate.** `fjs/dev/module.f.mjs:81-83`. The new proof `loadModuleMapReportsUnreadableRoot` pins the `readdir` half; nothing pins `loadFile`. Rewriting it so an import failure becomes `ok` — no new function, no new branch, so the coverage numbers don't move:

```js
const loadFile = f =>
    mapStep(catchStep(import_(f), pureOk), m => [/** @type {const} */ ([f, /** @type {Module} */ (m)])])
```

```
npx tsc                -> 0
node fjs/module.mjs t  -> pass: 2952, fail: 0
npm run cov            -> 100.00 | 100.00 | 100.00, exit 0
```

Against a real tree holding one `.f.mjs` that won't parse, that mutant is the difference between

```
Unexpected identifier 'is'                    exit=1     # pristine
Number of tests: pass: 0, fail: 0, total: 0   exit=0     # mutant
```

— a silently green run over a broken source tree, which is the failure the JSDoc at `:101-105` argues against ("a map missing the file it was asked about is worse than no map at all"). The `readdir` side *is* caught: swallowing there fails the new proof (2951/1). The asymmetry is only in which half got one.

A sibling of the existing proof closes it. I confirmed this fixture gives `error(["ioError",{"message":"'a.f.ts' is not a JsModule"}])` on pristine head and `ok` under the mutant:

```js
loadModuleMapReportsUnimportableModule: () => {
    const [, result] = virtual({ ...emptyState, root: { 'a.f.ts': [] } })(loadModuleMap({}))
    assert(result[0] === 'error', result)
},
```

Verified otherwise: `tsc` 0 both sides, `prepack` clean, `fjs t` 2952 pass / 0 fail against 2951 on base — the +1 is the new proof, nothing dropped. `cov` 100/100/100, `ci-update` clean. Declaration emit differs in exactly one file: `loadModuleMap`'s return becomes `IoEffect<…, NotImplemented | IoError>` with `LoadModuleOperations` **unchanged**, so the operation set didn't widen and the #1629 finding doesn't recur; `emergent_testing`'s `.d.mts` has no diff at all, confirming its 4-line change is a pure call-site retype. The changelog declares this completely and its `Changelog:` section is byte-identical to the entry file. End-to-end `fjs t` over `badparse`/`noperm`/`missing` exits 1 on both sides — the exit code is unchanged; head just replaces an uncaught-exception stack dump with a one-line message on stderr. `oneGoodOneBad` running 2 imports on head vs 1 on base is `allOk`'s documented "every effect still runs", not lost short-circuiting. Other mutants: swallowing the `readdir` error, and reverting either `testAll`'s or `register`'s io `step` to `rawStep`, all died — the last two via `cov` rather than `fjs t`. `fjs/dev` is genuinely fully migrated (zero `unwrapStep`, zero `throw` in production code) and `io-effect-migration.md`'s remaining list matches the grep exactly.

Unrelated to this diff: `allOk`'s "first error in list order" is still unpinned — flipping `okList` to keep the last error passes at 100/100/100, and it does change which message `fjs/dev` surfaces. `fjs/dev`'s contract doesn't depend on which failure wins, so this PR is unaffected; it stays the `fjs/effects/node` gap from #1629.

sergey-shandar:
Addressed in `2c723a7` — the mutation finding was right, and the proof you specified is in.

`loadModuleMapReportsUnimportableModule` lands next to the `readdir` one in `fjs/dev/module.f.mjs`, using your fixture (`'a.f.ts': []` is content rather than a module, so the virtual runner's `import` refuses it).

I checked it kills the mutant rather than merely passing: replacing `loadFile` with `mapStep(catchStep(import_(f), () => pureOk({})), …)` — every import failure turned into a success — fails this proof and only this proof (2952/1). Before it, that mutant passed the whole gate, as you said.

`npx tsc` clean, `fjs t` 2953/0, `npm run cov` 100/100/100.

---

o2alexanderfedin (approved):
Approving — the finding is closed, and the new proof pins the path its comment claims rather than a path that merely looks like it.

The verbatim `loadFile` mutant now dies on an **assertion**, not on the coverage gate:

```
node ./fjs/module.mjs t -> exit 1
  proof.loadModuleMapReportsUnimportableModule(): error
  Number of tests: pass: 2952, fail: 1, total: 2953
```

That distinction was the whole point of the finding — the mutant type-checks and adds no branch, so `cov` stayed at 100/100/100 and could never see it.

I checked that the fixture reaches `loadFile` rather than being filtered earlier, since a proof pinning the wrong path would leave the finding open. `virtual({...emptyState, root: { 'a.f.ts': [] }})(loadModuleMap({}))` returns `["error",["ioError",{"message":"'a.f.ts' is not a JsModule"}]]`, and that string is emitted at exactly one place in the repo — `fjs/effects/node/virtual/module.f.mjs:131`, inside the `import` handler. Control: the same tree with a real module (`'a.f.ts': () => ({})`) gives `["ok",{"./a.f.ts":{}}]`, and `shouldLoad('./a.f.ts')` is `true` while a non-matching `'a.txt': []` gives `["ok",{}]`. So discovery passes the file through and the import is what fails.

Real-tree probe, one `.f.mjs` containing `const x is 1`:

| | stderr | exit |
|---|---|---|
| head | `Missing initializer in const declaration` | 1 |
| mutant | `Number of tests: pass: 0, fail: 0, total: 0` | 0 |

— the silently-green run reproduced, and now stopped in the suite before it could ship.

Three further mutants survive, none of which I'd raise as a defect. One keeps erroring but loses the file identity (message degrades to `load failed`); the sibling `loadModuleMapReportsUnreadableRoot` asserts exactly as loosely, so that's the module's existing convention, not something this commit introduced. The other two swallow selectively — keyed on the file's index, or on `files.length > 1` — and each needs a deliberately introduced index/count branch, unlike the earlier mutant which needed none. Worth knowing one fixture change kills both: a two-file root whose *second* entry is content. The current code is correct either way — `{a.f.ts: fn, b.f.ts: []}` errors naming `b.f.ts` and `{a.f.ts: [], b.f.ts: fn}` names `a.f.ts`, so position doesn't matter.

Verified: `tsc` 0, `fjs t` 2953 pass / 0 fail against 2952 at the previous head — delta exactly +1. `cov` 100/100/100, `prepack` and `ci-update` clean. The delta is insertions only in `fjs/dev/module.f.mjs`, so no changelog change is needed, and the body's `Changelog:` section is still byte-identical to the entry file.

One nit on the body, not the code: its Validation paragraph still reads "`fjs t` 2952 pass / 0 fail" and "`loadModuleMapReportsUnreadableRoot` was added anyway" — it's 2953 and two proofs now.

PR #1631, opened 2026-08-17
djs: the parser reads modules, and JSON has its own reader

sergey-shandar:
Fixes `fjs/djs/todo/parser-statement-whitelist.md`, deleted here.

A statement begins with `import`, `const`, or `export` and never with a value. `parseInitialOp` used to fall through and read a leading value as the module's body, so `42`, `[1,2]`, and `{"a":1}` all parsed as modules — but a JSON document is not a FunctionalScript module: as JavaScript `{"a":1}` does not parse at all (`{` opens a block, and `"a":1` is no labeled statement), and `[1,2]` is an expression statement exporting nothing, so an import of one fails with *"does not provide an export named 'default'"*. The parser was giving those texts a value no engine gives them. The three keywords are a whitelist that grows with the language, not a closed set.

The statements are ordered too: every `import` before every `const`, and one required `export default` last. The order needs no new parser state — an `import` is an error once the module has a constant, which the parser's own state already records.

**JSON keeps its meaning, through its own reader.** `fjs compile x.json` now goes to `fjs/media/json`, so a JSON document still compiles to a module denoting the same value, with `__proto__` spelled the JavaScript way on the way out. That removes the JSON reading from the parser: `parseJsonFromTokens`, the `_Language` field, and the `pushStringKey`/`pushIdKey` split are gone, leaving one `pushPlainKey` that refuses `__proto__` in both plain spellings. A JSON document also needs no AST and no evaluation — it imports nothing and names nothing — so the JSON path is a read and a parse, nothing more.

Two consequences worth stating plainly, both intended:

- a `.json` file is JSON and nothing more — no `bigint`, `undefined`, comments, identifier or computed keys;
- a JSON file cannot be imported at all, since it is not a module. `with { type: "json" }` is what will make it importable (`spec/todo/2140-import-attributes.md`), and that clause is now the whole feature rather than a spelling for one.

Errors from the JSON reader carry no position, so `compile` names the file being compiled instead of printing `undefined:undefined:undefined` — the visible half of `fjs/djs/todo/parse-error-location-format`, whose remaining half (naming the file that actually failed, and giving `fjs/media/json` real positions) stays filed.

The spec follows the code: `spec/README` §1 and `1000-json` no longer claim a JSON document is a valid module — what the two languages share is values, not texts — §2.1 states the whitelist and the order, and `2480`'s "JSON is a subset except here" section becomes "the one key the two languages read differently".

Checks: `npx tsc` clean, full suite passing, `npm run cov` at 100% lines/branches/functions.

Changelog:
- **BREAKING CHANGES:** `djs`: the parser reads FunctionalScript modules only.
  A statement begins with `import`, `const`, or `export` and never with a
  value, so a JSON document no longer parses as a module; imports come before
  constants, and `export default` is required and last
- **BREAKING CHANGES:** `djs`: `fjs compile` reads a `.json` input with
  `fjs/media/json` instead of the module parser, so a `.json` file is JSON and
  nothing more — no `bigint`, `undefined`, comments, or identifier keys — and
  it cannot be imported until the language has `with { type: "json" }`. The
  `parser` export `parseJsonFromTokens` is gone with it
- `djs`: an error with no token to point at names the file being compiled
  instead of printing `undefined:undefined:undefined`

https://claude.ai/code/session_0152yFio8m9UwqeNwSENqQoL


o2alexanderfedin (approved):
Approving. Everything the todo promised landed and is proved — six mutants applied, six died, including restoring the bare-value fall-through in `parseInitialOp` (7 fail), `pushPlainKey` accepting `__proto__` (6 fail), routing a `.json` root back through `transpileModule` (3 fail), and the emitter unification that this line of work exists to prevent (4 fail).

No `__proto__` regression across the 122-line parser rewrite. `.json` → `.f.js` emits `export default {["__proto__"]:{"evil":1},"ok":2}`; `import()`ed in real node it gives `keys ["__proto__","ok"]` with the prototype still `Object.prototype` and a plain data descriptor — semantically identical to `JSON.parse` of the `.json` output. Both round-trips byte-identical, and the computed-key grammar is unmoved (`["constr"+"uctor"]`, `[name]`, `[1]`, `["a","b"]` all rejected).

One doc contradiction, in a file this PR doesn't touch.

**`README.md:9` still asserts the claim this PR set out to delete.** Line 8 establishes the subset-of-texts sense — "[JSON] and [JSON5] as subsets of JavaScript" — so line 9's "JSON is also a subset of FunctionalScript" says exactly what `spec/README.md:50-52` now denies: "A JSON document is **not** a FunctionalScript module, though. A module is a sequence of statements, and no statement begins with a value." The root README links to `./spec/README.md` five lines down, so a reader meets both. The deleted todo's task list named `spec/README` §1, `spec/1000-json` and `spec/2480` — all three are rewritten correctly and consistently; the front page was just outside the list. The value-level claim ("every JSON value is a FunctionalScript value") would be true and is presumably what's meant.

Judgment call, not a factual break: `spec/2480:75-77`'s "every other JSON document denotes the same value in both" reads literally as every JSON document having a meaning as a module, which is now false for all of them. Context makes the intended reading clear and it agrees with `spec/README.md:55-56`.

Verified otherwise: consequence 3 landed — all nine DJS extensions in a `.json` root now rejected (`bigint`, both comment forms, `undefined`, `const`, `export default`, identifier key, computed key, `import`). Error positions did not land for `fjs/media/json`, but nothing claims they did: `.json` errors print `./err.json - error: unexpected token` with no line/col, `parse-error-location-format.md` stays `open` and retitled to the surviving problem, and its quoted `errorLocation` snippet is byte-identical to the code. Importing a `.json` gives `plain.json:1:1 - error: unexpected token`, pinned by `jsonImportRejected` at `proof.f.mjs:172`; bare-value and empty `.f.js` inputs give intelligible messages too. `fjs/media/json` edge cases behave (`-0`→`0`, `1e400`→`null`, duplicate keys last-wins, lone surrogate→replacement; correctly rejects trailing comma, leading zero, `+1`, `NaN`, `Infinity`, single quotes, raw tab, trailing junk) — the 2000-level nesting stack overflow reproduces identically on `origin/main`, so it's pre-existing. Declaration-emit diff removes only `parseJsonFromTokens` (declared) plus internals; `parseFromTokens` and `transpile` unchanged. `grep` finds no dangling references to `parser-statement-whitelist`, `_Language`, `pushStringKey`, `pushIdKey` or `parserFor`, and `spec/todo/2140` correctly describes the interim state. Body's `Changelog:` section is byte-identical to `changelog/unreleased/1631.md` — the drift from the last two PRs is gone. Green: `tsc` 0, `fjs t` 2955 pass / 0 fail, `cov` 100/100/100, `prepack` and `ci-update` clean.

sergey-shandar:
Fixed in a25ef79 — `README.md:9` said "JSON is also a subset of FunctionalScript", which this branch makes false. It now reads:

> Every JSON value is a FunctionalScript value; a JSON document is data rather than a module, and `fjs compile` reads one with its own reader ([JSON](./spec/1000-json.md)).

I swept the rest of the tree for the same claim; the only other match is `spec/README.md:62`, which says DJS is the data *subset of FunctionalScript* — a different and still-true statement.

---

sergey-shandar:
Correction to my previous comment — the fix is one word per line, not a rewrite. `f96fba0` (amended, replacing a25ef79):

```diff
-- [JSON](https://en.wikipedia.org/wiki/JSON) and [JSON5](https://json5.org/) as subsets of JavaScript.
-  JSON is also a subset of FunctionalScript.
+- [JSON](https://en.wikipedia.org/wiki/JSON) and [JSON5](https://json5.org/) as subsets of JavaScript values.
+  JSON is also a subset of FunctionalScript values.
```

The subset claim was right all along; it is about values, not documents, which is exactly the distinction this PR draws.

---

o2alexanderfedin (approved):
Approving. The README finding is resolved, and `f96fba08` fixes it better than my suggestion did — moving "values" onto line 8 makes the whole bullet a claim about values, so there's no longer a text-level sense for line 9 to inherit. That was the actual defect; my version only corrected the second half.

Verified at `a25ef793` (the code is untouched between the two — the delta is `README.md` only): every JSON value kind compiles as `.json` and is rejected as `.f.js`, not just objects —

| text | as `.f.js` | as `.json` |
|---|---|---|
| `{"a":1}` | exit 1, `unexpected token` | exit 0 → `{"a":1}` |
| `[1,2]` | exit 1, `unexpected token` | exit 0 → `[1,2]` |
| `42` / `"hello"` / `true` / `null` / `-42.5` | exit 1, `unexpected token` | exit 0, round-trips |

— which is exactly why the value-level claim is the right one. The separate reader is structural, not just behavioural: `transpiler/module.f.mjs:22` imports `parse as jsonParse` from `media/json`, used in `transpileJson` at `:121`, disjoint from the module path's `tokenize` + `parseFromTokens`. And `{"__proto__":{"x":1},"y":2}` emits `export default {["__proto__"]:{"x":1},"y":2}` as a module, `{"__proto__":…}` as JSON, with the module recompiling byte-identically to the original — the interesting case still holds.

No fourth place states the old form: `grep -rn "subset of FunctionalScript"` returns only `spec/README.md:62`'s "DJS is the data subset of FunctionalScript", a different and correct claim.

Dropping the `spec/1000-json.md` link is fine — that file independently says the same thing, so nothing is orphaned.

The `spec/2480:76-77` wording I mentioned is untouched and I'd let it go rather than spend a round trip: the same file already says "a JSON document is not a module" at `:55` and "The disagreement is about a *text*, not a value" in the next paragraph, so no reader is misled. Worth a tweak whenever that file is next open.

Green at `a25ef793`: `tsc` 0, `fjs t` 2955 pass / 0 fail, `cov` 100/100/100, `prepack` and `ci-update` clean.

PR #1633, opened 2026-08-17
protocol/mcp: migrate session state and the stdio transport to IoEffect

sergey-shandar:
Stage 4 of the io-effect migration, the last of the queued consumer modules, following #1629, #1630 and #1632.

## Three answers at a protocol edge, not one

The plan predicted "a JSON-RPC error response for a session failure". Doing it showed the rule is really about what the protocol gives you to answer *with*, and the three cases here differ:

- **A request** whose session slot cannot be read answers `-32603`. Deliberately not the `-32002` `notInitialized` the gate would otherwise fall through to: that tells a client to re-run a handshake it has already run, so it would loop forever on a fault that is the server's.
- **A notification** has no response frame at all, so the failure stays silent and the session simply stays gated — the client observes it on its next call.
- **The transport** has no frame either, but it does have a caller, so a stdin that cannot be read ends the loop and propagates. That is what lets `fjs/mcp` drop the raw-transport bridge #1632 added and compose `stdioTransport` directly.

## A discard fixed on the way through

The `initialize` transition wrote the new session state through a `() =>` continuation that ignored the result, so a session which failed to record the transition still replied with a **successful handshake** and then rejected every later call as `notInitialized`. It answers `-32603` now.

## Validation

`npx tsc` clean, `fjs t` 2969 pass / 0 fail, `npm run cov` 100/100/100, `fjs ci` leaves the tree clean.

The coverage gate caught the three new `internalError` branches at 95.71%, so `sessionStateFailure` covers all three plus the notification path. The transport's propagation adds no branch, and is proven anyway in two parts: `readFailurePropagates` for the first read, and `readFailureAfterALinePropagates` for the loop's recursive tail — a tail wrapped in `catchStep(…, () => pureOk(undefined))` passes the first and fails the second.

## Stage 4 after this

The `unwrapStep` grep returns exactly one implementation site: `fjs/emergent_testing`'s registered callback, the considered panic at the external-framework boundary. The second sweep's remaining hits are all non-IO — `throw` on invalid data (`prime_field`, `btree`, `bnf` left recursion) and runner-internal invariants — which is what FunctionalScript reserves `throw` for. One task is left in the consumer-migration todo before Stage 4 can be closed: `fjs run`'s bare `unwrap` of `import_`.

Changelog:
- **BREAKING CHANGES:** `protocol/mcp`: a session slot the runner cannot reach
  answers JSON-RPC `-32603` instead of panicking, and a failed `initialize`
  transition answers it too rather than a successful handshake it did not record.
- **BREAKING CHANGES:** `protocol/mcp/stdio`: `stdioTransport` answers an
  `IoEffect`. A stdin that cannot be read ends the loop and hands the failure
  back, where it used to panic; EOF is still a clean shutdown.


o2alexanderfedin (commented):
The migration itself is correct — I couldn't break it. Nineteen real `node fjs/module.mjs mcp` sessions, base vs head, came back byte-identical on stdout, stderr and exit (normal handshake, malformed JSON, valid-JSON-not-JSON-RPC, unknown method, unknown tool, missing and extra params, bare notification, empty line, call-before-init, double `initialize`, a 200 KB `id`, truncated line at EOF, CRLF framing, three lines in one write, ENOTDIR and EACCES stores). Framing, partial reads and the oversized-response fallback are untouched, and the path leak stays closed — `{"text":"io error: EACCES"}` via `errorSummary`, with `errorMessage` remaining only at `emergent_testing:233` and `exitStep`, both stderr. Five mutants died; a sixth was killed by `cov`.

Three things.

**No changelog entry and no `Changelog:` section, but this changes both the public API and client-visible protocol behavior.** The PR body is empty and `changelog/unreleased/1633.md` doesn't exist. `CONTRIBUTING.md:197-202` makes the section mandatory either way — "so a forgotten entry is a visible omission rather than a silent one" — and `AGENTS.md:87-92` requires the file when behavior or the public API changes. Both apply, by measurement:

```diff
-export declare const stdioTransport: <O extends Operation>(handler: Step<O>) => Effect<Read | Write | O, void>;
+export declare const stdioTransport: <O extends Operation>(handler: Step<O>) => IoEffect<Read | Write | O, void, NotImplemented | IoError>;
```

| scenario | `origin/main` | this PR |
|---|---|---|
| `initialize`, slot read fails | throws | `-32603` |
| `initialize`, slot write fails | **`ok` handshake** | `-32603` |
| `tools/list`, slot read fails | throws | `-32603` |
| `notifications/initialized`, read fails | throws | `null`, session stays gated |
| stdin read fails | throws | stderr line, exit 1 |

A server that used to die now answers `-32603`, and a handshake that used to falsely succeed now fails. #1629, #1630 and #1632 each declared exactly this shape as `**BREAKING CHANGES:**`.

**The title is the branch name.** `CONTRIBUTING.md:178` wants `<topic>: <short description>`, imperative and lower-case after the colon; "Claude/io protocol mcp" has neither part. The branch commit already carries a conforming one — `protocol/mcp: migrate session state and the stdio transport to IoEffect`.

**`readFailurePropagates` pins only the *first* read, and the coverage gate can't see the gap.** The transport is a loop and its new JSDoc promises "the failure ends the loop and travels to whoever started the server", but `stdio/proof.f.mjs:78` fails `read` on the very first call, leaving the recursive tail untested. Wrapping `handleLine`'s tail in `rawMapStep(…, () => ok(undefined))` — no new branch, and the added lambda runs on every line so it stays covered — swallows every stdin failure after line one:

```
tsc 0 | fjs t 2968 pass / 0 fail | cov 100.00 | 100.00 | 100.00
```

Driving the real `stdioTransport` with a runner that serves one complete line then fails: clean head gives `["error",["notImplemented","read"]]`, the mutant gives `["ok",null]` — a silent exit 0 on dead stdin. The implementation is right; one assertion that serves a line *then* fails closes it.

Verified otherwise: `tsc` 0 both sides, tests 2968 pass / 0 fail against 2963 on base — +5, exactly the new proofs. `cov` 100/100/100, `prepack` and `ci-update` clean. Declaration diff touches only `stdioTransport`; `mcpStep`'s type is unchanged and `protocol/mcp/module.f.d.mts` is byte-identical. Todo files are accurate in both directions — `unwrapStep` outside the effects layer is now exactly `emergent_testing:130`, which is what `io-effect-migration.md:330-333` claims, while `:324` still reads "In progress" and `io-effect-consumer-migration.md:139` still lists the open `fjs` item, matching the surviving `fjs/module.f.mjs:59`. Nothing checked off that isn't done.

sergey-shandar:
All three addressed in `df1798a`.

**The coverage gap was real, and the mutant you described is the one that got through.** `readFailurePropagates` fails on the very first read, so the loop never recurses and the tail was never exercised. `readFailureAfterALinePropagates` feeds one complete line, lets `echoStep` answer it, and fails the read after that — and it asserts stdin was drained first, so it cannot silently degenerate back into the first-read case. Checked against exactly the mutant you named: wrapping the tail in `catchStep(stdioTransport(handler), () => pureOk(undefined))` leaves `readFailurePropagates` passing and fails the new one (2968/1). Coverage stays 100/100/100 — as you noted, this gap was invisible to the gate, since the recursion adds no branch of its own.

**Changelog** — `changelog/unreleased/1633.md` added, with a matching `Changelog:` section in the body. Two entries, both breaking: the `-32603` answers, and `stdioTransport`'s shape.

**Title** — now `protocol/mcp: migrate session state and the stdio transport to IoEffect`, matching the commit. The PR was opened from the branch name before I had a number to write the entry against; the body is rewritten to the squash-commit form too.

`npx tsc` clean, `fjs t` 2969/0, `npm run cov` 100/100/100.

---

o2alexanderfedin (approved):
Approving. All three are closed, and I checked the proof against my own mutant rather than the one the body names.

**The recursive tail is genuinely pinned.** `rawMapStep(…, () => ok(undefined))` on `handleLine`'s tail (`stdio/module.f.mjs:128`) — the no-new-branch swallow that previously passed `tsc`, 2968/0 and 100/100/100 — now fails `readFailureAfterALinePropagates` (2968 pass / 1 fail). The `catchStep(…, () => pureOk(undefined))` variant the body cites dies on the same proof, so that claim is accurate too. A tail that propagates but stops consuming dies via `multipleLines` and `loopContinuesAfterOversizedResponse`.

Two mutants survive and neither is worth closing. A depth-gated swallow (from the 3rd iteration on) is unrealizable — the tail is a single unconditional `() => stdioTransport(handler)` with no per-iteration state, so discriminating iteration 2 from 3 means adding a counter and a branch; the proof pins both structural positions the code actually distinguishes, base and recursive, and induction over the uniform tail covers the rest. The other re-tags the error to a different `NotImplemented` value, which the existing `readFailurePropagates` doesn't pin either.

**The changelog is byte-identical to the body's section** (452 bytes, same sha256, trailing newline included), conforms to `changelog/README.md`, and — checked by an independent probe rather than by its own proof — both headline claims hold:

| case | base | head |
|---|---|---|
| `initialize`, slot read fails | threw | `-32603`, not `-32002` |
| `initialize`, slot write fails | **successful handshake** | `-32603` |
| `tools/list` gate, read fails | threw | `-32603` |
| `notifications/initialized`, read fails | threw | `null`, session stays gated |

The entry doesn't call out the notification case separately, but "instead of panicking" covers the panic removal and "answers `-32603`" is scoped to paths that answer — a notification has no response frame — so within the ~250-char budget it's accurate, and the body covers it in full.

Body claims check out: the todo's own canonical `unwrapStep` grep returns exactly one implementation site, `emergent_testing/module.f.mjs:130`; deleting `sessionStateFailure` drops branch coverage to exactly the 95.71% the body cites, confirming it's the sole cover for those three `internalError` sites plus the notification path; and the consumer-migration todo's one remaining migration task really is `fjs run`'s bare `unwrap` of `import_`.

Verified otherwise: `tsc` 0, tests 2969/0 against 2968 at the previous head and 2963 on base, `cov` 100/100/100, `prepack` and `ci-update` clean. Across 1168 exported declarations exactly one changed — `stdioTransport` — with `casMcpServer` byte-identical to base. Path leak stays closed: a real stdio session with CAS objects chmod 000 mid-session answers `{"text":"io error: EACCES"}`, stdout byte-identical to base, with no hit for the home path, `/private`, `.cas` or the tmpdir name.

For the record and not worth acting on: the new title is 71 chars, so 79 with the appended ` (#1633)`, against `CONTRIBUTING.md:182`'s 72 including it. I'm not raising it — 42 of the last 200 subjects on `main` exceed 72, including `effects/io: … (#1615)` and `effects: … (#1607)` from this same series, so the practice is clearly not to enforce it. The part that mattered, the branch name as a title, is fixed.

PR #1634, opened 2026-08-17
todo: add `fjs web` static file server issue

sergey-shandar:
Files the design for a `fjs web` command that serves a directory over HTTP, starting with static pages such as `index.html`. Covers the CLI shape, the `fjs/web` module split into a pure path resolver, a socket-free request handler, and a thin `main`; the extension-to-media-type table that byte sniffing cannot replace; the 128 KiB `readFile` cap and its `413`; and the virtual-runner work needed to prove `createServer`/`listen`/`forever`.

Changelog: none

Claude-Session: https://claude.ai/code/session_01XkXZE13sebCYtACKZt7z28

### DIFF
--- a/fjs/todo/web-static-server.md
+++ b/fjs/todo/web-static-server.md
@@ -1,0 +1,129 @@
+## `fjs web` static file server
+
+**Priority:** P3
+**Status:** open
+
+### Problem
+
+There is no way to look at the pages this repository generates. `fjs/website/module.f.mjs`
+writes an `index.html`, and [generate-website](../website/todo/generate-website.md) plans
+`README.md` conversion, `main.css`, per-module `page.f.mjs` demos, and a browser test
+runner — all of which need a local HTTP server to be viewed at all. Today the only
+options are opening `file://` URLs (which breaks anything relying on an origin) or
+installing an unrelated third-party server.
+
+The effects to build one already exist and have no consumer: `CreateServer`, `Listen`,
+and `Forever` are declared in `fjs/effects/node/types.ts`, implemented by the Node
+runner (`fjs/effects/node/module.mjs`), and exported as `createServer`, `listen`, and
+`forever` from `fjs/effects/node/module.f.mjs`. Nothing in the repository calls them,
+so the HTTP side of the effect layer is unproven by any real program.
+
+### Proposal
+
+Add a `fjs web` command that serves a directory over HTTP. The first version does one
+thing: map a request path to a file on disk and return its bytes, so `index.html` and
+its assets load in a browser.
+
+#### CLI
+
+```
+fjs web [root] [port]
+```
+
+`root` defaults to `.` and `port` to `8080`. Both are positional because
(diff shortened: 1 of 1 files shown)


o2alexanderfedin (approved):
Approving. I checked every factual claim the doc makes about the current code rather than reading past them, and they all hold — the three HTTP effects are real and really unused (`CreateServer`/`Listen`/`Forever` at `effects/node/types.ts:190,194,202`, `Forever` in `NodeOp` at `:311`, implemented in the Node runner at `module.mjs:302,320,325`, exported at `module.f.mjs:266,273,278`, and a repo-wide grep outside `fjs/effects/node/` finds only prose); they really are `todo` in the virtual runner (`virtual/module.f.mjs:410-412`) while `readFile`/`stat` are implemented there, which is what makes `respond` provable; `readFile`'s 128 KiB cap is stated at `types.ts:72` exactly as quoted; `fjs/path` exports `normalize`/`concat`/`isProperPrefix`; the `commands` entry shape matches `module.f.mjs:22-65` verbatim; `web`/`w` are free against the live CLI; and `fjs/cli`'s "no named option yet" matches `options-edsl.md:7-9`. The `fjs/media/type` characterization is right too — the magic-byte table can't separate html/css/js because plain UTF-8 all lands in one bucket, per `detect-json.md:11-18`.

No effect-vocabulary problem: the proposal adds nothing to `NodeOp`. It fills three already-declared slots in the *virtual* runner, so nothing frozen widens. Path traversal isn't silent either — `resolve` rejects escapes via `isProperPrefix`, 400 is in the response table, and a traversal case is required in Tasks.

Conventions check out: `## Title` / `**Priority:** P3` / `**Status:** open` matches the template at `todo/README.md:62-85` with both values in the allowed sets, Problem/Proposal/Tasks/Related matches, and `fjs/todo/` is the right home — `todo/README.md:12-19` reserves it for cross-cutting concerns, `fjs/web/` doesn't exist yet, the work touches `fjs/media/type`, `fjs/module.f.mjs` and the virtual runner, and `fjs/todo/66g-fjs-run-commands.md` sets the precedent for a top-level command living here. All five relative links resolve and each target says what the referrer claims, including `requestlistener-stateful.md`, which does pose the question the doc's "unblocked by, but first real consumer of" framing attributes to it. Title conforms and `Changelog: none` is correct for a `todo/`-only PR.

One note, not a defect: **`fjs/emergent_testing/todo/browser-testing.md` already plans a static server** — "loopback static serving" in its shared controller, an `fjs browser-test serve` subcommand, and "the static server rejects paths outside its root" among its fixtures (`:92`, `:101-103`, `:117`). Not a duplicate filing of `fjs web`, but the two overlap on exactly this capability and this traversal rule, and neither links the other. Nothing in `todo/README.md` requires the cross-link; flagging so the overlap is a decision rather than an accident.

Green: `tsc` 0, `fjs t` 2963 pass / 0 fail, `ci-update` clean. Skipped `cov`/`prepack` — the PR touches exactly one `.md` file and no code.

PR #1635, opened 2026-08-17
fjs: report a bad `run` target instead of panicking

sergey-shandar:
The last site the migration's sweeps found, and the change that closes Stage 4.

## The command

`fjs run` unwrapped `import_`, so a file that would not import panicked out of the effect runner. The `assert` beside it did the same when the module imported but exported no `main`. The filed task named only the first, but they are one defect with two spellings: both are the user's command line rather than a program error, and both now answer on `stderr` with exit `1`, formatted `<file>: <reason>` — the same answer an unknown command a few lines up already gave.

`runImportError` was a `throw`-test asserting the panic, so it becomes an ordinary proof; `runNotANodeProgram` covers the second path.

## Stage 4 closes with it

The completion criterion was run rather than assumed, all three sweeps:

- **`unwrapStep`** — one implementation site left, `fjs/emergent_testing`'s registered callback, the considered panic at a foreign boundary.
- **bare `unwrap` / `throw`** — remaining hits are assertions inside `fjs/dev`'s inline proofs and `canonicalHash`'s decode of an already-validated hash, a data invariant its own comment already justified. No IO panic left.
- **published types** — `Reporter`, `Cas.list`, `Evo`, `stdioTransport` and `loadModuleMap` all carry channels now; `mcpStep` and `ToolEntry.handle` answer failures in their own protocol vocabulary, which *is* a channel.

So `fjs/effects/todo/io-effect-consumer-migration.md` is deleted, per the rule about deleting an issue in the pull request that fixes it. Its findings are folded into the parent plan's "What the migrated modules showed" rather than dropped — including the one most worth keeping: the `unwrapStep` grep could not see three of the sites this migration fixed, and no grep can see an interface whose *type* leaves implementations no way to report a failure. `Reporter` and `Cas.list` were both that, and both sat in modules already ticked off as migrated.

Stage 4 is marked **Done** and Stage 5 now reads **Ready**. One Stage 4 item moved rather than closed: revisiting `okStep`, `IoResult` and the redundant adapters is Stage 5's, since `okStep`'s only remaining caller is the Io `step` itself and inlining it belongs with the retirement of the raw layer.

## Validation

`npx tsc` clean, `fjs t` 2970 pass / 0 fail, `npm run cov` 100/100/100, `fjs ci` leaves the tree clean. Re-run after merging `main` (which now carries #1633).

Changelog:
- **BREAKING CHANGES:** `fjs`: `run` reports a target it cannot use instead of
  panicking. A file that will not import, or one that imports without exporting
  a `main` function, is now a message on `stderr` and exit `1`.


o2alexanderfedin (approved):
Approving. This closes the series cleanly, and the two mutants of the shape that got through on #1630 and #1633 both die here.

**The swallow mutants are dead.** Making the import-failure branch return `pure(0)` — no new branch, no new function — fails `runImportError` (2969/1); the same on the not-a-NodeProgram branch fails `runNotANodeProgram`. Dropping the file name from the message dies too, as does exiting 1 with the wrong reason. All seven mutants kept `tsc` clean, so the type checker isn't what caught them.

Panic is gone on every path the changelog names, confirmed against the real binary base vs head:

| | base | head |
|---|---|---|
| nonexistent path | uncaught `triggerUncaughtException` + `['ioError',…]` dump | `<file>: Cannot find module '…'` |
| `.txt` / directory | uncaught `ERR_UNKNOWN_FILE_EXTENSION` / `ERR_UNSUPPORTED_DIR_IMPORT` dump | `<file>: Unknown file extension ".txt"` / `Directory import … is not supported` |
| parse failure | uncaught `['ioError',{message:'Unexpected end of input'}]` | `<file>: Unexpected end of input` |
| no `main` export | uncaught `['not a NodeProgram','<file>']` | ``<file>: not a NodeProgram — no exported `main` function`` |

A program's own `throw` still surfaces as a stack trace on both sides, which the entry doesn't claim otherwise. Worth noting the exit code was already 1 (node's uncaught-exception exit), so the entry's "exit 1" is accurate but the real change is the message — no behavior a script keys on has moved.

**The deleted todo's items are all accounted for**, which was the thing most worth checking. `grep -rn io-effect-consumer-migration` repo-wide returns zero hits. Of its three open items: the `fjs run` `unwrap` is done (verified by the CLI matrix); "delete this file once all three sweeps come back clean" is done and I re-ran all three sweeps independently; the per-PR validation item is carried as `io-effect-migration.md:352`. Its "what the migrated modules settled" findings are all carried into the parent's equivalent section. Nothing dropped silently.

**The end state is what the todo claims.** `unwrapStep` in implementation code is exactly one site — `emergent_testing/module.f.mjs:130`, with the documented foreign-boundary comment at `:126` — and the other nine hits are proofs. Bare `unwrap(` in production is `cas/evo:102`'s `canonicalHash` on an already-validated hash, self-justified. Every surviving `throw` is a data invariant or runner internal (`prime_field`, `btree`, `bnf`, tokenizers, the virtual runner), not an IO panic. Stage 4 has no unchecked boxes and its one live item was explicitly *moved* to Stage 5 with a note rather than quietly closed.

Three cosmetic things, none worth acting on. `fjs run` with no argument reaches the new branch and prints `undefined: Cannot read properties of undefined (reading 'includes')` — still better than base's uncaught dump, and a pre-existing missing-arg gap rather than anything this PR introduced. `proof.f.mjs:70` asserts `stderr.includes('not a NodeProgram')`, so dropping the `<file>: ` prefix survives — but `runImportError` does pin that prefix, and the siblings in this file assert only `stderr !== ''`, so the looser form is the house convention. And `io-effect-migration.md:325-330`'s "Migrated:" list doesn't name the `run` command, though it appears lower down in the grep-insufficiency bullet, so nothing misleads.

Verified: `tsc` 0 both sides, tests 2970/0 against 2969 on base — the +1 is `throw.runImportError` becoming two proofs, nothing dropped. `cov` 100/100/100 with `fjs/module.f.mjs` itself at 100/100/100, `prepack` and `ci-update` clean. Declaration diff across 263 files is one hunk in `fjs/proof.d.mts` for the proof rename; `fjs/module.d.mts` is byte-identical, so no signature or op set moved. Title conforms, and the body's `Changelog:` section is byte-identical to the entry file.

PR #1636, opened 2026-08-17
effects: rename IoEffect to Effect and the representation to RawEffect

sergey-shandar:
Stage 5 of the io-effect migration. The fallible abstraction gets the good name and a default error channel; the `Pure | Do` representation becomes `RawEffect`.

## The premise did not survive the sweep

This stage was written as "retire the raw public `Effect`", keeping the representation under an internal name "rather than maintaining two public effect abstractions". That is not reachable. Of **251** raw `Effect<…>` annotations, **201 have a payload that is not a `Result`**, and several are public types that can never carry an error channel:

- `List<O, T>` is `RawEffect<O, Next<O, T>>` — the payload is a cons cell
- `Program<O>` answers an exit code
- `Step<O>` and `ToolEntry.handle` answer a JSON-RPC response and a tool result, which **are** their error channels

Giving those a channel would put an `ok(…)` wrapper on every stream cell and every tool answer, for a failure that cannot happen. So `RawEffect` is public and load-bearing rather than an implementation detail, and the two names now divide by meaning instead of by layer: `Effect` can fail, `RawEffect` cannot. The todo records this as a supersedure of the original plan, not a quiet deviation.

## The default was added last, on purpose

`Effect<O, T, E = NotImplemented>` makes `Effect<O, T>` legal *and fallible* — which is exactly what an un-migrated infallible site looks like. So the rename ran with the third parameter **required**: every one of those 201 sites became a compile error rather than silently acquiring an error channel it never wanted, and the default went in only once `tsc` was green.

This matters beyond this PR, because the plan's own safety argument was wrong. "`tsc` catches stragglers, since an un-migrated `Effect<O, IoResult<T>>` double-wraps" is true for the 50 fallible sites and was never true for the other 201. That is written into the todo for the next rename described as mechanical.

## Also in here

- **`okStep` is removed.** It was a step adapter whose only implementation consumer was the Io `step`'s body, one indirection away; that branch is written there now. Its proofs went with it — the Io `step`'s own already cover both, which the coverage gate confirms.
- **`fjs/AGENTS.md` §3.4 and `fjs/effects/io/README.md`** are rewritten for the new names, including which of the two to reach for and why.

## Validation

`npx tsc` clean, `fjs t` 2968 pass / 0 fail, `npm run cov` 100/100/100, `fjs ci` leaves the tree clean.

## Deliberately not in here

**Design docs still spell the old names.** Around thirty `todo/*.md` files plus `spec/todo/io-effects.md` contain `Effect<…>` in *proposed* code. Which type each occurrence means takes per-occurrence judgment, and they describe future work rather than current behaviour, so they are filed as a follow-up task in the migration todo rather than swept in silently.

**`fjs/effects/io/` did not move.** Folding it into `fjs/effects/` would collide with the raw module's `step` / `mapStep` exports — a separate structural change, not part of a rename.

Changelog:
- **BREAKING CHANGES:** `effects`: `IoEffect` is renamed `Effect`, with `E`
  defaulting to `NotImplemented`, and the `Pure | Do` representation is renamed
  `RawEffect`. Both stay public: `Effect` for what can fail, `RawEffect` for
  what cannot — a `List` cell, a `Program`'s exit code, a tool result.
- **BREAKING CHANGES:** `effects`: `okStep` is removed. It was a step adapter
  whose only caller was the Io `step`, which now writes that branch itself.


o2alexanderfedin (commented):
The rename is **provably exact**. I applied the inverse renaming to the head sources and emitted declarations and diffed against base: 4 of 264 `.d.mts` and 10 of 448 sources differ, and every surviving hunk is JSDoc prose, the declared `okStep` removal, the declared `E = NotImplemented` default, or an artifact of my own rename hitting an English article. **Not one type signature changed** — `step`'s emitted line is byte-identical across the `@type` → `@template`/`@param`/`@returns` rewrite.

The swap's hazard didn't materialize in code. Only two two-argument `Effect<>` uses exist at head and both are inside prose comments (`effects/io/types.ts:58,61`), so no annotation anywhere leans on the new default and it cannot have silently handed anything an error channel. A type-level probe confirms the raw sites stayed raw — `Program`, `List`, `ToolEntry.handle`, `Step` and `Handle` all answer `RawEffect` — and an intentionally wrong assertion in the same probe fails to compile, so it has teeth. 20 CLI invocations and four stdio MCP sessions are byte-identical base vs head, exit codes and stderr included.

Every finding is a stale doc, and all of them are in files this PR already edited.

**`fjs/effects/io/README.md:5-7` is the one that matters — a site that should have become `RawEffect` and didn't.** Line 3 was correctly updated to `` `Effect<O, T, E>` is `RawEffect<O, Result<T, E>>` ``, but two lines later the same paragraph still says "the raw `Effect<O, T>` remains the low-level representation both it and the raw combinators are built from." At head `Effect<O, T>` is the *fallible* type, so this sentence names the new type while calling it the raw representation — the exact meaning inversion the swap makes dangerous, in the layer's own README.

**The same file documents a symbol this PR deletes.** `7790210b` removes the exported `okStep`, but `:33` still says "wrap the continuation in `okStep`" and `:89-97` keeps a section whose body claims "Io `step` is raw `step` over `okStep`". At head `step` writes `r => r[0] === 'error' ? pure(r) : f(r[1])` itself. The migration todo got its `okStep` item updated; the README didn't.

**And `:71-73` now contradicts a paragraph twelve lines above it.** The `resultStep` section still reads "from stage 5, when the raw representation goes private, this is the public spelling of that instantiation" — but this PR *is* stage 5, and you rewrote `:58-62` to say the opposite ("Both are public and both are load-bearing … `RawEffect` for the representation").

**`io-effect-migration.md`'s Stage 5 checklist contradicts its own new prose.** The stage is headed "**Done.**" while `:426` leaves "Revisit okStep" unchecked even though `:439` checks off "`okStep` was the only one, inlined into the Io `step` that called it" — same work, opposite marks. `:431` and `:432-434` are unchecked for two things the new prose at `:396-406` explicitly decided against, the second one deciding the opposite outcome ("`RawEffect` is public and load-bearing, not an implementation detail"). `:449-454` also still carries the "`tsc` catches stragglers" paragraph verbatim, which `:417-421` now says "was only ever true for the 50 fallible sites … never true for the other 201", with nothing marking it superseded. The rest of this file is the honest part of the PR — it records the supersedure well; the checklist just wasn't reconciled with it.

Minor: `fjs/AGENTS.md:259,269,278`'s three `Cont` examples quote the type verbatim as `(_: Pr<O, O[0]>[1]) => Effect<O, T>`, but `effects/types.ts:73-74` now defines `Cont` with `RawEffect<O, T>`. Legal because `E` is defaulted, and denoting a different type. §3.4 of the same file was updated.

Verified otherwise: `tsc` 0 both sides; tests 2968/0 against base 2970/0 — the delta is exactly `proof.okStep.ok()` and `.error()`, 2 removed and 0 added across the full test-name sets, both declared. Worth stating plainly that this therefore **isn't a pure rename**: it bundles the `okStep` retirement, which is disclosed in the changelog, the body and the todo, and the inlined body is character-for-character the removed adapter. `cov` 100/100/100 confirms Io `step`'s own proofs cover the retired ones. `prepack` clean both sides at 264 files each, `ci-update` clean. The changelog exists, declares both breaking changes, and its 460 bytes are identical to the body's section. `IoEffect` survives only in prior PRs' changelog entries — correctly not rewritten — and in `todo/` design docs the body defers with a tracked follow-up.

sergey-shandar:
All five addressed in `9e30a6b`. Documentation only — no signature or behaviour change, and the inverse-rename property you verified still holds.

**`io/README.md`** had three of them, and they were the same mistake three times: I renamed the lines that *named* the type and left the ones that talked about it.

- `:5-7` — the paragraph renamed line 3 and then said "the raw `Effect<O, T>` remains the low-level representation" two lines later. Now `RawEffect`, and it says the raw type stays public for the computations that cannot fail rather than implying it is a leftover.
- `:33` and `:89-97` — both pointed at `okStep`, which this PR removes. The `:33` mention is dropped (the hazard it described is real without naming the adapter); the `:89-97` section was *about* `okStep` and is rewritten as "The error types are unioned, not unified", which is the property that actually survived, with a sentence recording where the adapter went.
- `:71-73` — promised the representation goes private in stage 5, which `:58-62` now contradicts. Rewritten, and I found the same stale promise in the migration todo's "Target composition" section, which you did not flag; that is fixed too, with a pointer to where the premise changed.

**`io-effect-migration.md`** — the three stage 5 boxes are checked, and one of them is reworded: it said "retire the old public `Effect<O, T>`", which spells the old raw type with the new name and is exactly the confusion at issue. The `tsc`-catches-stragglers paragraph is replaced rather than deleted, because the correction is the part worth keeping: it was true for the fallible sites and false for the ~200 infallible ones, and what actually made the sweep safe was requiring the third parameter until `tsc` was green.

**`fjs/AGENTS.md`** — all three `Cont` samples now quote `RawEffect<O, T>`.

`npx tsc` clean, `fjs t` 2968/0, `npm run cov` 100/100/100, `fjs ci` clean. The fixes are also merged forward into `claude/io-stage6`, which is stacked on this branch, so the stale prose does not come back with it.

---

o2alexanderfedin (commented):
All five are closed, and the replacements are accurate against the code rather than merely different — I checked `io/README.md:88-98`'s new "the error types are unioned, not unified" section against `io/module.f.mjs:128-132`, which does write `r => r[0] === 'error' ? pure(r) : f(r[1])` itself. The merge is clean: `git diff 40ffa7d5 bee3d45c -- spec/` is empty, so #1637's work is untouched, and the only `fjs/` deltas vs main are the rename.

Two one-line leftovers, both the same class as before — prose that survived in a file the fix commit didn't visit.

**`fjs/effects/io/module.f.mjs:181-185` is the third copy of the sentence finding 3 was about.**

```
 * escape from the layer rather than as the deliberate both-branches case. From
 * stage 5, when the raw representation goes private, this is the public
 * spelling of that instantiation.
```

The fix removed this claim from the README and rewrote it in the todo, but left it in `resultStep`'s JSDoc. This PR *is* stage 5 and it decided the opposite — `README.md:58-62` "Both are public and both are load-bearing", `io-effect-migration.md:432-436` "`RawEffect` … stays public". So the module JSDoc contradicts both, about the same symbol, and it ships: it lands in the published `module.f.d.mts:171`.

**`fjs/types/result/module.f.mjs:75` points at the deleted `okStep`.**

```
 * sibling of `fjs/effects`' `okStep` — which is the same combinator for a step
```

This PR removes that export. It's production JSDoc rather than a `todo/*.md`, so the deferral at `io-effect-migration.md:446-449` doesn't cover it. The referent is now the Io `step`.

Verified otherwise: the `okStep` grep returns 22 hits, all in deferred `todo/` docs or the two correct past-tense mentions (`io/README.md:96`, `io/module.f.mjs:114`), except the one above. `IoEffect` survivors are exactly the migration todo and the historical changelog entries for 1599/1604/1629/1630/1633/1636 — nothing reintroduced by the merge. `tsc` 0 both sides, which is itself evidence no un-renamed type reference survives since the raw type no longer answers to `Effect`. Tests 2968/0 against base 2970/0, and the test-name diff shows the delta is exactly `proof.okStep.ok()` and `.error()` with no other name changed. `cov` 100/100/100, `prepack` and `ci-update` clean, changelog byte-identical to the body's section.

Not from this PR, flagging only so it isn't mistaken for a regression: `io/README.md:139-155` still says "No Io `historyStep`" and that `foldStep`/`forEachStep` variants "arrive when real consumers require them", while `io/module.f.mjs` exports all three at `:275`, `:301`, `:322`. Identical on `origin/main` — the rename only changed `IoEffect`→`Effect` inside that text.

sergey-shandar:
Both fixed in `ee7f044`. Documentation only again.

**`io/module.f.mjs:181-185`** — that was the *third* copy of the same sentence, after the README's and the todo's. Now: "This is the spelling that says the both-branches case was meant", with no claim about a privacy that is not happening.

**`types/result/module.f.mjs:75`** — reworded rather than just dropping the name, since the cross-reference is still true: `okThen` is the pure sibling of the branch `fjs/effects/io`'s `step` now writes itself.

I then grepped for `okStep` across the repo rather than stopping at the two you named, and found **six `todo/*.md` files still citing it as an existing export** — `effect-list-fold`, `fold-stream-combinator`, `map-step-combinator`, `step-continuation-operation-union`, `../node/todo/ornotfound-combinator`, `../../cas/todo/write-closed-helpers`. Those are inside the design-doc sweep already deferred to its own task, so I left them there, but I have written into that task that they are a *different* problem from the rest of it: the others are stale names that substitute mechanically, while these point at something deleted, so each needs a decision about what its proposal meant. That way the follow-up cannot quietly treat them as a find-and-replace.

`main` had moved twice while I was working, so this also carries two merges of it. Re-validated after each: `npx tsc` clean, `fjs t` 2968/0, `npm run cov` 100/100/100, `fjs ci` clean.

Merged forward into `claude/io-stage6` (#1639), which re-validates at 2971/0 and 100/100/100.

---

o2alexanderfedin (approved):
Approving. Both leftovers are closed and the replacements are accurate against the code, not just different.

`grep -n 'goes private\|stage 5, when'` now returns zero hits repo-wide, and the published `module.f.d.mts:171` carries the new sentence with none of the old one — so the claim that shipped is gone from the artifact too. The `types/result` replacement checks out against the source: `io/module.f.mjs:128-132` writes `r => r[0] === 'error' ? pure(r) : f(r[1])` and `types/result:99`'s `okThen` is `r => r[0] === 'error' ? r : f(r[1])` — the identical branch modulo the `pure` lift, so "pure sibling" is exactly right.

One thing on the new note itself: **the enumeration at `:449-455` is one short.** Its antecedent set is "the `todo/*.md` under `fjs/effects`, `fjs/cas`, `fjs/protocol/mcp` and friends, **plus `spec/todo/io-effects.md`**" — and that last file also points at `okStep` as an existing export, present tense:

> The remaining combinators are `step` specializations … `mapStep` …, `foldStep` and `forEachStep` (sequential iteration), **`okStep` (the `Result` short-circuit)**.

`spec/todo/io-effects.md:113-117`. Its siblings in that sentence all still exist; `okStep` doesn't. Same dangling-reference class as the six, in a file the bullet explicitly names, so a sweep driven by the enumeration would skip it. "Seven of them", adding `spec/todo/io-effects.md`.

The six-file claim itself is true — all six exist, both relative paths resolve from `fjs/effects/todo/`, and each references `okStep` present-tense (`effect-list-fold:105`, `fold-stream-combinator:7,121`, `map-step-combinator:16,113`, `step-continuation-operation-union:30`, `node/todo/ornotfound-combinator:22,56`, `cas/todo/write-closed-helpers:66`). The full 27-hit grep reconciles: 9 in those six, 2 correct past-tense, 5 append-only changelog records, 9 stage-narrative or Stage-5-record lines in the migration doc — leaving `spec/todo/io-effects.md:117` above and `io-effect-migration.md:526`'s own Related index, which I've raised on #1639 rather than counting twice.

Green: `tsc` 0, `fjs t` 2968 pass / 0 fail, `cov` 100/100/100, `prepack` and `ci-update` clean.

PR #1637, opened 2026-08-17
spec: consolidate the language specification into one document

sergey-shandar:
## Summary

This PR consolidates the FunctionalScript language specification by moving content from individual feature documents into the main `spec/README.md`, creating a comprehensive single-source reference for the currently supported language features.

## Key Changes

- **Consolidated specification**: Merged 12 individual feature specification files into the main README:
  - `1000-json.md` → JSON Input section
  - `2110-default-export.md` → Exporting a Value section
  - `2120-const.md` → Shared Values, Constants section
  - `2130-default-import.md` → Importing Other Modules section
  - `2210-block-comment.md` → Comments section
  - `2310-undefined.md` → Supported Value Types section
  - `2320-bigint.md` → Supported Value Types section
  - `2410-identifier-property.md` → Property Keys section
  - `2420-line-comment.md` → Comments section
  - `2430-trailing-comma.md` → Arrays section
  - `2470-computed-property.md` → Property Keys section
  - `2480-proto-property-key.md` → The `__proto__` Key section

- **Reorganized README structure**:
  - Added "Principles" section explaining compatibility with JavaScript
  - Added "Command Line" section with file types and output details
  - Added "Supported Value Types" section covering all data types
  - Added "Module Structure" section explaining statement ordering
  - Added "Roadmap" section pointing to future features

- **Updated cross-references**: Fixed all internal links throughout the codebase to point to the new consolidated sections in the main README instead of individual files

- **Improved documentation clarity**: Expanded explanations with more examples and context while maintaining technical accuracy

## Implementation Details

- The main README now serves as the authoritative specification for currently-supported language features
- Individual feature documents were removed to avoid duplication and maintain a single source of truth
- All references in code comments, other documentation, and todo files were updated to link to the appropriate sections in the main README
- The structure follows a logical flow: principles → command line → language features → roadmap

Every statement in the new document was verified against the current compiler rather than carried over from the old text. That surfaced one discrepancy, now documented instead of repeated: the old README said `bigint` and `undefined` "are not available" in JSON output, but the `.json` writer emits the module spellings `34n` and `undefined`, producing a file that is not valid JSON. The spec now states the actual behavior and links the tracking issue, `fjs/djs/todo/json-bigint-serialization.md`.

https://claude.ai/code/session_01DWqGqw5dNiP4Pdfdxm8yz9

Changelog: none


o2alexanderfedin (approved):
Approving. I audited all twelve deleted files against the new document line by line rather than trusting line counts, and nothing was lost.

The one I checked hardest is `2480-proto-property-key.md`, since its qualifiers were added deliberately across #1626 → #1628 → #1631 and a shortened restatement would have quietly undone that work. All survive: "error **in a FunctionalScript module**" is preserved by placement inside the module chapter with the JSON-input case spelled out separately; "the extension of each file **named on the command line** picks the language" is verbatim and still bold; "the only spelling whose evaluation reproduces the property", the JSON emitter keeping the plain key, the imported-file rule, and "`{__proto__: 1}` is an error whatever the input file is called" are all intact. `2470`'s computed-key restriction survives and is still enforced — `["constr"+"uctor"]`, `[name]`, `[1]`, `["a","b"]`, `{3e7: true}` rejected; `["a"]`/`["a-b"]` accepted.

I re-ran the normative claims against the real compiler rather than only checking the text moved. The `proto.json → a.f.js → out.json` loop is byte-identical under `cmp`, both bad spellings error with the documented `path:line:column` format, sorted-keys-by-UTF-16-code-unit holds (`"10"` before `"2"`), infinity→`null` holds, and the const inlining/hoisting sample reproduces exactly.

**On `50861ae0`'s new sentence** — "the ❌ rows … are also what a `.json` output cannot carry" — I checked this specifically, because a flat reading of it would reassert the very falsehood this PR corrects. The writer does emit them:

```
export default 34n                        → 34n
export default { "a": 34n, "b": undefined } → {"a":34n,"b":undefined}
```

both exit 0. But the sentence links to `#output`, and that section says the `.json` writer "currently emits the module spellings `34n` and `undefined` anyway, producing a file that is not valid JSON; it should reject the value instead", with the tracking todo. So "cannot carry" reads as capacity-of-the-format with the caveat one click away — materially different from the old README's unqualified "not available", which left a reader thinking the compiler dropped them. Not a reintroduced falsehood, and I wouldn't block on it. The reference row is right too: a shared const expands to two independent copies in `.json`, so the file genuinely carries no reference.

`fe414920` is purely positional plus one additive clause, and both its code samples compile — `export default 5` → `5`, the object sample → `{"name":"fjs","tags":["data","config"]}`. The second new paragraph on references is accurate, including that sharing is by value not name: `const s = "hi"` used twice plus a bare `"hi"` collapse into one const.

Zero dangling references — `grep -rn` for all twelve deleted filenames returns nothing — and every anchor resolves in both directions: 20 headings, no duplicate slugs, all intra-document links, and every inbound `README.md#…` from `spec/todo/*` (10 files), `spec/todo/README.md`, `vm-command-format.md`, `todo/edag-stage1-discussion.md`, `fjs/djs/README.md` and the three JSDoc references, including the new `#output` and `#shared-values-constants`. The three `.mjs` files changed six JSDoc lines and zero executable lines, so `Changelog: none` is right.

Two notes.

**The title doesn't conform to `CONTRIBUTING.md:178`** — `<topic>: <short description>`, imperative, lower-case after the colon. "Consolidate spec documentation into main README" has no topic prefix and is capitalized, and it becomes the squash commit. Your own branch commit already has the right form: `spec: consolidate the language specification into one document`.

**"DJS" loses its only definition in `spec/`.** Base §2 opened with "DJS is the data subset of FunctionalScript"; the new document never uses the term, while `spec/todo/README.md:17` still titles a section "DJS" and `spec/todo/1010-undefined-property.md:40` says "the `undefined` value in DJS" pointing at `../README.md#supported-value-types`, which no longer defines it. The anchor resolves; only the term is now undefined.

Green: `tsc` 0, `fjs t` 2970 pass / 0 fail at base parity, `ci-update` clean.

sergey-shandar:
Both notes acted on.

**Title** — retitled to `spec: consolidate the language specification into one document`, matching `CONTRIBUTING.md:178`. 61 characters including the ` (#1637)` GitHub appends.

**DJS** — you're right that the term lost its only definition: `grep -c DJS spec/README.md` was 0 while `spec/todo/` still uses it in eight places, including the section title at `spec/todo/README.md:17` and the `1010-undefined-property.md:40` link you cite. Fixed in 9941f1a by defining both layers in the introduction — DJS as the data subset whose values serialize without additional run-time information, FJS as DJS plus functions, and everything the compiler accepts today being DJS. Put in the introduction rather than under `#supported-value-types` so the name is defined once before any section uses it; the inbound link from `1010` is about `undefined`, which still lives in that section.

On the `.json` sentence: your reading is the intended one, and thank you for testing it rather than assuming. "Cannot carry" is about the format, with `#output` one click away stating what the writer does today and that it should reject instead.

---

o2alexanderfedin (approved):
Both notes are closed. Re-approving at `9941f1a3`.

The DJS/FJS paragraph goes further than what I flagged and is accurate against every existing use. `fjs/djs/README.md:1-8` gives the same two properties and the same expansion of the acronym, `:16`'s "a DJS module denotes a **graph**" is verbatim the new line's claim, and `fjs/djs/todo/157.md:6` is compatible rather than contradictory — 157 states DJS's lower bound (⊇ JSON) where this states its upper bound (no functions). The FJS half compresses `spec/todo/README.md:51-53` exactly, and agrees with `serialization.md:14`, `io-effects.md:162-164`, `vm-command-format.md`'s DJS-vs-`function` command split, and `2460`'s "JSON ⊂ DJS ⊂ FS lattice".

"Everything the compiler accepts today is DJS" holds under test — every function form is rejected:

```
export default () => 1              → 1:16 unexpected token
export default { "f": () => 1 }     → 1:23 unexpected token
export default function () {…}      → 1:16 const not found
const f = () => 1 / { f }           → 1:11 unexpected token
export default { "x": 1 }           → exit 0
```

The final clause is the load-bearing one: it licenses `spec/todo/README.md`'s `## 2. DJS` bucket listing unimplemented items, which would otherwise look like it contradicted "this document specifies what the compiler accepts today". With it, the section reads correctly.

Everything from the earlier passes still stands: all twelve deleted files audited line by line with every `__proto__` qualifier intact, the computed-key restriction still enforced, zero dangling references, every anchor resolving both directions, and the three `.mjs` files changing six JSDoc lines and zero executable lines. Green: `tsc` 0, `fjs t` 2970 pass / 0 fail, `ci-update` clean.

PR #1638, opened 2026-08-17
path: file an issue for the lost root of an absolute path

sergey-shandar:
Adds `fjs/path/todo/absolute-path-root-segment.md`. No code changes.

Found while verifying the specification consolidation (#1637) against the real compiler.

## The bug

`parse` splits on `/` and drops every empty segment, so the leading empty segment that *is* the POSIX root disappears with them:

```js
normalize('/a/b/m.f.js')            // 'a/b/m.f.js'  — absolute became relative
concat('/a/b/m.f.js')('..')         // 'a/b'
concat('//srv/share/m.f.js')('..')  // 'srv/share'   — UNC root lost too
normalize('/')                      // ''
```

`foldNormalizeOp`'s `case '': case '.':` is right for an interior empty segment — `a//b` is `a/b` — and wrong for the first one, which is the root marker rather than a missing name. A Windows drive letter survives only because `C:` happens to be a non-empty segment.

`fjs compile` is where this becomes visible. The transpiler resolves each import as `concat(concat(path)('..'))(importPath)`, so naming the input absolutely turns every import into a path resolved against the current directory:

```sh
$ cd /tmp/x && fjs compile m.f.js out.json     # m.f.js imports ./lib.f.js
$ cat out.json
[{"port":8080}]

$ fjs compile /tmp/x/m.f.js out.json           # the same module, named absolutely
/tmp/x/m.f.js - error: file not found
```

The file named in that message exists — the one that was not found is `tmp/x/lib.f.js`, under the current directory. A module with no imports compiles fine from an absolute path, because the file named on the command line is read with the path as given and only imports go through `concat`.

The limitation is already worked around at the one other call site that meets absolute paths, `fjs/effects/node/module.mjs`: `v.includes(':') || v.startsWith('/') ? v : concat(process.cwd())(v)`. That `startsWith('/')` test exists because `concat` would eat the root. `fjs/cas` and `fjs/effects/node/virtual` also parse absolute paths and both happen to be unaffected — each compares or slices two paths that went through the same function, so the missing root cancels out.

## The issue file

Proposes splitting the root off before folding the segments and reattaching it afterwards, with `..` not allowed to escape a root that exists (`/a/../..` is `/`, while `a/../..` stays `..`). It also flags the open API question: `parse` returns `readonly string[]` and is exported, so the root has to reach its three callers somehow, and either way of doing that is a breaking change to a small API.

The misleading error message is not part of this issue — that half is already tracked by `fjs/djs/todo/parse-error-location-format.md`, which the file links.

Changelog: none

https://claude.ai/code/session_01DWqGqw5dNiP4Pdfdxm8yz9

---
_Generated by [Claude Code](https://claude.ai/code/session_01DWqGqw5dNiP4Pdfdxm8yz9)_

### DIFF
--- a/fjs/path/todo/absolute-path-root-segment.md
+++ b/fjs/path/todo/absolute-path-root-segment.md
@@ -46,8 +46,8 @@ The file named in that message exists; the one that was not found is
 imports compiles from an absolute path, because the file named on the command
 line is read with the path as given and only imports go through `concat`.
 
-The limitation is already worked around at the one other call site that meets
-absolute paths, [`fjs/effects/node/module.mjs`](../../effects/node/module.mjs):
+The limitation is already worked around at the one other call site it breaks,
+[`fjs/effects/node/module.mjs`](../../effects/node/module.mjs):
 
 ```js
 const s0 = v.includes(':') || v.startsWith('/') ? v : concat(process.cwd())(v)
--- a/fjs/path/todo/absolute-path-root-segment.md
+++ b/fjs/path/todo/absolute-path-root-segment.md
@@ -1,0 +1,102 @@
+## Keep the root of an absolute path
+
+**Priority:** P3
+**Status:** open
+
+### Problem
+
+[`parse`](../module.f.mjs) splits a path on `/` and drops every empty segment,
+so the leading empty segment that *is* the POSIX root disappears with them:
+
+```js
+normalize('/a/b/m.f.js')      // 'a/b/m.f.js'   — absolute became relative
+concat('/a/b/m.f.js')('..')   // 'a/b'
+parse('/a/b')                 // ['a', 'b']
+concat('//srv/share/m.f.js')('..')  // 'srv/share'  — UNC root lost too
+normalize('/')                // ''
+```
+
+`foldNormalizeOp` treats `''` as noise:
+
+```js
+case '': case '.': { return state }
+```
+
+That is right for an interior empty segment — `a//b` is `a/b` — and wrong for
+the first one, which is not a missing name but the root marker. A Windows
+drive letter survives only because `C:` happens to be a non-empty segment.
+
+`fjs compile` shows the consequence. The transpiler resolves an import with
+`concat(concat(path)('..'))(importPath)`
+([`fjs/djs/transpiler/module.f.mjs`](../../djs/transpiler/module.f.mjs)), so
+naming the input by an absolute path turns every import into a path resolved
+against the current directory instead:
+
+```sh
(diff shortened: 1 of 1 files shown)


o2alexanderfedin (approved):
Approving. Every factual claim reproduces exactly, which is the thing that matters in a bug report — a misstated repro sends the fixer the wrong way.

```
normalize('/a/b/m.f.js')          → "a/b/m.f.js"
concat('/a/b/m.f.js')('..')       → "a/b"
parse('/a/b')                     → ["a","b"]
concat('//srv/share/m.f.js')('..')→ "srv/share"
normalize('/')                    → ""
```

All return normally — no throw, no error value — so "lost" and "silent" are accurate at the `fjs/path` layer. The contrast cases hold too: `normalize('C:/a/b')` keeps the drive letter, `normalize('a//b')` correctly drops the interior empty segment.

The `fjs compile` scenario reproduces verbatim, including the non-obvious part — *which* file goes missing. From an unrelated cwd, `compile <abs>/m.f.js` fails with `file not found`; after creating `./tmp/x1638/lib.f.js` **under the cwd**, the same command succeeds and emits `[{"port":8080}]`. The import really is resolved against the current directory.

The impact audit checks out in both directions, which is the part most bug reports get wrong. `djs/transpiler:50-51` is affected as claimed; `effects/node/module.mjs:112` is verbatim the quoted workaround; and `fjs/cas` and `effects/node/virtual` are correctly excluded for exactly the reason given — both sides of their comparisons go through the same function. `fjs/dev` imports only `join`/`relativize`/`toPosix`, none of which fold segments, so omitting it is right rather than an oversight. "Three callers" of `parse` is accurate.

No existing proof pins the current behavior — `normalizeTest`/`concatTest` inputs are all relative, and the `'/abs/root'` case in `joinTest` goes through `join`, which preserves roots — so the file is right that a fix wouldn't have to change one.

Conventions and hygiene are fine: header values are in `todo/README.md`'s tables, sections match the template, `fjs/path/todo/` satisfies the placement rule, all four relative links resolve, and no duplicate exists anywhere in `todo/`, `spec/todo/` or the `fjs/*/todo/` trees. `Changelog: none` is right for a `todo/`-only PR. Green: `tsc` 0, `fjs t` 2970 pass / 0 fail, `ci-update` clean.

One optional wording note: `:50` calls `effects/node/module.mjs` "the one other call site that meets absolute paths", and `:58` then says `fjs/cas` and `effects/node/virtual` also parse them. The second sentence resolves the first, so it reads fine in sequence — mentioning only in case you meant "the one other *affected* call site".

PR #1639, opened 2026-08-17
effects: a runner may omit an operation and answer NotImplemented

sergey-shandar:
Stage 6, the last of the io-effect migration.

**Stacked on #1636.** Until that merges, the diff here shows Stage 5's rename as well; afterwards it reduces to the Stage 6 commits alone.

## The mechanism is not quite either shape the plan predicted

The stage was written expecting either "a partial operation-map type whose constraint guarantees every operation's return admits `error(NotImplemented)`" or "a separate Io-aware `match`". Sketching all three showed they collapse.

`match` cannot construct the answer *however* it is constrained, because `R` is the **runner's** wrapper — `Promise<…>` for an async loop, `(state) => [state, …]` for a state-threading one — not the operation's return type. A type-level constraint still needs a value that knows how to build an `R`, and only the runner has one. So `partialMatch(commands, onMissing)` takes the injector, and each runner writes its own once, beside the loop that fixes the shape; `partialRun` in `fjs/effects/mock` is the only one today.

## Two consequences the plan did not anticipate

**Partiality is opt-in.** `OperationMap` stays total, so a runner that merely *forgets* a handler is still a compile error — which is what should happen to the Node runner. `PartialOperationMap` is for a runner that is *meant* to lack operations.

**A missing handler is two different things, and types are erased.** An omitted `readFile` and a garbled `readFilee` reach the interpreter as the same failed lookup, so telling "declared but unimplemented" (recoverable) from "never a command at all" (a malformed node, still a panic) needs `O`'s commands as data. `CommandSet<O>` is a `Record<O[0], null>` for exactly that reason: a record is checked for **completeness**, so a command added to `O` and forgotten there is a compile error, where an array literal would only have its members checked and drift silently. It also verified the 28-key `NodeOp` list rather than leaving it to a careful reading of the union.

## The consumer

`fjs/effects/node/virtual` had five `todo` handlers — `exec`, `createServer`, `listen`, `forever`, `test` — that existed only to satisfy a total map and threw when reached. That is "this runner does not implement this operation" spelled as a panic. They are simply absent now.

`asyncRun` and the Node runner are **not** made partial: they implement all of `NodeOp`, so partiality would trade a compile-time guarantee for a runtime answer nothing asks for. The todo records that as a deliberate scope decision rather than a skipped task.

## Validation

Three proofs cover the new paths — `unimplemented.declaredButAbsent`, `unimplemented.throw.undeclaredCommand`, and `unimplemented.programChoosesAFallback`, the last being the stage's actual claim: a program catches an unavailable `exec` and writes instead, and the run completes.

The module header's "three functions discriminate `Pure` from `Do`" still holds — `match` and `partialMatch` are two entry points sharing `_matchWith`, so the count did not move.

Changelog:
- `effects`: a runner may implement part of an operation set. `partialMatch` and
  the mock `partialRun` answer a declared-but-unimplemented command with
  `error(notImplemented(command))` through the ordinary continuation, so a
  program can recover; a command outside the operation set still panics.
- **BREAKING CHANGES:** `effects/node/virtual`: `exec`, `createServer`,
  `listen`, `forever` and `test` are no longer implemented. They used to throw
  when reached and now answer `NotImplemented`.


o2alexanderfedin (commented):
The mechanism is sound and I couldn't break it. Seven mutants, seven dead — including the omitted-op detection answering `ok`, reporting the wrong op name, and a present-but-failing op made to look omitted. Worth noting **every one of those mutants still reported 100/100/100 coverage**, so the gate saw none of them; the proofs are what caught them.

Both risks I went looking for turned out not to exist. A misspelled op key is a compile error, not a silent omission — `readFilee` gives `TS2561 … Did you mean to write 'readFile'?`, and dropping a key from `nodeCommandSet` gives `TS2741`, so the `CommandSet` completeness argument really does pin all 28 `NodeOp` keys. And the mock change weakens no proof: the five affected ops were `todo` before, i.e. they threw, so anything reaching one already failed loudly. `partialRun`'s only caller is `virtual`, and the total `run` is behaviour-preserving with its four users untouched.

Behaviour is byte-identical for total runners — 19 CLI invocations, 5 compile runs including emitted file contents, and four stdio MCP sessions all match, with zero absolute paths on any protocol stdout. Declarations are additive only; `NodeOp` and `virtual: RunInstance<NodeOp, State>` are unchanged, so no consumer-supplied runner has to handle a new op.

Two doc lines, both present-tense claims that this PR makes false.

**`fjs/effects/todo/io-effect-migration.md:525-526`** still says:

> Until Stage 6 lands, `NotImplemented` may exist in the type model while current runners remain total over their declared operation maps.

This PR *is* Stage 6. The sentence sits immediately after the `## Stage 6` section headed "**Done.**" with all eight boxes `[x]`, contradicts the bullet three lines above it ("opt in where a runner has operations it genuinely lacks, **which today is the virtual one**"), and the file header at `:4-6` says "all six stages have landed".

**`:554`'s `### Related` index** reads "`fjs/effects/module.f.mjs` — raw `step`, `okStep`, `match` (whose missing-handler `assert` Stage 6 reworks)". Both halves are stale: `okStep` was removed by #1636 and isn't among the module's ten exports, and the `assert` rework landed in *this* PR as `partialMatch`/`PartialOperationMap`, so "Stage 6 reworks" is future tense about done work. A Related index describes current state rather than history, so unlike `:140` — which reads as Stage 3 narrative and I'd leave alone — this one needs updating alongside `:525`.

Verified otherwise: `tsc` 0, `fjs t` 2971 pass / 0 fail (+3 over #1636, exactly this PR's stage-6 proofs), `cov` 100/100/100, `ci-update` clean. Omitted ops answer through the ordinary continuation — all five formerly-`todo` virtual ops give `["error",["notImplemented","<op>"]]`, a present-but-failing `readFile` gives its `ioError` and isn't conflated, an undeclared op still throws, and the total `match` still panics. The increment since my last pass is fully accounted for: #1636's two fixes verbatim, the `okStep` note, and #1638's todo file byte-identical to main.

sergey-shandar:
Both fixed in `8f42176`. Documentation only.

**`:525-526`** — the caveat that `NotImplemented` exists in the type model with nothing producing it. Replaced rather than deleted, because the interesting half survives the stage: a runner is no longer *obliged* to be total, and `asyncRun` and the Node runner stay total by choice rather than by the type system's insistence.

**The Related index** — now points at `match` and `partialMatch` with the rework in the past tense, and records that `okStep` used to be listed there and where it went, rather than silently dropping the line.

**That caveat appears twice**, and you found one. The other is at `:140-142`, inside the Stage 3 section — same sentence, same claim, now false the same way. It is fixed here too, kept in the past tense since it sits in a section describing what Stage 3 established. I checked the rest of the file for other future-tense stage claims: the only remaining one is `:131`, which is a counterfactual about the rejected alternative ("Stage 6 *would then* have no typed channel"), correctly hypothetical and left alone.

Thanks for the mutation run — the omitted-op-answering-`ok` and wrong-op-name mutants are the two I most wanted killed, and your point that the gate saw none of the seven at 100/100/100 matches what this migration kept running into: coverage never had an opinion about any of the real defects in it.

`#1636` merged while I was working, so this branch now carries `main` and the PR reduces to the Stage 6 commits. Re-validated after that merge: `npx tsc` clean, `fjs t` 2971/0, `npm run cov` 100/100/100, `fjs ci` clean.

---

o2alexanderfedin (approved):
Approving. Both are closed, and I checked the replacement claims against the code rather than just confirming the old text was gone.

**Five is the right number.** Iterating all 28 `nodeCommands` through `virtual(emptyState)`, exactly five answer `['error', ['notImplemented', c]]` — `createServer`, `exec`, `forever`, `listen`, `test` — and `git diff` against main shows precisely those five `todo,` lines removed. Same set the runner's own comment names.

**"Total by choice, not by the type system's insistence" is literally true**, and I proved the second half rather than taking it. `asyncRun` uses `match(map)` and `ToAsyncOperationMap<O>` is a non-optional mapped type: deleting the `exec` handler from `runNodeEffect` gives `TS2741: Property 'exec' is missing … required in type 'ToAsyncOperationMap<NodeOp>'`, and renaming it gives `TS2561 … Did you mean to write 'exec'?`. So the exhaustiveness check the sentence credits them with is real and still enforced. `partialMatch`'s only caller remains `partialRun`, whose only caller is the virtual runner.

The `### Related` rewrite checks out too — `module.f.mjs:385-388` keeps `match`'s `assert(false, command)` while `:412-416` gives `partialMatch` the `assert(commands.includes(...))` then `onMissing(command)` split, both over a shared `_matchWith`; and `io/module.f.mjs:127-129` records the `okStep` inlining with `step` at `:140-144` carrying the inlined body.

Two things I looked at and decided are *not* problems. `mock`'s own `run` is a third total runner the sentence doesn't name, but it's total by choice with its own exhaustiveness check, and the paragraph summarises a bullet already scoped to `NodeOp` runners — naming it wouldn't change any claim. And revising `:137-142` into past tense inside the Stage 3 block reads fine: that bullet was always an interim-state statement rather than Stage 3 work, so the fit of the lead-in is unchanged in kind.

I also swept the file for `Stage 6 | will | until | not yet | no runner | total | today | currently | remains | reworks` and checked every hit — no surviving present-tense claim that Stage 6 falsifies. `:497`'s "`partialRun` … is the only one today" is accurate.

Green: `tsc` 0, `fjs t` 2971 pass / 0 fail, `cov` 100/100/100, `ci-update` clean, and the delta since my last pass is that one `.md` file only.

For context, not against this PR: `spec/todo/io-effects.md:117` still lists `okStep` as a current combinator. Your own `:449-455` sweep item names that file explicitly, so it's tracked.

PR #1640, opened 2026-08-18
effects: spell the effect channel Effect, not RawEffect&lt;O, Result&lt;…&gt;&gt;

sergey-shandar:
**Stacked on #1639** (itself stacked on #1636). Until those merge, the diff here shows Stage 5's rename and Stage 6 as well; afterwards it reduces to the one commit `effects: spell the effect channel Effect, not RawEffect<O, Result<…>>`.

`RawEffect` is still all over the tree in a spelling that predates the `Effect` alias. Because `Effect<O, T, E>` **is** `RawEffect<O, Result<T, E>>`, every change here is type-identical — nothing changes at runtime, and the whole value of the change is that the annotation now says what the `Result` means.

## Not every `Result`-valued raw effect is an `Effect`

Which is why this is 25 sites and not the 42 a grep suggests. The rule, now written into `Effect`'s doc rather than left to be re-derived on the next sweep:

It is the **effect channel** — spell it `Effect` — when `E` carries `NotImplemented`, because only a runner that could not dispatch produces that error.

It is **returned data** — leave it `RawEffect` — when `E` is a domain verdict the caller asked for, one that has *already absorbed* whatever channel produced it. Two of those:

- `evo.revision`'s `Result<RevisionData, string>`: a runner failure, a missing shard, and a blob that is not a revision are all the same `string` to the caller. `Evo.add`'s doc already says the domain verdict must not be collapsed into the channel; this is that rule applied to its sibling.
- the transpiler's `Result<Unknown, ParseError>`: a missing file is reported as a `ParseError` like any other parse failure.

Nine sites are left raw for that reason (five in `djs/transpiler`, four in `cas/evo`). Converting them would claim `string` or `ParseError` is what a runner answers when it cannot dispatch.

The remaining hits were prose describing the alias, or names that merely end in `Result` — `MatchResult`, `ToolsCallResult`, `ToolsListResult` — and one `OpResult` sitting in an operation's own return type rather than in an effect's payload.

## What was converted

`RawEffect<O, OpResult<T>>` → `Effect<O, T>` (the default channel); `RawEffect<O, IoResult<T>>` → `Effect<O, T, NotImplemented | IoError>`, the spelling `Effect`'s doc already uses as its example.

| file | n |
|---|---|
| `fjs/cas/module.f.mjs` | 7 |
| `fjs/cas/proof.f.mjs` | 3 |
| `fjs/cas/types.ts` — `Cas.write`, `Cas.list` | 2 |
| `fjs/cas/evo/proof.f.mjs` — `virtualOk` | 1 |
| `fjs/effects/node/module.f.mjs` — `all`, `both`, `writeFromStream`, `createServer` | 4 |
| `fjs/effects/node/types.ts` — `_WriteLoop`, `Console` | 2 |
| `fjs/effects/memory/module.f.mjs` | 3 |
| `fjs/media/type/module.f.mjs` — `detectStream` | 2 |
| `fjs/protocol/mcp/stdio/module.f.mjs` | 1 |

Also drops two claims in `effects/io/types.ts` that stages 3 and 4 had not happened yet.

## Validation

`npx tsc` clean, `fjs t` 2971 pass / 0 fail, `npm run cov` 100/100/100, `fjs ci` leaves the tree clean. Coverage is unmoved because no branch moved: these annotations are erased.

Changelog:
- `effects`: `Effect<O, T, E>` replaces `RawEffect<O, Result<T, E>>` wherever the
  `Result` is the effect channel — `cas`, `cas/evo`'s runner unwrap,
  `effects/memory`, `effects/node`'s `all`/`both`/`writeFromStream`/
  `createServer`/`Console`, `media/type`'s `detectStream`, and MCP stdio's
  response writer. The two spellings are the same type, so nothing changes at
  runtime; `Effect`'s doc now states the rule that decides between them.

### DIFF
--- a/changelog/unreleased/1640.md
+++ b/changelog/unreleased/1640.md
@@ -1,0 +1,6 @@
+- `effects`: `Effect<O, T, E>` replaces `RawEffect<O, Result<T, E>>` wherever the
+  `Result` is the effect channel — `cas`, `cas/evo`'s runner unwrap,
+  `effects/memory`, `effects/node`'s `all`/`both`/`writeFromStream`/
+  `createServer`/`Console`, `media/type`'s `detectStream`, and MCP stdio's
+  response writer. The two spellings are the same type, so nothing changes at
+  runtime; `Effect`'s doc now states the rule that decides between them.
--- a/fjs/cas/evo/proof.f.mjs
+++ b/fjs/cas/evo/proof.f.mjs
@@ -1,9 +1,8 @@
 /**
  * @import { Cas } from '../types.ts'
- * @import { RawEffect } from '../../effects/types.ts'
+ * @import { Effect } from '../../effects/io/types.ts'
  * @import { NodeOp } from '../../effects/node/types.ts'
  * @import { State } from '../../effects/node/virtual/types.ts'
- * @import { Result } from '../../types/result/types.ts'
  * @import { Vec } from '../../types/bit_vec/types.ts'
  * @import { Ok } from '../../types/result/types.ts'
  * @import { List } from '../../effects/list/types.ts'
@@ -78,7 +77,7 @@ const fixedCas = entries => ({
  * outer one is unwrapped here, the inner domain verdict is what each proof
  * goes on to inspect.
  *
- * @type {(state: State) => <T, E>(e: RawEffect<NodeOp, Result<T, E>>) => readonly [State, T]}
+ * @type {(state: State) => <T, E>(e: Effect<NodeOp, T, E>) => readonly [State, T]}
  */
 const virtualOk = state => e => {
     const [s, r] = virtual(state)(e)
--- a/fjs/cas/module.f.mjs
+++ b/fjs/cas/module.f.mjs
@@ -5,7 +5,7 @@
  *
  * @import { Sha2, State as Sha2State } from '../crypto/sha2/types.ts'
  * @import { Vec } from '../types/bit_vec/types.ts'
- * @import { RawEffect, Operation } from '../effects/types.ts'
+ * @import { Operation } from '../effects/types.ts'
  * @import { Effect, NotImplemented } from '../effects/io/types.ts'
  * @import {
  *  IoError,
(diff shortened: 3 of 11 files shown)


o2alexanderfedin (approved):
Approving. The risk in this one is that `E` defaults to `NotImplemented`, so a site whose old spelling named a different error type could silently lose it while the source diff still looks like pure respelling. It didn't happen anywhere.

I checked that by machine rather than by eye: 15 `Assert<Equal<headSignature, baseSpelling>>` assertions over every converted exported signature — `collectRead`, `casAddFile`, `Cas.write`, `Cas.list`, `all`, `both`, `writeFromStream`, `createServer`, `_WriteLoop`, `Console`, memory `create`/`read`/`write`, `detectStream` — compile at 0 errors, and flipping one to `IoResult<void>` produces `TS2344`, so the harness discriminates. The pattern holds throughout: `RawEffect<O, OpResult<T>>` → `Effect<O, T>` (taking the default, which is what `OpResult` already meant) and `RawEffect<O, IoResult<T>>` → `Effect<O, T, NotImplemented | IoError>` — always explicit, never the default. The one generic-`E` site, `virtualOk` at `cas/evo/proof.f.mjs:80`, keeps `<T, E>`.

The declaration diff across 356 emitted files touches 10, and every hunk is cosmetic alias respelling plus import churn. Every `O` union is character-for-character identical, so **no operation set widened** — no consumer runner has to handle a new op, which is the failure mode that bit an earlier PR in this series.

I skipped mutation testing, and the justification is stronger than "no branch moved": stripping comments from each of the seven touched `.f.mjs` files gives `SAME` against base for all seven, and the non-comment `*.f.mjs` diff is empty. Not a byte of executable code changed. The `.ts` changes are type-only and erased. For the same reason a path-leak regression isn't possible here, though I confirmed it anyway across four MCP sessions — including `chmod -R 500`, where `cas_add` answers `{"text":"write","isError":true}` with no path, identical on both sides.

`fjs/effects/io/types.ts`'s definition is untouched — `Effect<O, T, E = NotImplemented>` at `:91` is identical base and head, and stripping comments from the file gives a zero diff. Its +30 is the new "not every `Result`-valued raw effect is an `Effect`" paragraph plus dropping two stale stage-3/4 claims.

The body's "nine sites left raw" is exact: 4 in `cas/evo` (all `Result<…, string>`) and 5 in `djs/transpiler` (all `Result<…, ParseError>`), none carrying `NotImplemented` — so the rule the doc now states holds tree-wide with no exceptions.

Verified otherwise: `tsc` 0 both sides, tests 2971/0 on each with the full test-name sets diffing empty across all 2971 keys — none added, removed or renamed, and the `cas` proof hunks are `@type` annotations over byte-identical bodies. `cov` 100/100/100, `prepack` and `ci-update` clean. CLI differential identical apart from the worktree path inside a `Cannot find module` message. Changelog byte-identical to the body's section and conforming.

PR #1641, opened 2026-08-18
effects: stages 1–4 of retiring RawEffect, and the plan itself

sergey-shandar:
Stages **1 through 4** of a plan this PR also writes down, as `fjs/effects/todo/retire-raw-effect.md`, plus the one design decision that plan left open.

#1640 settled the rule — `Effect` against `RawEffect` is **composition against representation**, and the channel is where short-circuiting lives — but left the work it implies unwritten.

## Stage 1: `IoChannel`

`NotImplemented | IoError` was spelled out 51 times across 50 lines, 41 of them as an `Effect` channel. It now has a name beside `IoError` and `IoResult`, and `IoResult<T>` is `Result<T, IoChannel>`. Same type throughout.

The point is not brevity. An effect that does no IO *yet* is one added `readFile` away from doing some, and where each signature names its own errors that change walks up every enclosing signature — the cascade that sank `throws` clauses elsewhere, until engineers declared everything throwing rather than maintain it. This is that concession made deliberately, in one place. Not a ceiling: an operation with failures of its own still extends the channel.

## Stage 2: `unwrapStep` names what it panics on

It was generic in `E`, so it compiled however far a channel widened: one fallible read added upstream enlarged what every downstream call crashed on, with no diagnostic. Its doc offered greppability as the safeguard — but grepping finds sites, not sites whose *scope* grew since they were read. That is the `nothrow` defect in the one place where the consequence is a crashing program rather than a compile error.

It now takes a `summary: (e: E) => string`. A renderer for one channel cannot accept a wider one, so widening breaks at the site that chose to panic. `_UnwrapStepPinsItsChannel` in `effects/io/types.ts` fails if the argument ever goes back to accepting anything.

## Stage 3: a `Program`'s exit code is a `Result`

`RawEffect<O, number>` could not say which numbers meant failure, so nothing short-circuited on one.

`Program<O> = (options) => Effect<O, 0, number>`. The success type is the literal `0`, which is what makes `exitCode` a projection instead of a branch: `r[1]` is the code on either side. `errorExit` is `Effect<Write, never, number>`, so no value can be invented for *its* success branch.

## Stage 4: a stream fails at the cell, not at the item

`List<O, T> = RawEffect<O, Next<O, T>>` put the failure in `T`, so nine consumers hand-wrote the same `if (t === 'error') return …`. It could not be factored: a combinator generic in `T` cannot look inside it. `List<O, T, E> = Effect<O, Next<O, T, E>, E>` — `step` does it, and searching for that shape now returns nothing.

It also removed a constructible bad state: a failing producer owed a tail nobody would pull, and `Cas.read`'s doc had to *legislate* that an error item is never end-of-stream. Those are different shapes now, so the rule enforces itself.

## The open decision: `cas/evo` flattens

`Evo.add` and `Evo.revision` nested a `Result` inside the effect's value, arguing that folding it into the channel would report "invalid parent hash" the way it reports a runner that cannot dispatch `memWrite`.

The consumers do not support that. Both branches of the MCP handler produced an `isError` result; the only thing the layering decided was **which renderer ran**, and a union decides that just as well. `EvoError` is a tagged tuple beside `NotImplemented` in one `EvoChannel`. Nothing continues past a rejection to run further effects, which is the one behaviour nesting bought — while `resolveParents` carried two hand-written `[0] === 'error'` tests doing what the Io `foldStep` does for free.

## Three review rounds

Every substantive finding was correct. The ones worth naming here:

- **Two comments claimed a protection the compiler does not give**, and the reviewer showed it by compiling. `() => pure(0)` no longer fits a `Program`, but `() => pureOk(0)` does — so `fjs/module.f.mjs`'s comment guards a mistake that is still available, and says so. `errorExit`'s `T = never` gives a continuation nobody reaches, not a compile error.
- **One sentence lived in four files.** "Genuinely cannot fail — a `List` cell, a `Program`'s exit code…" was removed from `effects/io/types.ts`, then found again in `fjs/AGENTS.md`, then again in `effects/io/README.md`. Three rounds, three copies: grepping for a claim is not grepping for a name.
- **`io-effect-migration.md` recorded the opposite decision as `[x]`** — keep `RawEffect` public because "`List`, `Program`, `Step` and `ToolEntry.handle` … none of them can fail". Two of the four have channels now. That file keeps its decision and gains what changed, rather than being rewritten to look prescient.
- **I invented a merge race.** I had described one commit as work #1640 merged out from under. It was authored eighteen minutes *after* that merge, on an already-merged branch — never in a PR, never reviewed. It arrives here as new content, and the changelog no longer narrates the race.
- **Counts corrected**: 11 `unwrapStep` sites (10 proofs), not 14/13. 51 occurrences across 50 lines, not 50. Stage 4 was #1615/#1629/#1630/#1632/#1633 — 44 files as a union, +1250/−483; my figures used the wrong PR set and summed per-PR file counts, which double-counts.

Also closed: the last spelled-out union outside a doc (`_WriteLoop` returned it while taking `IoChannel` on the same line), six sites documenting the error-item model, a test named for a payload that is now a stream failure, and three todos resting on that model.

## What became of stages 5–7

Continued in #1642, stacked on this:

- **Stage 5** was nearly complete already. Most remaining raw sites are mock runners in proofs (the representation layer) or **absorb points**, where a module converts a channel into its own vocabulary — `protocol/mcp` most clearly, since the protocol *is* its error channel.
- **Stage 6** split. Requiring a `Result` from every `Operation` landed, latching the invariant that a runner may decline any command. **Deleting `RawEffect` did not**: ~69 payloads are permanently infallible, so removing the alias would wrap them in an `ok` nobody reads. That conclusion lives in `RawEffect`'s doc now.
- **Stage 7** deleted the plan file.

## Validation

Changelog:
- `effects/node`: `IoChannel` names the standard IO error channel
  (`NotImplemented | IoError`), which was spelled out 51 times across 50 lines. `IoResult<T>`
  is now `Result<T, IoChannel>`. Same type throughout — an IO-touching effect
  says it fails the way node IO fails, so gaining a new way to do so no longer
  walks up every enclosing signature.
- **BREAKING CHANGES:** `cas/evo`: `Evo.add` and `Evo.revision` no longer nest a
  domain `Result` inside the effect's value. Both answer
  `Effect<…, T, EvoChannel>`, where `EvoChannel` is the new tagged `EvoError`
  or `NotImplemented`. A rejected revision now short-circuits `step` like any
  other failure, and `evoSummary` renders either case.
- **BREAKING CHANGES:** `effects/io`: `unwrapStep` takes a second argument, a
  `summary` for the channel it panics on. It was generic in the error type, so
  it compiled however far a channel widened and quietly enlarged what it
  crashed on; a renderer written for one channel does not accept a wider one,
  so that is now a compile error at the site that chose to panic. `errorSummary`
  is the renderer for the node channel.
- **BREAKING CHANGES:** `effects/node`: a `Program` answers
  `Effect<O, 0, number>` rather than `RawEffect<O, number>` — `ok(0)` for
  success, `error(n)` for failure. A bare number could not say which codes meant
  failure, so nothing short-circuited on one and a continuation that dropped a
  failure and answered `0` type-checked. `errorExit` is
  `Effect<Write, never, number>`; the new `exitCode` reads the code from either
  branch.
- **BREAKING CHANGES:** `effects/list`: a `List` cell can fail.
  `List<O, T, E> = Effect<O, Next<O, T, E>, E>`, so a producer that cannot
  deliver fails the stream instead of yielding an error *item* followed by a
  tail nobody pulls. `Cas.read` is `(hash) => List<O, Vec, IoChannel>`. Six
  consumers hand-rolled the same short-circuit because the failure sat in `T`,
  where a generic combinator could not see it; `step` does it now.
- `djs/transpiler`: the five `ParseError`-valued signatures say
  `Effect<ReadFile, T, ParseError>` rather than the equivalent
  `RawEffect<ReadFile, Result<T, ParseError>>`. The same type, spelled as what
  it means: a parse failure is a failure, so it belongs in the channel.


o2alexanderfedin (commented):
The `cas/evo` change is behaviourally sound — I couldn't make it drop work, swallow an error, or leak a path. What's left is documentation that still describes the API it breaks.

Findings below are measured at `6ede4944` unless noted; I re-checked each against `9a7db457`, which landed while I was reviewing and whose own content (`unwrapStep names the errors it panics on`) I haven't reviewed yet.

**`fjs/effects/io/types.ts:62-70` still argues against what this commit does, and cites the example it just deleted.** Confirmed present at `9a7db457`:

> It is ordinary returned data — leave it `RawEffect` — when `E` is a domain verdict … : `evo.revision`'s `Result<RevisionData, string>` … Collapsing the second kind into the channel would report "invalid parent hash" the way it reports a runner that cannot dispatch `memWrite`;

`evo.revision`'s `Result<RevisionData, string>` no longer exists, and "invalid parent hash" now does travel the channel beside `notImplemented`. This is the published `types.d.ts` for the whole effect API — either the rule or its worked example needs to move in this PR.

**`fjs/cas/evo/README.md:26-27` documents the pre-change signatures**, also confirmed at `9a7db457`:

```ts
add: (rev: RevisionData) => Effect<O | MemOp, Result<Hash, string>>
revision: (hash: Hash) => Effect<O | MemOp, Result<RevisionData, string>>
```

Its "Failure reporting" section still says the message "comes back as `error(message)` … so a transport can surface it to the caller directly" — the payload is an `EvoError` tuple now and the transport renders it through `evoSummary`, which is exactly what you changed `fjs/mcp/evo/module.f.mjs` to do. `AGENTS.md:210` puts architectural choices in the module README, and this is the one doc describing what the BREAKING entry breaks.

**`fjs/cas/evo/proof.f.mjs:76-80`** — in the file this commit rewrote, `virtualOk`'s JSDoc still says "`evo.add`'s two `Result`s are exactly that split: the outer one is unwrapped here, the inner domain verdict is what each proof goes on to inspect." One `Result` now, and the diff deleted every assertion that inspected an inner verdict.

**`retire-raw-effect.md` still never acknowledges #1636**, confirmed at `9a7db457`: no occurrence of `1636`, `public`, `load-bearing` or `supersede`. Its title remains the direct negation of `io-effect-migration.md:434`'s `[x] … **superseded**` and `:436-438`'s `[x] Keep the Pure | Do representation … **public rather than internal**, because List, Program, Step and ToolEntry.handle are all built on it`.

**Several of the doc's measurements don't reproduce.** Recomputed at `6ede4944`: "all 27 operations" is 28, and `_AnyOp` isn't the only non-`Result` exception (`effects/proof.f.mjs:27`'s `_AddOp`). "40 across 23 files" — the 40 holds, the file count is 18. "five PRs, 46 files, +962/−555" — five PRs right, but 44 files / +1250/−483. Stage 1's "all 50 spelled-out unions collapsed onto it" — 51 existed, 50 collapsed, and `effects/node/types.ts:173`'s `_WriteLoop` still spells the union out 106 lines below the new alias. "The other 13 sites are proofs" — 11 `unwrapStep` call sites, 10 of them proofs. "all 45 occurrences" of effect-`List` — 51, and this commit added one. The status line still says "two decisions still open" while the section now holds two `[x]` and no `[ ]`.

Also minor: `effects/todo/step-continuation-operation-union.md:21` lists `cas/evo/module.f.mjs:456` among "six call sites"; this commit deletes it, leaving five.

Good news on the parts that mattered most. Five of six mutants died, including making the domain rejection answer `ok` (2962/9), removing the short-circuit from `resolveParents` (2935/36), and letting the `IoError` ride the channel untranslated (2970/1). The sixth survives at 2971/0 and 100/100/100 — `evoSummary = e => e[1]`, bypassing `errorSummary` on the `NotImplemented` arm — but it's **inherited, not introduced**: the covering proof at `mcp/evo/proof.f.mjs:72` asserts `includes('memRead')`, which the raw payload satisfies, and it's byte-identical on base.

No caller loses work. Both production callers behave identically on rejection, and where the raw combinators previously "continued", the continuation only re-tested and returned — the operation traces are identical. `IoError extends EvoChannel` is `false`, so no host error can reach `evoSummary` by type, which makes the `catchStep` translation in `addRevision` the single funnel. MCP sessions byte-identical head vs base with no path in any response, including `chmod 555` and `chmod 000` stores. Proof keys identical at 2924 both sides, and the assertions got stronger — `assertEvoError` pins the tag where the old pairs didn't. Gates green both sides: `tsc` 0, 2971/0, `cov` 100/100/100, `prepack` and `ci-update` clean. The body's `Changelog:` section was missing the entire BREAKING bullet at `6ede4944`; `9a7db457` fixed that and it's byte-identical now.

o2alexanderfedin (commented):
Behaviour is clean across all four changes — six mutants, six dead; every exit code, CLI transcript and MCP response byte-identical to base; no operation union widened; no path on a wire. What's left is claims. Two of them are comments that promise a protection the compiler doesn't give, which is the kind that costs someone the guard later.

**`fjs/module.f.mjs:40-44` says a mistake is no longer available. It is.**

> `Program`'s exit code is a `Result` now, so a chain that drops the failure and answers `ok(0)` no longer type-checks — this comment records why the shape is what it is, rather than guarding a mistake that is still available.

At head, with generated `.d.mts` files cleared:

```js
handler: ({ home }) => step(casMcpServer(home), () => pureOk(0))   // compiles, 0 errors
handler: ({ home }) => step(casMcpServer(home), () => pure(0))     // TS2322
```

`step(e, () => pureOk(0))` is `RawEffect<O, Ok<0>>`, and `Ok<0>` is assignable to `Result<0, number>`. Control: at base the `pure(0)` form compiles clean, so the harness discriminates and the commit really did close the bare-number spelling. But the exact spelling the comment names as dead is the one that lives, and the last clause invites deleting the `exitStep`. Your commit message and `retire-raw-effect.md:125` both say `pure(0)` and are correct — only this comment respells it.

**`fjs/effects/node/module.f.mjs:431-433`** — "It never succeeds, and the type says so … a chain that continues past this is a compile error rather than a continuation nobody reaches." All three compile at head:

```js
ioStep(errorExit('boom'), v => pureOk(v))
ioStep(errorExit('boom'), () => log('after'))
rawStep(errorExit('boom'), r => pure(r))
```

`T = never` makes the parameter `never`, which is exactly "a continuation nobody reaches". The signature is still an improvement; the sentence claims more than it buys.

**`e9c1ef4d` leaves six sites documenting the error-item model it removed**, two inside blocks it edited. `fjs/cas/module.f.mjs:314` still says a `ListEffect` of "`ok` items" — contradicted by the two lines this commit added just below it ("A failed read fails the stream. It used to be yielded as an error *item*"). `fjs/media/type/module.f.mjs:252-253` still says "A read `error` item short-circuits into the `IoResult` error", in the JSDoc block whose `@param`/`@type` lines this commit rewrote. Also `media/type/README.md:57`, `cas/proof.f.mjs:304-306` (test still named `casWriteErrorItemAborts`, payload now `pureError`), `cas/evo/proof.f.mjs:47`, `cas/cli/proof.f.mjs:56-57`.

**The #1636 contradiction is now a factual error, not just an unacknowledged reversal.** `io-effect-migration.md:436-438` records as decided and `[x]`: keep `RawEffect` public "because `List`, `Program`, `Step` and `ToolEntry.handle` are all built on it **and none of them can fail**". `Program` gained a channel in `28919999`; `List` gains one in `e9c1ef4d`. Both stated reasons are false at head, and that file is untouched by this PR. `retire-raw-effect.md` has one `1636` hit (the rename credit at `:218`) and zero for `public`/`supersede`/`load-bearing`, so the two plan files still assert opposites.

**`effects/node/types.ts:173`** — `e9c1ef4d` edited this exact line to take `List<O, Vec, IoChannel>` and left the return spelled out as `NotImplemented | IoError`, which is the one real leftover behind "all 50 spelled-out unions collapsed onto it" (`retire-raw-effect.md:100`). `77d74c53` changed `:70` to 51 but left `:91` and `:100` at 50.

Counts still not reproducing, all re-measured at `e9c1ef4d`:

- `:4` "two decisions still open" — both boxes (`:78`, `:86`) are `[x]`; zero `[ ]`.
- `:122` "The other 13 sites are proofs" — 11 `unwrapStep` sites, 10 of them proofs. Measured at `b55c7366`, `08b6e3a6`, `9a7db457`, `28919999` and `e9c1ef4d`; never 13 at any of them.
- `:65-66` "40 across 23 files" — eight different countings at the plan's base give 34/18, 38/18, 42/19, 53, 57/20, 78/24. None is 40/23.
- `:44-45` "46 files, +962/−555" — five PRs is right (#1615, #1629, #1630, #1632, #1633); the union is 44 files, +1250/−483.
- `:222-223` "stage 3 would subsume" — the `List` stage was renumbered 3→4 in `ef524e28`. Now that stage 4 is `[x]`, neither todo it names was closed: `fold-stream-combinator.md:12` still describes folding `List<O, IoResult<Vec>>`, a type that no longer exists.
- `step-continuation-operation-union.md:21` still lists `cas/evo/module.f.mjs:456` / `Result<Hash, string>` among "six call sites"; `6ede4944` removed that cast and the line is an object-literal field now.

The counts that **do** reproduce exactly: 24 importing files, 51 occurrences across 50 lines (the doubled one is `cas/module.f.mjs:207`), 41 as a channel, 45 effect-`List` occurrences, 28 operations. Both corrections in `77d74c53` are right.

On the lost commit — the recovery is complete and main is not inconsistent: `changelog/unreleased/1640.md` on main never claimed `djs/transpiler`, and `prepack` at `b55c7366` exits 0, so the loss is inert. One framing correction, though. `b4556592` was authored 01:49Z, thirteen minutes *after* #1640 merged at 01:36Z, and `refs/pull/1640/head` is still `c0eb1d89`. So it isn't recovered reviewed work — it never was in #1640 and no reviewer saw it. Arriving here is right; it's new content, not restored content.

Minor: `changelog/unreleased/1641.md` and the body's section are the same 2311 bytes but not byte-identical — the `djs/transpiler` item is 5th in the file and 6th in the body. And that item names a PR number and explains a merge race, which reads as process history rather than a release note.

One gap worth knowing about rather than fixing here: `exitCode(await runNodeEffect(…))` at `effects/node/module.mjs:385` is the single place a `Result` becomes a process exit code, it sits outside the `**/module.f.mjs` coverage glob, and no proof reaches it — a mutant that drops the code survives both `fjs t` and `cov` at 100/100/100. I covered it empirically instead: 33 in-tree CLI cases plus the four `fjs r` module shapes all map correctly. Related, and pre-existing on base: `fjs t` derives its own exit status from the code under test, so mutating `exitCode` or `exitCodeStep` prints failures while still exiting 0 — `npm test` would go green with 35 failing tests. `npm run cov` is what actually kills those.

Verified green at `e9c1ef4d`: `tsc` 0, 2971 pass / 0 fail, `cov` 100/100/100, `prepack` 0, `ci-update` clean. Proof keys identical base vs head across 1871 unique names — nothing added, removed or renamed — and `effects/node/proof.f.mjs:126-138` is strengthened, asserting both tag and code without going through `exitCode`. Mutants killed: publishing a mid-stream failure (2969/2), `collectRead` swallowing a stream failure (2966/5), a missing shard reading as an empty blob (2963/8), `writeLoop` truncating on failure (2970/1), `detectStream` classifying a failed read (2968/3), `cas add` of a missing file succeeding (2970/1). `_UnwrapStepPinsItsChannel` is not vacuous — rewriting its parameter to `(e: unknown) => string` fires `TS2344`.

o2alexanderfedin (commented):
Both comment fixes are right, and I probed them rather than reading them — every probe on a tree with the generated `.d.mts` files deleted, with `pure(0)` as the negative control to prove `tsc` was reading source:

- `step(casMcpServer(home), () => pure(0))` → `TS2322: Type 'number' is not assignable to type 'Result<0, number>'`
- `step(casMcpServer(home), () => pureOk(0))` → 0 errors

which is exactly what `fjs/module.f.mjs:43-48` now says. `errorExit`'s three clauses check out too: a continuation past it compiles, `step(errorExit('boom'), t => pureOk(t))` is `Effect<Write, never, number>` so the `never` parameter can't yield a value, and driving it at runtime never fires the continuation (`["error",1]`). One edge the sentence survives but readers might trip on: at the *composition* level a success type can be invented — `step(errorExit(s), () => pureOk(0))` type-checks as `Effect<Write, 0, number>`. Runtime never delivers it, and you scope "the success branch" to `errorExit`'s own `T`, so the claim stands.

Also confirmed: this commit changes no executable code. All six `.f.mjs` files comment-stripped and diffed against `e9c1ef4d` are byte-identical, no proof renamed, no assertion weakened.

**The title undersells it — there's a fourth file.** `fjs/effects/io/README.md:7` still reads "`RawEffect<O, T>` … stays public for the computations that genuinely cannot fail" — the same sentence you deleted from `io/types.ts` in this PR and from `fjs/AGENTS.md` in this commit. At head it contradicts `io/types.ts:50-51` ("the division between them is not 'fallible' against 'infallible' — it is *composition* against *representation*") and your new `fjs/AGENTS.md:769-771`. That README is untouched by the entire PR. Weaker second instance at `fjs/effects/module.f.mjs:17` ("the ones here stay for the infallible compositions").

**`io-effect-migration.md` is now self-contradicting against its own sibling.** `:411-412` "the two names now divide by meaning: `Effect` can fail, `RawEffect` cannot", and `:435-438`'s recorded reason for keeping `RawEffect` public — "`List`, `Program`, `Step` and `ToolEntry.handle` are all built on it and **none of them can fail**". At head `List<O, T, E> = Effect<O, Next<O, T, E>, E>` and `Program<O> = (options) => Effect<O, 0, number>`; two of the four are `Effect`s with channels. And `:409-410` ("giving those a channel would put an `ok(…)` wrapper on every stream cell … for a failure that cannot happen") is answered head-on by `retire-raw-effect.md:73` ("There is no infallible stream, so a fallible cell costs no wrapper anywhere"). `retire-raw-effect.md:218-221` links this file as Related, so a reader is actively sent to the wrong version.

**`fjs/media/type` still documents the removed model, in a block this PR edited.** `module.f.mjs:252-253` says "A read `error` item short-circuits into the `IoResult` error" directly above the signature the PR rewrote to `@param {List<O, Vec, IoChannel>}`; the body at `:262-268` is a two-case loop with no item inspection. And `README.md:57-58` documents types that no longer exist — `detectStream(stream) // Effect<O, IoResult<DetectMeta>>`, where head is `(stream: List<O, Vec, IoChannel>) => Effect<O, DetectMeta, IoChannel>`. As spelled, `Effect<O, IoResult<DetectMeta>>` would mean a doubly-wrapped `RawEffect<O, Result<Result<DetectMeta, IoChannel>, NotImplemented>>`.

**`effects/node/types.ts:173`** still returns `NotImplemented | IoError` while the *same line* takes `List<O, Vec, IoChannel>` — which reads as an oversight rather than a choice, and it's the last thing standing between `retire-raw-effect.md:101` and being true. Measured at head: exactly two spelled-out occurrences outside `.md` remain — this one and prose at `io/types.ts:81`.

**`retire-raw-effect.md:4` "two decisions still open"** — the section at `:77` holds two items, both `[x]`, zero `[ ]`.

**The changelog gap widened rather than closed.** File is 2324 bytes, body's section 2311. Item 1 now differs in *text*: the file says "spelled out 51 times across 50 lines", the body still says "at 50 sites" — this commit edited only the file. Items 5 and 6 are still swapped between the two. Since the body becomes the squash message, they need to match.

Minor, introduced here: `fjs/cas/cli/proof.f.mjs:56` reads "comes back as **an** stream failure" — leftover `an` from "an error item".

Withdrawing one from my last review: I flagged "40 across 23 files". Re-measuring gives 34–41 occurrences across 18–24 files depending on whether `@import` lines count, so both 23 and 24 sit inside the spread and I can't call your correction wrong.

Closed and verified: `cas/module.f.mjs:314`, `cas/evo/proof.f.mjs:47`, `cas/cli/proof.f.mjs:56` (modulo the typo); the surviving "error item" mentions in `cas/types.ts:31`, `effects/list/types.ts:39` and `retire-raw-effect.md:150` are all correct past-tense framing. `unwrapStep` is 11 sites, 10 proofs — your new text is exact. 51 occurrences across 50 lines at the PR base reproduces exactly. Both `fold-stream-combinator.md` and `effect-list-fold.md` now carry accurate notes; I checked all four consumers the first names and each did lose the error-item branch. `fjs/AGENTS.md`'s factual claims all hold — the `E = NotImplemented` default, `RawEffect` as `Pure | Do`, the `match`/`runPure`/`do_` locations, and the MCP absorb point — with no contradiction against root `AGENTS.md` or `CONTRIBUTING.md`.

Gates at `706ab7f3`: `tsc` 0, 2971 pass / 0 fail, `cov` 100/100/100, `prepack` 0, `ci-update` clean.

o2alexanderfedin (commented):
Everything else I raised is closed and verified against the code. One new claim came in wrong, though, and it's the correction itself.

**`step-continuation-operation-union.md:18-22` — "Five call sites, not six" is also false; it's three.** The `cas/evo` row you removed was right to go. But two more rows were already dead before this PR, and the same measurement that caught `evo` catches them:

| Row | Status at `2840de56` |
| --- | --- |
| `fjs/cas/module.f.mjs:348` | the file is 346 lines. That cast lived in `casUpload` and went in `fafa4b85` (#1601) |
| `fjs/effects/proof.f.mjs:85` | the cast wrapped `okStep`, retired in `7790210b`; `grep -rn okStep fjs/` gives 0 code hits |
| `fjs/djs/module.f.mjs:41` | alive, at `:53`, now `Effect<_CompileOp, 0, number>` |
| `fjs/djs/transpiler/module.f.mjs:103` | alive, at `:96`, now `Effect<ReadFile, Unknown, ParseError>` |
| `fjs/emergent_testing/proof.f.mjs:448` | alive, at `:488`, now `Effect<_RegisterMockOps \| Readdir \| Import, 0, number>` |

Three surviving casts against five claimed. The two dead rows aren't this PR's doing — they were stale at the base — but the blockquote is a new positive count, so it inherits the problem rather than fixing it. The three live rows' line numbers were already stale; their *spellings* went stale in this PR, which added the third type parameter. Also `:15` and `:45` still say "six call sites" and "the six sites" underneath the correction, and `:34` still cites `okStep` as an existing wrapper.

Everything else checks out, and I verified the replacements against the code rather than just confirming the old text was gone:

- `io/README.md:6-9` and `effects/module.f.mjs:17-18` now state composition-against-representation, consistent with `io/types.ts:50-51`, `fjs/AGENTS.md:768-769` and `effects/types.ts:16-20`. The supporting claim holds: `list/types.ts:51` is `List<O,T,E> = Effect<O, Next<O,T,E>, E>`, `node/types.ts:396` is `Program<O> = (options) => Effect<O, 0, number>`.
- `io-effect-migration.md:415-426` is accurate on all four names — `List` and `Program` carry channels, `Step` and `ToolEntry.handle` are still raw for the reason given — and the `ok(…)`-wrapper contradiction with `retire-raw-effect.md:75` is resolved in place rather than deleted, which is the right call for a historical doc.
- `media/type/module.f.mjs:252-254` matches the real signature and the two-case loop at `:262-268`; `media/type/README.md:57-58` now reads `Effect<O, DetectMeta, IoChannel>`.
- `effects/node/types.ts:173` is the `_WriteLoop` fix and it is a pure respelling: `Eq<IoChannel, NotImplemented | IoError>` passes while both negative controls (`Eq<IoChannel, NotImplemented>`, `Eq<IoChannel, IoError>`) error `TS2322`, so the harness discriminates. `:103`'s "all 50 collapsed" is now true — base had 51 occurrences across 50 non-`.md` lines, head has 2, both non-annotations.
- `retire-raw-effect.md:4-5` fixed, and the new stat reproduces exactly: #1615+#1629+#1630+#1632+#1633 = +1250/−483 across 44 unique files. Residual: the heading at `:79` still reads "Decisions still open" above two `[x]`, though the Status line now corrects it.
- The merge-race story is gone tree-wide (2 grep hits, both unrelated). One detail off in the commit message only: `05ca7ce1` says "18 minutes after the merge at 01:31Z", which is `b55c7366`'s commit timestamp; `merged_at` is 01:36:14Z, so 13. Same conclusion either way.

Code: three of the four `.f.mjs` are comment-only; `cas/proof.f.mjs` has one executable change, the `casWriteErrorItemAborts` → `casWriteStreamFailureAborts` rename with a byte-identical body. I mutated the path that test guards anyway — `cas/module.f.mjs:210`, `fail(curPath, r[1])` → `publish(state, offset, curPath)`, no new function or branch — and it dies 2969/2 on `casWriteStreamFailureAborts` and `mainAddMissingFile`. Prepack declaration diff is 3 hunks, all JSDoc prose plus that rename; no exported signature changed, no `O` union widened.

Changelog is now byte-identical to the body's section — same md5, 33 lines, 6 items, same order — so both divergences I raised are closed, and item 1's "51 times across 50 lines" measures correctly at base.

Gates at `2840de56`: `tsc` 0, 2971 pass / 0 fail, `cov` 100/100/100, `prepack` 0, `ci-update` clean.

PR #1642, opened 2026-08-18
effects: finish the io-effect migration — stage 5 onward, and where it stops

sergey-shandar:
**Stacked on #1641.** Until that merges the diff shows its stages too.

> ⚠️ **The two approvals here predate most of this PR.** They cover `b9a70653`/`51929923` (the transpiler) and `2bfe25d3` (a merge). Since then it has gained two core changes — `effects/io`'s fold signature and `Operation`'s return constraint — plus the conclusion that `RawEffect` should *not* be deleted, and a design-doc sweep. Those are unreviewed.

This carries `retire-raw-effect.md`'s stages 5–7 to the end, and then its sibling's last open item.

## `djs/transpiler`: the first parse error is the channel

`ParseContext.error` held "the first parse error encountered, or `null` while everything is clean" — a hand-rolled error channel. Two places set it and three tested it before doing any work:

```js
if (context.error !== null) { return pure(context) }                        // foldNextModuleOp
if (contextWithImports.error !== null) { return pure(contextWithImports) }  // transpileWithImports
if (context.error !== null) { return pure(error(context.error)) }           // transpileModule
```

Those three vanish — `step` does them, so a `ParseContext` that exists is one that is still good. Two more conditionals go for different reasons: `transpileWithImports`'s `parseModuleResult[0] === 'ok'` wrapper, because `step` only calls it on success, and the two `readUtf8File` error branches, which collapse into one shared `notFound`. (Six conditionals in total; a reviewer rightly flagged "5 error tests" as ambiguous, since one of the six tests ok-polarity.)

`notFound` is where `IoChannel` stops: however a read failed — missing file, unreadable, a runner without `readFile` — a transpiler's answer is the same `ParseError`, so translating once at the edge keeps `ParseError` the whole channel behind it. That collapse is not new; base did the same thing at both call sites.

## Naming intent where both branches matter

`compile`, `ci` and `run` said raw `step` while inspecting both branches. They now say `resultStep` — the same function, but the name states it and the narrower type checks it. The two MCP tool registries look identical and are not: their handlers leave the Io layer to produce a `ToolsCallResult`, and `resultStep` rejects them. They keep raw `step`.

## The Io fold takes an `Effect`

`foldStep`/`forEachStep` took a raw effect for `items`, documented as: a held list has no error channel, so requiring one "would mean lifting at every call site to express something no consumer has". That is the argument this migration rejected everywhere else, and the cost it names is `pureOk` instead of `pure`. What it did not count is what the raw version charged the *fallible* producer:

```js
exitStep(ioStep(c.list(), hashes => forEachStep(pure(hashes), …)))   // before
exitStep(forEachStep(c.list(), …))                                   // after
```

`fjs/cas/cli` now imports nothing from the raw module — the first consumer off it entirely.

## Stage 6 split, and half of it is a "no"

**Requiring a `Result` from every `Operation` landed.** A runner may decline any command — `partialMatch` answers `error(notImplemented(command))` through the command's own *output* — so an operation whose return admitted no error was a hole in that mechanism, with nowhere to put the refusal. All 28 already complied; it is a rule and a latch now, so a new operation cannot be declared infallible, and `do_` produces an `Effect` by construction rather than by each author remembering.

**Deleting `RawEffect` did not, and the migration is what shows why.** The plan's argument was that removal is free once nothing infallible is left to wrap. Everything the plan asked for landed — and the infallible payloads did not go away. They sorted into two *permanently* infallible kinds:

- **The representation**, generic over its payload by definition — the runners, `match`, `runPure`, the raw combinators. **53 uses.** `<T>(e: RawEffect<O, T>) => …` cannot be `Result`-valued without ceasing to be generic.
- **Absorb points**, where a module converts a channel into its own vocabulary — **11 MCP protocol values**, where the protocol *is* the error channel and a notification has no frame to put an error in, plus 7 `void`.

That is 71 payloads gaining an `ok` nobody reads, to remove one alias. `RawEffect`'s doc now carries the conclusion and its evidence, where the next person to have this idea will find it. With nothing actionable left, `retire-raw-effect.md` is deleted.

## The design-doc sweep

`io-effect-migration.md`'s last open item, and the same class of defect three review rounds kept finding.

`okStep` was cited by seven documents as an existing export. It is not renamed, it is gone — inlined into the Io `step`, whose body was its only caller — so substitution was never the fix. Each citation meant the same thing, the `Result` short-circuit, and that still exists: it is what the Io `step` *is*.

`IoResult<T>` needed no substitution at all; it survived and still names an operation's return, so the strategy and staging-lease tables were right as written. What was genuinely stale is narrower: `List<O, IoResult<Vec>>` and the two-parameter `Effect<O, IoResult<T>>`, which would now mean a doubly-wrapped `Result`.

Two judgement calls: `todo/inline-type-casts.md` is left alone on its own instruction (an audit snapshot; a re-audit rather than a partial edit refreshes it), with its two moved rows noted beside the nine already recorded as moved. And `io-effect-migration.md` is marked done but **not deleted**, which its own status line used to instruct — `fjs/effects/io/README.md` and `spec/todo/io-effects.md` both cite it, so it is the record now rather than an open task.

## Validation

Changelog:
- **BREAKING CHANGES:** `djs/transpiler`: `ParseContext` no longer carries an
  `error` field. The first parse error is the effect's channel
  (`Effect<ReadFile, ParseContext, ParseError>`), so `step` short-circuits
  instead of three separate places testing `context.error` before doing any
  work, and a context that exists is a context that is still good.
- `djs`, `ci`, `fjs`: three entry points that observe both branches of a
  `Result` say `resultStep` instead of the raw `step`. Same function, but the
  name states the intent and the narrower type checks it.
- **BREAKING CHANGES:** `effects/io`: `foldStep` and `forEachStep` take an
  `Effect` for `items` rather than a raw effect, so a producer that can fail
  feeds the fold directly. Lifting a held list is `pureOk` instead of `pure` —
  the same one call — where the raw version made the fallible producer wrap the
  whole fold in a step whose only job was to unwrap and re-wrap the list.
- **BREAKING CHANGES:** `effects`: an `Operation`'s return type must be a
  `Result`. A runner may decline any command — `partialMatch` answers
  `error(notImplemented(command))` through the command's own output — so an
  operation returning something with no error case was a hole in that
  mechanism. Every operation in the tree already complied; this makes it a rule
  and a latch, so a new one cannot be declared infallible.
- `docs`: the design docs no longer cite `okStep`, an export removed when it was
  inlined into the Io `step`, nor spell types the migration replaced —
  `List<O, IoResult<Vec>>` and the two-parameter `Effect<O, IoResult<T>>`, both
  of which would now mean a doubly-wrapped `Result`. This closes the last open
  item of `io-effect-migration.md`.


o2alexanderfedin (approved):
Approving this PR's own two commits, measured against `e9c1ef4d`. My open comment on #1641 stands and this isn't a review of the stack beneath it.

The thing worth proving here is that deleting three guards doesn't let work continue past a failure, and it doesn't. I checked that by construction rather than by reading: 23 scenarios run through a recording interpreter comparing both the result **and the exact `readFile` op sequence**, base vs head — byte-identical. When an import fails to parse, ops stop at `[main.f.js, a.f.js]` and the sibling is never read; at depth 3 with a later sibling, `[main, a, b, c]` and `d.f.js` is never read; circular-then-missing stops at `[main, a]`. So the short-circuit really did move into `step` rather than going missing.

Error identity survives in both directions: first import parse-errors + second missing → the parse error; first missing + second parse-errors → `file not found`. First error wins, identically at both revisions.

Six mutants, six killed — `notFound` answering `pureOk('')` (3 fail), the circular-dependency error (1), the fold's failure swallowed in `transpileModule` (7) and in `foldNextModuleOp` (7), the deleted `transpileWithImports` guard (3), and `parseModule` swallowing the parser's own error (4). Worth noting for the record: `npm run cov` on the circular-dependency mutant printed **100.00 / 100.00 / 100.00** and still exited 1 — the coverage numbers are blind to it, the failing test is what catches it. Every one of these is caught by a real proof rather than by the gate.

I looked hard at the `notFound` collapse, since answering `file not found` for an EACCES or a runner without `readFile` would be an accuracy loss. It isn't a regression: base already collapsed both call sites the same way (`if (result[0] === 'error') return pure(error({message: 'file not found'}))`), and EACCES, a missing `readFile`, and the `.json` variants all yield `file not found` at both revisions. No `IoChannel` distinction is lost here because none was ever exposed. `transpile` is the only export, `ParseContext` has no importer outside the module, and neither is reachable from an MCP tool.

Differentials byte-for-byte across all three revisions: 18 CLI invocations incl. `c` and `r` failure modes (2724 bytes identical), `fjs i` in a scratch tree covering both branches of the `resultStep`-swapped `ci` (21855 bytes), and a real stdio MCP session (4377 bytes, no host path on the wire). Declarations: `types.d.ts` loses exactly `readonly error: ParseError | null` and its now-unused import, nothing else; `transpile` still emits `(path: string) => Effect<ReadFile, Unknown, ParseError>`, so **`O` did not widen**. The second commit changes no declaration file at all.

I checked the second commit's claim instead of taking it — swapping raw `step` for `resultStep` in the two MCP tool registries gives 88 lines of `tsc` errors (`Result<any,any>` not assignable to `ToolsCallResult`'s `StructTs<…>`), so "`resultStep` rejects them" holds.

Gates at `51929923`: `tsc` 0, 2971 pass / 0 fail, `cov` 100/100/100, `prepack` 0, `ci-update` clean — same five, same numbers at `b9a70653`. Changelog byte-identical to the body's section. Net −3 lines checks out (58 added / 61 deleted across the two source files), as do "two places set it" and "three tested `context.error`".

One arguable count: "5 error tests" — the commit deletes six conditionals (base `:42, :50, :59, :76, :98, :119`). If `:50`'s `parseModuleResult[0] === 'ok'` is deliberately excluded as ok-polarity, five is right; otherwise six.

o2alexanderfedin (approved):
Re-approving after the merge. This head only brings in #1641's latest commit, and I checked the one place that could have gone wrong rather than assuming a merge is a no-op.

Both branches had edited `fjs/module.f.mjs`: #1642 renamed `step` to `resultStep`, #1641 rewrote the `mcp` handler comment. The resolution keeps both — #1641's second paragraph is byte-identical, and its first paragraph carries #1642's rename applied to the one line that names the function. Nothing dropped, nothing duplicated.

The rewritten comment is still true at this head, probed on a worktree with zero generated `.d.mts` files present:

- `resultStep(casMcpServer(home), () => pure(0))` → `TS2322: Type 'number' is not assignable to type 'Result<0, number>'`
- `resultStep(casMcpServer(home), () => pureOk(0))` → 0 errors

which is exactly what it now claims.

Everything I approved at `51929923` is intact: all six of the other files are byte-identical, and the only movement in `retire-raw-effect.md` is a hunk-header offset from #1641's two added lines upstream. From the other direction, `51929923..2bfe25d3` touches exactly #1641's 13 files and nothing else — no evil merge.

Gates at `2bfe25d3`: `tsc` 0, 2971 pass / 0 fail, `cov` 100/100/100, `prepack` 0, `ci-update` clean. CLI differential over 13 invocations is byte-identical apart from one stack-trace line number shifting `:460` → `:463`, which is #1641's +3-line edit to that file. Changelog still byte-identical to the body's section, same order.

o2alexanderfedin (commented):
The `Operation` constraint is the right call and it genuinely bites — I probed it rather than reading it, on a tree with the generated `.d.mts` deleted and with a wrong control (`const x: string = 42` → `TS2322`) to prove the harness discriminates. A `(x: number) => number` operation gives `TS2344: Type 'BadOp' does not satisfy the constraint 'Operation'`. Across ten candidate return types it rejects `number`, `unknown`, `void` and `readonly[string, unknown]`, and accepts `Result`, `Ok`, `Error` and `readonly['ok'|'error', number]`. No structural escape, so "it is also a latch" holds.

Three things are still wrong, all claims rather than code. Findings measured at `f6222eb7`.

**`changelog/unreleased/1642.md:18` and `fjs/effects/types.ts:18` are disproven by this PR's own diff.** Both say every operation already complied — the entry as "Every operation in the tree already complied", the JSDoc as "already returned `OpResult<…>` or `IoResult<…>`". Of the 32 operation declarations at head, two did not: `fjs/effects/proof.f.mjs`'s `_AddOp` and `_AnyOp` returned bare `number`, and this PR rewrites them:

```
-/** @typedef {readonly['add', (a: number, b: number) => number]} _AddOp */
+ * @typedef {readonly['add', (a: number, b: number) => Ok<number>]} _AddOp
- * @typedef {readonly[string, (a: number) => number]} _AnyOp
+ * @typedef {readonly[string, (a: number) => Ok<number>]} _AnyOp
```

The commit message is straight about this ("Two proof-local operations that returned bare numbers … now answer `ok`"); the two texts that ship aren't. Separately, `Ok<number>` is neither an `OpResult` nor an `IoResult`, so the JSDoc's more specific wording doesn't cover them even now.

**`fjs/effects/io/README.md:61-62` now contradicts its own opening paragraph.** The header fix landed — `:7-9` reads "The division is *composition* against *representation*, not fallible against infallible — a `List` cell and a `Program`'s exit code were once named here as computations that cannot fail, and both carry channels now." Sixty lines later the file still says `RawEffect` is used "for a computation with no failure to report — a `List` cell, a `Program`'s exit code, an MCP tool result." Only the third is true. It matters more than a stray sentence because `io-effect-migration.md`'s new blockquote points readers at this file as the authority.

**`fjs/cas/cli/module.f.mjs:56-58` is an orphaned comment that says the opposite of the code below it.** `6459df56` added the new explanation and left the old one on top:

```js
// The fold's `items` is deliberately a raw effect, so a fallible
// producer composes with `step` ahead of it rather than the fold
// growing an error channel of its own.
// `c.list()` feeds the fold directly: the step that used to sit
// here existed only to unwrap the list so `pure` could wrap it again.
return exitStep(forEachStep(c.list(), j => log(vecToCBase32(j))))
```

A tree-wide sweep for that argument finds this one occurrence; every other "raw effect" mention in `fjs/` is accurate.

Low confidence, take or leave: `types.ts:14-15` and the entry both say "`partialMatch` answers `error(notImplemented(command))`", but `effects/module.f.mjs:412-416` shows it delegating to a caller-supplied `onMissing` — only `effects/mock/module.f.mjs:36` builds that answer, and `mock/module.f.mjs:27-30` says outright that `partialMatch` *cannot*, "because the shape of an answer is the runner's". And "lets `do_` produce an `Effect` by construction and `Cont` return one" — `do_` does, but `Cont<O, T>`'s `T` is unconstrained; only its parameter became a `Result`.

Closed since my last pass, verified: the body no longer frames this as stage 5 "recorded in the plan"; `changelog/unreleased/1642.md` and the body's section are now byte-identical at 1739 bytes, five items, same order, with all three breaking changes declared; `io/README.md`'s opening claim is fixed; and `io-effect-migration.md`'s superseded rationale is handled well — a blockquote correcting the two examples in place rather than rewriting the record, which is the right treatment for a historical doc. `grep -rn 'retire-raw-effect'` returns 0 hits tree-wide, so deleting the plan left nothing dangling.

Behaviour is unchanged. A recording runner over 22 scenarios across all five changed call sites — `ci/nix`, `cas/cli`, `djs/transpiler`, `cas/evo`, `cas`, each with empty input, clean finish and mid-stream failure at several positions — gives identical results *and* identical operation sequences, `diff` empty. The semantics that matter still hold at head: `cas` gc with a failing `rm` keeps sweeping (`…rm,rm…`), the `randomInt` fold short-circuits on the third call, `nix` stops at the first failure.

Mutants: five killed, one survivor that isn't yours. Swallowing the `items` channel in `foldStep` — the path this commit newly enables — dies 2970/1; dropping its per-item short-circuit dies 2962/9; the transpiler's circular-dependency error dies 2970/1; swallowing `c.list()`'s channel at the one fallible `cas/cli` site dies 2970/1. The survivor is `ci/nix`'s `writeFlake` with `step` → `resultStep`, which drops the mkdir→writeFile short-circuit and holds 2971/0 at 100/100/100 — but `writeFlake`'s body is byte-identical at base, so it's a pre-existing untested claim behind `nixFlakes`' "stopping at the first failure", not a regression here. Worth a todo.

Declarations: 356 files compared base vs head, exactly three differ, all intended — `Operation`'s return, `foldStep`/`forEachStep` taking an `Effect`, and the two proof ops. No `O` operation union widened. The absorb points the new argument names all check out (`Handle`, `Step`, `toolsList`/`toolsCall`, `_errResponse(id)(internalError)`, `decodeRevisionBlob`). "Roughly seventy" non-`Result` payloads measures 94 of 101 — right order of magnitude.

Gates re-run by me at `f6222eb7` rather than inherited: `tsc` 0, 2971 pass / 0 fail, `cov` 100/100/100 all files, `prepack` 0, `ci-update` 0 with a clean tree.

PR #1644, opened 2026-08-18
0.46.0

sergey-shandar:
Release 0.46.0.

- Bump `package.json` and `package-lock.json` from 0.45.0 to 0.46.0.
- Rename `changelog/unreleased/` to `changelog/0.46.0/`, keeping the 63 entry
  files as they are apart from the two corrections below. Git does not track
  empty directories, so the next pull request that adds an entry recreates
  `changelog/unreleased/`.

The minor position moves (0.45 → 0.46) because the release ships
`**BREAKING CHANGES:**` entries — 29 of the 63, from #1565 through #1643.

`fjs/ci/config/module.f.mjs` keeps its `functionalscript` pin at `0.45.0` by
design: it names the published release the generated smoke tests bootstrap
from, not the in-repo version.

Two entries are corrected before the release makes them immutable, both for
claims that a later pull request in the same unreleased window reversed:

- `0.46.0/1595.md`'s third bullet said `bnf/ll1` keeps the nested repetition
  shape, named a `descentDivergence` proof group, and said both READMEs make an
  AST belong to the backend that produced it. #1611 reversed all three — the
  group is `descentEquivalence`, and the READMEs now say the AST is one
  contract shared by both backends. Since `proof.f.mjs` files ship under the
  `**/*.mjs` glob, the entry pointed a reader at a named artifact that is in
  the package under a different name. The bullet now states the interim shape
  in the past tense and defers to the entry above it.
- `0.46.0/1643.md`'s last bullet cited `changelog/1636.md` and
  `changelog/1642.md`, paths that exist under no layout, and named `#1636`
  inline, which `changelog/README.md` § Entries forbids. It now says "the
  entries below".

This is the treatment #1643 already gave `1636.md` for `RawEffect`: a name
created and deleted inside one unreleased window is withdrawn before the
release publishes it.

`npm test` passes: 2968 tests, 0 failures.

Release-manager check for every update from `main` — and again right before
merging: `changelog/unreleased/` must be empty. A pull request merged after the
rename puts its entry file back there, and an update from `main` carries it into
this branch outside `changelog/0.46.0/`; move any such file in before merging.

Changelog: none


o2alexanderfedin (commented):
The mechanics are exactly as described — I verified rather than assumed. `git diff -M100%` gives 63 `R100` lines and nothing else besides the two version files, and independent blob-hash comparison of all 63 entries before and after finds 0 mismatches and 0 missing. Every count in the body reproduces: 63 entries, 29 carrying `**BREAKING CHANGES:**`, the breaking subset genuinely spanning #1565 to #1643, and 2968 tests / 0 failures. `changelog/unreleased/` is empty at head with `origin/main` still at the merge-base, so nothing is stranded — the failure your release-manager note warns about hasn't happened. `package-lock.json` is bumped in both places. The `0.45.0` pin is correct and self-documenting, and the eleven `0.45.0` hits in `.github/workflows/ci.yml` regenerate identically from it.

One thing is worth fixing before this becomes immutable.

**`changelog/0.46.0/1595.md:12-19` is false at 0.46.0, and nothing in the release retracts it.** The third bullet says, present tense:

> `bnf/ll1` **keeps the nested repetition shape** … the new **`descentDivergence` proof group** pins all six cases side by side, and **both READMEs now say an AST belongs to the backend that produced it**

#1611 reversed all three inside this same window:

- `descentDivergence` does not exist. It is `descentEquivalence` at `fjs/bnf/ll1/proof.f.mjs:499`; `git grep descentDivergence` outside `changelog/` returns zero hits, and `c1019409` contains the `-    descentDivergence: {` line.
- Both READMEs say the opposite now — `fjs/bnf/README.md:13` is headed "## The AST is one contract", and `fjs/bnf/descent/README.md:38-40` calls it "one contract shared by both backends".
- `1611.md:1-3` says the AST "now matches `bnf/descent`'s … a flat node per `Repeat`".

`proof.f.mjs` files ship under the `**/*.mjs` glob, so a reader is pointed at a named artifact that's in the package under a different name. `1611.md` renders above `1595.md` and does correct the *substance* of the AST divergence, but it never mentions `descentDivergence`, so that name is announced and never withdrawn anywhere in the release.

This is the same shape as the case #1643 already fixed — `1643.md:77-79` rewrote `1636.md` so it stopped announcing `RawEffect` as a type that stays public, on the grounds that it was created and deleted inside one unreleased window. `1595`/`1611` is that pattern and was missed. Editing an unreleased entry is allowed, and this PR is the last gate.

Minor, inherited from #1643 rather than introduced here: `1643.md:77` and `:80` cite `changelog/1636.md` and `changelog/1642.md`, paths that exist at no point — they were `unreleased/` and are now `0.46.0/`. That bullet also names `#1636` inline, which `changelog/README.md` § Entries forbids.

The rest of the released set is coherent, which I checked by extracting all 369 backticked identifiers across the 63 entries and resolving each against the code at head. `descentDivergence` is the only unretracted absence. The `RawEffect` case reads correctly now — `1636.md:3-5` says outright that the name "is deleted again before this release ships … nothing in the published API is called `RawEffect`", and its "the entries above refer to" is right under descending render. The `IoEffect`, `okStep`, `parseJsonFromTokens`, `Evo.add`, `unwrapStep` and `foldStep`/`forEachStep` chains all narrate forward correctly, and `1643.md`'s 13 + 5 = 18 exports match the 18 `export const` in `fjs/effects/module.f.mjs` name for name.

Also confirmed nothing unrecorded ships: of the 85 squash-merged PRs in the window, 23 have no entry, and reading every title they're all `todo/`, docs, spec, proof-only, `tsconfig` or CI — legitimately `Changelog: none`. Gates at `df56ba49`: `tsc` 0 with the control firing, 2968 pass / 0 fail, `cov` 100/100/100 over 2906, `prepack` 0, `ci-update` clean. `Changelog: none` is right for this PR, and the bare-version title matches `CONTRIBUTING.md:184` and prior release commits.

sergey-shandar:
Both fixed in `df7cabc`. I reproduced each finding before changing anything.

**`1595.md` third bullet.** Confirmed on all three counts: `git grep descentDivergence` outside `changelog/` returns nothing, `fjs/bnf/ll1/proof.f.mjs:499` declares `descentEquivalence`, and `fjs/bnf/README.md:13` ("## The AST is one contract") with `fjs/bnf/descent/README.md:38-40` ("one contract shared by both backends") say the opposite of what the bullet claims. The bullet was true when written and false by the time the release ships.

Rewritten to state the interim shape in the past tense and defer to the entry above rather than contradict it:

> `bnf/ll1`: a `Repeat` compiled back to right-recursion here — the dispatch model inlined a nullable item's first set and had nowhere to put a looping frame — leaving six ways this backend's AST differed from `bnf/descent`'s, one of which lost grouping outright. All six are gone again before this release ships: the entry above makes the AST one contract across backends, and the proof group pinning the cases side by side is `descentEquivalence`

Naming the group that exists withdraws the one that doesn't, so nothing is announced and left standing. Keeping the bullet rather than deleting it follows the `1636.md`/`RawEffect` precedent — the per-pull-request record stays, the false claim doesn't.

**`1643.md` last bullet.** Also confirmed: `changelog/1636.md` and `changelog/1642.md` name no path under either layout — those entries were `unreleased/` and are now `0.46.0/` — and the inline `#1636` is a pull-request number inside an entry, which `changelog/README.md` § Entries rules out. Now reads "the entries below", the idiom `1636.md` already uses in the other direction under descending render. The substance is unchanged and still accurate: `1642.md:18-20` does carry the corrected arithmetic.

Both edits are confined to those two bullets. `npm test` at the new head: 2968 pass, 0 fail. `changelog/unreleased/` is still absent and `origin/main` is still at the merge-base `b232131`, so nothing is stranded — I'll re-check both right before merge.

One thing I noticed and deliberately did not touch: `1554.md`, `1556.md`, `1559.md` and `1562.md` end with an inline `[#NNNN](url)` self-link, the pre-convention form. They're harmless and were reviewed as-is in their own pull requests, so stripping them here would widen a release diff for cosmetics. Happy to do it in this pull request if you'd rather they land clean.

---

o2alexanderfedin (commented):
The `1643.md` fix is fully correct and the `1595.md` retraction is right about everything except one number it introduced.

**`changelog/0.46.0/1595.md:14-15` — "six ways … All six are gone" reproduces neither count.** Measured three ways:

- **As recorded when #1595 shipped** (`616ed20d`): the `descentDivergence` group pinned 6 cases, but its own `variant` case is commented "A variant is the one shape both agree on" — so 5 divergences. `bnf/ll1/README.md` there says "Six grammars… **Only the variant agrees**… One mechanism explains **all five**", and #1611's commit message says "This closes **all five** recorded divergences at once."
- **As pinned now**: `descentEquivalence` holds 8 cases (`fjs/bnf/ll1/proof.f.mjs:499-535`), with the comment "only `variant` always agreed" — 7 divergences.
- **Empirically**, running both backends over all 8 grammars: pre-#1611, **7 of 8 diverge**, only `variant` agrees — the two nullable-prefix cases diverged too (`((("A") ("A")) ("5"))` vs `("A" ("A" *()) ("5"))`, and `(() ("5"))` vs `("5")`). At head, 8 of 8 agree.

So it's 5 or 7, never 6. The old bullet actually kept these apart correctly — "one of **five ways** … all **six cases**" — and the rewrite collapsed the two into a single wrong number. Either "seven ways … all seven are gone" (matching the head proof group's own comment and the measurement) or "five recorded ways" (matching #1611's framing) would be true.

Everything else in that bullet is right: `descentEquivalence` is the real name, `descentDivergence` now appears nowhere in the tree including `changelog/`, `nesting` did lose grouping outright (pre-#1611 it produced an AST byte-identical to `leadingItem`'s), and the past-tense framing holds — `bnf/ll1/README.md:26-33` says a `Repeat` is one flat node, and both READMEs say the AST is one contract. "The entry above" correctly resolves to `1611.md` under the descending rule.

`1643.md` is clean. `1643` is the highest entry in the release, so 1636 and 1642 genuinely render below it; `grep 'changelog/'` across `changelog/0.46.0/` returns nothing, and the inline `#NNNN` is gone. The operation recount reproduces exactly — six operations declared inside proofs at `bd4f90f7^`, four returning a bare `Result` (`io/proof.f.mjs:16,22` and `io/types.ts:95,97`), two returning bare `number` (`effects/proof.f.mjs:27,40`), both rewritten by that same commit, and every host operation on `OpResult`/`IoResult`.

No collateral damage: the commit touches only those two files, and blob-hash comparison against `b2321311:changelog/unreleased/` confirms the other **61** entries are still byte-identical, so the pure-rename property holds. `changelog/unreleased/` is still empty, `origin/main` hasn't moved, and 63 / 29 / #1565–#1643 all still check out. Gates at `df7cabc1`: `tsc` 0 with the control firing, 2968 pass / 0 fail, `cov` 100/100/100 over 2906, `prepack` 0, `ci-update` clean, both version files at 0.46.0 in all three places.

One observation, not a finding, only because you applied the standard yourself to `1636.md`: `1628.md:10` announces `parseJsonFromTokens` in the present tense and `1631.md:8` — rendering directly above it — says it "is gone with it". Same introduce-and-remove-in-one-window shape, but unlike `RawEffect` it's retracted by name immediately above, so the published section reads coherently. I'd leave it.

The wider sweep is clean: 418 backticked tokens across the 63 entries, 262 identifier-shaped, resolved against the code at head; the 9 unresolved are all legitimate history whose replacement exists.

sergey-shandar:
You're right, and the diagnosis is exact: my rewrite collapsed two numbers the original bullet had kept apart. Fixed in `dea5cac`.

I verified both counts rather than taking them on faith:

- `git show 616ed20d:fjs/bnf/ll1/proof.f.mjs` — `descentDivergence` at :464 holds six cases, and `variant` at :471 is commented "A variant is the one shape both agree on". Five recorded divergences, six cases. So the original's "one of five ways … all six cases" was precise, and "six ways" reproduces neither half of it.
- At head, `descentEquivalence` (`fjs/bnf/ll1/proof.f.mjs:499-535`) holds eight `bothBackends` cases and is commented "These are the cases that used to diverge … only `variant` always agreed". Seven.

The bullet now reads:

> … leaving **seven** ways this backend's AST differed from `bnf/descent`'s, one of which lost grouping outright. All **seven** are gone again before this release ships: the entry above makes the AST one contract across backends, and the `descentEquivalence` proof group pins all **eight** cases side by side

I took head's accounting over #1611's "five recorded" for two reasons. The bullet names `descentEquivalence`, so a reader who opens it counts those eight and finds the two numbers consistent — the same reason the earlier fix mattered at all. And seven is the true count rather than the recorded subset: your measurement of the two nullable-prefix cases shows they diverged before #1611 without being pinned yet, so "five" would understate what actually differed. Restoring the ways-vs-cases split keeps the original's precision instead of trading one wrong number for another.

`nesting` losing grouping outright is confirmed at `616ed20d:494-500` — "Grouping is *lost*, not reshaped … byte-identical ASTs".

On `parseJsonFromTokens`: agreed, and I've left it. `1631.md` retracts it by name in the entry rendering directly above `1628.md`, so the section is self-correcting where `RawEffect` was not — that's the distinction that made `1636.md` need an edit, and it doesn't apply here.

`npm test` at `dea5cac`: 2968 pass, 0 fail. The commit touches only `1595.md`, three lines. `changelog/unreleased/` is still absent and `origin/main` is still at `b232131`; I'll check both again right before merge.

---

o2alexanderfedin (approved):
Approving. Both counts are right, and I re-derived them from scratch rather than checking the text against my own earlier measurement — running both backends over all eight grammars at `616ed20d` and comparing `showAst` outputs directly:

| case | pre-#1611 |
|---|---|
| `leadingItem`, `skippedOption`, `takenOption`, `repetition`, `nesting`, `repetitionInNullablePrefix`, `emptyRepetitionInNullablePrefix` | diverge |
| `variant` | agree |

Seven of eight. The control matters here: the same probe at head gives 0 diverged / 8 agreed, with every AST equal to the string pinned in `proof.f.mjs`, so "All seven are gone" is measured rather than asserted. Eight cases in the group at `:499`, `variant` the sole always-agreeing one, matching its own comment at `:498`.

The five-versus-seven tension resolves cleanly, and in your favour. The old README tabulated six grammars of which five diverged, and #1611's message says "closes all five **recorded** divergences" — *recorded* is the operative word, because #1611 added two cases the old table never listed, and both of those diverged pre-#1611 as well. So five is what was written down at the time and seven is how many ways the ASTs actually differed across the eight cases now pinned. The bullet claims the latter, which is what I measured, and the old six-row table maps exactly onto the first six rows above.

Rest of the bullet holds: `descentEquivalence` is real at four sites, `descentDivergence` has zero hits tree-wide, and "lost grouping outright" is `nesting` — its pre-#1611 ll1 output is byte-identical to `leadingItem`'s, so grouping was destroyed rather than reshaped.

One thing I looked at and decided is fine: the entry rendering *directly* above 1595 is 1596, with 1611 fifteen further up. But that's your established idiom — `1636.md:3` says "the entries above refer to" about 1640–1643 with 1637–1639 in between — so "the entry above" reads correctly.

Release mechanics still hold at this head: 63 entries, 61 blob hashes byte-identical to their `unreleased/` originals with only the two deliberately-corrected files differing, `changelog/unreleased/` absent, `origin/main` still at the merge-base so nothing is stranded, 29 breaking entries spanning #1565–#1643, and 0.46.0 in `package.json` and both places in `package-lock.json`.

Final coherence sweep over all 63 entries — 577 backticked occurrences, 418 unique, resolved against an identifier set built from all 353 shipped files plus a filesystem check for the 66 path-shaped tokens. Every unresolved token is a removal or rename statement rather than a present-tense advertisement: `casUpload` "is removed", `fjs/effects/io/` "is gone", `Eff` "is removed", and so on. `1636.md`'s "nothing in the published API is called `RawEffect`" holds — the only occurrences left are JSDoc history. `IoEffect` survives in a `todo/` file the `files` globs don't ship, and the entries naming it are corrected by `1636.md` rendering above them, the same structure as the 1628/1631 case.

Gates at `dea5cac2`: `tsc` 0 with the control firing at `proof.f.mjs(590,7)`, 2968 pass / 0 fail, `cov` 100/100/100, `prepack` 0, `ci-update` clean.

Good release. The entries are worth the care given they freeze on merge.

PR #1645, opened 2026-08-19
types/rtti: restore validate, the verbatim reader

o2alexanderfedin:
@sergey-shandar — this is the module you deleted in #1624, back with the one part of the case the retirement did not cover.

## The gap

0.43.1's `validate` returned `ok(value)` — the value it was handed. `parse` rebuilds. Verified against `827dba0`:

```js
const schema = { a: number, b: option(string) }
parse(schema)({ a: 1, extra: 'x' })     // ['ok', { a: 1, b: undefined }]
validate(schema)({ a: 1, extra: 'x' })  // ['ok', { a: 1, extra: 'x' }]  (this PR)
```

Every declared key is present in what `parse` builds (an absent optional becomes a materialized `undefined`), every undeclared key is dropped, and a container result is never reference-identical to its input. For "read this value as `T`" that is right. For "is this value a `T`?" it rewrites the thing being asked about, and the schema form offers no opt-out.

## Why a consumer needs the verbatim form

A downstream Form 1040 engine keeps its values as JSON documents in a content-addressed store:

- Rebuilding changes the bytes, so it changes the hash — the document that comes out of a check is not the document that went in.
- Absent is not present-with-`undefined`. An absent box means the payer did not report that figure; `{ b: undefined }` says something else and serializes differently.

`retire-validate.md` read the objection to deleting `validate` as a cost objection — two traversals, a discarded intermediate — and answered it with the one-pass JSON reader (`media/json/todo/rtti-parse.md`). For text in, that answer is strictly better and I am not arguing with it. It does not reach a value already in memory whose identity is load-bearing, and its closing line — "neither has a reason to return its input unchanged" — is the sentence a content-addressed consumer contradicts.

The precedent is inside #1624: `dev/package_json`, its one non-drop-in site, ended up doing "parse for the check, return the value it was given, with the cast the successful parse justifies". That is this function hand-rolled at one call site, with a cast. (#1625 removed that module for having no consumers, not because the shape was wrong.) A consumer whose rules forbid casts, `any`, and hand-written type predicates cannot write that line.

The data form's `validate` is not a substitute: it returns `ResultE` with the payload erased to `Unknown`, so recovering `Ts<T>` needs a cast; it requires `toData` first; and it is closed on tuples (`data-form-open-tuples.md`), so it rejects values `parse` accepts.

## What is here

`fjs/types/rtti/validate/{module.f.mjs,types.ts,proof.f.mjs}` — the deleted module restored, minus two things:

- **No `sizeOk`.** #1622's exact-length check does not come back. Tuples and structs are open, exactly as `parse` has them; this restores a reader, not the closed model.
- **No re-export block.** The four `common` symbols it used to re-export stay in `common`; nothing imported them.

It shares the kernel, not the pass: `visit` for schema recognition, `orVisit`, the primitive checks, and `eachEntry` in the no-accumulator mode its own JSDoc describes — "a caller whose whole question is 'did every entry succeed?' passes `undefined`/`acc => acc` instead and pays no allocation per entry". The schema form has had no such caller since #1624; `data/module.f.mjs`'s `validate` is the same shape over `Data`, and this is its thunk-form twin.

Additive: `parse` is untouched in behaviour, signature and exports; no importer moves back; the data form is untouched. Two dangling `../validate` references become true again with no edit — `common/types.ts`'s header and its `Validate<T>` doc, and `data/module.f.mjs`'s "the counterpart of `../validate`".

## What the proofs pin

Three leaves are the whole reason the function exists:

- `verbatim.absentOptionalStaysAbsent` — `!('b' in out)`, beside the `parse` call showing `'b' in out`. (`out.b === undefined` would not tell them apart.)
- `verbatim.undeclaredMemberSurvives` — the extra is in the result, on structs and tuples both.
- `verbatim.referenceIdentity` — `Object.is(result, input)` over array, record, tuple, struct, `or`, `unknown`, and a nested container; the mirror of parse's `freshArray`.

One more guards the risk you flagged by deleting the module — two walkers drifting. `sameAcceptanceAsParse` runs a 25-row table asserting `validate(s)(v)[0] === parse(s)(v)[0]`, and on error that both report the same `{ path, message }`; the four openness rows from `retire-validate.md` are in it. If the readers ever disagree about what a `T` is, it fails.

If you want them to share the container skeleton as well, the hoist-on-second-consumer rule now applies to it — glad to do that next; kept out of here for one-feature/minimal-change.

## Docs

`rtti/README.md`'s module list said "there is no separate validator that returns its argument unchanged", which this makes false. It gains a "The two schema-form readers" section (which reader for which job, why `validate` is not a `parse` with the result thrown away, and a pointer to the one-pass JSON reader for the cost case), a `validate` column on the openness table, and a note in the #1622 guard that the restored module has no length check either. Every example re-verified by execution.

Left alone: three `todo/` asides saying the two-walker duplication "was resolved by deleting validate" (`export-node-accessors`, `kindset-eliminator`, `proof-shared-asserts`). Say the word and I will repoint them.

## Checks

`npx tsc` clean. `npm test` 3032/3032. `npm run cov` 2970/2970 at 100% line/branch/function, `validate/module.f.mjs` included. Deno and Bun are not installed in this environment, so those runners and `npm run update` did not run; `package.json` is untouched, so the lockfiles and the generated workflow are unaffected — `npm run ci-update` produces no diff.

Changelog:
- `types/rtti`: new `validate/module.f.mjs`. `validate(schema)(value)` accepts
  what `parse` accepts but returns the original value rather than a
  reconstruction — undeclared members survive, absent optional members stay
  absent, and the result is reference-identical to the input.


fjs/types/rtti/validate/types.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Type-level API for RTTI validation.
+ *
+ * `Validate<T>` and `Result<T>` are the shared consumer signatures — `parse`
+ * uses the same pair, under the name `Parse<T>` — so a caller can swap the two
+ * readers without changing a type annotation.
+ *
+ * @module
+ */
+
+export type { Path, Result, Validate, ValidationError } from '../common/types.ts'
```
sergey-shandar: Don't re-export types. We don't need the `validate/types.ts`.

sergey-shandar (changes requested):

sergey-shandar (approved):

o2alexanderfedin:
## Downstream result: this PR is the last thing between `fjs-dev/finance` and a green 0.46.0

Context for the review, since it may change how the trade-off reads. We migrated the whole engine to 0.46.0 today. Numbers, all measured:

| | |
|---|---|
| `tsc` errors | 1526 → 629 (specifier rename) → 282 → **0** |
| tests | 2505, **2478 pass** |
| still failing | **27 — all of them this issue, none anything else** |

**0.46.0 was a good release to consume.** The new Effect system deleted code rather than adding it: a `try` we had carried since 0.41.0 is gone, because a refusal that arrives as `error(notImplemented)` can be checked as a value instead of caught. It also exposed a latent mislabel — a missing-blob failure had been surfacing to clients as a *policy refusal* because it travelled as a throw through that same `try`. Three `any`s went with it. The error channel made our own code read better: the success path became the only path a reader follows.

The 27 are the single place where `parse` cannot stand in for the old `validate`, and it is not about cost. Our values are **JSON documents already in a content-addressed store**. A rebuilt struct is a different value: an absent optional box means *the payer did not report that figure*, which is not the same fact as `undefined`, and re-serialising it changes the bytes, which changes the hash, which breaks byte-identical rerun of pinned programs. We cannot reach for the cast that `dev/package_json` used — this project forbids casts, `any` and non-null assertions outright, and both agents that hit this stopped and recorded rather than writing one.

So the choice we are left with, if `validate` stays deleted, is: fork the dependency, or weaken a rule that exists to keep a tax engine honest. Neither is good, and this PR is small and additive.

Happy to change anything about the shape — the name, the location, or making it a documented `common`-level export rather than a restored module — if a different form fits `retire-validate.md`'s intent better. The behaviour is the only part we need.

/cc @sergey-shandar

PR #1621, opened 2026-08-16
ci: bump functionalscript bootstrap version and nixpkgs snapshot

sergey-shandar:
## Summary
This PR updates the FunctionalScript dependency to version 0.45.0 across all CI workflows and configuration files, along with updating the nixpkgs commit reference.

## Key Changes
- **FunctionalScript version bump**: Updated from 0.44.0 to 0.45.0 in:
  - CI workflow configuration (.github/workflows/ci.yml) - 11 occurrences across npm, deno, and bun package managers
  - Module configuration (fjs/ci/config/module.f.mjs)
  
- **Nixpkgs update**: Updated the nixpkgs commit reference from `9f78f44a87948854445dae0b6bf82b2e87e4efb5` to `02e08985a27c65ffd33d434eeb2e660a2e4dc84d` in:
  - Module configuration (fjs/ci/config/module.f.mjs)
  - Generated Nix flake files for Node.js 22, 24, and 26

## Implementation Details
The changes ensure consistency across all CI environments (npm, deno, bun) and development shells by using the same FunctionalScript version and nixpkgs commit throughout the project configuration.

Changelog: none

https://claude.ai/code/session_01KRZ9i3sKgJShU8ssvqUs9b

PR #1646, opened 2026-08-19
0.46.1

sergey-shandar:
Release 0.46.1.

- Bump `package.json` and `package-lock.json` from 0.46.0 to 0.46.1.
- Rename `changelog/unreleased/` to `changelog/0.46.1/`, keeping its one entry
  file exactly as it is. Git does not track empty directories, so the next pull
  request that adds an entry recreates `changelog/unreleased/`.

The patch position moves because the release ships no `**BREAKING CHANGES:**`
entry: `1645.md` adds `types/rtti`'s `validate` module, which is a new feature.
Pre-1.0 the minor position moves only for a breaking change, so a feature-only
release is a patch — the `changelog/README.md` § Breaking changes and versioning
table's middle row.

The entry resolves against the code at head: `fjs/types/rtti/validate/module.f.mjs`
exists and exports `validate` at :216.

`fjs/ci/config/module.f.mjs` keeps its `functionalscript` pin at `0.45.0`. It
names the published release the generated smoke tests bootstrap from rather than
the in-repo version, and moving it regenerates CI, so it is a change to CI rather
than part of a release.

`npm test` passes: 3032 tests, 0 failures. `npm ci` reinstalls cleanly, so the
lockfile is in step with the bumped `package.json`.

Release-manager check for every update from `main` — and again right before
merging: `changelog/unreleased/` must be empty. A pull request merged after the
rename puts its entry file back there, and an update from `main` carries it into
this branch outside `changelog/0.46.1/`; move any such file in before merging.

Changelog: none

---
_Generated by [Claude Code](https://claude.ai/code/session_01JQYHF7SXSDT22CLcmz3nSq)_


o2alexanderfedin (approved):
Checked this at head `e2e47855` in a clean worktree. The versioning claim holds: `changelog/README.md:98-102`'s middle row gives `0.Y.(Z+1)` for a feature-only pre-1.0 release, `1645.md` carries no `**BREAKING CHANGES:**` prefix, and 0.46.0 was the previous version — so 0.46.1 is right.

The rename is byte-identical (both paths hash to `1c0ddc59`) with nothing tracked left under `changelog/unreleased/`; both lockfile version fields moved alongside `package.json` (`:3` and `:9`); `fjs/types/rtti/validate/module.f.mjs:216` exports `validate` exactly as stated; and leaving the pin at `fjs/ci/config/module.f.mjs:28` matches how #1521 and #1621 moved it in their own PRs rather than in a release — neither the 0.46.0 nor the 0.45.0 release commit touched it.

Gates: `npm ci` clean, `npm test` 3032 pass / 0 fail, `npm run cov` 2970 pass at 100% line/branch/function, `npm run prepack` and `npm pack --dry-run` clean (`functionalscript-0.46.1.tgz`, 622 files), `npm run ci-update` leaves the tree unchanged. The `tsc` runs were preceded by deleting every `*.d.mts` and validated with a negative control that fired.

PR #1659, opened 2026-08-21
edag: add comparison operators (===, !==, >, <, >=)

sergey-shandar:
## Summary
- Add five binary comparison nodes to `fjs/edag`, all following the same `Phantom`+`Check3`-pinned pattern as `add`/`sub`: `Eq` (`['===', exp, exp]`), `Neq` (`['!==', exp, exp]`), `Gt` (`['>', exp, exp]`), `Lt` (`['<', exp, exp]`), and `Ge` (`['>=', exp, exp]`). All five are wired into the `exp` union / `Exp` type — 21 members at this head, 21 distinct tags, no collisions — and each has a full `proof.f.mjs` section (ok/nested, missing-operand error, extra-tail-ignored, wrong-tag error) mirroring `add`'s exactly.
- Fixed the `@import` block's `Neq` line along the way, which was missing its leading `*` (same recurring typo the `Own`/`neg` lines had earlier in this stack).
- No `<=` (a natural sixth) yet — not added since it wasn't asked for; happy to add it the same way.

## Test plan
- [x] `npm run cov` — 3114/3114 tests, 100% lines/branches/functions

Changelog:
- `edag`: add `Eq` (`===`), `Neq` (`!==`), `Gt` (`>`), `Lt` (`<`), and `Ge` (`>=`)
  comparison nodes, all now in the `exp` union

### DIFF
--- a/changelog/unreleased/1659.md
+++ b/changelog/unreleased/1659.md
@@ -1,0 +1,2 @@
+- `edag`: add `Eq` (`===`), `Neq` (`!==`), `Gt` (`>`), `Lt` (`<`), and `Ge` (`>=`)
+  comparison nodes, all now in the `exp` union
--- a/fjs/edag/module.f.mjs
+++ b/fjs/edag/module.f.mjs
@@ -26,6 +26,7 @@
  *  Eq,
  *  Neq,
  *  Gt,
+ *  Lt,
  * } from './types.ts'
  * @import { Phantom } from '../types/phantom/types.ts'
  */
@@ -72,7 +73,8 @@ import {
  *  typeof frame,
  *  typeof eq,
  *  typeof neq,
- *  typeof gt
+ *  typeof gt,
+ *  typeof lt
  * ]}
  */
 export const exp = () => (['or',
@@ -95,6 +97,7 @@ export const exp = () => (['or',
     eq,
     neq,
     gt,
+    lt,
 ])
 
 /** @typedef {Assert<Check<Exp, typeof exp>>} _ExpAssert */
@@ -358,3 +361,12 @@ const _gt = /** @type {const} */(['>', exp, exp])
 export const gt = _gt
 
 /** @typedef {Assert<Check3<Gt, typeof _gt, typeof gt>>} _Gt */
+
+// <
+
+const _lt = /** @type {const} */(['<', exp, exp])
+
+/** @type {Phantom<typeof _lt, Lt>} */
+export const lt = _lt
+
+/** @typedef {Assert<Check3<Lt, typeof _lt, typeof lt>>} _Lt */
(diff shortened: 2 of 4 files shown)


o2alexanderfedin (commented):
All five are properly pinned — I ran the full battery per node rather than spot-checking, since this stack has shipped an unpinned node more than once. For each of `eq`/`neq`/`gt`/`lt`/`ge`: deleting it from the `exp` union reddens its `ok` and `extraTailIsIgnored`, and retagging it in module + types together does the same. Both `Check3` halves bite for all five — the raw-thunk mutation gives exactly one error at that node's own `_X`, and the phantom-`T` mutation reddens its `_X` plus the usual cascade (I confirmed against `add` that the cascade is pre-existing to the design, not something these five introduce).

`['>=', 1, 2]` genuinely doesn't fall through to `gt` — deleting `ge` from the union reddens `ge.ok`, so the three comparison tags are behaviorally distinct, not just textually. Suite is 3114/3114, exactly +20 over the merge base for 5 nodes × 4 leaves. `cov` exit 0 at 100.00, `ci-update` no diff, 18/18 checks. The changelog is byte-identical to the body's section, no `**BREAKING CHANGES:**` marker is owed, and `proof.f.mjs:2`'s header is still exactly true — `comma` remains the only union member without a section, and all five new ones have one.

One thing:

**The body's "16 distinct tags total (no collisions)" is the pre-PR number.** That's the count at the merge base `164e2853`; at this head `exp` has **21** members and 21 distinct tags — `[] {} args Number String . () .() own + - neg , => frame === !== > < >=` plus `undefined` inside `primitive`. The "(no collisions)" half is right. Since the body becomes the squash commit, the wrong count would land in `main`'s history.

Not raising, for the record: the bot's changelog request was correct when posted and is resolved by `bb1740ea`. And `types.ts:2-5`'s header enumeration was already stale by eight kinds on `main` — this PR makes it thirteen, but it isn't this PR's defect.

o2alexanderfedin (approved):
Count corrected and verified independently — I re-derived it by importing `exp()` and walking the union rather than reading the source, so it isn't just checking the new number against the one that produced it. 21 members: 20 tagged tuples (`[] {} args Number String . () .() own + - neg , => frame === !== > < >=`) plus `undefined` from `primitive`'s `or`. 21 distinct tags, zero duplicates. Both figures in the body are right.

Dropping the enumeration from `types.ts`'s header is the better fix — it removes the thing that drifts rather than resyncing it, and I checked the premise holds: `export type Exp` really is the next declaration (line 11, after only the blank line and the `// exp` marker), and no other prose enumeration of kinds survives anywhere in `fjs/edag/`. The `@import` block in `module.f.mjs` lists all 21, but that's type-checked rather than prose, so it can't go stale silently. No JSDoc tag was removed with the comment — `types.ts` has none anywhere in the file.

Gates unchanged: `tsc` clean, 3114/3114, `cov` exit 0 at 100.00, `ci-update` no diff.

Approving. Thanks for fixing the `types.ts` header too — that one predated this PR and I'd said as much.

PR #1660, opened 2026-08-21
edag: consolidate unary/binary ops into UnaryOp/BinaryOp

sergey-shandar:
## Summary
- **Consolidates ~13 individually-defined node kinds into two generic, id-parameterized schemas.** `Add`, `Sub`, `Neg`, `Eq`, `Neq`, `Gt`, `Lt`, `Ge`, `Own`, `Fn`, `Call`, and `StringCast` are gone as standalone types/exports; in their place, `UnaryOpId = 'String' | 'Number' | 'neg' | '!' | '~'` + `UnaryOp = readonly[UnaryOpId, Exp]`, and `BinaryOpId` (a 24-member literal union) + `BinaryOp = readonly[BinaryOpId, Exp, Exp]`. `NumberCast` stays as its own type/export — `index`'s definition still needs it standalone (`or(numberCast, string, number)`), even though `'Number'` is also a `UnaryOpId` member.
- **This also lands ~16 operator ids that had no node at all before**, in one shot: `!`, `~`, `<=`, `*`, `/`, `%`, `**`, `&`, `|`, `^`, `<<`, `>>`, `>>>`, `&&`, `||`, `??` — the full unary/arithmetic/bitwise/logical set from `edag-stage1-discussion.md`'s Operators table, now all reachable through `unaryOp`/`binaryOp`.
- **Fixed a build-breaking dangling import**: `module.f.mjs` still imported `StringCast` from `types.ts`, which no longer exports it post-consolidation.
- **`proof.f.mjs` is rewritten to match.** Its ~13 old sections named node kinds this PR deletes and only exercised 10 of `binaryOp`'s 24 ids — the new ones had no coverage at all, and deleting any of them from `binaryOpId`/`unaryOpId` left `tsc`/the suite/`cov` all green (verified this against `'**'` and, after `!`/`~` landed, `'~'` too, before writing/updating the fix each time). Replaced with `unaryOp:`/`binaryOp:` sections that loop over an explicit id list per kind, so each id's removal reddens its own assertion, plus an `unknownIdIsRejected` case on both (checking `unaryOpId`/`binaryOpId` directly and through a full `exp` value) — previously nothing pinned that the id schema constrains membership rather than behaving like a bare `string`. Also fixed three stale `call` references left over from the consolidation (`module.f.mjs`'s doc comment, `phantom/types.ts`'s `Phantom` example, and `rtti/todo/close-type.md`).

## Test plan
- [x] `npm run cov` — 3069/3069 tests, 100% lines/branches/functions

Changelog:
- `edag`: `Add`, `Sub`, `Neg`, `Eq`, `Neq`, `Gt`, `Lt`, `Ge`, `Own`, `Fn`, `Call`, and
  `StringCast` are replaced by two generic, id-parameterized node kinds — `UnaryOp`
  (`String`/`Number`/`neg`/`!`/`~`) and `BinaryOp` (`=>`, `own`, `()`, `===`, `!==`, `>`,
  `>=`, `<`, `<=`, `+`, `-`, `*`, `/`, `%`, `**`, `&`, `|`, `^`, `<<`, `>>`, `>>>`, `&&`,
  `||`, `??`) — adding `!`, `~`, `<=`, `*`, `/`, `%`, `**`, `&`, `|`, `^`, `<<`, `>>`,
  `>>>`, `&&`, `||`, and `??` to the `exp` union for the first time

No `**BREAKING CHANGES:**` marker: per the established precedent on #1658/#1656, nothing
in `fjs/edag/` has ever been published (`npm pack functionalscript@latest` has zero
`fjs/edag/*` entries), so renaming/consolidating its exports within this same unreleased
cycle breaks no shipped consumer.

### DIFF
--- a/changelog/unreleased/1660.md
+++ b/changelog/unreleased/1660.md
@@ -1,6 +1,6 @@
 - `edag`: `Add`, `Sub`, `Neg`, `Eq`, `Neq`, `Gt`, `Lt`, `Ge`, `Own`, `Fn`, `Call`, and
   `StringCast` are replaced by two generic, id-parameterized node kinds — `UnaryOp`
-  (`neg`/`String`/`Number`) and `BinaryOp` (`=>`, `own`, `()`, `===`, `!==`, `>`, `>=`,
-  `<`, `<=`, `+`, `-`, `*`, `/`, `%`, `**`, `&`, `|`, `^`, `<<`, `>>`, `>>>`, `&&`, `||`,
-  `??`) — adding `<=`, `*`, `/`, `%`, `**`, `&`, `|`, `^`, `<<`, `>>`, `>>>`, `&&`, `||`,
-  and `??` to the `exp` union for the first time
+  (`String`/`Number`/`neg`/`!`/`~`) and `BinaryOp` (`=>`, `own`, `()`, `===`, `!==`, `>`,
+  `>=`, `<`, `<=`, `+`, `-`, `*`, `/`, `%`, `**`, `&`, `|`, `^`, `<<`, `>>`, `>>>`, `&&`,
+  `||`, `??`) — adding `!`, `~`, `<=`, `*`, `/`, `%`, `**`, `&`, `|`, `^`, `<<`, `>>`,
+  `>>>`, `&&`, `||`, and `??` to the `exp` union for the first time
--- a/changelog/unreleased/1660.md
+++ b/changelog/unreleased/1660.md
@@ -1,0 +1,6 @@
+- `edag`: `Add`, `Sub`, `Neg`, `Eq`, `Neq`, `Gt`, `Lt`, `Ge`, `Own`, `Fn`, `Call`, and
+  `StringCast` are replaced by two generic, id-parameterized node kinds — `UnaryOp`
+  (`neg`/`String`/`Number`) and `BinaryOp` (`=>`, `own`, `()`, `===`, `!==`, `>`, `>=`,
+  `<`, `<=`, `+`, `-`, `*`, `/`, `%`, `**`, `&`, `|`, `^`, `<<`, `>>`, `>>>`, `&&`, `||`,
+  `??`) — adding `<=`, `*`, `/`, `%`, `**`, `&`, `|`, `^`, `<<`, `>>`, `>>>`, `&&`, `||`,
+  and `??` to the `exp` union for the first time
--- a/fjs/edag/module.f.mjs
+++ b/fjs/edag/module.f.mjs
@@ -28,6 +28,7 @@
  *  Gt,
  *  Lt,
  *  Ge,
+ *  BinaryOpId,
  * } from './types.ts'
  * @import { Phantom } from '../types/phantom/types.ts'
  */
@@ -382,3 +383,14 @@ const _ge = /** @type {const} */(['>=', exp, exp])
 export const ge = _ge
 
 /** @typedef {Assert<Check3<Ge, typeof _ge, typeof ge>>} _Ge */
+
+const binaryOpId = or(
+    '===', '!==', '>', '>=', '<', '<=',
+    '+', '-', '*', '/', '%', '**',
+    '&', '|', '^', '<<', '>>', '>>>',
+    '&&', '||', '??'
+)
+
+/** @typedef {Assert<Check<BinaryOpId, typeof binaryOpId>>} _BinaryOpId */
+
+const binaryOp = /** @type {const} */([binaryOpId, exp, exp])
--- a/fjs/edag/module.f.mjs
+++ b/fjs/edag/module.f.mjs
@@ -220,22 +220,6 @@ export const propertyCall = _propertyCall
  * @typedef {Assert<Check3<PropertyCall, typeof _propertyCall, typeof propertyCall>>} _PropertyCall
  */
 
-// Negation (aka a unary minus) — a word tag, `"neg"`, not `"-"`'s unary
-// arity (an earlier draft overloaded `"-"` the way JS itself does; see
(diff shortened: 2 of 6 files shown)


o2alexanderfedin (commented):
`_UnaryOp` is genuinely fixed — all three drifts now error at `:228` (`UnaryOp` gaining a third operand, the schema gaining one, and `UnaryOp` losing its operand), and I ran the counterfactual to be sure the commit is what did it: reverting just the typedef to `Check<UnaryOp, typeof unaryOp>` while keeping the drift applied puts `tsc` back to exit 0. The `BinaryOpId` realignment is whitespace only — all four member lists across both heads are identical at 24, same order. Behavior is unchanged from my A/B: only the intended new ids flipped, arity and bogus-id rejection preserved.

Before the rest — one correction to my last review, in your favour. I said the old proof sections "no longer describe what's being validated", echoing your framing. That's too harsh: deleting each pre-existing id reddens exactly its section (`'+'`→`add.*`, `'=>'`→`fn.*`, `'()'`→4 leaves, and so on for all 13). They still pin membership and arity for every id that existed before. The gap is narrower than either of us said.

Four things still open:

1. **All 14 new ids are unproven.** `proof.f.mjs` exercises 10 of the 24 `BinaryOpId` members as real values; the unexercised 14 are exactly the set your changelog calls new. Deleting `'<='`, `'*'` and `'??'` from both `binaryOpId` and `BinaryOpId` leaves `tsc` at 0 and the suite at 3114/0 — the gate structurally can't see them, since ids are `or` arguments rather than lines or branches.

2. **Nothing pins id rejection.** Adding `'nosuchop'` to both `binaryOpId` and `BinaryOpId` — a real widening of `validate(exp)` — leaves `tsc`, the suite and `cov` all green. The existing `error:` cases pin only near-misses (`'ownz'`, `'(x)'`, `'.(x)'`), not the id set as a whole. Now that ids are data, this is the check that stops the union drifting arbitrarily wide.

3. **`proof.f.mjs:2` is now false** — it still says "one section per node kind … Exception: `comma`", but the kinds are `unaryOp`/`binaryOp` (no sections) while eleven sections name kinds this PR deletes. The sections themselves still test real behavior; it's the header claim that's wrong.

4. **Three stale `call` references**, not two — I missed one last time: `module.f.mjs:37`, `fjs/types/phantom/types.ts:32`, and **`fjs/types/rtti/todo/close-type.md:135`** ("`args`, `propertyAccessor`, `call`, and `propertyCall` are fixed-arity nodes"). All three were correct before this PR.

Gates green at `e111c4e0`: `tsc` clean, 3114/3114, `cov` exit 0 at 100.00, `ci-update` no diff. The changelog is accurate — its 12 removed types match the symbol diff exactly, and its 14 new ids match the unexercised set exactly.

o2alexanderfedin (commented):
First, `e1a3f9cb` closed all four items from the previous round — verified by mutation, not by reading. 27/27 operator ids load-bearing, id rejection pinned, the `proof.f.mjs` header now accurate (every `exp` member has a section except `comma`, exactly as documented), all three stale `call` refs fixed with correct replacements, and nothing lost in the rewrite.

`a979ba34` adds `'!'` and `'~'`, which is right — `todo/edag-stage1-discussion.md`'s Operators table names exactly those two at arity 1, and the reorder is inert (all five variants are disjoint const primitives, so first-match order can't change the outcome). But the proof list didn't come along:

1. **`fjs/edag/proof.f.mjs:46`** — `unaryOpIds` is still `['neg', 'String', 'Number']`. Deleting `'!'` (or `'~'`, or both) from `fjs/edag/module.f.mjs:220` **and** `fjs/edag/types.ts:78` leaves tsc at 0 and the suite at 3069/3069. Both new ids are unproven; the commit is a no-op to every gate. The single-file TS2344s at `_UnaryOpId`/`_UnaryOp` only prove the two lists mirror each other.
2. **`fjs/edag/proof.f.mjs:158` and `:174`** — "these three" / "its three members". Five now.
3. **`changelog/unreleased/1660.md:3`** — `UnaryOp` is `(`neg`/`String`/`Number`)`; `!` and `~` are missing there and from the trailing "adding … to the `exp` union for the first time" list. The PR body's `Changelog:` line is byte-identical today, so both need the same edit to stay so; the Summary bullets (`UnaryOpId = 'neg' | 'String' | 'Number'`, "~14 operator ids" — now 16) go into the squash commit too.

The binary enumeration is still exact at 24.

o2alexanderfedin (approved):
All three items closed, verified by mutation.

- **`'!'` and `'~'` proven.** Deleting either — or both, i.e. a full runtime revert of `a979ba34` — from `module.f.mjs:220` and `types.ts:78` now reddens `proof.unaryOp.ok`. Leaf count stays 3069, which is right: the ids feed a `for…of` inside one leaf, so each adds zero leaves.
- **Counts fixed**, and every other enumeration in `proof.f.mjs` checks out — `:119` "three accepted shapes" matches `index`, `:195` "fourteen" is correctly scoped to `binaryOp` (24 − 10 pre-PR nodes) and doesn't double-count the two new unary ids, `:197` "twenty-four" matches.
- **Changelog and PR body byte-identical** (510 bytes both), unary and binary enumerations match `types.ts` in set, count and order across all four files, and the Summary bullets are corrected (`~16 operator ids`, all 16 confirmed absent at the merge base).

No regression: the other 25 ids still redden on deletion, `unknownIdIsRejected` still fires when either id set is widened to `string`, and the `_UnaryOpId`/`_UnaryOp`/`_BinaryOpId`/`_BinaryOp` guards still produce TS2344 in both directions — no `Check`-vs-`Check3` tautology. All four gates green.

PR #1661, opened 2026-08-21
bnf/ll1: add an SSN grammar proof

sergey-shandar:
## Summary
- Adds `proof.ssn` to `fjs/bnf/ll1/proof.f.mjs`: a small LL(1) grammar for a Social Security Number, `ddd-dd-dddd` — `[ws, repeat(3)(d), ws, '-', ws, repeat(2)(d), ws, '-', ws, repeat(4)(d), ws]` where `d = range('09')` and `ws = repeat0Plus(' ')` — exercising the same `parser`/`expect` pattern the file's other tests already use. The `ws` tolerance is load-bearing: it's what lets `'  123  - 45 - 6789 '` accept.
- This started as an uncommitted, unnamed scratch test (`tn`/`tfn`) with a real bug: `expect('123-34-5678', false)` — but `'123-34-5678'` is a validly-shaped `ddd-dd-dddd` string, and actually parses `ok`. Confirmed by running the grammar directly outside the test harness before touching anything. The grammar itself was already correct; only the scratch test's own expectation was wrong, and it never made it into a commit — so there's no `false`→`true` diff here, only the finished, correct `ssn` proof.
- Renamed to `ssn`, with `'123-45-6789'` as the primary (accepting) example, and rounded out coverage with a few more negative cases: too short, no dashes, two different wrong-grouping shapes (`ddd-dddd-dd`, `ddd-ddd-dddd`), and a letter where a digit is required.
- Also extracts `isMatchSuccess`/`expectMatch` (module-scope helpers introduced in #1664 for the `ssn` proof) and reuses them in the four other sections of this file that were duplicating the same success-check/`expect` pattern, per review feedback.

## Test plan
- [x] `npm run cov` — 100% lines/branches/functions (`bnf/ll1` at 100%)

Changelog: none — test-only change, no source behavior or public API is touched.

### DIFF
--- a/fjs/bnf/ll1/proof.f.mjs
+++ b/fjs/bnf/ll1/proof.f.mjs
@@ -3,6 +3,7 @@
  * @import { RuleSet } from '../data/types.ts'
  * @import { CodePointMeta } from '../descent/types.ts'
  * @import { Rule as FRule } from '../types.ts'
+ * @import { Match } from './types.ts'
  * @import { MatchResult } from './types.ts'
  */
 
@@ -17,6 +18,15 @@ import { deterministic, showAst } from '../testlib.f.mjs'
 
 /** @type {(cp: CodePoint) => CodePointMeta<unknown>} */
 const mapCodePoint = cp => [cp, undefined]
+
+/** @type {(mr: MatchResult) => boolean} */
+const isMatchSuccess = ([, success, remainder]) => success && remainder?.length === 0
+
+/** @type {(m: Match) => (s: string, success: boolean) => void} */
+const expectMatch = m => (s, success) => {
+    const mr = m('', toArray(stringToCodePointList(s)))
+    assertEq(isMatchSuccess(mr), success, mr)
+}
 
 /**
  * One grammar matched by both backends, which must build the same AST for it:
@@ -588,17 +598,11 @@ export const proof = {
     ssn: () => {
         const ws = repeat0Plus(' ')
         const d = range('09')
-        const r = (/** @type {number} */n) => repeat(n)(d)
         const ssn = /** @type {const} */([
-            ws, r(3), ws, '-', ws, r(2), ws, '-', ws, r(4), ws])
+            ws, repeat(3)(d), ws, '-', ws, repeat(2)(d), ws, '-', ws, repeat(4)(d), ws])
 
         const m = parser(ssn) // must not throw 'can not merge'
-
-        /** @type {(s: string, success: boolean) => void} */
-        const expect = (s, success) => {
-            const mr = m('', toArray(stringToCodePointList(s)))
-            assertEq(mr[1] && mr[2]?.length === 0, success, mr)
-        }
+        const expect = expectMatch(m)
 
         expect('123-45-6789', true)
         expect('  123  - 45 - 6789 ', true)
--- a/fjs/bnf/ll1/proof.f.mjs
+++ b/fjs/bnf/ll1/proof.f.mjs
@@ -251,13 +251,7 @@ export const proof = {
         },
         () => {
             const m = parser(option('a'))
-
-            const isSuccess = (/** @type {MatchResult} */mr) => mr[1] && mr[2]?.length === 0
-            /** @type {(s: string, success: boolean) => void} */
-            const expect = (s, success) => {
-                const mr = m('', toArray(stringToCodePointList(s)))
-                assertEq(isSuccess(mr), success, mr)
-            }
+            const expect = expectMatch(m)
 
             expect('a', true)
             expect('', true)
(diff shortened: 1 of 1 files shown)


o2alexanderfedin (commented):
The code is sound. I verified the central claim independently — running the grammar outside the harness, `'123-34-5678'` gives `ok:true` with the input fully consumed, so the old `expect(..., false)` was indeed the bug and the grammar was right. Every new assertion is load-bearing: all 8 grammar mutations I tried go red (each `r(n)` arity, each `'-'`, widening `range('09')` to letters, removing `ws`), and flipping any of the 7 expectations at `:603-609` reddens. Each negative case fails for the reason its comment gives — `'123-3456-78'` is 11 chars like a valid SSN and dies at the grouping boundary, not on length. `Changelog: none` is correct; `changelog/README.md` and `AGENTS.md:88-92` both name test-only changes explicitly.

Two things in the title and body, which `CONTRIBUTING.md:171-176` makes part of the deliverable:

1. **The body's grammar isn't the committed one.** Summary bullet 1 gives `[d, d, d, '-', d, d, '-', d, d, d, d]` over `range('09')`. `fjs/bnf/ll1/proof.f.mjs:588-593` commits `[ws, r(3), ws, '-', ws, r(2), ws, '-', ws, r(4), ws]` with `ws = repeat0Plus(' ')`. They differ behaviorally: `'  123  - 45 - 6789 '` is accepted by the committed grammar and rejected by the body's — so the described grammar would fail this PR's own test at `:604`. The whitespace tolerance isn't mentioned at all, and `:604` is the only thing keeping `ws` load-bearing.

2. **"fix a wrong test expectation" isn't in the diff.** The buggy `expect('123-34-5678', false)` was in an uncommitted scratch file; the lone `-1` is the import at `:11`. Bullet 3 also says `'123-34-5678'` was "kept as the primary example", but that string isn't in the committed file — the primary positive is `'123-45-6789'` at `:603`. Someone reading the squash commit will look for a `false`→`true` change and find none.

Title only, no code change: `bnf/ll1: add an SSN grammar proof` fits, with the `'123-34-5678'` story framed as pre-history.

o2alexanderfedin (commented):
The refactor is behavior-preserving and the code is fully proven. `isMatchSuccess` keeps both halves of the old predicate, and I probed them separately — neutering the remainder check reddens, and so does a tautological predicate, so the proof can't pass on partial consumption. All 8 grammar mutations redden (each `repeat(n)` arity, both `'-'`, widening `range('09')` to letters, dropping `ws`, and changing `ws`'s character), and all 7 expectation flips redden. Gates green, `bnf/ll1` at 100%, count 3065 = main's 3064 + the one `ssn` leaf.

`#1664` is fine — it targeted `rtti-bnf` as its base, so `8e9616ee` is a real two-parent merge of a stacked follow-up, not a stray squash. My earlier read of the `(#1664)` suffix was wrong.

Everything left is in the text that becomes the squash commit:

1. **Title** — "fix a wrong test expectation" still describes work not in the diff. The only deletion against main is the import line at `fjs/bnf/ll1/proof.f.mjs:12`; the buggy expectation only ever existed in the uncommitted scratch file.
2. **Bullet 1** — the stated grammar `[d, d, d, '-', d, d, '-', d, d, d, d]` isn't the committed one (`:601-602`), and it would reject `'  123  - 45 - 6789 '`, which `:608` asserts is accepted. The `repeat0Plus(' ')` tolerance is still undocumented — and it's the only thing keeping `ws` load-bearing.
3. **Bullet 3** — `'123-34-5678'` is described as "kept as the primary example" but appears nowhere in the file; the primary positive is `'123-45-6789'` at `:607`.
4. **Test plan** — `3070/3070` is stale; it's 3065.
5. Since `#1664`'s content reaches main only through this squash, the body should mention `isMatchSuccess`/`expectMatch` (`:22-29`). `Changelog: none` stays correct for the combined change — `#1664` is test-only too.

Non-blocking: `expectMatch` has one caller, and four sibling sections (`:255`, `:282`, `:318`, `:585`) still define the predicate inline.

o2alexanderfedin (approved):
All five items closed, and the helper refactor is proven safe.

The four replaced predicates are character-for-character the same expression as `isMatchSuccess` — no site checked a different field, used `==`, dropped the remainder check, or compared against a different success value, and `[, success, remainder]` is exactly `mr[1]`/`mr[2]`. `:278` and `:310` keeping their inline `expect` isn't a half-conversion; `expectMatch` hard-codes the entry name `''`, and those two need `'value'` and `ml(entry, …)`.

Proven by mutation rather than by reading: 41 mutants across all five affected sections, 41 red, no survivors. Neutering the helper reddens across sections both ways — dropping the remainder half reddens four sections, and a tautological predicate reddens all five. So the shared helper is load-bearing everywhere it landed, including in the four pre-existing sections.

On the body: the stated grammar is now character-identical to `:585-587`, and the `ws` claim holds — weakening `ws` to `option(' ')` reddens exactly `'  123  - 45 - 6789 '`, and substituting the previously-documented grammar reddens the same case with the whole string unconsumed. `'123-45-6789'` really is the primary at `:592`, the negative-case list matches one-for-one, the Test plan's three claims all verify, and the `#1664` attribution is right down to the count — exactly four other sections carried a copy of that check. `git log -S'123-34-5678'` against main returns nothing, so bullet 2's "never made it into a commit" is accurate.

Gates green, 3065 = main's 3064 plus the one `ssn` leaf, `bnf/ll1` at 100%. `Changelog: none` correct for the combined change.

PR #1666, opened 2026-08-21
edag/todo: propose optional chaining HCF design

sergey-shandar:
## Summary

- make ordinary `.` / `?.` produce plain property values and introduce `.this` / `?.this` only where receiver HCF is required by an immediate call computation
- retire `.()` as a breaking EDAG vocabulary change in favor of composing `.this` with `()`
- represent optional-chain HCF structurally with continuations and a scoped `['it']` placeholder
- make `['it']` valid only inside an owning optional continuation, with identity-aware validation for ambiguous cross-binding sharing
- make `.this` / terminal `?.this` placement a public EDAG validation invariant, not only a compiler convention
- document grouping, optional-call, nested-chain, and argument-evaluation edge cases

This PR only rewrites the design TODO; implementation remains in the task list.

Changelog: none

### DIFF
--- a/fjs/edag/todo/optional_chaining.md
+++ b/fjs/edag/todo/optional_chaining.md
@@ -188,12 +188,12 @@ carries `this = a` into the call.
 after a call:
 
 ```js
-// f?.(...args)
-['?.()', f, args]
+// f?.(...callArgs)
+['?.()', f, callArgs]
 
-// f?.(...args).x
-['?.()', f, args,
-    ['.', ['it'], 'x'],
+// f?.(...callArgs).x
+['?.()', f, callArgs,
+    ['.', ['it'], x],
 ]
 
 // a?.b?.(...c).d
@@ -213,7 +213,7 @@ the root of the expression:
 // a?.b().c.d
 ['?.this', a, b,
     ['.',
-        ['.', ['()', ['it'], args], c],
+        ['.', ['()', ['it'], ['[]', []]], c],
         d,
     ],
 ]
@@ -223,8 +223,6 @@ the root of the expression:
     ['()', ['.this', ['it'], c], d],
 ]
 ```
-
-Here `args` denotes the complete argument-array operand for `b()`.
 
 #### Continuation scope and DAG identity
 
--- a/fjs/edag/todo/optional_chaining.md
+++ b/fjs/edag/todo/optional_chaining.md
@@ -46,11 +46,16 @@ Make ordinary property access return an ordinary value with no receiver HCF:
 ```
 
 Add `.this` for the exceptional case where the property result must carry
-`this = object` into an immediately following `()` or `?.()`:
+`this = object` into an immediately following call computation:
 
 ```js
 ['.this', object, property]
 ```
+
+Grouping parentheses are transparent for this rule. The relevant question is
+whether the property result is immediately consumed as the callee of `()` or
+`?.()` in the same source expression, not whether the source tokens are
+textually adjacent.
 
 Examples:
 
@@ -72,7 +77,7 @@ const x = ['.', a, b]
 ['?.()', ['.this', a, b], c]
 ```
 
-This replaces `.()`:
+This retires `.()` as a breaking EDAG vocabulary change:
 
 ```js
 // old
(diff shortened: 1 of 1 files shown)


o2alexanderfedin (commented):
The note is factually sound. Every claim about the current system checks out — `.`/`()`/`.()` at `module.f.mjs:226,243`, the `index` restriction matching `edag-stage1-discussion.md:245`, `args` as the whole argument array (`:182-189`), `['%']` defaulting to `['.', ['args'], 0]` against `:693-714`, and the new-scope rule against `:365-370`. Removing `.()` doesn't contradict `:179-181`'s agreed point either, since `.this` + `()` still makes receiver-preserving call representable — only the spelling changes. Every JS example reproduces in node, including the behavioural ones: `null?.b(arg())` evaluates `arg` zero times, `(null?.b)(arg2())` evaluates it once and then throws, and `a?.b.c(...d)` really does take `a.b` as receiver. `todo/README.md` format is followed, links resolve, and dropping `unbind` leaves no dangling references anywhere.

Six things, the first two mechanical:

1. **No `Changelog:` section in the body** — mandatory per `AGENTS.md:85-92`; `Changelog: none` is the right value for a `todo/`-only change (`changelog/README.md:48-52`).
2. **Title** — `CONTRIBUTING.md:178-182` wants `<topic>: <short description>`, lower-case after the colon. Something like `edag/todo: propose optional chaining HCF design` fits at 54 including the suffix.
3. **`:106-110`, `:242` — `%` is already `op2Id`'s binary remainder** (`module.f.mjs:330`), and the note doesn't say so while proposing to add it to `op0`. The collision is real but currently masked: I added `'%'` to `op0Id` and probed — `['%', a, b]` still parses as remainder, but *only* because `_exp` lists `op2` before `op0` (`:70-80`). Flipping that order reparses it as `['%']` and silently drops both operands. `['%', x]` also goes from rejected to accepted. `close-type.md:134-138` is the settled fix — one sentence pointing there would close it.
4. **`:85-86` contradicts its own example at `:64-65`.** The rule says `.this` applies only when the member expression is "immediately followed by" `()` or `?.()`, but `(a.b)(...c)` — where `)` intervenes — is given `.this`. "Modulo grouping parentheses" covers it, and your Problem section at `:29-30` already establishes that parens are transparent here.
5. **`:112-113`, `:185-193`** — `['%']` would be the first node whose value depends on where it sits *within one scope*, which sits against `README.md`'s "memoizes every node by identity within one invocation" and the intra-scope aliasing rule at `:901-904`. Your own example at `:185-190` has two structurally identical `['%']` leaves that must evaluate differently. Task 2 partly covers it; naming the interaction would help whoever implements.
6. **`:267-268`** — Task 3 says "compiler/interpreter plans", but `.()` is also documented in `fjs/edag/README.md:51,62,66`, `fjs/djs/todo/interpret-edag.md:54,99`, `todo/edag-spec.md:139`, and `edag-stage1-discussion.md` — including under "Agreed points (not under discussion)" at `:174-181` and "Decided" at `:935`. Worth naming so those don't get missed.

Minor: `:182` quotes `'x'` where the other examples use bare metavariables, and `:178,203,213` use bare `args` as a metavariable in a file that also discusses the `['args']` node.

sergey-shandar:
> 3. **`:106-110`, `:242` — `%` is already `op2Id`'s binary remainder** (`module.f.mjs:330`), and the note doesn't say so while proposing to add it to `op0`. The collision is real but currently masked: I added `'%'` to `op0Id` and probed — `['%', a, b]` still parses as remainder, but _only_ because `_exp` lists `op2` before `op0` (`:70-80`). Flipping that order reparses it as `['%']` and silently drops both operands. `['%', x]` also goes from rejected to accepted. `close-type.md:134-138` is the settled fix — one sentence pointing there would close it.

Good catch! Let's replace it with something else.

> 4. **`:85-86` contradicts its own example at `:64-65`.** The rule says `.this` applies only when the member expression is "immediately followed by" `()` or `?.()`, but `(a.b)(...c)` — where `)` intervenes — is given `.this`. "Modulo grouping parentheses" covers it, and your Problem section at `:29-30` already establishes that parens are transparent here.

It should state immediately followed by computation.

> 5. **`:112-113`, `:185-193`** — `['%']` would be the first node whose value depends on where it sits _within one scope_, which sits against `README.md`'s "memoizes every node by identity within one invocation" and the intra-scope aliasing rule at `:901-904`. Your own example at `:185-190` has two structurally identical `['%']` leaves that must evaluate differently. Task 2 partly covers it; naming the interaction would help whoever implements.

Like `['frame']` and `['args']` depend on the content, in particular, on the nearest `['=>']` parent. We may remove the association `['%']` with `['.', ['args'], 0]`.

> 6. **`:267-268`** — Task 3 says "compiler/interpreter plans", but `.()` is also documented in `fjs/edag/README.md:51,62,66`, `fjs/djs/todo/interpret-edag.md:54,99`, `todo/edag-spec.md:139`, and `edag-stage1-discussion.md` — including under "Agreed points (not under discussion)" at `:174-181` and "Decided" at `:935`. Worth naming so those don't get missed.

We will retire the `.()` operator. It will be a breaking change.

o2alexanderfedin (approved):
All six closed.

Renaming the placeholder to `['it']` dissolves the `%` problem rather than papering over it — no `%` remains in the file at all. I probed all five proposed tags (`it`, `.this`, `?.`, `?.this`, `?.()`) with `parse(exp)` at arities 1-4: every one is `no match`, so nothing collides, and `['%', 1, 2]` still parses as remainder with no ambiguity left to reason about.

`:55-58` fixes the `.this` rule the right way — "whether the property result is immediately consumed as the callee … not whether the source tokens are textually adjacent" — and it now holds for every `.this`/`?.this` example in the file, including `:73-74` where plain `.` is correctly used instead.

`#### Continuation scope and DAG identity` at `:227-249` addresses the memoization point head-on rather than deferring it: it names the conflict at `:229`, requires lexical binding, declares ambiguous cross-binding sharing invalid EDAG, and ties it to the existing `['args']`/`['frame']` scope rule — with a Task to back it. Task 4 at `:282-286` now names all five files; I re-grepped `.()` across every `*.md` at this head and the hits land in exactly those five and nowhere else.

Every JS example reproduces, including the evaluation-count ones (`null?.b(arg())` zero times, `(null?.b)(arg())` once then throws) and the receiver checks. Every claim about the current schema checks out, tags are spelled uniformly, the summary table agrees with the prose, links resolve, and the `args`→`callArgs` rename removes the clash with the `['args']` node.

Title and `Changelog: none` both conform; `todo/edag: …` and `todo: …` have precedent on main, and `fjs/edag/todo/` is a real directory. Gates green at 3068, matching main.

PR #1667, opened 2026-08-21
edag: document spread in the README, open an operations/HCF todo

sergey-shandar:
Documentation-only follow-up to #1665, which added the `spread` node.

## Summary
- **Covers `spread` in the README's prose and caveats.** #1665 added the node table row; these are the two things a table row can't carry:
  - The call-arguments paragraph now distinguishes the two senses of "spread" that coexist after #1665 — `f(...xs)` still needs no `...` node (`['()', f, xs]` passes the whole array through), while the mixed form `f(a, ...b)` is *newly* expressible as `['()', f, ['[]', [a, ['...', b]]]]`. Both shapes verified against `validate(exp)`.
  - A new caveat records the runtime constraint the shape-only schema cannot express, and which differs by container: in an array the operand must be iterable (`[...1]`, `[...null]`, `[...{a: 1}]` all throw), in an object anything goes with primitives and `null` contributing nothing (`{...null}` is `{}`). Object spread also copies own enumerable properties *through* getters — the opposite of `own`, which reads the descriptor's value and never calls one. Every claim checked in real Node rather than asserted from memory.
- **Adds `fjs/edag/todo/operations.md`**, a working note on hidden control flow (HCF) in chained operations — the `this`-binding and optional-chaining short-circuit cases where the same EDAG shape has to produce different results depending on syntax, tabulated with concrete examples.

## Test plan
- [x] `npm run cov` — 3068/3068 tests, 100% lines/branches/functions

## Notes for review
`fjs/edag/todo/operations.md` is an in-progress design note. Two things reviewers will likely raise, left deliberately for a follow-up rather than pre-empted here:
- It doesn't yet follow the `todo/` issue format (h2 title, `**Priority:**`/`**Status:**`, `### Problem`/`### Tasks`/`### Related`) from [todo/README.md](https://github.com/functionalscript/functionalscript/blob/main/todo/README.md).
- It overlaps `fjs/edag/todo/optional_chaining.md`, whose single open task is exactly this investigation. Whether the two merge, or `operations.md` supersedes it, is an open call.

Changelog: none


o2alexanderfedin (commented):
The call-arguments paragraph checks out on both shapes, and "newly expressible" is right: against the pre-#1665 schema `['()', f, ['[]', [a, ['...', b]]]]` is `no match`, while at this head `parse` preserves `['...','x']` verbatim, so the `spread` alternative of `items` really is what matched. No single-vs-array slip either — the inner array is there, and the control `['[]', a]` correctly fails.

The getter half of the caveat is exactly right, and I checked both directions: `{...{get g(){…}}}` invokes the getter once and yields a plain data property, while `getOwnPropertyDescriptor(o,'g')?.value` invokes it zero times. Non-enumerable own properties are excluded from the spread but still readable by `own`. The iterability claims all reproduce.

Three things:

1. **`fjs/edag/README.md:97-98`** — "with primitives and `null` simply contributing nothing (`{...null}` is `{}`)" is false for strings. `{...'ab'}` is `{"0":"a","1":"b"}`, and `['{}', [['...', 'ab']]]` validates today, so an executor implementer following this caveat would get it wrong. Every other primitive does contribute nothing — number, boolean, bigint, symbol, plus `null`/`undefined` — so it's just strings that need exempting.
2. **`fjs/edag/todo/operations.md:22,23`** — the tag is spelled `'.?'` there but `'?.'` at `:7`, and `optional_chaining.md` uses `?.` throughout. `.?` isn't the JS spelling either.
3. **`fjs/edag/todo/operations.md:7-8`** — `['?.', a, b]` and `['?.()', a, b]` sit in an unmarked "Syntax | EDAG" table beside `'.'` and `'()'`, which do exist. Neither tag is in the schema, and it reads against `optional_chaining.md:8` ("The EDAG operation vocabulary (`.`, `()`, `.()`) has **no representation** for JavaScript's optional chaining"). That file guards its equivalent tables under "Not a proposal"; this one has no such marker. Flagging the factual contradiction only — not the overlap you already called out.

Everything in `operations.md` reproduces in node, including `([].at)(0)` preserving the receiver and `(undefined?.b).c` throwing, and both tables render.

Bookkeeping: the Test plan's 3065 is right for the branch as pushed, but you're 3 commits behind main and `edd3361a` adds three leaves — what lands will be 3068. The test-merge into `3c36c6bd` is clean and keeps main's `[]`-suffix paragraph correctly.

o2alexanderfedin (commented):
The merge is clean — `operations.md` untouched, and the README picks up main's `[]`-suffix paragraph and `items[]`/`properties[]` cells exactly once each, no conflict residue. Gates green, still docs-only.

The count question resolves itself: it's **3068** here now, so the body's 3065 just needs updating.

The three items from my last review are unchanged:

1. `fjs/edag/README.md:104-105` — `{...'ab'}` is `{"0":"a","1":"b"}` and `['{}', [['...', 'ab']]]` validates, so "primitives … simply contributing nothing" still needs a string exemption. The array half at `:102-103` is exactly right.
2. `fjs/edag/todo/operations.md:22,23` — `'.?'` vs `'?.'` at `:7`.
3. `fjs/edag/todo/operations.md:7-8` — `['?.', …]`/`['?.()', …]` still unmarked. Worth noting this is judged against main, where `optional_chaining.md:8` still says the vocabulary has no representation for optional chaining; `#1666` would change that, but it isn't merged.

sergey-shandar:
All three addressed in 9fb80bc2, each reproduced before fixing.

**1. String exemption** — you're right, and the failure mode you describe is the real one: `['{}', [['...', 'ab']]]` validates, so the caveat as written would have sent an executor implementer the wrong way on strings. I checked each primitive individually — number, boolean, bigint, symbol, `null`, and `undefined` all contribute nothing; string is the sole exception, contributing its indices. The caveat now says that, and keeps the array half unchanged since it was already correct (strings being iterable, `[...'ab']` works too).

**2. `.?` → `?.`** — fixed in both combination rows. `?.` matches the first table, `optional_chaining.md`, and JS itself; `.?` was simply a typo.

**3. Unmarked `?.`/`?.()`** — added a note under the first table stating that only `.` and `()` exist in the schema, and that `?.`/`?.()` are candidate tags nothing implements. Phrased to stay true against main, where `optional_chaining.md` still says the vocabulary has no representation for optional chaining — so it won't contradict either that file or #1666 if that lands.

**Bookkeeping** — 3068 confirmed locally; PR body updated.

Still open, and still yours to call rather than something I'd pre-empt: `operations.md` doesn't yet follow the `todo/` issue format (h2 title, `**Priority:**`/`**Status:**`, `### Problem`/`### Tasks`/`### Related`), and its relationship to `optional_chaining.md` — merge, supersede, or keep separate — is undecided.

o2alexanderfedin (approved):
All three closed.

The rewritten caveat holds exhaustively — I ran `{...x}` for every type it names: number, boolean, bigint, symbol, `null` and `undefined` each contribute zero own enumerable properties, and a string contributes exactly its indices. It also dodges the trap the new wording could have introduced: `length` is *not* copied, because on a boxed string it's `enumerable:false`. `{...'ab'}` is exactly `{0:'a',1:'b'}`, and `{...Object('ab')}` agrees. The getter rewording is still accurate against `module.f.mjs:313` and the `ownJs` proof.

`.?` is gone from `operations.md` entirely — all four occurrences are `?.`/`?.()` now — and the new marker at `:10-12` resolves the contradiction with `optional_chaining.md:8` by saying the same thing outright.

Two things you may or may not care about, neither blocking:

- `fjs/edag/todo/operations.md:10` — "Only `.` and `()` exist in the schema today" is exactly right about the four tags in the table above it, but read as a claim about the schema it omits `.()` (`types.ts:77`), which `optional_chaining.md:8` does list. "Of the tags above, only…" would make it precise.
- The body's Test plan still says 3065; it's 3068 here now, same as main.

Gates green, docs-only, 100% coverage.

PR #1669, opened 2026-08-23
djs/tokenizer: derive the tag vocabulary from the grammar

Trinidadec:
Closes the `vocabulary-single-source` todo.

The operator vocabulary and the trivia character lists were each spelled out two or three times, and the copies had already drifted. The todo documents the live case: a phantom `<<<=` operator survived unnoticed because `operatorTags`' hand-copy never matched the tag the grammar actually emitted.

The descent parser emits a variant branch's key as its tag, so the grammar already *is* the vocabulary:

- `operator` moves to module scope — it captured nothing from `buildToken` — and `operatorTags` becomes `new Set(Object.keys(operator))`.
- The whitespace and newline characters become one `wsChars` / `nlChars` string each, feeding both the grammar's `ws` / `newLine` rules and the `wsTags` / `nlTags` membership sets behind `triviaKind`.
- `filterFunc` delegates its four trivia cases to `triviaKind` instead of listing the characters a third time.

Membership is a `Set` rather than `String.includes` deliberately: `scanFunc` calls `triviaKind` with the initial `''` tag, and `'\n\r'.includes('')` is `true`, which would have classified the start state as a newline.

## The `'/'` TODO was stale

`operatorTags` carried a TODO claiming `'/'` made multi-char line comments fall through as two `'/'` operator tokens, because the `oneline` rule consumed only one non-newline character. It named two possible fixes — and the first had already been applied. The rule reads `repeat0Plus(remove(unicodeRange, newLine))` today, so the comment outlived what it described and sent a reader after a bug that isn't there.

Checked before removing it: `//ab\n` yields one `//` token, inner slashes stay inside the comment (`//a//b\n`, `///a\n`, `/*a//b*/`), and `a/b` still tokenizes `/` as an operator. The TODO is replaced with what is actually true — `'/'` is a member because division is an operator — and three proof cases now pin that behavior so the claim is checked rather than asserted.

## Verification

- `npx tsc` — clean apart from two pre-existing errors in `fjs/effects/node/module.mjs` and `tsconfig.json`, both untouched here.
- `fjs t` — 3071 pass, 0 fail (3068 on `main`, plus the three new comment/division cases).
- Diffed the derived tag set against the hand-written one it replaces: 55 keys, identical in both directions. A silent difference there is exactly the failure mode this change is about, so it seemed worth checking rather than trusting the test suite alone.

The todo file is deleted in this PR per the repo's done-issue rule, with its rationale captured in JSDoc on `wsChars`, `operator`, and `operatorTags`. Its one inbound cross-reference in `fjs/media/json/todo/bnf-grammar-single-owner.md` is repointed at the code rather than left to rot.

Related: #1670 adds the DJS parser token-stream adapter. The two are independent and can land in either order.

Changelog: none


sergey-shandar:
@codex review

sergey-shandar (approved):

PR #1671, opened 2026-08-23
bnf: choose a TerminalRange representation for bigint symbols

Trinidadec:
Fills in the [terminal-range-representation](https://github.com/functionalscript/functionalscript/blob/main/fjs/bnf/todo/terminal-range-representation.md) investigation with a recommendation. Documentation only — no code changes. It is the root of the blocked chain: `terminal-range-representation` → `bigint-symbols` → `utf8-token-symbols` → `new-parser`.

Per [AGENTS.md](https://github.com/functionalscript/functionalscript/blob/main/AGENTS.md), a non-trivial feature needs a concrete design in its issue first, so this is up for review before `bigint-symbols` is implemented against it.

## Proposal

A packed bigint with fixed 257-bit halves and an order-preserving `value + 1` endpoint encoding. 257 bits because `encodeTerminal` reaches `2^256`, which needs bit 256 set.

## What actually decides it

Not the codec — the IR. `bnf/data` discriminates its four rule kinds **by JavaScript type alone**: a number is a `TerminalRange`, an array a `Sequence`, an object a `Variant`, a string a `Repeat`, so no shape has to be probed.

That makes a structural range a much larger change than it looks. `[start, end]` collides with `Sequence`, `{ start, end }` with `Variant`, a string form with `Repeat`. Any of them forces every rule kind to become explicitly tagged — rewriting `RuleSet`, both backends, `toData`, and every serialized grammar. A bigint keeps the discriminator a one-word change (`typeof rule === 'number'` → `'bigint'`) at the six sites that test it.

**Structural forms are rejected on that ground alone**, before size or speed enter into it.

Between the two remaining scalar forms, canonicity decides. Fixed-width packing is canonical by construction. A varint pair is not: `0x80 0x00` and `0x00` denote the same value, so one range has several spellings unless canonicity is separately specified *and validated on decode*. For content-addressed data that means several addresses for one grammar — a defect the fixed-width form cannot express.

## Measurements

Measured rather than assumed. Decode throughput, and `rangeDecode`'s share of a real parse (instrumented, tokenizing 24,576 characters of DJS):

| representation | decode | share of parse time |
| --- | --- | --- |
| current, 24-bit in a `number` | 32.6 M ops/s | 10.7% |
| packed bigint, 257-bit halves | 22.0 M ops/s | 15.9% |
| canonical varint pair | 15.4 M ops/s | 22.7% |

Serialized size of the 55 terminal ranges in the DJS tokenizer grammar:

| representation | chars | vs current |
| --- | --- | --- |
| current | 512 | 1x |
| packed bigint | 4356 | 8.5x |
| canonical varint | 221 | 0.4x |

The 8.5x size cost is the one real argument against the recommendation. I accepted it because the absolute number is small — under 4KB for a whole grammar — and grammars are build-time artifacts, not payloads. The varint form wins size exactly where it is cheap to lose, and loses canonicity and hot decode speed where they are expensive.

## An unrelated finding worth acting on first

Instrumenting `rangeDecode` turned up that it runs **28.6 times per input character**: `descent` decodes `ruleSet[name]` on every terminal match attempt, re-decoding the same handful of rules throughout a parse. That is 10.7% of parse time today.

Decoding each terminal once at parser construction — beside the `emptyTags` pass that already walks the whole `RuleSet` — removes it from the hot path entirely. Worth doing **before** the bigint migration: it is a win at the current representation too, and it is the difference between the bigint move costing +5pp and costing nothing. Recorded as follow-on work on `bigint-symbols` rather than done here.

## Verification

Documentation-only, so no behavior to test. The measurement instrumentation was applied to `fjs/bnf/module.f.mjs` temporarily and reverted; `npx tsc` shows only the two pre-existing errors in `fjs/effects/node/module.mjs` and `tsconfig.json`, and the working tree is otherwise unchanged.

Changelog: none

### DIFF
--- a/fjs/bnf/todo/terminal-range-representation.md
+++ b/fjs/bnf/todo/terminal-range-representation.md
@@ -1,7 +1,7 @@
 ## Investigate TerminalRange representation
 
 **Priority:** P3
-**Status:** open
+**Status:** open — investigation complete, proposal awaiting review
 
 ### Problem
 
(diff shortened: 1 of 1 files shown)


Trinidadec:
Superseded by #1672: the registered `token_symbol` alphabet is kept, so `bigint-symbols` is on hold and this investigation has no consumer.

The conclusion, for whenever it revives: a packed bigint with fixed 257-bit halves and a `value + 1` endpoint encoding. Structural forms are rejected because `bnf/data` discriminates rule kinds by JavaScript type alone, so a tuple or record collides with `Sequence`/`Variant`/`Repeat`. Varint is rejected on canonicity — one range would have several spellings, so content-addressed data would have several addresses.

Measured: decode 32.6 M ops/s today vs 22.0 packed vs 15.4 varint; the DJS grammar's 55 terminal ranges serialize to 512 chars today vs 4356 packed vs 221 varint.

The one finding here that is independent of the migration and still worth doing: `descent` calls `rangeDecode` 28.6 times per input character, re-decoding the same rules throughout a parse — 10.7% of parse time. Hoisting it into parser construction is a win at the current representation.

PR #1675, opened 2026-08-23
edag: pin optional chaining against its naive desugaring

sergey-shandar:
Adds a helper and two cases to `chainsJs` in `fjs/edag/proof.f.mjs`, running the JS the chain nodes are built around — the `ownJs`/`chainsJs` pattern.

## Summary
- **`a?.at` is not `a !== null && a !== undefined ? a.at : undefined`.** `desugarOptionalAt` is that desugaring, guarding both nullish values exactly as `?.` does. It is faithful on every input it returns: `undefined` for `null` and for `undefined`, and for anything else the *same function* `a.at` denotes — `desugarOptionalAt([42]) === [42].at`. So no value it produces is wrong.
- **What it loses is the receiver.** `chainsJs.throw.desugaredOptional` calls the last of those values, and it throws, where `(a?.at)(0)` is `42` (`chainsJs.receiver`). Lowering an optional link to a conditional would therefore lose `this` — which is why `(a?.b)(d)` keeps `?.b` as a step of the call's own `lambdas`, `['()', a, [['|?.', 'b']], d]`, rather than completing an `['?.', …]` node that hands on an ordinary value.
- **The assertions live in the ok case, `chainsJs.desugaredOptional`, not in the throw case.** A `throw` case passes on any exception, so a failing `assertEq` placed there is masked — verified by injecting a deliberately false one and watching the suite stay green. The ok case also takes every branch of the helper.

## Test plan
- [x] `npm test` — 3099/3099
- [x] Coverage gate green — 100% lines/branches/functions
- [x] Verified in node against `(a.at)(0)` and `(a?.at)(0)` controls, both `42`, and `null?.at`/`undefined?.at`, both `undefined`
- [x] Mutation-verified in a clean worktree: dropping the `null` guard reddens `chainsJs.desugaredOptional`; binding the receiver (`o.at.bind(o)`) reddens both `chainsJs.desugaredOptional` and `chainsJs.throw.desugaredOptional`, since the ok case pins the function identity the bind changes

Changelog: none


sergey-shandar:
All four findings addressed in 34cd044b — the consolidation removes the section rather than patching it, so three of them stop existing.

**Duplication (codex, `:529`)** — agreed, and done exactly as suggested: only the conditional-desugaring case survives, now `chainsJs.throw.desugaredOptional`, placed beside `detachedReceiver` since it is the same receiver rule reached by a different route. Its comment ties it to the node shape it constrains. The other six are gone; `ok[0]`/`ok[2]` were also duplicates of each other.

**`ok[1]` asserting nothing (codex, `:533`)** — correct, it passed on any non-throwing result. Rather than adding the `assertEq`, the case is dropped: `chainsJs.shortCircuit` already pins that exact spelling as `assertEq(u?.at.name, undefined)`.

**Unused `y` (`:554`) and always-true guard (`:559`)** — both gone with the removed cases. The surviving case avoids the second by construction: the guard takes its operand as a parameter, so it is not statically known, and the case `return`s rather than assigning to a discarded binding.

Net diff against main is now +11 lines, all one case.

Verification: `desugar([42])(0)` throws while both `(a.at)(0)` and `(a?.at)(0)` return `42`, so the throw is receiver detachment rather than anything about `a`. Mutation-checked — binding the receiver in the desugaring (`o.at.bind(o)`) stops the throw and reddens exactly this case and nothing else. Gates green at 3098/3098.

sergey-shandar:
Fixed in 29152546 — the branch really was unproved, thanks. I took a different placement than suggested, for a reason worth recording.

**Asserting inside the `throw` case would not have worked.** A `throw` case passes on *any* exception, so a failing `assertEq` there throws and reads as the expected failure. I checked rather than assuming: injecting a deliberately false `assertEq(desugar(undefined), 12345)` into the throw case left the suite green at 3098/0. Branch coverage would have been satisfied while the assertion proved nothing.

**What landed instead.** The desugaring is now a module-level `desugarOptionalAt`, with both branches asserted from a new `chainsJs.desugaredOptional` **ok** case, where a throw reddens:

```js
desugaredOptional: () => {
    assertEq(desugarOptionalAt(undefined), undefined)
    assertEq(desugarOptionalAt([42]), [42].at)
},
```

`throw.desugaredOptional` keeps only the call, `desugarOptionalAt([42])(0)`.

The second assertion also sharpens what the case is for: the desugaring returns the *same function* `a.at` denotes (both are `Array.prototype.at`), so neither branch produces a wrong value — the receiver is the only thing lost. That is the whole argument for `?.` not lowering to a conditional.

**Mutation-verified both ways.** Breaking the nullish branch (`: undefined` → `: null`) reddens exactly the ok case and nothing else — the check the masked placement could not have made. Binding the receiver (`o.at.bind(o)`) still reddens the throw case.

Also moved the new case *above* the "call counterpart of `throw.groupedOptional`" comment, which documents the `throw` block and had ended up separated from it.

Gates green: `tsc` clean, 3099/3099, `ci-update` no diff.

sergey-shandar:
@codex review

sergey-shandar:
Fixed in dad7815d. Good catch — this was a real bug in the helper, not just an incomplete assertion.

`?.` short-circuits on **either** nullish value, but my guard tested `undefined` only, so `desugarOptionalAt(null)` fell through to `null.at` and threw, where `null?.at` is `undefined`. The helper was therefore not the desugaring of anything, which contradicted its own comment — the one claiming `undefined` on a nullish input.

Now:

```js
const desugarOptionalAt = o => o !== null && o !== undefined ? o.at : undefined
```

with the `null` case asserted alongside the other two, so every branch is still taken:

```js
desugaredOptional: () => {
    assertEq(desugarOptionalAt(null), undefined)
    assertEq(desugarOptionalAt(undefined), undefined)
    assertEq(desugarOptionalAt([42]), [42].at)
},
```

The `&&` spelling rather than `!= null` matches the house style in `fjs/asserts/module.f.mjs:67` (`a !== null && a !== undefined`); loose equality is not used anywhere in `fjs/`.

Mutation-verified: restoring the `undefined`-only guard reddens exactly `chainsJs.desugaredOptional` and nothing else, so the `null` assertion is load-bearing — the reported bug is now caught rather than merely fixed.

Gates green: `tsc` clean, 3099/3099, `ci-update` no diff.

sergey-shandar:
@codex review

o2alexanderfedin (commented):
The description still describes commit `34cd044b`, not the head. Since title+body become the squash commit, three of its claims land on `main` as false.

**The desugaring it quotes is the one the head commit removed as a bug.** The Summary says `a?.at` is not `a !== undefined ? a.at : undefined`, but `fjs/edag/proof.f.mjs:92` is `o !== null && o !== undefined ? o.at : undefined`, and the helper's own JSDoc at `fjs/edag/proof.f.mjs:84-86` says a guard on `undefined` alone "would not be the desugaring of anything" — that is exactly why `dad7815d` exists. The commit message would assert the buggy form as the point of the change.

**It says "adds one case to `chainsJs.throw`"; the diff adds two cases and a helper** — `chainsJs.desugaredOptional` at `fjs/edag/proof.f.mjs:509`, `chainsJs.throw.desugaredOptional` at `fjs/edag/proof.f.mjs:531`, and `desugarOptionalAt` at `fjs/edag/proof.f.mjs:92`. The Summary never mentions the ok case, which is where the null/undefined/receiver-identity assertions actually live.

**Both test-plan numbers are stale.** At head `npm test` is 3099/3099 and `npm run cov` reports 3037 tests (base `f1dd81d3`: 3097 and 3035) — 3098 is the `34cd044b` count. And "reddens exactly this case, nothing else" is no longer true for the `o.at.bind(o)` mutation: it reddens two, `chainsJs.desugaredOptional` and `chainsJs.throw.desugaredOptional`, because `fjs/edag/proof.f.mjs:512` asserts `desugarOptionalAt([42]) === [42].at`.

sergey-shandar:
All four corrected — the body is rewritten against the head commit rather than patched. Thanks for catching that it was still describing `34cd044b`; as the squash commit it would have put the removed bug on `main` as the point of the change.

1. **The quoted desugaring** is now the one that exists: `o !== null && o !== undefined ? o.at : undefined`. The old text quoted the `undefined`-only form that `dad7815d` removed precisely because it is not the desugaring of anything.
2. **"one case to `chainsJs.throw`" → a helper and two cases.** The Summary now leads with the ok case, `chainsJs.desugaredOptional`, since that is where the null/undefined/identity assertions live, and says why they are there rather than in the throw case.
3. **Numbers refreshed**, with one discrepancy I could not reproduce. `npm test` is 3099/3099 — confirmed. But the coverage gate reports **3099** for me too, not 3037; I ran it at `dad7815d` in a clean `git worktree` (my working tree has untracked scratch files that break discovery, so I isolated rather than measure through them). Since I can't reproduce 3037 and you can, the difference looks environmental, so the body now claims the coverage gate is green at 100% lines/branches/functions and does not print a second, disputed count. If you'd rather it carried your number, say so.
4. **The mutation claim is corrected, and you're right about why.** `o.at.bind(o)` reddens both cases, because the ok case pins `desugarOptionalAt([42]) === [42].at` and a bound function is a different reference. Re-verified in the worktree — exactly those two, `fail 2`. The body now states both, and separately notes that dropping the `null` guard reddens only the ok case.

Unrelated, since it cost me the local gates: the "3037 vs 3099" investigation surfaced that `npm test` fails outright on an unpushed working-tree file — `fjs/edag/todo/vm.f.mjs:42` has `RequiredMap<Op0Id, unknown}` with no closing `>`, so `tsc` stops with TS1005 and the coverage runner discovers 1 test instead of 3099. Untracked and not part of this PR, so I left it alone.

o2alexanderfedin (approved):
Re-reviewed at `f707780b`; the three stale-description findings are fixed, and the delta since `dad7815d` is comment-only (the other files in that range came in via the `main` merge). Reproduced the masking claim independently rather than taking it on trust: a deliberately false `assertEq(0, 1)` inside `chainsJs.throw.desugaredOptional` leaves the suite green at 3099/3099 with the case still reporting `# EXPECTED TO THROW`, while the same assertion in the ok case `chainsJs.desugaredOptional` reddens it — so the stated reason for the assertions' placement holds, and `fjs/emergent_testing/module.f.mjs:369-370` inverts the result for `throws` exactly as described. Both stated mutations land as claimed: `o.at.bind(o)` reddens exactly `chainsJs.desugaredOptional` and `chainsJs.throw.desugaredOptional` (`fail 2`), and dropping the `null` guard reddens only the ok case (`fail 1`); the ok case does take all three helper branches, `desugarOptionalAt([42]) === [42].at` holds in node, and `['|?.', 'b']` matches the two-element `LambdaPropertyAccessor` at `fjs/edag/types.ts:81`. Gates green — `tsc` clean, 3099/3099, coverage exit 0 at 100/100/100, `ci-update` no diff.

PR #1679, opened 2026-08-24
djs/parser: recover the token-symbol input layer

Trinidadec:
Recovers the `new-parser` token-symbol input layer, which was merged as #1670 but never reached `main`.

## What happened

#1670 was stacked on #1672, targeting `bnf/keep-registered-token-symbols` rather than `main`. The two merged in the wrong order:

- **19:12:25** — #1672 merged into `main`.
- **19:33:55** — #1670 merged into `bnf/keep-registered-token-symbols`, 21 minutes later.

By then that base branch had already delivered its content to `main` and was not being merged anywhere again, so #1670's code landed on a branch with no route onward. GitHub retargets a stacked PR when its base merges, but only while the PR is still open — this one was merged into the stale base instead.

The result: `main` carries #1672's todo changes, and the task list even shows the token-layer tasks ticked, while none of the code they describe is present. `splitEof`, `ordinaryTokenNames`, `tokenEncoding`, `tokenToSymbol` and their proofs are all absent from `main`.

## What this restores

A merge of `bnf/keep-registered-token-symbols` onto current `main`, clean and with no conflicts. The diff against `main` is exactly the four parser files:

- `fjs/djs/parser/module.f.mjs` — the EOF adapter, the 26-name alphabet, the encoding, and the token-to-symbol mapping
- `fjs/djs/parser/types.ts` — `_FramingKeyword`, `_OrdinaryTokenName`
- `fjs/djs/parser/proof.f.mjs` — the framing-keyword and lexical-error parity proofs
- `fjs/bnf/todo/new-parser.md` — the task ticks

Nothing else moves. An earlier `git diff main..branch` looked alarming — it appeared to revert EDAG and nanvm work — but that was only the branch being 38 commits behind; the three-way merge takes `main`'s side on all of it, and the stat above confirms it.

## Verification

- `npx tsc` — clean apart from the two pre-existing errors in `fjs/effects/node/module.mjs` and `tsconfig.json`.
- `fjs t` — 3116 pass, 0 fail, with all 18 recovered proofs running.

## Note for next time

Stacked PRs need the base to merge **first**, then the stacked PR retargets to `main` and merges second. Merging the stacked PR into an already-merged base strands it silently — nothing fails, and the todo ticks make it look done.

Changelog: none


MaryManzhula:
@codex review

sergey-shandar (approved):

PR #1685, opened 2026-08-24
edag/amnesia: evaluate chains — `()`, `?.`, and `?.()`

sergey-shandar:
## Summary

The three nodes that own a `lambdas` were unevaluated: `?.` and `?.()` were
`todo`, and `()` called `todo()` whenever its chain operand was non-empty, so
only a plain `f(...args)` worked. All three now execute, and nothing in
`fjs/edag/amnesia/module.f.mjs` is `todo` any more.

A chain carries two things no `exp` value can (`../README.md`, "Chains"): the
receiver a property step hands to a following call, and the short-circuit an
optional link starts. Both live in one walk, `applyLambda`, over a state that
says what is known about the current value so far:

| state | meaning |
| --- | --- |
| `undefined` | the region short-circuited on a nullish link |
| `readonly [any]` | a bare value, no receiver |
| `{ obj, prop }` | an accessor, left unresolved so `this` survives the call |

Keeping the accessor unresolved is the point: `value` reads `obj[prop]`, while
`call` invokes `obj[prop](...args)` so a method gets its receiver. Each step
id is one case — `|.` makes an accessor, `|()` calls and clears the receiver,
and `|?.`/`|?.()` are those two behind a nullish guard that returns before
evaluating its own operand.

Every node is then that walk plus what it does with the result:

```js
'()':    call (property(applyLambda(i, c,                             [i(b)])),  () => i(d))
'?.':    value(property(applyLambda(i, [['|?.', index], ...lambdas],  [i(a)])))
'?.()':  value(property(applyLambda(i, [...lambdas, ['|?.()', args], ...tail], [i(a)])))
```

`property` turns a short-circuited region back into the value `undefined`, and
the one character of difference is the whole distinction between the nodes:
`()` ends its region at the parentheses, so that `undefined` is what gets
called and the node throws, while `?.` and `?.()` own their regions, so it is
the node's value. `u?.b(d)` is `undefined`; `(u?.b)(d)` throws.

An empty `lambdas` needs no special case — `reduce` hands the seed back — so
`f(...args)` is the chain of no steps.

## Two bugs found by proving it

**A receiver invented for a bare value.** `call`'s bare-value branch spelled
the call `p[0](...)`, which is a *method* call on the one-element state array,
so a value with no receiver ran with that wrapper as `this`. A detached host
method then read from the wrapper instead of throwing: `((a.at)(0))(0)`
evaluated to `Array.prototype.at`. Now called through a temporary, which
carries a comment saying why it cannot be inlined back.

**`?.` dropped its own index.** The handler destructured `index` and never
used it, so only the trailing `lambdas` ran: `a?.b` evaluated to `a`, and
`[1, 2, 3]?.[1]` to the array. The nullish path looked correct, which is what
hid it — the failure needs a property actually present. That step is now the
first of the region.

## Proofs

`fjs/edag/amnesia/proof.f.mjs`, grouped by node:

- **`chain`** (`()`) — a property step and its receiver, a call step clearing
  it, both optional steps, all three `index` forms, and composed steps.
  `chain.throw` holds every nullish case, because under `()` a short-circuit
  always ends in calling `undefined`.
- **`optionalPropertyAccessor` / `optionalRegion`** (`?.`) — the node's own
  step, then the half `()` cannot state: the short-circuit *returns*, so the
  skip is directly observable. A skipped step's index, a skipped call's
  arguments, and the node's own index all stay unevaluated.
- **`optionalCall` / `optionalCallRegion`** (`?.()`) — the receiver through
  both spellings of the first `lambdas`, the second one running on the call's
  result, and the three ways into the short-circuit.
- Mirrors under `throw`: operands that *are* evaluated on a present value, and
  `?.()` guarding nullish rather than non-callable.

Cases pair across nodes wherever the two readings differ — `u?.b(...c)` is
`undefined` next to `(u?.b)(...c)` throwing. Amnesia produces the specified
throw on every host, so these can be stated at all, where `../proof.f.mjs`'s
`chainsJs` has to leave them out: JavaScriptCore carries the short-circuit
through the parentheses.

`fjs/edag/amnesia/module.f.mjs` is at 100% lines, branches, and functions.

## Documentation

`amnesia/README.md`'s "Not implemented" section now records that every node
evaluates, and no longer claims `(u?.b)(d)` is a corner inherited from the
host — the walk decides it on every host. `edag/README.md`'s note on that
boundary points at the proof that pins it. `fjs/edag/todo/lambda.md`, which
held the design, is deleted per `AGENTS.md` §1.

## Testing

- `npx tsc` — clean.
- `npm test` — 3181 pass, 0 fail.
- `npm run cov` — `all files | 100.00 | 100.00 | 100.00`.
- `npm run ci-update` — no drift.
- Mutation checks, since a proof that cannot fail proves nothing. Restoring
  `p[0](...f())` reddens the two detached-receiver cases; restoring the
  dropped `?.` index reddens three; and for `?.()`, dropping the `|?.()` step
  reddens four, weakening it to `|()` reddens the short-circuit section, and
  reordering it before the first `lambdas` reddens two.


Trinidadec (commented):
delete lambda.md

Trinidadec (approved):

o2alexanderfedin (commented):
The implementation is well-proven — I built 9 mutations and every one reddens a specific new proof, including all five results you list. Three things to fix.

**1. `Property` and `Hcf` need the `_` prefix.** Both land in the emitted declaration:

```ts
export type Property = readonly [any] | { readonly obj: any; readonly prop: any; };
export type Hcf = undefined | Property;
```

They're the only contract delta vs `main` (`vm`'s signature is unchanged). `AGENTS.md:250` requires the underscore for implementation-only typedefs, `_Func1` in this same file follows it, and `43ca8f71` already applied that fix here once.

Worse, `Property` collides with the exported `Property` in `fjs/edag/types.ts` (`readonly[':', Exp, Exp]`) — two unrelated public `Property` types in `fjs/edag`, and this module imports `Properties` from that file. `_Accessor` / `_Callee` would kill the collision as well as the visibility problem. Both names are new here, so nothing was previously exported under them.

**2. No changelog entry and no `Changelog:` section.** `AGENTS.md:88` makes the section mandatory in every PR description, and this changes published behavior — `vm` on `['?.', …]`, `['?.()', …]` and `['()', …]` with non-empty `lambdas` went from `todo()` to evaluating, in a module shipped via `files: ["**/*.mjs"]`. Precedent: `1677.md`, `1673.md`. `changelog/unreleased/` has 16 files, none for 1685.

**3. Optional steps read the accessor twice.** `option` evaluates `value(hcf)` for the nullish check, then `g()` re-reads `obj[prop]` through `lazyDot`/`call`. Measured against this head with a getter counter on a host `frame` object:

| expression | getter reads | JS |
|---|---|---|
| `frame?.g?.h` | 2 | 1 |
| `(frame.g?.m)()` | 2 | 1 |
| `frame.o?.m()` | 2 | 1 |

Reachable through `Context.frame`/`args`, which take arbitrary host values. Raising it because your own `detachedReceiver` fix in this PR holds to exactly this host-observable standard. Fix shape: have `option` read once and pass the value on — `|?.` becomes `option(obj => ({ obj, prop: f(e) }))`, and the `|?.()` branch then invokes the already-read callee with `Reflect.apply(callee, hcf.obj, f(e))` rather than re-reading through `call`.

**Minor:** the description says `fjs/edag/todo/lambda.md` "is deleted per AGENTS.md §1", but that file never existed on `main` — it was added and removed inside the branch (`4e9dcd60`). The diff is 4 files with no deletion.

**Verified clean:** `npm test` 3181 pass / 0 fail, `npx tsc` clean. Removing the `hcf === undefined` early-out is caught by tsc rather than a proof (`TS2345`), which is the better outcome. `@module` survives declaration emit; no `any`/`elided` in `vm`'s signature. All doc citations resolve (`README.md:91`, `proof.f.mjs:469/477/490`, `amnesia/proof.f.mjs:330`). The one new case I couldn't redden with a non-contrived mutation is `throw.evaluatedArgument`.

PR #1691, opened 2026-08-25
dev: build allFiles effects declaratively

sergey-shandar:
### Motivation

- Replace an imperative, accumulator-reassigning directory-entry loop in `allFiles` that repeatedly copied the growing effect array and violated `AGENTS.md` preferences for declarative code.
- Use the repository's typed `fromEntries` helper instead of destructuring the native `Object.fromEntries` to keep typed helpers consistent across the codebase.

### Description

- Rewrote `allFiles` in `fjs/dev/module.f.mjs` to map each directory entry to zero-or-one effects using `d.flatMap`, preserving hidden-file and `node_modules` filtering while avoiding repeated array copies.
- Added an import for `fromEntries` from `fjs/types/object/module.f.mjs` and removed the previous `Object.fromEntries` destructuring.
- Removed the completed issue file `fjs/dev/todo/dev-allfiles-flatmap.md` as the change implements that todo.

### Testing

- Ran `npx tsc` and it succeeded.
- Ran `fjs test` and the test suite passed (3228 tests passed, 0 failed).
- Ran `cargo test`, `cargo clippy`, and `cargo fmt -- --check` and the Rust tests and checks passed.
- Ran `npm run update` but `deno` was unavailable in the environment so the Deno-install step could not run; no generated artifacts were left uncommitted.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a8dd18264b4832ba45bfc5e2dd0ead6)

Changelog: none


sergey-shandar:
@codex review

o2alexanderfedin (commented):
Behaviour-preserving — verified where it counts. `allFiles` drives test discovery, so a refactor that quietly finds *fewer* files still passes green. Ran `loadModuleMap` against the real tree on both heads: **261 keys on `main`, 261 here, `diff` byte-identical** — same paths, same order, which also pins the `Object.fromEntries` → typed `fromEntries` swap as key-order-preserving. `npm test` 3228/0 on both; every emitted `.d.mts` byte-identical between the two (no drift, no `any`, `@module` intact); `npm run cov` 100%.

**One thing to fix:** `Changelog: none` is a bullet inside `### Description`, followed by `### Testing`. `AGENTS.md:88-93` asks for a description *ending* in the section, and the body becomes the squash commit — #1687/#1688/#1689 all end with it as a trailing block. The content is right; only the placement.

**Worth a line of proof:** deleting the `if (name.startsWith('.')) { return [] }` guard **survives** the full suite — 3228/0, unchanged. No proof root has a dot-named entry, and on the real tree `.git`/`.github` hold nothing matching `shouldLoad`, so the branch is executed under `cov` (hence 100% branches) but nothing pins it. Pre-existing — `main`'s `continue` had the same gap — but since this PR rewrites exactly that loop, `'.hidden.f.ts': []` in `allFilesFindsFunctionalScript`'s root would close it in one line. The other two guards are pinned: dropping the `node_modules` skip fails `allFilesSkipsNodeModules`; inverting `predicate(file)` aborts the runner.

Equivalence otherwise holds per-entry: same three guards in the same order, `flatMap` preserves iteration order, empty dir yields `allOk()` both sides, recursion unchanged, and read/import failures still propagate (`loadModuleMapReportsUnreadableRoot` / `…UnimportableModule` pass). `readdir` is declared `readonly Dirent[]` (`effects/node/types.ts:122`), so `.flatMap` is contract-guaranteed rather than relying on the runner returning an Array. The O(n²) `result = [...result, x]` accumulator is genuinely gone, and no dangling reference to the deleted todo remains.

Not verified: symlink handling — reasoned only. The `i.isFile` guard is byte-for-byte the same decision as before, so whatever `readdir` reports is treated identically on both heads, but no proof covers it on either side.

Move the `Changelog:` to the end and this is an approve.

sergey-shandar:
@codex review

o2alexanderfedin (commented):
The guard is pinned now — deleting `if (name.startsWith('.')) { return [] }` fails `allFilesFindsFunctionalScript` (3227/1) where it previously survived at 3228/0. One line, and it closes a gap that coverage couldn't see: the branch was executing under `cov`, just not being asserted on.

Delta is that one line. Discovery still 261 keys, identical to `main`. `npm test` 3228/0, tsc clean.

Still open: `Changelog: none` is a bullet inside `### Description`, with `### Testing` after it. `AGENTS.md:86-93` wants the description to *end* in the section, since it becomes the squash commit — #1687/#1688/#1689 all place it as a trailing block. Move it after `### Testing` and this is an approve.

o2alexanderfedin (approved):
Fixed — `Changelog: none` is now the last line of the body.

The delta is a pure rebase: diff-of-diffs against the merge base is byte-identical to `7b689f26`'s content, so the code is unchanged and the earlier verification carries over — hidden-file guard pinned (deleting it fails `allFilesFindsFunctionalScript`), discovery still 261 keys matching `main`, declarations byte-identical. `npm test` 3256/0, same as `main`.

Approving.

PR #1695, opened 2026-08-25
cas: reuse collectRead in round-trip proofs

sergey-shandar:
### Motivation

- The CAS round-trip proofs duplicated a three-case stream-fold implementation (`drain`) that the module already exposes as `collectRead`, causing proof code to bypass the production abstraction. 
- Reusing the production helper reduces duplication and prevents divergence between production code and its proofs.

### Description

- Replace the locally inlined `drain` implementations with `collectRead(c.read(hash))` in both the single-chunk and multi-chunk CAS round-trip proofs in `fjs/cas/proof.f.mjs`.
- Compare the returned `Vec` directly (use the `collectRead` result) instead of concatenating a locally-collected chunk list.
- Remove the now-completed `fjs/cas/todo/proof-drain-collectread.md` issue file.

### Testing

- Ran `npx tsc` and it succeeded.
- Ran the FunctionalScript proofs for CAS via the test runner (`node ./fjs/module.mjs t ./fjs/cas`) and the suite passed (no failures reported).
- Ran `cargo clippy` and `cargo fmt -- --check` for the Rust side and both checks succeeded.
- Ran `git diff --check` and it reported no whitespace or merge-marker issues.

Changelog: none

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a8df0c12cdc832bbd34f30f7297aa2f)


o2alexanderfedin (approved):
Reuse is assertion-preserving, verified rather than assumed.

Old `drain` accumulated chunks then compared `msb.listToVec(chunks)` to the content; `collectRead` concatenates the same chunks and compares the same bytes — identical assertion, and it additionally enforces the `maxLength` cap. No error-branch loss: both stop at the first failed cell via `ioStep`, and the `assert(readResult[0] === 'ok', …)` guard is intact. Proof count unchanged at 3256; both call sites converted, none removed.

The check that matters for a proof-reuse change is whether the reusing proofs still catch corruption. Two mutations: duplicating each chunk in `fileCas.read` (the module under test, not the helper) → 40 failures; dropping the concat inside `collectRead` → 35. Both reverted. `npx tsc` clean, `npm test` 3256/0. `Changelog: none` is right for a proof-only change and the section is present.

Approving.

PR #1697, opened 2026-08-25
effects/node/virtual: centralize entity kind checks

sergey-shandar:
### Motivation

- The virtual filesystem implementation repeatedly re-derived the same three-way entity discrimination (`Vec[] | Dir | JsModule`) at many call sites, producing duplicated logic and scattered casts.  
- Centralizing the checks makes intent explicit, removes fragile inline tests and `as` casts, and reduces cognitive overhead when reasoning about file/dir/module branches.  
- The change is a small, safe refactor that preserves behavior while improving readability and maintainability.  

### Description

- Add three named narrowing predicates: `isBinFile`, `isJsModule`, and `isDir` next to the `Entity` import in `fjs/effects/node/virtual/module.f.mjs`.  
- Replace the dozen+ ad hoc structural checks across traversal, file reads/writes, directory listing, rename, byte I/O, and stat handling with calls to the new predicates, and update messages/guards accordingly.  
- Remove the completed todo file `fjs/effects/node/virtual/todo/name-entity-kind-discrimination-once.md`.  

Changelog: none

### Testing

- `npx tsc` — passed.  
- `npx fjs test` — passed (test suite run; 3,256 tests passed in this environment).  
- `cargo clippy` — passed.  
- `cargo fmt -- --check` — passed.  
- `git diff --check` — passed.  
- `npm run update` — attempted; Node/npm stages ran but the environment lacked `deno` so the full update step could not complete (environment warning).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a8e0a1fae78832b86bcfca14be6a0c1)


o2alexanderfedin (approved):
The refactor is sound. I checked it as a behavior-preservation claim rather than reading it: extracted every old and new predicate expression across the 12 call sites and compared them over the full `_Entity | undefined` domain. Eight of the ten distinct conditions are identical on every input. Two diverge only on `undefined`, and both are unreachable — `readdir`'s `isFile` is guarded by `if (content === undefined) { continue }` on the line above, and `insertEntityAt`'s `entityIsDir` takes its entity from `extractEntity`'s ok payload, which returns `enoent` for an undefined entry. `null` agrees everywhere too.

Since no proofs changed, I mutation-tested to confirm the existing ones actually cover the touched branches. All three mutants die: dropping the `JsModule` exclusion from `isDir` → 23 failures, neutering `isBinFile` → 140, neutering `isJsModule` → 23.

Deleting the `instanceof Array` vs `Array.isArray` comment is right, not doc drift — I rewrote `isBinFile` to use `Array.isArray` and `tsc --noEmit` still exits 0, so the explicit `@returns {entity is readonly Vec[]}` predicate makes the old rationale moot. The todo deletion is warranted; all three of its tasks are done and no `as Dir` / `as readonly Vec[]` casts remain.

`npm test` 3256/0, `tsc --noEmit` clean after purging stale `.d.mts`.

**One thing to fix before merge:** the `Changelog:` section isn't last. The body runs Motivation → Description → `Changelog: none` → `### Testing` → trailer. `### Testing` is a content section, not a trailer, so per CONTRIBUTING.md:186 it belongs above the changelog. The content is correct — `none` is right for an internal refactor. It matters only because the body becomes the squash commit and CONTRIBUTING says not to repair it in the merge box.

Pre-existing and not caused by this PR, but now slightly worse: `dir-spine-descend.md` and `resolve-file-helper.md` quote the `Array.isArray` snippets this PR rewrote, and `node-program-options-factory.md` cites `module.f.mjs:434`/`:442` where the real lines are now 475/483 (they were already stale at 460/468 on base). Also, the deleted todo's "Related" note about the `result[0] === 'error'` duplication / `propagateError` idea goes away with the file and isn't captured anywhere else.

Approving on the code; please reorder the body first.

sergey-shandar:
@codex review

o2alexanderfedin (approved):
This closes exactly the caveat I put on my last approval ("equivalent absent cross-realm arrays"), and the gap was worse than I assumed. I built a real `node:vm` array and ran the new proof's scenario on both commits:

```
Array.isArray(foreign)   = true
foreign instanceof Array = false

base 0c21ef0c:  rename("myfile","mydir")  ->  ["ok", undefined]
head 0d357cf9:  rename("myfile","mydir")  ->  ["error", ["ioError", {"message":"'mydir' is a directory"}]]
```

The base doesn't merely misclassify — it **silently succeeds**. The foreign array falls through to `isDir`, so `rename` treats a file as a directory and reports `ok`. That's a wrong answer, not a missing error. Only `isBinFile` needed the flip; `isDir` derives from it and inherits the fix, and `isJsModule`'s `typeof === 'function'` was already cross-realm-safe.

Mutation: reverting to `entity instanceof Array` fails exactly one test, `renameForeignArrayFileOntoDirectory`, and nothing else — so the proof kills the mutant with no collateral.

Worth knowing, not a defect: the proof doesn't construct a real cross-realm array. It uses `Reflect.construct(Array, [], Object)` as a surrogate, and the comment says so. I checked the surrogate against an actual `vm` array on all three discriminating observables (`Array.isArray` true, `instanceof Array` false, `typeof` `'object'`) and it's faithful for these predicates while keeping the proof `vm`-free.

The fix is complete within the file — all thirteen entity discriminations now route through the three predicates, no raw checks left. The `instanceof` uses elsewhere aren't gaps: `effects/node/proof.f.mjs:190-193` is deliberate with its own comment about narrowing a `readonly` array out of the union, `instanceof Promise` is documented at `types.ts:303`, and `toIoError`'s `instanceof Error` degrades via `String(e)`.

`npm test` 3257/0 (one more than the body's 3,256 — the new proof), tsc clean, no stale declarations to purge.

**The body ordering from my last review is still unfixed** — `Changelog: none` still sits above `### Testing` rather than last before the trailer (CONTRIBUTING.md:186). Same ten-second edit; flagging it once more only because the body becomes the squash commit. Secondary: this commit does change observable behavior, which sits a little oddly with `Changelog: none`, though the behavior belongs to an internal fixture that isn't exported, so `none` is defensible. The order is the actual violation.

Approving.

PR #1700, opened 2026-08-25
fjs: document file-type naming conventions

sergey-shandar:
### Motivation

- The repository lacks a single canonical description of filename conventions for FunctionalScript modules, proofs, host-side modules, and type-only companions, which causes confusion when introducing new conventions. 
- The open todo asked whether to introduce a `node.app.*` application suffix and where the canonical doc should live, so the design needs to be resolved before any implementation work. 

### Description

- Updated `fjs/todo/document-file-type-naming-conventions.md` to clarify current conventions and to recommend documenting them in `fjs/README.md` adjacent to the CLI and `main`-export guidance. 
- The change distinguishes authored FunctionalScript modules (`module.f.mjs`) from host JavaScript modules (`module.mjs`) and explicitly identifies `types.ts` as the type-only companion, and it clarifies proof file spellings and that `.f.js` is a future planned spelling. 
- The file records a decision to not introduce a `node.app.*` filename suffix and to keep the existing named `main` export convention for applications, plus a short checklist for adding the final docs and removing the todo. 
- Changelog: none

### Testing

- Ran `git diff --check` and inspected the resulting diff, which reported no formatting or whitespace errors and shows the intended content changes. 
- Verified repository file patterns with `find` and `rg` to ensure the documented spellings match existing files under `fjs/`. 
- Performed regular VCS operations (`git add` / `git commit`) to create the change; those commands succeeded. 
- The TypeScript build and test suites were not run because this is a documentation/design change only and touches no executable code.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a8e203eeca4832b96952f450597e168)

### DIFF
--- a/fjs/todo/document-file-type-naming-conventions.md
+++ b/fjs/todo/document-file-type-naming-conventions.md
@@ -21,8 +21,9 @@ and `main`-export conventions:
   spelling authors use today.
 - `module.mjs` — vanilla JavaScript host integration. It may use capabilities
   outside FunctionalScript and may run effects at import time.
-- `types.ts` — authored, type-only TypeScript companion. It has no runtime
-  representation and is imported only with `import type` or JSDoc `@import`.
+- `types.ts` — authored, type-only TypeScript module. It may stand alone when a
+  declaration has no runtime representation, or accompany a runtime module,
+  and is imported only with `import type` or JSDoc `@import`.
 
 ### `proof.*` — a module that proves other modules
 
--- a/fjs/todo/document-file-type-naming-conventions.md
+++ b/fjs/todo/document-file-type-naming-conventions.md
@@ -13,17 +13,26 @@ the file types or points to the other.
 Document the file-type conventions in `fjs/README.md`, next to the existing CLI
 and `main`-export conventions:
 
-### `module.*` — a module
+### `module.*` — a package entry point
 
-- `module.f.mjs` — authored FunctionalScript module: pure by construction and
-  safe to bulk-load. `.f.js` is the planned spelling for parser-compatible,
-  dependency-closed groups once authored-`.f.js` package support exists, not a
-  spelling authors use today.
-- `module.mjs` — vanilla JavaScript host integration. It may use capabilities
-  outside FunctionalScript and may run effects at import time.
-- `types.ts` — authored, type-only TypeScript module. It may stand alone when a
-  declaration has no runtime representation, or accompany a runtime module,
-  and is imported only with `import type` or JSDoc `@import`.
+- `module.f.mjs` — a package entry point authored in FunctionalScript: pure by
+  construction and safe to bulk-load.
+- `module.mjs` — a package entry point authored as vanilla JavaScript host
+  integration. It may use capabilities outside FunctionalScript and may run
+  effects at import time.
+
+The `module` basename marks the package entry point, not every module. Other
+authored FunctionalScript sources use a descriptive basename with the same
+`.f.mjs` marker, such as `bnf.f.mjs` and `json.f.mjs`; other host JavaScript
+sources similarly use `.mjs`. `.f.js` is the planned FunctionalScript spelling
+for parser-compatible, dependency-closed groups once authored-`.f.js` package
+support exists, not a spelling authors use today.
+
+### `types.ts` — a type-only module
+
+An authored, type-only TypeScript module. It may stand alone when a declaration
+has no runtime representation, or accompany a runtime module, and is imported
+only with `import type` or JSDoc `@import`.
 
 ### `proof.*` — a module that proves other modules
 
--- a/fjs/todo/document-file-type-naming-conventions.md
+++ b/fjs/todo/document-file-type-naming-conventions.md
@@ -3,7 +3,12 @@
 **Priority:** P3
 **Status:** open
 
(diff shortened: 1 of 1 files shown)


o2alexanderfedin (approved):
Checked every rule against the tree rather than reading the prose, and found **zero counterexamples** — which is unusual for a convention document. Notably: all 93 `.ts` files under `fjs/` are named `types.ts`, with no exceptions, and a scripted check confirms not one of them has a runtime export. `module.mjs` is 5 files, all genuinely host-side (`types/result/module.mjs` uses `try`/`catch`, which FunctionalScript forbids). Proof spellings are exactly `proof.f.mjs` ×124 and `proof.mjs` ×1, and that one — `effects/node/memory/proof.mjs` — does import `asyncRun` from the host module, so its stated rationale holds. Dropping `proof.ts` / `proof.js` / `proof.mts` is a fix rather than a loss: all three have zero instances. `NodeProgram = Program<NodeOp>` is verbatim at `effects/node/types.ts:418`, and the code fence matches `fjs/module.mjs` byte for byte.

**One thing worth acting on: `fjs/AGENTS.md` §3.5 (lines 935-952) already owns three of these bullets, close to verbatim** — that `types.ts` carries no runtime implementation, that `.f.mjs` consumes it through JSDoc `@import` while TypeScript uses `import type`, and that compiler support doesn't gate the `.f.mjs` → `.f.js` rename.

Your own Proof checklist has the right instinct — link to the existing compiler, testing and `fjs run` docs "instead of duplicating their details" — but it omits §3.5, which is the actual existing owner. Adding §3.5 to that list would close the loop; otherwise the new `fjs/README.md` section becomes a second copy of a rule that already exists, and the repo's `CLAUDE.md` is explicit about where that leads ("two copies drift"). Relatedly, the unchanged line 6 — "There is no single place that defines them" — is now roughly half false given §3.5.

Not contradicted, just outside the doc's scope (it doesn't claim exhaustiveness): `types/bigint/benchmark.mjs`, `emergent_testing/all.test.mjs`, and six fixture `.f.mjs` that are neither `module` nor `proof` (`m.f.mjs`, `input.f.mjs`, `example.f.mjs`, `bnf.f.mjs`, `json.f.mjs`, `testlib.f.mjs`) backing the README compile examples.

Discoverability is fine — this edits an existing todo, linked from two siblings, and the anchor `#proof--a-module-that-proves-other-modules` still resolves after the edit.

`npm test` 3266/0, tsc clean, the one relative link resolves.

**The PR body needs a fix before merge, and it's two violations rather than one.** `- Changelog: none` is a bullet inside `### Description` rather than a standalone section, and `### Testing` then follows it. CONTRIBUTING.md:186-188 wants a `Changelog:` section that is the last section before an optional trailer block. The content is right; promoting it to its own section after `### Testing` and before the `[Codex Task]` trailer fixes both. Flagging it because the body becomes the squash commit — this is the third PR this week with the same shape, so it may be worth fixing in whatever template generates these.

Approving on the content.

sergey-shandar:
@codex review

sergey-shandar:
@codex review

o2alexanderfedin (approved):
Both points from my last review are resolved.

**§3.5 is now named and linked** — `[fjs/AGENTS.md §3.5](../AGENTS.md#35-functionalscript-module-rules)`. The citation is accurate: `fjs/AGENTS.md:935` is exactly `### 3.5 FunctionalScript module rules`, and the anchor is the correct GitHub slug, matching the style AGENTS.md uses for its own internal links.

**The half-false line is gone**, and its replacement is true. I checked both halves of the new claim: `fjs/README.md` has the `fjs run` and `Convention: export const main` sections and contains zero references to AGENTS.md; `fjs/AGENTS.md` links `./fsc/README.md`, `./effects/README.md` and `../changelog/README.md` but never `./README.md`. So "neither gives a reader a short map of the file types or points to the other" holds exactly.

The one substantive rewording — `.f.js` timing, from "after the compiler's extension migration" to "parser-compatible, dependency-closed groups once authored-`.f.js` package support exists" — is a precision improvement backed by the tree rather than invented: `AGENTS.md:951-952` says compiler support does *not* gate the rename while parser coverage and package support do, `ci/todo/f-js-package-support.md:88` has the first rename blocked by package support, and `fsc/README.md:13` confirms `.f.js` in stage 1 is generated rather than authored — which is why "not a spelling authors use today" stays right.

The `70fe56d3` follow-up refines `types.ts` from "has no runtime representation" to "may stand alone when a declaration has no runtime representation, or accompany a runtime module". That's compatible with the tree — all 93 `types.ts` files still have no runtime exports.

Re-swept at the new head: `module.f.mjs` 131, `proof.f.mjs` 125, `types.ts` 93, `module.mjs` 5, `proof.mjs` 1, and zero tracked `*.f.js`, `node.app.*` or `*.f.ts`. Still no counterexample. Links resolve, no line-number citations to go stale, `npm test` 3266/0, tsc clean.

**The PR body is unchanged on the one thing I asked for.** `- Changelog: none` is still the fourth bullet inside `### Description`, and `### Testing` still follows it before the `[Codex Task]` trailer; CONTRIBUTING.md:186-188 wants a standalone `Changelog:` section, last before the trailer. The content is right — this only needs the body edited, not a new commit. Third time raising it, and I'd guess it's the template rather than you: several PRs this week have produced the identical shape, so fixing whatever generates these bodies would close it once instead of per-PR.

Approving.

sergey-shandar:
@codex review

PR #1703, opened 2026-08-26
cli: render command help only when needed

sergey-shandar:
### Motivation

- Avoid constructing the aligned help table on the common successful dispatch path by separating presentation from routing, reducing unnecessary work and keeping `dispatch` focused on lookup and routing.

### Description

- Hoist the help-table formatting into a pure module-scope `renderHelp` helper in `fjs/cli/module.f.mjs` and call it only on the missing-command, explicit-help, and unknown-command branches; preserve existing outputs and behavior.
- Remove the resolved todo file `fjs/cli/todo/dispatch-help-rendering.md`.

### Testing

- Ran the repository checks: `npx tsc` succeeded, `node ./fjs/module.mjs test` ran the full FunctionalScript test suite (3,272 passed, 0 failed), `cargo clippy` and `cargo fmt -- --check` succeeded, and `git diff --check`/`npx tsc` showed no issues; `npm run update` ran but Deno was not available in the environment so its install step could not execute (no tracked changes resulted).  Changelog: none

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a8e47dc5a70832bb1c9cd948398368e)


o2alexanderfedin (approved):
The laziness is real, which is the thing worth checking on a change like this — "only when needed" is easy to claim while still evaluating eagerly. `renderHelp` isn't exported, so I instrumented it in-tree and drove it through the virtual effects runner with nested command tables:

```
                        base   head
flat success   echo hi     1  ->  0
alias success  e hi        1  ->  0
nested         sub ping    2  ->  0
depth-3        g m leaf    3  ->  0
help sub                   2  ->  1
sub help                   2  ->  1
no args / help / bogus     1  ->  1
```

Zero renders on every success path, exactly one on every help path. Note the base cost scaled with nesting depth — `dispatch` recurses per level and each level rebuilt the table — so head removes work that grew with the tree, not just work that happened early.

Output is byte-identical across 36 invocations, stdout, stderr and exit code captured separately (md5 `1e0965cd…` on both). The matrix came from the two registries rather than guesswork: no-args, `help`/`h`/`?`, `--help`, `-h`, unknown command, unknown flag, `help <x>` for every registered name plus `bogus`/`help`/`toString`, the `cas` subtree including `cas help add`, and the four arity errors. 22 exits 0, 14 exits 1, same on both.

End-to-end still works, not just help: `compile`/`c` to `.json` and `.mjs` produce identical bytes, `run` behaves the same on args / bare / missing file / non-program module, and a full `cas add` → `list` → `get` roundtrip under an isolated `HOME` gives the same hash and restores the content.

Four mutations, three caught. The one that survived — dropping `padEnd(width)` alignment — is a pre-existing assertion gap by construction, since no proof file changed and the expression moved verbatim; my byte-diff is what covers formatting there. Worth recording the reason I checked at all: at base those `renderHelp` lines ran on *every* dispatch and were trivially covered, while at head they only run on help paths, so coverage could have regressed silently. It didn't — neutering `renderHelp` still fails 6 proofs.

`npm test` 3272/0 on both, tsc clean, `git diff --check` clean, and the implementation matches the deleted todo's proposal verbatim.

Two things to fix, neither in the code:

**This PR introduces a dangling link.** `fjs/cli/todo/options-edsl.md:71` still points at the file being deleted:

```
- [dispatch-help-rendering](./dispatch-help-rendering.md) — the help
  rendering that declared options must plug into.
```

**The `Changelog:` placement, in a third variant.** Here it's appended inline at the end of a bullet inside `### Testing` — `...(no tracked changes resulted).  Changelog: none` — so it isn't a standalone section at all. AGENTS.md §5 wants the description to end in a `Changelog:` section. This is a different shape from the two I've flagged on sibling PRs this week (the `### Testing`-after-Changelog form, and the bullet-inside-Description form), which reinforces that it's worth fixing whatever generates these bodies rather than each PR in turn. The content `none` is correct — output is byte-identical and this is an internal refactor plus a todo deletion.

Approving on the code.

PR #1705, opened 2026-08-26
mcp: deduplicate cas_get metadata mapping

sergey-shandar:
### Motivation

- Remove duplicated verdict→wire mapping logic in `cas_get` to make the code clearer and avoid subtle inconsistencies. 
- Make the `cas_get` handler reuse a single adapter for streaming, dialect-refined, and inline-content paths. 
- Close the local todo that described this internal duplication now that it is addressed.

### Description

- Add a curried `toMeta` adapter that converts a media-type detector verdict into the `_Meta` wire shape given a per-request `uri` and reuse it across the `cas_get` handler. 
- Replace three local copies of `mime_type → mimeType`, `Number(length)`, `type`, and `uri` assembly with `toMeta(uri)(...)`, keeping the wire format unchanged. 
- Reuse the adapter for the streaming verdict, the buffered dialect-refine path, and the inline-content path so the mapping logic lives in one place. 
- Remove the completed todo file `fjs/cas/todo/66n-cas-get-response-shape.md`.

### Testing

- Ran `git diff --check` to validate formatting and whitespace (passed). 
- Ran `npx tsc` for TypeScript checks (passed). 
- Ran `node ./fjs/module.mjs test` to exercise the FunctionalScript proofs (ran; no failures observed). 
- Ran Rust checks (`cargo clippy`, `cargo fmt -- --check`) (ran; no failures observed). 
- Ran `npm run update` which progressed through dependency steps but stopped at `deno install` because Deno is not available in the environment (partial; informational only).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a8e670b1188832bbcf04e8d7ed52930)


o2alexanderfedin (approved):
Verified the response shape end to end through the MCP layer rather than at the function boundary — `mcpStep` + `casMcpHandlers` on the virtual node runner, the same path `fjs/mcp/proof.f.mjs` uses, with the harness kept outside both worktrees so no stray `proof` export joined the runner. **51 `cas_get` calls, 103 JSON-RPC responses, byte-identical** between base and head: empty blob (both spellings), single byte, unicode, escape-heavy text, all three dialects, a dialect-shaped-but-odd revision, PNG and ZIP magic, NUL-bearing valid UTF-8, invalid UTF-8, 90 KB text and binary, text at exactly `maxLengthBytes` and one byte over, multi-chunk text-then-binary, a three-chunk blob for chunk-boundary agreement, plus absent/malformed/wrong-length/empty hash and a non-boolean `content`.

**Two of the three sites were textually identical; the third wasn't, and that's worth knowing.** The metadata-only refine at `:247` had **mixed provenance**:

```js
// base:  length and type from the STREAMING verdict, mimeType from the buffered re-detect
return pureOk(okResult(toJson({ ...meta, mimeType: refined.mime_type })))
// head:  all three from refined
return pureOk(okResult(toJson(meta(refined))))
```

So that site genuinely changed rather than being merged verbatim. It's still behaviour-preserving, and I checked why instead of assuming: `detectDialect` takes `length`/`type` straight from `detectVec`, and `detectVec` and `detectStream` are the same machine — both `finish(…push…)` with per-byte steps, so folding chunk-by-chunk equals folding the whole `Vec`. The one theoretical gap, a ragged chunk getting zero-padded by `u8List`, is closed empirically: `fileCas.read` always issues whole-byte reads. Key order survives either way. Arguably the new form is better — it can no longer pair a mimeType from one verdict with a length from another.

Five mutations, all caught, but the split matters. Renaming `mimeType` and dropping `uri` are caught by **tsc**, which aborts before any proof runs — that's gate coverage, not demonstrated behavioural coverage. The two that reach the suite (`meta(refined)` → `meta(detected)` at each site) each fail **exactly one** proof. So the path whose provenance this PR actually changed is pinned by a single proof — covered, but thinly. My harness diverged from base under all five, which doubles as a negative control that the byte comparison discriminates at all.

The todo is genuinely resolved: task 1 matches the proposed signature essentially verbatim, task 3 is what I verified above. Task 2 said to *evaluate* folding the two `collectRead` + `detectDialect` arms and apply only "if it stays readable" — declining is legitimate under that wording, though nothing records that the evaluation happened. No dangling links: `git grep -i -e 66n -e cas-get-response-shape` at the base finds only the todo itself, every other hit being a hex bigint literal.

`npm test` 3311/0, tsc clean, and `toMeta` is exactly what `AGENTS.md:755` prescribes — captures lifted into leading curried parameters and hoisted to module scope.

**One thing to fix: the body has no `Changelog:` section at all.** It ends `### Testing` bullets → `------` → the Codex trailer. That's a fourth distinct shape this week, after the `### Testing`-after-Changelog form, the bullet-inside-Description form, and the inline-at-the-end-of-a-bullet form — which I think settles that the body template is the thing to fix rather than each PR. The content `none` is right: 51 byte-identical cases prove nothing protocol-visible changed.

Minor: the description says "replace **three** local copies", but only two were the duplicated verdict→wire mapping the todo described; the third was the mixed-provenance spread above.

Not verified: the comparison is at the collected-`Response` level rather than raw stdout through `stdioTransport` — that layer is untouched here and its proofs still pass. `content:true` at exactly `maxLengthBytes` was skipped since base64 inflation overflows the transport regardless of this PR; it's bracketed on both sides by cases that are identical.

Approving on the code.

PR #1706, opened 2026-08-26
djs/transpiler: use mapStep for context projection

sergey-shandar:
### Motivation

- Make an explicit pure projection where the transpiler maps a folded context into the updated `ParseContext` so the effect shape is clear and does not widen the operation set.  
- Avoid the `step(..., x => pureOk(...))` pattern that looks like an effectful continuation but is actually a pure projection, and record the conversion in the outstanding `map-step-combinator` TODO.  

### Description

- Replace the `step(..., context => pureOk(...))` pattern with `mapStep(...)` in `fjs/djs/transpiler/module.f.mjs` for the import-context projection and the module-to-context projection, preserving semantics while keeping the operation set unchanged.  
- Adjusted imports in `fjs/djs/transpiler/module.f.mjs` to use the consolidated `effects` export (`mapStep`, `pure`, `step`, etc.).  
- Return a plain `ParseContext` value from the projection continuation instead of wrapping it in `pureOk(...)`, since `mapStep` expects a projection that yields the pure value.  
- Update `fjs/effects/todo/map-step-combinator.md` to record that the `djs/transpiler` projection was converted and to clarify that nearby `pure(Result)` continuations remain channel constructors and must not be converted.  

### Testing

- Ran `npx tsc` (TypeScript check) and it succeeded.  
- Ran the FunctionalScript test suite via `node fjs/module.mjs test` and all tests passed (3313 passed, 0 failed).  
- Ran Rust checks: `cargo clippy` and `cargo fmt -- --check`, both succeeded.  
- Ran `npm run update`, which completed repository update steps but stopped partway where `deno` is required in this environment; the change itself was exercised by the TypeScript check and the full test suite above.  

Changelog: none

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a8e6fe30288832b842ea56a1544f964)

### DIFF
--- a/fjs/djs/transpiler/module.f.mjs
+++ b/fjs/djs/transpiler/module.f.mjs
@@ -23,8 +23,7 @@ import { concat as pathConcat } from '../../path/module.f.mjs'
 import { parseFromTokens } from '../parser/module.f.mjs'
 import { parse as jsonParse } from '../../media/json/module.f.mjs'
 import { run } from '../ast/module.f.mjs'
-import { pure } from '../../effects/module.f.mjs'
-import { catchStep, foldStep, mapStep, pureError, pureOk, step } from '../../effects/module.f.mjs'
+import { catchStep, foldStep, mapStep, pure, pureError, pureOk, step } from '../../effects/module.f.mjs'
 import { readUtf8File } from '../../effects/node/module.f.mjs'
 
 /**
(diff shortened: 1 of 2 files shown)


o2alexanderfedin (approved):
No revert — the branch is just cut immediately before #1705 landed. The PR's only non-merge commit touches exactly `djs/transpiler/module.f.mjs` and `effects/todo/map-step-combinator.md`, and `refs/pull/1706/merge` against `main` shows only those two files, keeping #1705's `toMeta` rewrite and the deleted CAS todo. (Flagging it because the raw `main..head` diff makes it look otherwise.)

Behaviour is byte-identical: 30 invocations per side, 118 artifacts, `diff -r` clean. 12 success and 18 error cases, including circular dependency, self-import cycle, transitively-missing import and a syntax error inside an imported module — so the error branch through the converted site really was exercised, with identical text and exit codes.

Reducing both spellings to `resultStep` shows why: on error the identical tuple is passed through by identity in both, no rebuild and no retag, so a `ParseError` reaches `errorLocation` unchanged, and `g` never runs on the error branch either way. `mapStep` is also right on typing grounds — the projection issues no commands, so the operation set doesn't widen the way `step` would allow.

Three mutations of the rewritten site, all caught (29 / 1 / 29 failures). `npm test` 3313/0 on both sides, tsc clean, `Changelog: none` in the correct shape.

One small correction for the body, since it becomes the squash message: it says both "the import-context projection and the module-to-context projection" were converted, but `transpileModule` was already `mapStep` on `main` — this converts one site. The todo edit is honest, and its claim that the remaining two `pure(Result)` sites are channel constructors rather than `mapStep` candidates checks out — `mapStep` there would double-wrap into `ok(Result)`.

Approving.

PR #1711, opened 2026-08-26
edag: restructure chain nodes as a grammar with three lambda types

sergey-shandar:
## Summary

Implements `fjs/edag/todo/chain-node-grammar.md`. The flat `lambdas` array is replaced by a recursive grammar: four chain nodes and three mutually recursive continuation types, so the uniqueness of a chain spelling is enforced by the schema rather than by a lowering convention plus a validation pass.

## The two bits

A chain carries exactly two kinds of hidden control flow, and the continuation types are the states of those two bits:

- **P** — a receiver is live
- **O** — a short-circuit region is open

|  | outside an option | inside an option |
| --- | --- | --- |
| **receiver live** | `propertyLambda` | `optionPropertyLambda` |
| **value only** | *(an `exp`)* | `optionLambda` |

Three types and not four: `00` is the definition of a node boundary, so the fourth cell is an ordinary `Exp`.

## Nodes

```
['()',   exp, exp]                        // call, no receiver, no region
['.',    exp, index, propertyLambda]
['?.',   exp, index, optionPropertyLambda]
['?.()', exp, exp,   optionLambda]
```

A method call is the property-access node *owning* its call, which is what carries the `this` binding: `a.b(...c)` is `['.', a, 'b', ['|()', c, null]]`, while `(0, a.b)(...c)` is `['()', ['.', a, 'b', null], c]`.

## Continuations

Every step tag carries the `|` prefix, and every production has uniform arity with an explicit `null` terminator:

```ts
type PropertyLambda =
    | null
    | readonly['|()',   Exp, null]
    | readonly['|?.()', Exp, OptionLambda]

type OptionLambda =
    | null
    | readonly['|()', Exp,   OptionLambda]
    | readonly['|.',  Index, OptionPropertyLambda]

type OptionPropertyLambda =
    | null
    | readonly['|()',   Exp,   OptionLambda]
    | readonly['|.',    Index, OptionPropertyLambda]
    | readonly['|?.()', Exp,   OptionLambda]
    | readonly['|!()',  Exp,   null]
```

Both are correctness requirements, not readability ones, and the todo document called the first a prerequisite blocker:

- **The prefix.** Unprefixed, `['()', f, null]` is simultaneously a well-formed call node — call `f` with `null` as its arguments — and a well-formed `optionLambda` — call the chain's value with `f` as its arguments, and stop. The two readings have the same length, so `close` cannot separate them; only disjoint vocabularies can. Pinned by `lambdas.tagsAreDisjoint`.
- **Uniform arity.** A two-element terminal handed a real continuation would validate as the terminal with the rest silently dropped. Pinned by `lambdas.terminalsAreExplicit`.

`|!()` is new — the call a group puts *outside* an optional region, `(a?.b)(...c)`, and the one step a short-circuit does not skip. `|?.` is gone: a guarded property access always starts a `?.` node.

## What stops being expressible

The four duplicate families are unspellable, not rejected — one `unspellable` proof each:

| family | why |
| --- | --- |
| `a?.b?.c` | no lambda has a `?.` production |
| `a.b(...c)?.d` | `propertyLambda`'s `\|()` is terminal |
| `(a?.(...b))(...c)` | `optionLambda` has no `\|!()` |
| `a?.b(...c)?.(...d)` | `optionLambda` has no guarded step |

Every tuple in the schema is now `close`d, so "exactly one spelling" is literal rather than "up to trailing elements", and the static types match the runtime ones exactly — no `TupleTs` approximation. `['args', 'extra']` no longer validates.

## Where the evidence comes from

`|!()` is the production V8 and JavaScriptCore disagree about, so no JavaScript oracle establishes it on every runner. It is pinned by evaluating the node instead — `optionRegion.throw.closeStepOnUndefined` — alongside its non-throwing twin.

One gap is stated rather than closed, and it is stated in the code. `a.b(...c)` against `(a?.b)(...c)` on a nullish base differ only in whether the arguments ran; both throw, this language has no mutation for a skipped operand to record itself with, and a `throw` case is pass/fail rather than payload-inspecting (`fjs/AGENTS.md` §1.5). So the order is carried structurally: `callProperty` takes the argument *node* and evaluates it inside the call expression, where JavaScript's own order applies. Review of this PR found the earlier thunk-based spelling admitted a surviving mutant; this is the fix.

## Also

- Supersedes `chain-nodes.md`; both todo files are deleted, and their laws, the parenthesis-law asymmetry, and the engine split move into `fjs/edag/README.md` § Chains.
- `Index` kept over `Exp` in the naming positions — the two overlap (`['Number', e]` is both a `numberCast` and an `op1`), so widening buys a second spelling of every computed key and no expressive power.
- Stale `lambdas` spellings updated in `fjs/djs/todo/` and `todo/edag-stage1-discussion.md`.
- `npx tsc` clean, declaration emit clean, 3383/0, and `fjs/edag/**/module.f.mjs` at 100% line/branch/function coverage.

Changelog:

- **BREAKING CHANGES:** `edag`: chains are a grammar, not an array of steps. `lambdas` is gone; `.` and `?.` carry an `index` plus a continuation, `?.()` a continuation, and `()` is now `['()', exp, exp]` with no chain at all — so `a.b(...c)` is `['.', a, 'b', ['|()', c, null]]`. Three continuation types (`propertyLambda`, `optionLambda`, `optionPropertyLambda`) make the duplicate spellings unspellable rather than merely discouraged.
- **BREAKING CHANGES:** `edag`: new `|!()` step — the call a group puts outside an optional region, `(a?.b)(...c)` — replacing `|?.`, which no longer exists: a guarded property access always starts a `?.` node.
- **BREAKING CHANGES:** `edag`: every node tuple is `close`d, so a trailing element is now a validation error — `['args', 'extra']` no longer validates.

https://claude.ai/code/session_01Y5eGe4abCH5yqLQJtRhz43


o2alexanderfedin (commented):
The design holds. I checked the central claim independently — the one #1707 failed — and **minimality is structural here**: all 6324 derivable terms (≤5 ops) validate against the real `validate(exp)`, and all 15 non-minimal or stale spellings I could construct are rejected *by the schema*, not by a runtime pass. The bijection reproduces exactly: 172 / 1044 / 6324 terms, same number of distinct behaviour signatures, 0 duplicates, 0 holes, against independently generated JS. Behaviour vs merge-base and vs real V8: 1406 cells, 0 divergences in all three directions. `npm test` 3381/0, `tsc` clean. Doc deletion is clean — no dangling links, unresolved link targets went 33 → 24.

(Harness self-test, since this trap has fired twice before: nullish at depth 0 only manufactures 64/646/5016 phantom collapses, and dropping `this` identity manufactures 34/234/1490. The zeros above are with the full basis.)

Three things:

**1. The PR body has no `Changelog:` section.** CONTRIBUTING makes it mandatory and it becomes the squash commit on `main`. There is a prose `## Breaking Changes` heading with different wording and no `**BREAKING CHANGES:**` markers, so the marker never reaches `main`. `changelog/unreleased/1711.md` itself is correct — three items, all correctly prefixed; I confirmed the closedness item empirically (`['args', 'extra']` no longer validates). This is the same body defect #1707 had.

**2. The body documents the pre-fix design.** It spells the lambda types with unprefixed tags — `PropertyLambda: null | ['()', exp] | ['?.()', exp, optionLambda]`, `['!()', exp]` — which is the two-element terminal and the colliding vocabulary the deleted doc itself called a prerequisite blocker (`['()', f, null]` validating as both a node and a lambda). Shipped code is `['|()', exp, null]`, `['|.', …]`, `['|!()', exp, null]`. As written the permanent commit message describes a design that cannot be implemented. Also: title should be `edag: restructure chain nodes …` — the branch commit `53077ae6b` is correctly prefixed, the PR title isn't.

**3. One surviving mutant, and the deleted doc predicted it.** Making `argsOf` eager passes the entire suite — 3381/0, 0 failures. It is not equivalent: it differs from head on 27 of 1406 corpus cells, always by evaluating an argument list V8 leaves untouched (`a.b(...c)` on nullish `a` logs `a1`; V8 logs nothing). It contradicts `argsOf`'s own docstring — "JavaScript reads the callee before it evaluates the arguments" — so the invariant is documented but unproven.

`chain-node-grammar.md` recorded exactly this as an open task and the file is deleted with it undone: *"a proof comparing values and throw kinds alone would pass while the distinction silently disappeared."* Both sides exist at head (`chain.throw.propertyOnUndefined`, `optionRegion.throw.closeStepOnUndefined`) but both use `noArgs`, so nothing observes whether the arguments ran.

The obvious fix doesn't work — `boom` is `null['x']`, so both readings raise `TypeError` and `ev()` is payload-blind per AGENTS.md §1.5. What does kill it, verified: an argument operand with a distinguishable throw payload, e.g. `['own', ['{}', []], 1]`, plus a proof inspecting the caught value. Head then gives `TypeError` for `['.', undef, 'at', ['|()', mark, null]]` and the `own` payload for `['?.', undef, 'at', ['|!()', mark, null]]`; the eager reading gives the `own` payload for both.

Worth knowing: the other 11 mutants all died, but M5–M11 (receiver binding, guard placement, short-circuit propagation) were each killed by exactly one assertion. The margin there is one proof wide.

Two smaller items:

- **`chain.throw.optionalCallOnAbsentProperty` has no head counterpart under any name.** Head keeps only the non-throwing skip variant (`optionRegion.skips`); the throwing close-after-guard-on-absent-property is gone. #1707 dropped a family silently, so flagging it — if it's deliberate, fine, but assertion budgets otherwise line up (`edag/proof` 142 → 176, `amnesia/proof` 150 → 150 byte-identical despite the rewrite).
- `optionPropertyLambda`'s `|?.()` production reached through a `|.` step is unpinned — no `['|.', …, ['|?.()', …]]` shape exists anywhere under `fjs/edag/`.

Nits: `proof.f.mjs:604` cross-references `chain.throw.closeStepOnUndefined`, but the export is `optionRegion.throw.…` (missed rename; `README.md:263` and `amnesia/proof.f.mjs:359` get it right). `README.md:233` family-4 row is labelled `a?.b(...c)?.d` but the stated reason and `unspellable.guardedStepAfterCall` both cover `?.(...d)` — the property variant is blocked by family 1. `module.f.mjs:408,426-428` `dot`/`optionDot` examples put a bare `exp1` in the index slot; `index` is `numberCast | string | number`, so a computed key must be `['Number', e]` (`['.', a, ['args'], null]` is rejected). `closed.extraTailIsError` covers a trailing element on `|()` but not on `|.`, `|?.()` or `|!()`.

sergey-shandar:
All addressed in `ca702fd`, except that I took a different route on the mutant — details below.

**1 & 2 — PR title and body.** Both fixed. You were right about the cause: the body was written from the todo document's *pre-fix* grammar, so it shipped the unprefixed tags and the two-element terminal that the document itself called a prerequisite blocker. The body now states the prefix and the uniform arity as the correctness requirements they are, names the proofs that pin each (`lambdas.tagsAreDisjoint`, `lambdas.terminalsAreExplicit`), and carries a `Changelog:` section with the three `**BREAKING CHANGES:**` entries verbatim from `changelog/unreleased/1711.md`. Title is `edag: restructure chain nodes …`.

**3 — the surviving mutant.** Reproduced: eager `argsOf`, 3381/0. Real defect, thank you.

I did not take the proposed fix, because it isn't available under the repo's own rules: `fjs/AGENTS.md` §1.5 forbids `try`/`catch` in `.f.mjs` outright, and the runner's `throw` key is pass/fail — it never exposes the caught value to an assertion. So there is no way to write "and the payload was the `own` error" in a proof here. Nor is the distinction reachable any other way: both readings throw, and with no mutation a skipped operand has nothing to record itself with.

So I made the order structural instead of a laziness convention. `callValue`/`callProperty` now take the argument **node** and evaluate it themselves, inside the call expression:

```js
const callProperty = (f, obj, prop, e) => obj[prop](...argsOf(f, e))
```

`obj[prop]` is the callee expression, so JavaScript's own order puts the read before the argument list, and `argsOf` loses its thunk — the eager mutation is no longer expressible as an innocuous edit, only as a deliberate hoist against an explicit docstring. Verified out-of-band against the two terms you named: head gives `TypeError` for `['.', undef, 'at', ['|()', mark, null]]` and the `own` payload for `['?.', undef, 'at', ['|!()', mark, null]]`, with `mark` = `['own', ['{}', []], 1]`.

The residual gap is now stated where it lives rather than left implicit — in `callProperty`'s JSDoc and in "Where the host engines disagree" in `README.md`, both of which say plainly that the order is unprovable here and that taking an evaluated array would leave every test passing. I'd rather record that than claim a proof that does not exist; if you want the payload-inspecting proof, it needs a runner or `asserts` change first, and that is its own PR.

**The dropped family.** Not deliberate — restored, in both spellings, since the parenthesis law has two sides here and they should agree: `optionRegion.throw.closeStepAfterPropertyGuard` (`(a.absent?.(...b).m)(...d)`, the step form) and `…callOfSkippedGuard` (`(a.absent?.(...b))(...d)`, across a node boundary). The step form turned out to be load-bearing beyond the family: it is the third entry to `skip` — from a `propertyLambda`-opened region — and nothing else covered it.

**`|?.()` through a `|.`.** Pinned both ways: `lambdas.optionPropertyLambda` and `chains.callsAgainstTheRegion` for the shape, and `optionDot` for the execution (`a?.b.c?.(...d)`).

**Nits.** All four fixed: the `optionRegion.throw.` cross-reference, the family-4 row (relabelled `a?.b(...c)?.(...d)`, noting the property variant is family 1), the `dot`/`optionDot` examples (a computed key is `['Number', exp]`, with the rejection called out), and `closed.extraTailIsError` now covers a trailing element on `|.`, `|?.()` and `|!()`, not only `|()`.

Coverage on `fjs/edag/**/module.f.mjs` is still 100/100/100; suite 3383/0 on node, and on bun the only two failures are the `fjs/fsc` pair that also fails on `main`.

---

o2alexanderfedin (approved):
Approving. The structural fix does what it says, and I like that it says what it doesn't do.

Deleting the thunk seam means the order now comes from JavaScript's own evaluation of `obj[prop](...)`. My original mutant is simply gone — `argsOf` *is* eager now, there is nothing left to force. The eager *semantics* is still spellable with a hoist, and it survives (2 spellings tried, both `tsc` exit 0, both 3383/0, both differing from head on 814 of 2938 corpus cells) — but that residual is declared verbatim in `callProperty`'s JSDoc and in the README rather than papered over, which is the right way to ship an invariant that can't be proven under a payload-blind `ev()`.

One datum worth keeping: the same hoist inside `callValue` is behaviourally **equivalent** (0/2938), because its callee arrives already evaluated. Only `callProperty` carries the invariant.

Behaviour is preserved: 2938 cells, 0 divergences head-vs-previous. The 278 vm-vs-V8 cells are identical at both SHAs and are the documented double-read in guarded calls — collapsing the immediately-repeated identical property read leaves 0. Harness self-tests all manufacture collapses (removing the operand trace hides the eager mutant entirely, removing `this` identity hides the receiver mutant entirely), so the zeros aren't vacuous.

Nothing thinned: case-label census identical, `edag/proof` 77→80 / 77→80, `amnesia/proof` 92→93 / 27→29, leaves 38→40, nothing dropped. The neighbour battery I was worried about all still dies. All nine new rows die individually — including `callOfSkippedGuard`, which turns out to be the only proof pinning that `()` over a short-circuited `.` throws.

All five carried items fixed, and the body is now correct: `Changelog:` present, last section before the trailer, three items byte-identical to the file, the prose `## Breaking Changes` heading gone, the design spelled with the shipped `|`-prefixed tags, title prefixed.

Two notes, neither blocking and neither yours to fix here:

`.` evaluates its base before its index only by JS argument order, undocumented — reversing it survives at head **and** at `1c971fbc5`, and the handler is byte-identical across the delta, so it is pre-existing. It is the same unprovability class you just documented for arguments (`?.` is killed instead, because there the index must not run at all). The same one-sentence JSDoc treatment would close it.

`callProperty`'s JSDoc says "Both helpers take the operand and evaluate it themselves… so that JavaScript's own order applies." For `callValue` that's vacuous, per the equivalence above — one word.

PR #1732, opened 2026-08-27
types/rtti: make a bare Struct or Tuple schema closed

sergey-shandar:
Stage 1 of the closed-by-default proposal
(`fjs/types/rtti/todo/option-as-omission.md`, which keeps only stage 2 after
this).

A bare `Struct` or `Tuple` now admits the members it declares and no others.
`close` is gone: the bare form *is* `rest(c, never)`, so it needs no spelling,
and that also removes the `close(c, rest?)` sentinel, which used `undefined`
for "no undeclared member" and so collided with `undefined` the const. `rest`
takes both arguments; `open(c)` is `rest(c, unknown)` and carries a `const`
signature of its own — without it `open([42])` widens to `Type[]` while every
other assertion still passes, so `proof.f.mjs` pins it.

## Why

The open default had to be defended twice in prose — a "This is deliberate;
please do not 'fix' it" section in the README and a "Do not add a length check"
paragraph in `validate` — against a contributor who kept arriving at the
opposite. It also made `validate`'s success cast **unsound**:
`validate([42])([42, 'extra'])` handed back a two-element array whose static
type said `.length` was `1`. `parse` escaped it by rebuilding; `validate`
could not. The consumers voted by kind, too: `edag` wrote `close` on all 21
nodes of its ADT, and the protocol modules never wrote it once.

## What the readers gained

- The tuple length check returns, now stated by the model rather than inferred
  from `Ts<>` — which is what made #1622's version wrong at the time.
- `undeclaredMembers` decides a container's members by what an index **reads**,
  not by whether it is an own entry, so an index the prototype supplies is held
  to the schema. All three readers share the walk, `array(t)` included.
- The array kind bounds its length when the rest admits nothing, decided by
  `emptyRest` in the data form: `toData(rest(c, r))` denoting the same set as
  `toData(c)`, compared with `subset` both ways. That closes a live divergence
  where `array(or())` accepted a hole-only array on the thunk readers and not
  on the data one.

`RestTs` renders the tail a stated rest implies, `...(Ts<R> | undefined)[]` —
a hole past the prefix is no member, so a reader skips it and the index reads
`undefined`. Two renderer divergences are documented and pinned rather than
closed: `rest([42], [or()])`, where the printer recognizes the empty rest
semantically and a syntactic test cannot; and `rest([], r)`, where the printer
sees one node for two spellings. Both leave the type renderer *wider* than the
schema, which is the direction a success cast can afford.

## Consumers

`edag` drops its 21 `close(...)`. The protocol, media and `dialectEntry`
schemas say `open`, each with the reason written beside it, so the next reader
does not unwrap one and silently break forward compatibility. `ci/common`,
`mcp/cas` and `mcp/evo` record the audit's other answer: a vocabulary the
module itself defines stays closed.

## Review fixes

- A canonical numeric key at or above `2 ** 32 - 1` is not an array index —
  assigning `a['4294967295']` creates an ordinary property and leaves `length`
  alone. Reading it as one put it past the length bound *and* past the
  non-index filter, so it rode through a closed container.
- The member walk is bounded by what the value and its prototypes carry rather
  than by `length`. Materializing `0 … length - 1` turned a
  `new Array(2 ** 32 - 1)` into billions of iterations before any check could
  reject it — measured at 11 s for `2 ** 26`, now 1 ms for the full range.
- That walk then deduplicated with a linear search per index, which made every
  `array(t)` read quadratic — 829 ms at 40 000 elements against 3 ms at 1 000.
  `[[OwnPropertyKeys]]` already yields an array's indices ascending and
  without repeats, so the own names are the answer unless the prototype chain
  supplies an index the value does not; the merge runs only then. 31 ms at
  40 000.
- `array(t)`/`record(t)` walk members the same way as everything else, which
  restores the acceptance agreement an own-entry walk had broken for an
  inherited index.
- The fixtures for the two cases above need in-place mutation, which
  `fjs/AGENTS.md` §3.1 forbids in authored FunctionalScript — and the subset
  has neither mutation nor prototype writes, so these values are reachable
  only from plain JavaScript. `shouldLoad` in `fjs/dev/module.f.mjs` makes a
  plain `proof.mjs` the opt-in home for that, so they live in
  `types/rtti/host.proof.mjs` with the three-reader rows that need them.
- Dropped an `any` cast in `media/module.f.mjs` whose TS2589 rationale was
  stale after the surrounding signature changed.

## Checks

`npx tsc` clean, `fjs test` 3465/3465, `npm run cov` 100% lines/branches/
functions, `cargo clippy` and `cargo fmt --check` clean, `npm run ci-update`
produces no diff.

Changelog:
- **BREAKING CHANGES:** `types/rtti`: a bare `Struct` or `Tuple` schema is
  **closed** — it admits the members it declares and no others, and a tuple
  bounds its `length`. `close(c, rest)` becomes `rest(c, r)`, whose second
  argument is required, and `open(c)` is the old bare form.
- **BREAKING CHANGES:** `types/rtti/common`: the `Visitor<R>` handler `close`
  is renamed `rest` and its second parameter is a `Type` rather than
  `Type | undefined`; `undeclaredEntries` becomes `undeclaredMembers`.
- **BREAKING CHANGES:** `types/rtti`: an array whose element set admits nothing
  is the *empty* array, on every reader — `array(or())` no longer accepts a
  hole-only array, which the data form already rejected.
- **BREAKING CHANGES:** `media`: `dialectEntry` takes a schema with a stated
  rest rather than a bare struct, so a registered dialect says `open`.
- `types/rtti`: an index a value's *prototype* supplies is a member like any
  other — held to a container's rest, and to `array(t)`'s element type — on all
  three readers; an own-entry walk skipped it.
- `types/rtti`: a canonical numeric key at or above `2 ** 32 - 1` is an
  ordinary property rather than an index, so a closed container rejects it;
  reading it as an index let it past both the length bound and the non-index
  check.
- `types/rtti`: reading an array's members is linear in its length again — the
  index walk added above deduplicated with a linear search per index, which
  cost every `array(t)` read 829 ms at 40 000 elements against 3 ms at 1 000.
- `types/rtti/ts`: `Ts<>` renders a stated rest's tuple tail as
  `...(Ts<R> | undefined)[]`, and the runtime printer prints the same; a bare
  tuple renders exactly, which makes `validate`'s success cast sound.
- `types/rtti/data`: new `equivalent` (mutual `subset`) and `emptyRest`, the
  criterion deciding whether a stated rest makes any difference to the
  canonical form.

https://claude.ai/code/session_015focSdZkmvit7uH6bsc8tx


o2alexanderfedin (commented):
Measured at `7b17ea666` against `origin/main` `159035786`.

The change itself is sound and proven. `npm test` 3459/3459 exit 0 vs main's 3447/3447 — **+12, no coverage drop** despite the 2046 deleted lines: of 42 vanished leaf names, every one maps to a rename (`close.*`→`closed.*`/`open.*`/`rest.*`, `undeclaredEntries`→`undeclaredMembers`) or a deliberate semantic inversion (`extraItemsAcceptedAndDropped`→`extraItemsRejected`), and the only name without a like-named successor, `ts/proof close.optionalPosition`, survives as `const.optionalTuplePosition` at `fjs/types/rtti/ts/proof.f.mjs:174` with the same expectation.

The inversion is real in both readers: bare `{a: number}` on `{a:1,b:2}` goes from `ok {"a":1}` to `error {"path":[],"message":"unexpected value"}`; bare `[number,string]` on `[1,'x',9]` from `ok` to `error`. Opt-in openness works — `open({a:number})` on `{a:1,b:2}` → `ok`, `rest({a:number}, string)` on `{a:1,b:'x'}` → `ok` and on `{a:1,b:2}` → `error {"path":["b"]}`.

Mutation-tested, both killed, and the two halves are pinned separately: weakening the undeclared-member check in `validate/module.f.mjs:198` and `parse/module.f.mjs:192` → 3441 pass / **18 fail** across parse, validate and edag; weakening only the tuple length bound → 3456 / **3 fail** (`closed.tuple.holePastThePrefixRejected`, `validate.beyondAClosedPrefix`, `validate.sameAcceptanceInTheDataForm`).

Two things to fix:

1. **The PR body has no `Changelog:` section.** `changelog/unreleased/1732.md` exists and is accurate — 7 items, 4 of them `**BREAKING CHANGES:**` — but `AGENTS.md:91` and `CONTRIBUTING.md:187` require the description to repeat that file item-for-item as its last section before any trailer block, and the body's prose "Breaking Changes" section is not it: the `common` `Visitor<R>` rename, the `array(or())` empty-array change and the `data` `equivalent`/`emptyRest` additions appear nowhere in the body. This repo squash-merges with the body as the commit message, so the gap lands in history permanently.

2. **An `any` escape hatch that is not needed**, `fjs/media/module.f.mjs:138`: `/** @type {ValidateE} */ (/** @type {any} */ (rttiParse)(type))`, with a comment claiming `Parse<Rest<S, Type>>` raises TS2589. It does not — removing the inner `any` and keeping the `ValidateE` cast gives `tsc --noEmit` exit 0 and `npm test` 3459/0, identical to the head. Drop the cast, or correct the comment.

sergey-shandar:
All five findings are addressed in `15a07bb`. Each was reproduced first.

**@o2alexanderfedin**

1. **Missing `Changelog:` section.** The description now carries it as its last section before the trailer, item-for-item with `changelog/unreleased/1732.md`, and the prose "Breaking Changes" section is gone so there is only one list to keep true. The title also now follows the `<topic>: <description>` form. Two items were added to the entry file for the fixes below, so both sides moved together.
2. **The `any` in `fjs/media/module.f.mjs`.** Correct — it was stale. The TS2589 was real under the earlier `@template {Rest<Struct, Type>} T` signature and stopped being reachable when that became `@template {Struct} S` with `@param {Rest<S, Type>} type`. The comment outlived the constraint it described. Cast dropped; `tsc` clean.

**Codex** — all three are real, and two of them are regressions this PR introduced.

1. **Index range.** `arrayIndex` is bounded by `2 ** 32 - 1` now. Measured before the fix: `a = [1]; a['4294967295'] = 2` left `a.length` at `1`, and bare `[number]` answered `ok` on all three readers — the key was past the length-bounded walk and past the non-index filter, so it was a member on neither path.
2. **Sparse-array cost.** Real, and mine: `[...Array(value.length).keys()]` cost 11 s at `2 ** 26` before any check could reject the value. `readIndices` now enumerates the own property names of the value and its prototype chain instead — every index that reads a value is an own property of something on that chain, so the walk is bounded by what is actually carried. `new Array(2 ** 32 - 1)`, which carries one own name, now answers in 1 ms.
3. **Uniform readers.** Also real and mine: sharing `undeclaredMembers` with the data form's `arraySetValidate` while leaving `array(t)`/`record(t)` on `Object.entries` split the three readers on an inherited index. They share the walk now — `array(t)` is `rest([], t)`, so one walk is the point rather than a coincidence.

Each fix is pinned: the index range and the bounded walk in `common/proof.f.mjs`, and rows in `validate/proof.f.mjs`'s three-reader acceptance table for the inherited index under `array(t)`, the out-of-range key, and a full-range sparse array.

`npx tsc` clean, `fjs test` 3462/3462, `npm run cov` 100% lines/branches/functions, `npm run ci-update` produces no diff.

---

sergey-shandar:
Both P1s fixed in `e264cab`. Both were mine, and the first was a regression the *previous* round introduced.

**Θ(N²) dedup** — confirmed before fixing: `validate(array(number))` over a dense array of N took 3 ms / 14 ms / 148 ms / 829 ms at N = 1 000 / 4 000 / 16 000 / 40 000. Since `array(t)` now shares the member walk, that was every array read in the tree.

The fix drops the dedup from the common path rather than making it cheaper. `[[OwnPropertyKeys]]` yields an array's integer indices ascending and without repeats, so the own names already *are* the answer — a merge is needed only when something on the prototype chain supplies an index the value does not, which is empty for an ordinary array. The `indexOf` dedup now runs over that set alone, bounded by the chain rather than by N. Same measurements after: 2 / 4 / 12 / 31 ms.

The `i in value` test went with it: a name reached by walking the value and its chain is a property of one of them, so the array reads at it by construction — that check was an extra chain lookup per element on every read.

**Mutation in the fixtures** — right, and the rule leaves no way to build these values inside the subset: they need `Object.setPrototypeOf` and a property assignment, and FunctionalScript has neither mutation nor prototype writes. That is also *why* they are worth pinning — a caller in ordinary JavaScript can hand one to an exported reader, and both are cases where reading a member by entry and reading it by index disagree.

So I took your second option. `shouldLoad` in `fjs/dev/module.f.mjs` already designates a plain `proof.mjs` as the opt-in home for impure host-JavaScript proofs, so the fixtures and the three-reader rows that need them moved to `fjs/types/rtti/host.proof.mjs`, and no `.f.mjs` file calls `Object.setPrototypeOf` any more. A shadowed prototype index joins them — the case separating "the chain names an index" from "the chain supplies one", which is where the merge decides whether it has anything to do.

One thing I did **not** change: the pre-existing `Object.assign([1], { foo: 2 })` fixtures in `common/proof.f.mjs`, `data/proof.f.mjs` and `validate/proof.f.mjs`. They predate this PR and moving them would widen it; if they should follow, that reads to me as its own change.

`npx tsc` clean, `fjs test` 3465/3465, `npm run cov` 100% lines/branches/functions, `npm run ci-update` produces no diff. The description's `Changelog:` section and the entry file both carry the perf item now.

---

o2alexanderfedin (approved):
Approved. Verified at `15a07bbf7`: both findings fixed, and fixed at the source. The body now ends with a `Changelog:` section that is byte-identical to `changelog/unreleased/1732.md` item for item (8 items, 4 breaking), and `fjs/media/module.f.mjs:135` is now `const v = /** @type {ValidateE} */ (rttiParse(type))` — the inner `any` and the stale TS2589 comment both deleted rather than re-justified.

Gates green and ahead of main: `npm test` 3462/3462 exit 0, `tsc --noEmit` exit 0 (main `159035786`: 3447/3447). The new commit's headline behaviour — `arrayIndex` rejecting keys `>= 2 ** 32 - 1` so they fall through as ordinary properties — is mutation-proven: relaxing the bound to `i <= 2 ** 32 - 1` is KILLED (3461 pass / 1 fail).

o2alexanderfedin (approved):
Approval still stands. Verified at `e264cabc0`: `npm test` 3465/3465 exit 0, `tsc --noEmit` exit 0 (main `159035786`: 3447/3447). The three leaf names that left `common/proof.f.mjs` and `validate/proof.f.mjs` all reappear under the new host-only `fjs/types/rtti/host.proof.mjs`, along with `shadowedIndexIsTheOwnOne` and `readersAgree` — no lost coverage, net +3. `Changelog:` still matches the entry file item-for-item including the new perf item, and the perf claim reproduces: `undeclaredMembers([], a)` is 3.4 ms at 1 000 and 42.8 ms at 40 000. Dropping `.filter(i => !Object.hasOwn(value, i))` is KILLED (3464/3465) by `shadowedIndexIsTheOwnOne`.

One coverage nit, not a defect: dropping the `.filter((i, at, a) => a.indexOf(i) === at)` dedup on `inherited` **survives** — 3465/3465. It is load-bearing with two prototype levels both supplying index `1` (`value=[42,,]` → `[0,99]` → `[0,77]`): head gives `[["1",99]]`, the mutant `[["1",99],["1",99]]`. Nothing in the suite distinguishes them because every fixture has a single prototype level carrying indices. Validation results are order- and multiplicity-insensitive here so no verdict changes, but a one-line row next to `shadowedIndexIsTheOwnOne` would pin it.

PR #1733, opened 2026-08-27
Document bun optional chain parentheses bug and tracking

sergey-shandar:
## Summary

This PR documents a known issue with how bun handles parentheses in optional chains, which causes incorrect behavior compared to V8 and the JavaScript specification. The issue affects both bun's runtime engine (JavaScriptCore) and its transpiler, and is tracked across multiple documentation files.

## Key Changes

- **Added blocking issue documentation** (`todo/blocked/bun-optional-chain-parentheses.md`): Comprehensive record of the bug including:
  - Problem description with concrete examples showing the divergence between V8 and bun
  - Distinction between the engine-level and transpiler-level defects
  - Trigger conditions for unblocking
  - Task checklist for tracking and fixing
  - Related documentation references

- **Updated EDAG documentation** (`fjs/edag/README.md`): Added cross-reference to the new blocking issue document in the "Where the host engines disagree" section, linking the parenthesis-ending-chain problem to its detailed tracking

- **Updated amnesia module documentation** (`fjs/edag/amnesia/README.md`): Added reference to the blocking issue in the explanation of how `skip` handles the `|!()` step, providing context for why the workaround is necessary

- **Updated proof module** (`fjs/edag/proof.f.mjs`): Enhanced comments explaining the two commented-out test cases to reference the blocking issue document and clarify what unblocks them

## Implementation Details

The changes maintain the existing workarounds while establishing a clear tracking mechanism for the bug. The blocking issue document serves as a single source of truth for:
- What versions of bun exhibit the problem
- Specific test cases that demonstrate the divergence
- Upstream issue references (oven-sh/bun#31812)
- Conditions for considering the issue resolved

https://claude.ai/code/session_01F3Ne51veqffkQNsKkXXUsK

### DIFF
--- a/fjs/edag/README.md
+++ b/fjs/edag/README.md
@@ -292,6 +292,9 @@ with `SyntaxError: Cannot use tagged templates in an optional chain`, where
 It is [oven-sh/bun#31812](https://github.com/oven-sh/bun/issues/31812), filed
 for the `new` sibling `new (baz()?.qux)()`; one root cause, the parenthesis
 ceasing to end the chain, so restrictions that hold inside it leak past.
+Both halves are tracked, with the versions they were measured on and what
+unblocks them, in
+[bun-optional-chain-parentheses.md](../../todo/blocked/bun-optional-chain-parentheses.md).
 
 ### The cost
 
--- a/fjs/edag/amnesia/README.md
+++ b/fjs/edag/amnesia/README.md
@@ -88,7 +88,9 @@ One corner where that would have been wrong is interpreted here instead. When
 `u` is nullish, `(u?.b)(d)` must throw: the parentheses end the optional
 chain, so `undefined` is what gets called. V8 does throw; JavaScriptCore
 (hence `bun test`) carries the short-circuit through the parentheses and
-evaluates to `undefined` (["Chains"](../README.md#chains)). That spelling is
+evaluates to `undefined` (["Chains"](../README.md#chains), tracked in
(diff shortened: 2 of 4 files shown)


o2alexanderfedin (approved):
Approved. Verified at `5993c1120`; `npm test` 3465/3465 exit 0 and `tsc --noEmit` exit 0, identical to `main` — expected, since the only non-markdown change is a rewrapped sentence inside an existing block comment in `fjs/edag/proof.f.mjs`, not a rewritten expression.

The spec reading is right: `(u?.b)` is a PrimaryExpression, so the chain does not extend past the parentheses and the call on nullish `u` must throw. Node reproduces the document's Node column exactly — all four rows `TypeError` — and I also confirmed `` (u?.b)`tag` `` and `new (u?.b)()` throw in both a module and `eval`, which matches the claim that the in-chain restrictions stop applying once parenthesized. Upstream tracking is cited (oven-sh/bun#31812) with an explicit open task to file the WebKit side. All 12 relative links and both anchors resolve, and the new file matches the shape of the 11 pre-existing `todo/blocked/` files. Changelog handling is correct per `AGENTS.md:92-96`.

For the record: **bun is not installed here, so the entire "bun (JavaScriptCore)" column is unverified** — the version claims and the `eval`/`new Function` rows under bun included. I checked the shape of the claim, not bun's behaviour. `bun = '1.4.0'` in `fjs/ci/config/module.f.mjs:31` does match.

PR #1736, opened 2026-08-27
Move rtti from fjs/types/ to fjs/ as a peer of djs

sergey-shandar:
## Summary

Relocates the `rtti` module from `fjs/types/rtti/` to `fjs/rtti/`, establishing it as a peer of `djs` rather than a member of the `types/` directory. This reflects the architectural reality that `rtti` is a consumer of `types/*` modules (not a member), and all its external consumers are peers of `types/`, not members.

## Key Changes

- **Directory move**: `fjs/types/rtti/` → `fjs/rtti/` (preserves git history with `git mv`)
- **Import path updates**:
  - External imports drop the `types/` segment: `../types/rtti/…` → `../rtti/…`
  - Internal imports re-anchor by depth rule: paths into `types/` gain a `types/` segment, paths out lose one
  - 30 code files updated across `media`, `protocol`, `mcp`, `ci`, `edag`, `emergent_testing`
  - 58 import lines within `rtti/` re-anchored
- **Documentation updates**:
  - 37 markdown links updated (36 drop `types/`, 1 gains `../`)
  - 15 outbound links from moved files lose one `../`
  - 16 non-link relative references in inline code/fences updated
  - 21 self-referential `fjs/types/rtti/…` paths converted to `fjs/rtti/…`
  - 14 prose mentions updated
  - one link outside the moved subtree missed by the initial pass, `fjs/types/todo/141.md`'s `[`rtti/data`](../rtti/data/module.f.mjs)`, re-anchored to `../../rtti/data/module.f.mjs`
- **Preserved**: `fjs/media/json/rtti/` remains in place (named after what it binds, no collision)

## Rationale

`rtti` is 5789 lines (38% of `fjs/types/`), has six modules and its own README tree. Zero imports come from within `types/`; all 64 references originate from peer directories. It consumes `types/*` modules the same way the rest of `fjs/` does, and its outward dependencies point to peers (`fjs/asserts`, `fjs/js/keywords`, `fjs/djs`), not down the hierarchy. Structurally, it belongs beside `djs` as "the types described over the data model."

## Breaking Change

This is a breaking change for downstream importers of `functionalscript/fjs/types/rtti/…` paths, since the package has no `exports` map and importers reach these modules by path. The move is its own PR per the one-move-per-PR rule, touching nothing else.

Changelog:
- **BREAKING CHANGES:** `fjs/types/rtti/` moved to `fjs/rtti/` — it is a peer
  of `djs`, not a member of `types/`. No code changed, only import paths;
  every importer inside and outside the repo needs the `types/` segment
  dropped from `functionalscript/fjs/types/rtti/…`.

https://claude.ai/code/session_01EJuNsKf9jpgULV67jFo2sy


o2alexanderfedin (commented):
Measured at `18b831007`. The move itself is clean: all 34 `fjs/types/rtti/**` files are detected renames, and filtering every diff line that mentions an rtti path segment out of `git diff -M -C` leaves **zero** content changes across the whole PR except one added prose block in `fjs/rtti/README.md`. `npm test` 3470/3470 exit 0 on both sides, `tsc --noEmit` exit 0 on both, and the sorted proof leaf names are byte-identical — the full-path diff is exactly 315 removed under `./fjs/types/rtti/` and 315 added under `./fjs/rtti/`. No behaviour smuggled in, no coverage lost. Todo hygiene is right too: `move-rtti-out-of-types.md` is deleted with its rationale preserved in the new README section, and `group-fs-subdirectories-by-concern.md` agrees rather than contradicting.

Two things to fix:

1. **One broken link, and it is this PR's doing.** `fjs/types/todo/141.md:34` writes ``[`rtti/data`](../rtti/data/module.f.mjs)``, which resolved to `fjs/types/rtti/data/…` before the move and resolves to nothing now. The file was not touched by the PR, so a `types/rtti` grep does not catch it — a repo-wide relative-link scan at both revisions produces exactly one new break and zero fixes, this one. `../../rtti/data/module.f.mjs` fixes it.

2. **The PR body has no literal `Changelog:` section.** `CONTRIBUTING.md:189` wants one as the last section before the trailer; the body instead mentions the changelog inside a bullet (`- **Changelog**: Added entry with…`) and ends at `## Breaking Change`. The entry file is correct — `**BREAKING CHANGES:**` is the right marker here, since the package has no `exports` map and importers reach these modules by path — so it is only the body that is owed. This repo squash-merges with the body as the commit message, so the gap lands in history.

o2alexanderfedin (approved):
Approved. Verified at `f1a93d0bf`: the link now reads `../../rtti/data/module.f.mjs` and resolves to a file that exists on disk, and a repo-wide relative-link scan finds 18 pre-existing breaks at the head and the same 18 on `origin/main` — zero broken at the head but not on main, so the fix introduced none of its own. The body now ends with a literal `Changelog:` section, byte-identical to `changelog/unreleased/1736.md` including the `**BREAKING CHANGES:**` marker. `npm test` 3470/3470 exit 0 and `tsc --noEmit` exit 0 on both sides, and the delta touches nothing but that one line.

PR #1763, opened 2026-08-28
ci: hand the packed tarball to CI as an artifact

sergey-shandar:
Step 2 of Stage 2's packed-artifact check, designed in [#1757](https://github.com/functionalscript/functionalscript/pull/1757). Step 1 (the job-ordering edge) merged as [#1762](https://github.com/functionalscript/functionalscript/pull/1762).

## Why

The package is built in CI and then thrown away: `node26` runs `npm pack` and nothing keeps the result. Checking the package as a consumer sees it requires a job with **no repository checkout**, and such a job can only receive the tarball through an artifact.

This adds the producing half only. The consuming job is the next step and deliberately is not here.

## The change

`actions/upload-artifact` pinned at `v7.0.1` (latest release), and an upload step in `node26` immediately after `npm pack`. The generated workflow gains exactly those eight lines.

Two decisions worth stating:

- **The artifact name is an exported constant** (`packageArtifact`), not a string literal. The consuming job downloads by name, so producer and consumer share one definition rather than two literals that can drift apart silently.
- **`if-no-files-found: error`.** The action's default is to warn and upload nothing. That would turn a packing failure into a missing-artifact failure in the consuming job — the wrong place, reported as the wrong cause. This is the same "red for the wrong reason" failure the ordering edge in #1762 exists to prevent.

## Proof

`packageArtifact` covers the properties rather than the shape, and I checked each against a mutant that breaks it:

| Property | Mutant | Caught |
|---|---|---|
| upload follows `npm pack` | move it before | ✅ `expected the upload to follow npm pack` |
| name matches the exported constant | rename to a literal | ✅ |
| `if-no-files-found: error` | drop the key | ✅ |
| exactly one job uploads | — | asserted |

Ordering is not cosmetic: uploading before packing ships an empty artifact, and `if-no-files-found: error` would then fire in the wrong job.

## Checks

`npx tsc` clean; full suite with coverage at 100%; `npm run ci-update` reproduces `.github/workflows/ci.yml` with no further diff.

No changelog entry: no behavior or public API change for package users — this is the repository's own CI generator.

https://claude.ai/code/session_01LqeS5t2ZKkSu3chRPMXR7n

---
_Generated by [Claude Code](https://claude.ai/code/session_01LqeS5t2ZKkSu3chRPMXR7n)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -509,6 +509,14 @@
         },
         {
           "run": "npm pack"
+        },
+        {
+          "uses": "actions/upload-artifact@v7.0.1",
+          "with": {
+            "name": "package-tarball",
+            "path": "*.tgz",
+            "if-no-files-found": "error"
+          }
         }
       ]
     },
--- a/fjs/ci/config/module.f.mjs
+++ b/fjs/ci/config/module.f.mjs
@@ -73,6 +73,8 @@ export const actions = /** @type {const} */({
     'actions/setup-node': 'v7.0.0',
     // https://github.com/marketplace/actions/cache
     'actions/cache': 'v6.1.0',
+    // https://github.com/marketplace/actions/upload-a-build-artifact
+    'actions/upload-artifact': 'v7.0.1',
     // https://github.com/marketplace/actions/setup-deno
     'denoland/setup-deno': 'v2.0.5',
     // https://github.com/marketplace/actions/setup-bun
--- a/fjs/ci/node/module.f.mjs
+++ b/fjs/ci/node/module.f.mjs
@@ -12,6 +12,12 @@ import { node } from '../config/module.f.mjs'
 import { install, test, ubuntuArm, uses } from '../common/module.f.mjs'
 import { nixInstall, nixVersionCheckStep } from '../nix/module.f.mjs'
 
+/**
+ * Name of the CI artifact carrying the `npm pack` tarball. The producing step
+ * is below; a consuming job downloads it by this name.
+ */
+export const packageArtifact = /** @type {const} */ ('package-tarball')
+
 /** @type {(v: string) => string} */
 export const major = v => v.split('.')[0]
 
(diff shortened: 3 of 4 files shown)


o2alexanderfedin (approved):
Approved.

Round-trip is clean — regenerating leaves `git diff` empty, which matters more here than last time since `ci.yml` is in the diff. The upload step is well-formed and correctly placed: `actions/upload-artifact@v7.0.1` exists as a tag, and it sits after `npm pack` in `node26Steps` with no `working-directory` override, so the glob looks where `npm pack` actually writes. `needs` is genuinely not used yet, matching your commit message. The `-2` lines are the two widened imports, not a deleted assertion. Gates `tsc` 0, `npm test` 3529/3529, zero leaves vanished and one added.

Four of the five properties kill under mutation — dropping the step, wrong `name`, `if-no-files-found: warn`, and ordering asserted directly.

One gap: **`path` is unpinned.** Changing `path: '*.tgz'` to `'*.wrong'` leaves `npm test` at 3529/3529, exit 0. Your own mutant list never claimed `path`, so this is a gap rather than a wrong claim, and `if-no-files-found: error` limits the blast radius — a glob matching nothing fails loudly in the producing job. What survives is a path matching the *wrong* files quietly. `assertEq(upload?.path, '*.tgz')` closes it.

sergey-shandar:
The `path` gap is real — reproduced it before fixing, and closed in b926e29.

Confirmed exactly as you described: changing `path: '*.tgz'` to `'*.wrong'` left `packageArtifact` passing and the suite green. My mutant list covered the step's presence, name, ordering and no-files behavior, and simply never touched `path` — so the property with the least coverage was the one nothing asserted.

Your bounding argument is right and worth making sharper, because it points at *why* this one matters more than it first looks. `if-no-files-found: error` catches a glob matching **nothing**. What survives is a glob matching the **wrong** files — uploaded quietly, under the correct artifact name, at the correct size. That is worse than an empty artifact: an empty one fails loudly in the producing job, whereas a wrong one flows to the consuming job in step 3, which then type-checks something real and unrelated and reports green. A check that validates the wrong artifact is indistinguishable from a working one until it matters.

`assertEq(upload?.path, '*.tgz')` added, with a comment recording that distinction so the next person doesn't drop it as redundant with `if-no-files-found`. The same mutant now fails.

Re-verified: `tsc` clean, coverage 100%, `npm run ci-update` produces no further diff — the fix is proof-only, so `ci.yml` is untouched.

---

PR #1769, opened 2026-08-28
fjs: restore `@module` where documentation is published

sergey-shandar:
Closes `fjs/todo/module-tag-restore.md`, filed in #1765.

#1756 stripped `@module` from 102 files and `eebc007` from 15 `private.ts`, on the reading that the tag marks a package entry point. It does not. `deno doc` emits `module_doc` only for a file carrying the tag and nothing at all for one without it, so the prose in those files was still in source and invisible to the tool. #1765 corrected §2; this brings the tree into line with it.

## The restore

106 files, two lines each and nothing else in the source diff:

|count|what|why|
|-|-|-|
|89|`types.ts`|the public type-level API — its emitted declarations are what a consumer reads|
|16|`private.ts`|contributor-facing prose (see below)|
|1|`fjs/emergent_testing/browser.mjs`|real prose: browser-native proof execution, and why it has no Node dependencies|

**Left untagged deliberately.** The 11 proof files: a proof documents a verification rather than an API, and no documentation build is pointed at proofs. And `fjs/bnf/testlib.f.mjs`, whose leading block holds only `@import` tags, so there is no prose for a tag to attach to.

**`private.ts` is tagged but not published.** `fjs/todo/separate-private-types.md` puts those types outside the public declaration closure and its Stage 2 drops their generated declarations from the package. The tag makes the prose reachable by `deno doc` for a contributor reading the sources; it does not make these files website input, and a public build must stay pointed away from them. §2 carries that distinction.

## Verified, not assumed

`deno doc --json` reports `module_doc` carrying a `module` tag for **all 106** — and for **none** of the controls left untagged (`testlib.f.mjs` and two proof files). The negative control is the point: without it, a check that always says yes is indistinguishable from a passing one, which is the failure §6 was written about.

104 of the 106 are byte-identical to their pre-strip originals, and no file in the #1756 set had drifted since.

## Two corrections to the issue as filed

Both were mine, in #1765.

- It said 16 `private.ts` were stripped by #1750. **15** were, by `eebc007`. The 16th, `fjs/rtti/parse/private.ts`, was created afterwards and never carried a tag — same fix, different provenance. The inventory was built from one commit's diff without asking whether the set was complete, which is the same mistake that missed `private.ts` entirely the first time.
- `fjs/asn.1/private.ts` had gained a `_Round8` type below its header since the strip. Reverting the file wholesale would have silently deleted it, so the tag was inserted instead; its header now matches the original exactly.

## The half that mattered more

The issue said the restore was only one part, and the other was that one rule lived in three documents and drifted in the copies nobody was editing.

`todo/migrate-typescript-to-mjs.md` is a live **P1** that stated the old rule in three normative places, including an acceptance criterion — as written, following it would have re-stripped these exact tags. Each now links §2 rather than restating it. Its declaration-emit measurement is a historical record and is untouched.

`fjs/AGENTS.md` §2's paragraph marking the tree non-compliant is replaced by the proof decision the issue asked to have recorded there rather than left in a todo. `fjs/todo/module-tag-restore.md` is deleted, its tasks done.

## Gates

`npx tsc` clean. Proof suite 3533/3533. Link check clean apart from two pre-existing `[#NNN](url)` template placeholders in `changelog/README.md` and `todo/changelog-website.md`.

No changelog entry: no runtime behaviour changes — the diff is JSDoc comments and documentation.

https://claude.ai/code/session_01T7vAocRuxfmWJDnujCoMup

---
_Generated by [Claude Code](https://claude.ai/code/session_01T7vAocRuxfmWJDnujCoMup)_


o2alexanderfedin (approved):
Approved.

The restoration matches the rule, and I checked all 106 files rather than a sample: every one is exactly `+2 -0`, every insertion lands inside the file's existing leading JSDoc block, and the restored text is byte-identical to the pre-strip originals — using `eebc0078a^` for the `private.ts` set rather than #1756's parent, which would have been the wrong pre-image. `fjs/asn.1/private.ts` is the one file whose content moved on since, and the tag went into the original header with the later `_Round8` untouched. No block reattached to the wrong declaration, which was the failure mode I was most worried about here.

89 `types.ts` + 16 `private.ts` + `browser.mjs`, zero `proof.*`, and `testlib.f.mjs` correctly left alone.

**A correction on my side.** I had been carrying an inventory saying 8 of the 11 `proof.*` files had prose and should therefore regain the tag. That was my error: `fjs/AGENTS.md` §2 — already merged in #1765, unchanged by this PR — excludes proofs outright ("a proof's documentation is not published, so the tag has nothing to attach it to"), with no prose condition. The 98-of-102 figure I verified earlier was a count of which files *had prose*, which is not the same question as which should be tagged, and I conflated the two across two reviews. This PR follows the actual rule.

Everything else holds. The emitted declarations differ in exactly those 106 files and in nothing but the two comment lines. Proof leaf-name sets at merge-base and head are identical, 3533/3533 both sides, `tsc` 0. Rule and sweep agree, and the two stale restatements in `jsdoc-verification.md` and `migrate-typescript-to-mjs.md` now link §2 instead of repeating it. No changelog needed, matching #1756 which shipped without one.

One thing worth stating plainly since it is the second sweep over these files: nothing guards this. There is no `@module` check in CI, this PR adds none, and it does not claim to — the tags were verified once and can drift again the same way.

sergey-shandar:
Thanks — and the `eebc0078a^` pre-image is the right one to have checked against. Using #1756's parent for the `private.ts` set would have compared them to a state that still had the tag for an unrelated reason, so a wrong restore would have looked correct.

**On your correction:** the prose count and the tag rule are different questions, and I made the same conflation in the other direction. My inventory in #1765 said 16 `private.ts` were stripped by #1750; 15 were, by `eebc0078a`, and the 16th (`fjs/rtti/parse/private.ts`) was created afterwards and never carried one. Both errors come from reading a commit diff as if it were the population.

**On nothing guarding this — that is deliberate, and recorded.** [`todo/jsdoc-verification.md`](https://github.com/functionalscript/functionalscript/blob/claude/review-todos-documentation-hrijwz/todo/jsdoc-verification.md) is the issue for it, at P4 on @sergey-shandar's call that reviewers do catch these — as you just did. Root [`AGENTS.md` §6](https://github.com/functionalscript/functionalscript/blob/main/AGENTS.md) is why this PR adds no check rather than an oversight: a `grep` for `@module` cannot tell a JSDoc tag from the same characters in a string, and the guard built that way in #1765 flagged the file whose assertions named it. §6's position is that an unenforced written rule beats machinery whose failures are silent.

This PR updated that issue to record the restore as the fourth pass over one tag with no check registering any of them. So the drift risk you name is the open question there, not a gap this PR leaves unstated — and "no tool is good enough, leave it to review" is written in as a legitimate outcome of that investigation.

Worth adding to it from this round: `deno doc --json` answered the question directly, cheaply, and with a negative control, using a tool the repository already pins. That is the most promising lead the issue lists, and it is now more than a guess.

---

PR #1773, opened 2026-08-29
ci: run Node 24 and Node 26 through their generated flakes

sergey-shandar:
Phase 3 of `fjs/ci/todo/66b-dockerfile-nix-integration.md`, two thirds done.
Node 24 and Node 26 install Nix through the pinned action and run their commands
through their own generated flakes; Node 22 is the last job on `setup-node`, and
is now as mechanical as the other two.

```sh
# node24
test "$(nix develop ./nix/node24 --command node --version)" = v24.19.0
nix develop ./nix/node24 --command npm ci
nix develop ./nix/node24 --command node --test

# node26 — regeneration and the drift check last, git is the runner's
test "$(nix develop ./nix/node26 --command node --version)" = v26.7.0
nix develop ./nix/node26 --command npm ci
nix develop ./nix/node26 --command npx tsc
nix develop ./nix/node26 --command npm run cov
nix develop ./nix/node26 --command npm pack
nix develop ./nix/node26 --command npm run ci-update
git add -A && git diff --cached --exit-code
```

Each job's commands and their order are unchanged, except that Node 26's
regeneration and drift check move to the end — the check compares the tree
against what the generator produces, so ending with it makes it the last word.
Nothing earlier steps leave behind is tracked: `*.tgz`, the declarations
`prepack` emits, and `/nix/*/flake.lock` are all ignored, and `tsc` runs with
`noEmit`.

## A CI step runs one command

Root `AGENTS.md` gains §7. The step is the unit CI reports on, so a job bundled
into `bash -c 'a && b && c'` collapses to one red result naming the wrapper
rather than the command that failed, and hides which commands ran at all. Two
commands are one step only when the second is meaningless alone — the two such
pairs in the generated workflow, `git add -A && git diff --cached --exit-code`
and `sudo apt-get update && sudo apt-get install -y …`, are named as qualifying.

Re-entering the shell per step costs nothing: a step names the flake only when
it needs a tool the flake pins, so `git` stays the runner's, and no flake has to
declare it. `nixDevelopAll`, which existed only to build the bundled form, is
removed.

## A flake is checked by the job that uses it

The `nix-flakes` job — a separate job that instantiated every generated flake
and compared the Node it provided to an expected version — is deleted, with
`nixVersionCheckStep`, `nodeNixVersionSteps` and `nodeNixFlakeJob`. That check
is now a step of each job that has a runtime to check: Node 24 and Node 26
assert what their flake provides, Node 22 what `setup-node` handed it, all
through one builder.

It is each job's **first** command, with nothing exempted — `npm ci` runs
`preinstall`/`install`/`postinstall` hooks from the project and its
dependencies, code that would otherwise execute on an unconfirmed runtime.
`nodeVersionChecks` in `fjs/ci/proof.f.mjs` pins that ordering. The platform
matrix keeps no check: its Windows jobs run `run` steps under PowerShell, where
this POSIX spelling would not survive.

What needs no Nix is checked by proofs over the generator's output instead:
`fjs/ci/proof.f.mjs` requires the written file to equal `flakeText(job)` whole
and the job's package attribute to follow its configured version, while
`fjs/ci/nix/proof.f.mjs` pins that text character for character — the accepted
commit and the job's default development shell included. One gap is recorded
rather than hidden: nothing evaluates the Node 22 flake until that job migrates.

## Node 22 loses `fjs test`

It carried `fjs test`, and the global install that put `fjs` on `PATH`, only
because Node 22 could not run `node --test`. It can, and the job already ran
both, so the pair is gone and Node 22 is `npm ci` and `node --test`.
`nodeVersionJobs` no longer takes the published version, which Node 22 was the
only reader of.

`nix/node22/flake.nix` loses the `shellHook` that existed for that install,
leaving no job declaring one — the generator still emits a hook for a job that
declares one, and `fjs/ci/nix/proof.f.mjs` holds that capability to its shape.
The published CLI is still exercised by the six platform jobs and by `deno` and
`bun`.

## Shorter paths

Generated flakes move one level up: `nix/generated/node24/flake.nix` is now
`nix/node24/flake.nix`, `generatedDirectory` is `nix`, and a step names
`./nix/node24`. `.gitignore` follows with `/nix/*/flake.lock`, which still
matches exactly the per-job locks and leaves a future hand-maintained
`nix/flake.lock` visible. `nix/` is therefore not a directory the generator owns
outright — `nix/README.md` sits beside the generated subdirectories, and
stale-output removal must delete the job directories it generated rather than
everything it finds.

## The `@typedef` gate is deleted

`node26` ran a `grep` for `@typedef` over every authored `.mjs`, which root
`AGENTS.md` §6 rules out. The rule it stood for is about scope — a file-scope
`@typedef` is prohibited, a function-local one allowed — and the pattern encoded
scope as indentation. Its own near-miss is the evidence: the file holding that
pattern escaped both branches only because both sit indented past what the
pattern accepts. `todo/jsdoc-verification.md` owns the replacement and records
this as its worked example; the rule stands, unenforced, which §6 calls the
better trade.

## Documentation and issues

`fjs ci` is documented as **not stable**: its output changes shape between
releases and the generator does not migrate a consumer's tree, which is the
answer to leftover `nix/generated/` files after an upgrade. `CONTRIBUTING.md`
is corrected to say pull requests land as merge commits rather than squashes —
so branch commit messages are permanent rather than working notes — and
`todo/commit-message-standard.md` records that its squash-only repository
settings were never applied. `node26-typedef-gate-reaches-consumers.md` becomes
`ci-generator-audience.md`, holding what survives its report.
`built-package-checks.md` is new: eight jobs install `functionalscript` from the
registry at a version that is not the commit under review, so the check runs
backwards.

## Checks

`npx tsc` clean, `fjs t` 0 failures, `npm run cov` 100%
lines/branches/functions, no drift from `npm run ci-update` — re-run after each
merge of `main` into this branch. CI has been green on every head since the
migrations landed, with both Nix jobs in line with their neighbours on time.

Diffing exported names against `main` gives exactly six public changes, all
named in the entries below.

Changelog:
- **BREAKING CHANGES:** `fjs/ci/nix`: generated flakes move one level up, so
  `nix/generated/node24/flake.nix` is now `nix/node24/flake.nix` and
  `generatedDirectory` is `nix`.
- **BREAKING CHANGES:** `fjs/ci/nix`: `nixDevelopAll` and `nixVersionCheckStep`
  are removed, with the `nix-flakes` CI job. A flake is checked by the job that
  uses it, and what it declares is asserted by proofs.
- **BREAKING CHANGES:** `fjs/ci/node`: `nodeNixVersionSteps` and
  `nodeNixFlakeJob` are removed with that job, and `nodeVersionJobs` takes no
  arguments — Node 22 was the only reader of the published version it was
  passed.
- `ci`: the Node 24 and Node 26 jobs run their commands through their generated
  Nix flakes, one step each, instead of installing Node with `setup-node`.
- `ci`: every Ubuntu Node job checks its runtime version before its first real
  command, and the Node 26 job runs `npm run ci-update` and its drift check
  last, after every other command.
- `ci`: the `node26` job no longer greps `.mjs` files for a file-scope
  `@typedef`. The rule stands; a text pattern cannot answer a question about
  scope, so it approximated one as indentation.
- `ci`: the Node 22 job drops `fjs test` and the global install feeding it,
  running `node --test` like the other Node jobs. Its generated flake loses the
  `shellHook` that existed for that install.


o2alexanderfedin (commented):
Reviewed the delta `c05b613ad..31952c46c006147dd7cc09fad733529aecbdfd1b`. The round-trip gate still ties `nix/<job>/flake.nix` back to config — `nixFlakes` writes in the same effect chain as the workflow and node26 still runs `ci-update` + `git diff --cached --exit-code` — so "use the flakes, do not check them" did not make them hand-maintained. The rename is clean: `git grep 'nix/generated'` leaves only the 0.40.0 changelog and this PR's own entry, and `npm pack --dry-run` confirms no flake ships, so it is not a breaking published move.

Two documents now say the opposite of what shipped.

**`fjs/ci/README.md:68-81`** describes the deleted `nix-flakes` job in the present tense, including "That check is the *only* place the expectation is written" and an instruction to "give each migrated job its own version check inside the `nix develop` invocation, or nothing ties the Nix runtime to the version the Windows and macOS jobs install." That is the repo's own written warning about exactly the cost this PR accepts, left standing next to the acceptance.

**`fjs/ci/config/module.f.mjs:39`** still reads "the generated flakes on the Nix jobs, which assert the version they actually get." At `c05b613ad` that was false for node24 only; deleting the `nix-flakes` job made it false for all three — in the comment a person bumping a Node version reads first.

For the record on scope: at the previous head the lost version check was node24's. It is now repo-wide. Mutating `node24: '24.19.0'` to `'24.99.0'` at this head regenerates nothing (`git diff --stat` shows only the config file, so the drift gate stays green too) and the suite passes 3542/3542. The `nixFlakes` proof pins `nodejs_${major}`, the commit and the shell attribute — major version only, which is why. node22's and node26's flakes are now built by nothing until those jobs migrate; node24's is at least exercised by its own `npm ci`. All disclosed in the commit message, both Nix todos and `nix/README.md`, with the mitigation assigned to the unimplemented `ci-nix-update` — so this is a note, not a blocker. The two documents above are.

Nix is not installed here, so every flake claim is read off the generator, not built. `tsc` 0, `npm test` 3542/3542, round-trip clean.

o2alexanderfedin (commented):
The version check is the right shape and the headline finding is closed. Confirmed by mutation at `858d41de6`: `node24: '24.99.0'` now regenerates `ci.yml:473` — it regenerated nothing before — and the same holds for `node22` and `node26`, so the fix is repo-wide, not node24-only. One builder, one source in `config/module.f.mjs`, so this is not the two-copies shape. The migrated job's check runs inside `nix develop` and is its first real step; `test "$(...)" = vX` with no `|| true` fails the job under GHA's `bash -e`. `nodeVersionChecks` is a real gate, not a tautology: breaking the generator to emit `v${version}x` fails it, binding node22's check to `node.node24` fails it, and moving the check after `fjs test` fails it too, so the "before the first real command" claim is enforced rather than asserted.

Nix is not installed here, so the step actually catching a wrong runtime on the runner is read, not measured.

**`fjs/ci/README.md:68-80` was not updated, and it is the one that misleads.** It still describes the `nix-flakes` job in the present tense, and "That check is the *only* place the expectation is written" now points at a job this PR deletes — while the expectation lives in three job steps. Lines 77-80 then instruct someone to "give each migrated job its own version check inside the `nix develop` invocation", which is exactly what this commit did. The PR does touch this file for the path rename, so the section was reachable.

Minor: the Progress note in `65z-ci-nix.md:8-12` still lists the migrated job's sequence as "`npm ci`, `node --test`" without the version check that is now its first real step.

`tsc` 0, `npm test` 3543/3543 at `858d41de6` (+1 leaf, `nodeVersionChecks`), round-trip clean. The commit since is one new todo file, `published-package-checks.md`, which I read but did not review in depth.

o2alexanderfedin (commented):
The proof change is a genuine strengthening and I verified it rather than reading it. I patched `writeFlake` to append a trailing line to the file it writes — chosen so every substring the old proof searched for survives intact — and ran the full suite at both heads: at `456e9b809` `fjs/ci/proof.f.mjs → nixFlakes` **passes** (fail: 2); at `acb2875e9` it **fails** (fail: 3). Equality rejects what `includes` accepted. Scoping it honestly: the suite was red at both heads because `fjs/ci/nix/proof.f.mjs` already caught this class from its literal fixtures, so what the change buys is closing the ci proof's false-accept surface, not new coverage.

Node 26's migration keeps all five gates, and moving the drift check last strengthens it — every earlier step's output is now inside the comparison. I checked the "nothing tracked is left behind" claim against `.gitignore`: the tarball, `prepack`'s declarations and `flake.lock` are all ignored, so it holds. `AGENTS.md`'s new wording is accurate — exactly two `&&`-steps exist in the generated workflow and they are the two it names. Only `nix/node22` is unevaluated now: the sole `nix develop` paths are `./nix/node24` and `./nix/node26`. Your `26.7.0 -> 26.99.0` mutation is still caught by the drift gate — regeneration produces a 16-line `ci.yml` diff including `= v26.99.0`, though whether the pinned snapshot rejects it inside `nix develop` is unverifiable here, since Nix is not installed.

My typedef-gate finding is closed by `d5f1c46af`. Three things left, all documentation:

**`66b-dockerfile-nix-integration.md:227` contradicts what shipped, under a heading that makes it worse.** The block is titled "Preserve the current command sequences and order" and records node26 as `npm ci`, `npm run ci-update`, the drift check, `npx tsc`, `npm run cov`, `npm pack`. The generated workflow now runs `npx tsc`, `npm run cov`, `npm pack`, `npm run ci-update`, then the drift check. The same file's Progress note, added in this delta, documents the new order correctly — so the file answers the question twice, differently. The block's `node22` entry is stale too: `functionalscript@0.38.0` and `fjs t` where the workflow has `0.47.0` and `fjs test`, and no version check. That block is the instruction sheet for the next migration.

**Both READMEs now name the wrong proof file.** `a31c603fd` moved the pinned-commit and `devShells.<system>.default` assertions out of `fjs/ci/proof.f.mjs` into `fjs/ci/nix/proof.f.mjs`'s literal fixtures — the ci proof's own new comment says so. But `fjs/ci/README.md:93` still says the three properties are "asserted by `proof.f.mjs`" and `nix/README.md:44` says "asserted by `fjs/ci/proof.f.mjs`". Only the `nodejs_<major>` one is still there. Both sentences were written or edited in this delta.

Minor: changelog entry 3 is now 301 characters against the ~250 guideline (the others are 145/215/196). It grew when the Node 26 migration was folded in; splitting the ordering sentence out would fix it.

Gates: `tsc` 0, `npm test` 3543/3543 at both heads, round-trip empty, leaf diff one 1:1 rename (`migratedNodeJob` -> `migratedNodeJobs`). Static assert calls in `fjs/ci/proof.f.mjs` go 82 -> 78, accounted for: the `includes` triple collapses into one stronger `assertEq`, and `migratedNodeJobs` drops two statements while looping over two jobs, so executed assertions rise 7 -> 8. Two vanish outright — the `nix-flakes`-absent check, still covered by the 13-job count, and `!hasRunInJob('node22', 'nix develop')`, now only partly covered.

o2alexanderfedin (commented):
The substance is sound. The `shellHook` removal takes its only reason with it — Node 22 no longer does the global install the hook existed to make writable — and the one vanished proof leaf, `nixFlakes.nodeShellHook`, asserted a fact that is now false. The capability it protected is still pinned: I changed the emitted attribute from `shellHook` to `shellHookX` and `flakeText.shellHook` goes red on its own, so the fixture outliving its user is deliberate rather than orphaned. All three new `matrixShape` asserts are live under separate mutations, the carried-forward `!hasRunInJob('node22', 'nix develop')` assert survived, and the `22.23.2 -> 22.99.0` mutation still fires `nodeVersionChecks` and only that. Node 22's version check is now literally the job's first `run` step. Leaf count 3543 -> 3542, the diff being exactly that one leaf and nothing else.

Two documentation items.

**`fjs/ci/README.md:101` says the platform jobs are "the only jobs that exercise the published CLI".** They are not: `deno` runs `deno run -A --minimum-dependency-age=0 npm:functionalscript@0.47.0 test` and `bun` runs `bunx functionalscript@0.47.0 test`, both from the registry. That is eight jobs, which is what `built-package-checks.md` says two files away in this same commit — "Eight CI jobs install `functionalscript` and run it". Dropping "the only", or scoping it to the Node jobs, reconciles them.

**Still open from the previous round:** `fjs/ci/todo/66b-dockerfile-nix-integration.md:37-42` continues to credit the accepted commit, `devShells.<system>.default` and the `nodejs_<major>` all to `fjs/ci/proof.f.mjs`. Both READMEs now split those correctly between that file and `fjs/ci/nix/proof.f.mjs`; this paragraph is the last holdout, and it sits in a file this commit edited by 41 lines.

Optional: `fjs/ci/proof.f.mjs:279-293`'s `globalInstall` predicate now filters nothing — no Node job has an `npm install -g` step any more — and its comment still describes that install as a live, deliberately allowed step whose relocation is "`built-package-checks.md`'s business".

`tsc` 0, `npm test` 3542/3542, round-trip empty. Nix is not installed here, so every flake claim is generator output and file text, never an evaluation.

o2alexanderfedin (approved):
Approved. All three items from the last round are closed, and the exemption removal is a strengthening rather than a coverage loss — under the old proof `nodeVersionChecks` passed with a global install inserted ahead of the check, under the new one it fails, while the `echo hi` control fails on both. Leaf names and assertion counts in `fjs/ci/proof.f.mjs` are identical across the delta.

o2alexanderfedin (approved):
Still approved. The concrete path checks out: `nix/generated/node24/flake.nix` was there at the merge-base and `nix/node24/flake.nix` is there now, `generatedDirectory` is `nix`, and the generated workflow runs `nix develop ./nix/node24`. Entry 1 goes 140 to 172 characters; the other six are untouched. `tsc` 0, `npm test` 3542/3542, round-trip empty.

PR #1778, opened 2026-08-29
djs/tokenizer: anchor unterminated comments at their opening

Trinidadec:
Fixes the concrete half of [error-position-range](https://github.com/functionalscript/functionalscript/blob/main/fjs/djs/tokenizer/todo/error-position-range.md): two unterminated constructs reported two different positions.

## What was wrong

```
"value            → 1:1    the opening quote
"unterminated     → 1:1
/* c              → 1:5    the end of input
/* abcdefghij     → 1:14
```

A string was anchored at its opening; a comment at wherever the input ran out. The comment's content is consumed before the `unterminated` tag is emitted, so the position lookup found nothing after the tag and fell back to the last position available. The proof covering it said so out loud — *"an unterminated block comment is the exception"* — which is a defect written down rather than a convention.

## What changed

Both now anchor at the construct's opening, which is what the reader is being told about: *this* `/*` was never closed. It is what TypeScript reports for the same input, and what the string case already did.

The search runs backwards from the error tag, so `/* ok */ /* bad` blames the **second** comment at 1:10 rather than the first.

Messages now use values that already existed in `_ErrorMessage` and were never emitted:

| input | before | now |
| --- | --- | --- |
| `/* c` | `invalid token` @1:5 | `*/ expected` @1:1 |
| `/* ok */ /* bad` | `invalid token` @1:16 | `*/ expected` @1:10 |
| `123abc` | `invalid token` @1:4 | `invalid number` @1:4 |

## What deliberately did not change

**A malformed number keeps a different anchor.** `123abc` points at the `a`, not at the `1`, because the offending character is what the reader has to change. Different error *kinds* warrant different anchors — the inconsistency was between two constructs asking the same question, not between these two.

**An unterminated string stays `invalid token`.** That path is a hard grammar failure, and the tokenizer cannot yet tell an unterminated string from a bad escape like `"a\qb"` — both fail the whole rule at the opening quote. Distinguishing them needs the grammar tagging in [error-message-specificity](https://github.com/functionalscript/functionalscript/blob/main/fjs/djs/tokenizer/todo/error-message-specificity.md); guessing would report the wrong cause, which is worse than reporting a vague one.

**The end of a span is untouched.** `error-position-range` keeps that half and stays open: nothing renders a span, and `errorLocation` prints `path:line:column` and would discard an end.

## Verification

- `npx tsc` — 0 errors.
- `fjs t` — 3544 pass, 0 fail.
- `npm run cov` — exit 0. Two fallback branches in the new helper were unreachable by construction (an `unterminated` tag exists only inside a `comment`, which always carries its opening `/`), so they are asserted rather than silently defaulted.

Four existing expectations encoded the old behaviour and are updated, including the one that named it "the exception".

Changelog:
- `djs/tokenizer`: an unterminated block comment is reported at its `/*` rather
  than at the end of input, matching an unterminated string. Unterminated
  comments now say `*/ expected` and malformed numbers `invalid number`, in
  place of a generic `invalid token`.


MaryManzhula:
@codex review

sergey-shandar (approved):

PR #1807, opened 2026-08-30
ci: run every job that can in the developer shell

sergey-shandar:
Seven generated flakes become three. `deno`, `bun`, `wasm` and Node 26 run their commands in `nix/dev` — the shell a developer enters — instead of each having one of its own.

## Why sharing is safe, and where it is not

The old rule was that jobs must not share, because a shell with five runtimes would let a job pass on whichever `node` reached `PATH` first. That risk is real, and it is **narrower than the rule it produced**: it applies only where a command resolves its runtime from `PATH`.

| job | command | runtime |
|---|---|---|
| `deno` | `deno task cov` | named |
| `bun` | `bun test --coverage` | named |
| `wasm` | `cargo test --target …` | named |
| `node26` | `tsc`, `npm run cov` | named / the shell's own `node` |
| `node22`, `node24` | `npm ci`, `node --test` | **whatever reaches `PATH` first** |

The first four cannot be handed the wrong runtime by what else is installed, so they share. The last two can, and one shell holds one `node` — so they keep a flake apiece carrying the single release each exists to test. Node 26 needs no flake of its own because the release it wants *is* the shared shell's.

That split is a proof, not a comment: `nixCoverage` asserts that exactly one declared flake is entered by more than one job, and that every other is entered only by the job named after it.

The payoff is bigger than the tidier tree — **the environment CI proves is now the one people work in**, rather than a fifth arrangement nobody uses.

## The `dev` job is gone

It existed for one reason: nothing else entered that flake, so it would rot until a developer's shell failed to build. Four jobs entering it on every pull request answers that better than one job asserting six versions. Between them they still assert all six — `node` and `tsc` from `node26`, `deno` from `deno`, `bun` from `bun`, both WASM runtimes from `wasm`.

13 jobs instead of 14.

## What I want reviewed

**The cost is per-job download.** A `deno` job that used to realise one package now realises the whole closure, Rust toolchain included, and there is no binary cache of our own yet. I have not measured it. `096-ci-caching.md` is where that goes, and this change makes it worth more — four jobs now want the same closure.

**One consequence for a project that is not this one.** `nixJobs` is a list rather than a function of the project, so a project without a `Cargo.toml` gets no `wasm` job and therefore nothing checking the two WASM runtimes in its shell. Same trade `ci-generator-audience.md` describes for every job this generator writes unconditionally, and recorded in `65z-ci-nix.md`.

**Four orphaned directories were deleted by hand** — `nix/{deno,bun,wasm,node26}`. The generator still cannot remove stale output; that is the open item in `65z-ci-nix.md` waiting on a recursive `rm` effect.

## Structure

`nixShell` moved to `fjs/ci/nix/module.f.mjs`, because the flake the jobs share is a property of the Nix layer rather than of any one job — and because `dev` importing from `bun`/`node`/`rust` while those imported an id back from `dev` would have been a cycle.

`fjs/ci/dev/module.f.mjs` is now the single environment declaration and has no steps. It still takes each tool from the module that owns the commands using it — `bunPin` from `../bun`, `wasmRust` and `wasmPackages` from `../rust` — so nothing here is a second copy. `takesEachToolFromItsOwner` asserts identity rather than equality, since a structurally equal copy would satisfy a comparison and still be a second place to edit.

## Verification

`tsc` clean, 3630/3630 proofs, 100.00 lines/branches/functions, no generator drift.

Not verified: that the shared shell builds under Nix — no Nix in this container, as on the previous PRs. Every job in this workflow evaluates it, so a failure is loud and immediate rather than latent.

Changelog:

- **BREAKING CHANGES:** `ci`: the exported `NixJob` declarations `denoNixJob`,
  `bunNixJob` and `wasmNixJob` are gone, and `nodeNixJobs` no longer carries the
  default Node version. The jobs those described now enter one shared shell, so
  `nixJobs` is three flakes rather than seven — `dev`, `node22` and `node24`.
  `wasmRust` and `wasmPackages` replace `wasmNixJob` for a caller that wanted
  the toolchain or the runtime attributes.
- `ci`: every canonical job but Node 22 and Node 24 runs its commands in the
  generated developer shell, `nix/dev`, rather than in a flake of its own. A
  command that names its runtime — `deno task cov`, `bun test`, `cargo test`,
  `tsc` — cannot be served the wrong one by what else is installed, so those
  jobs share; `npm ci` and `node --test` take whichever `node` reaches `PATH`
  first, so the two older Node versions keep a shell holding the single release
  each exists to test. CI now proves the environment developers work in.
- `ci`: the `dev` job is gone, leaving 13 jobs. It existed because nothing else
  entered that flake; four jobs now enter it on every pull request, and between
  them still assert all six versions it provides.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DSZbnNEKjvNRbPijo6wt61)_


o2alexanderfedin:
Merged before I finished, so this is a follow-up rather than a verdict. The architecture holds and the safety argument survives testing — I checked it from the generated YAML rather than the table, and each migrated job does assert its own runtime version as its first step in the shared shell. One real gap, now on `main`.

**The shared shell can silently lose `deno` or the compiler and the suite stays green.** `fjs/ci/dev/proof.f.mjs` holds `devNixJob` to its Node (`oneNodeAndItIsTheDefault`), its Rust and WASM runtimes and its Bun pin (`takesEachToolFromItsOwner`), and `git` (`shape`) — but the file does not contain the strings `deno` or `typescript` anywhere. Deleting `'deno'` from `devNixJob.packages` regenerates `nix/flake.nix` without `pkgs.deno` and leaves the suite at `pass: 3635, fail: 0`; deleting `typescript.attribute` does the same. Both mutations confirmed applied before trusting the green.

Those two are precisely the unversioned attributes, and `fjs/ci/deno/proof.f.mjs:48` already hands the job off:

> `../dev/proof.f.mjs` is where that shell is held to providing a `deno`.

It is not. Every `deno` assertion in `fjs/ci/**` is about the job's command lines — `deno task cov`, `deno install --frozen`, no `setup-deno` — never about the shell carrying a `deno`. The retired `fjs/ci/deno/proof.f.mjs::nixJob` used to assert `denoNixJob.packages[0] === 'deno'` and nothing replaced it. Two lines in `takesEachToolFromItsOwner` close it:

```js
assert(devNixJob.packages.includes('deno'), devNixJob.packages.join(' '))
assert(devNixJob.packages.includes(typescript.attribute), devNixJob.packages.join(' '))
```

Smaller, same visit:

- `nix/README.md:58-65` still explains a `case` line the generated `run` script no longer has — `b7bc81f3` replaced `$0`-derivation with a written-in path, and `runText`'s own doc comment was updated while the README was not. The same paragraph cites `./nix/deno/run`, which this PR deleted.
- `changelog/unreleased/1803.md:17` still says `nix develop ./nix/dev`. It is unreleased, so it ships naming a directory that is gone.
- `devSteps` is removed but not named under `**BREAKING CHANGES:**`, and `runText` changed from `string` to `(id: string) => string` without its new arity being listed. No `exports` map, so both are published breaks.
- `fjs/ci/todo/66h-ci-npm-global-install.md` argues against itself: "Why this still qualifies" bullet 1 still says only one consumer survives and the threshold no longer holds, under an `**Update:**` saying the second reappeared and it does.
- Three of `1807.md`'s seven entries run 438-576 characters against `changelog/README.md`'s ~250 per entry.
- The body says 3630 and the last commit message 3633; it is 3635.

Verified separately, since it was the other load-bearing claim: the platform jobs really do not load `node_modules` — I moved it away entirely and the suite still passed — and the removed install is not the published-CLI path, which still runs in all six. Both directions of that change are gated by real leaves.

PR #1809, opened 2026-08-31
emergent_testing: the browser's report is folded in FunctionalScript

sergey-shandar:
`reportOf` counted results, decided the run's pass/fail status and assembled the
report — arithmetic over values — and was impure for exactly one reason: it read
`navigator.userAgent` itself.

Asked precisely, that is not a computation at all. It is a string the host
knows, so the host passes it in alongside the wall clock, and the fold moves to
`browser/module.f.mjs`. Nothing was designed to make this possible: no new
operation, no interpreter, no decision about what a "run the suite" effect would
look like.

## Why this one first

`browser-runner-functional-script.md` measures the remaining debt as three
definitions. `reportOf` was the smallest and the only one that needed nothing
invented — `runBrowserProofs` owns the interpreter, and `startBrowserTestSources`
owns `import()` and the loading sequence, so both come with a design question
attached. Taking this one first also shrinks what those two have to carry.

The debt row is now ~134 lines of the file's 298, down from ~155 of 320, and the
todo records the sharper test the move suggests: not *"does it touch a host
object"* but *"which values does it need from one"*.

## The proofs no longer need a DOM

They ran only through `browser/proof.mjs`'s 524-line stand-in; they are now
ordinary `.f.mjs` proofs pinning what a fold owes a consumer:

- any failure fails the run; an empty run passed, because nothing failed;
- an override wins over the fold — that is the case it exists for, a run that
  never reached its leaves — but it does **not** suppress the counts, since
  totals disagreeing with `results` would read as an empty suite rather than a
  broken one;
- the host's two facts travel through untouched, which is the statement that
  this function reads no clock and knows no browser.

## Verification

`npx tsc` clean · `fjs t` 3640/3640 · `node --test` 3578/3578 · coverage 100% ·
`npm run ci-update` no diff.

## Changelog

- `emergent_testing`: new `browser/module.f.mjs` `reportOf` folds a browser
  run's results into its report — counts, status, and the host's own browser
  string and duration passed in rather than read

Not breaking: nothing moved off a path a consumer could import, and the report a
page publishes is unchanged field for field. It is an *addition* because
`reportOf` is exported so the host half can import it, and with no `exports` map
in `package.json` an exported name in a `.f.mjs` is deep-importable — the same
reasoning #1806 had to be told.


o2alexanderfedin (approved):
Approving. The move is behaviour-preserving and I checked it by construction rather than by reading the diff: the base's `reportOf` against the head's exported one over 160 cases — eight result fixtures crossed with five status arguments and four durations — gives zero differences, identical key order and identical `results` identity. `undefined ?? x` and `null ?? x` agree, so the `undefined` → `string | null` change is invisible. Regenerating at both commits leaves `index.html`, `_browser-test-entry.mjs` and `_browser-suite.mjs` byte-identical. No published break either: the old `reportOf` was module-private, and `browser/module.mjs`'s export list is unchanged, so nothing externally importable moved. Coverage is +5 leaves, none vanished, 3635 → 3640.

Two things.

**One of the five new proofs is missing its subject.** Replacing `results` with `[]` in the returned report passes **all five** `reportOf` proofs — they assert `status`, `totals.*`, `browser` and `duration`, and none of them looks at `r.results`. The impure `browser/proof.mjs` catches it, which is exactly the dependency this PR exists to remove: the premise is that this logic is now provable purely, and the one property that says the results survive the fold is the one the pure proofs do not check. `empty`'s own comment even reasons about `results` — "totals that disagreed with `results` would read as an empty suite rather than a broken one" — without asserting it. One `assertEq(r.results.length, …)`, or a structural check in `folds`, closes it.

Worth knowing while you are there: `folds` uses one passed and one failed, which is symmetric under a `passed`/`failed` swap, so that mutant is caught only by `overrideBeatsThePassingFold`'s asymmetric `{tests:1, passed:1}` fixture. That assertion carries more than its comment claims.

**The wrong ratio is still there, and it is my error you inherited.** `browser-runner-functional-script.md` still reads "is 524 lines — about seven times the next impure proof file in the repository, `effects/node/memory/proof.mjs` at 78". I gave you that figure and it was wrong: I globbed bare `proof.mjs` and missed every `*.proof.mjs`. By the repo's own definition — `shouldLoad` at `fjs/dev/module.f.mjs:43-47` — the next largest is `fjs/rtti/host.proof.mjs` at **379**, so the ratio is about **1.4×**, and `effects/node/memory/proof.mjs` at 78 is the *smallest* of them, not the next largest. This PR revised the paragraphs around it and is the natural place to fix it.

The numbers this PR does correct all check out: `module.mjs` is 298 lines, the DOM-adapter row measures 84 against the stated ~85, and dropping "majority" for "largest part" is right — 140 of 298 is a plurality. Ticking no task is also right: `proof.mjs` is still 524 lines and both host-touching functions remain, so task 1 is genuinely unfinished.

`tsc` clean at both commits, `pass: 3640, fail: 0` at head against 3635 at the merge base, `npm run website` exit 0 leaving no diff, changelog entry 204 characters.

sergey-shandar:
Both fixed in `2be74cc2`.

**1. The missing subject — and it was the PR's own premise that it broke.** You are right that this is the sharpest form of the finding: the claim is "this logic is now provable purely", and the single property saying the rows survive the fold was the one only the impure file checked. `empty`'s comment reasoning about `results` without asserting it makes it worse, not better.

`folds` now compares the rows structurally against the input; `empty` and `overrideBeatsThePassingFold` assert their length — the override case deliberately, since a run that failed before producing rows is exactly when a consumer needs whatever did land. Mutation-checked rather than assumed: with `results: []` in the returned report, `folds` and `overrideBeatsThePassingFold` both fail (23 failures overall, the rest being `browser/proof.mjs` as before); restored, 3640/3640.

Thank you for the note on `folds` being symmetric under a `passed`/`failed` swap — that mutant is indeed carried by `overrideBeatsThePassingFold`'s asymmetric fixture rather than by the proof that looks like it should catch it. Worth a separate look at whether the asymmetry should be stated where it is relied on.

**2. The ratio.** Checked against `shouldLoad` (`fjs/dev/module.f.mjs:43-47`) rather than taking either of our figures: `*.proof.mjs` files are impure proofs too, so the set is six files totalling 1,434 lines, the next largest is `rtti/host.proof.mjs` at **379** — about **1.4×**, not seven — and `effects/node/memory/proof.mjs` at 78 is the smallest, not the next.

Rather than swap in 1.4×, which no longer carries the paragraph's point, I replaced it with a measure that does and that I verified: 524 lines is the **largest** impure proof file in the repository and **more than a third of all impure proof code**. The comparison to `rtti/host.proof.mjs` stays, with what that file is for — building values FunctionalScript cannot express — since that is the contrast worth drawing: one of these two exists because the subset cannot reach the values, the other because the logic has not moved yet.

`tsc` clean, `fjs t` 3640/3640, `node --test` 3578/3578, coverage 100%, `ci-update` no diff.

---

PR #1811, opened 2026-08-31
nanvm: implement addition

sasha-gil:
Implement JavaScript addition as a primitive-first operation: concatenate when either primitive is a string; otherwise add matching numeric kinds and reject Number/BigInt mixing.

Adds shared JavaScript/Rust conformance cases for numeric addition, string concatenation, coercible arrays and objects, BigInt operands, and mixed-kind throws. The old direct `Unpacked` addition, which could panic on ordinary operands, is removed.

Validation: `npm run ci-update`, targeted `fjs/nanvm` proofs, `cargo test`, `cargo clippy`, and `cargo fmt -- --check`.

Changelog:
- nanvm: Add JavaScript-compatible addition for VM values.


o2alexanderfedin (commented):
Reviewed as the delta over this PR's base branch — `#1810` at `94a3b37dc739adf0828db6cb9c06c97bb18fb6a0`, which is also the merge base. Note the base branch has since moved to `7b690b47` ("cover subtraction edge cases"), which this head does not contain.

`+` is correct, and I checked it rather than reasoning about it. `any/add.rs:9-10` applies ToPrimitive to both operands left-then-right before either is inspected, `:11` tests the ToPrimitive *results* rather than the originals, `:14` is ToString left-first, `:18` ToNumeric left-first then the mixed-kind reject. Exercised with a temporary probe (confirmed it could go red, then reverted): `[] + []` → `""`, `[] + {}` → `"[object Object]"`, `1n + "2"` → `"12"` with no TypeError, `1n + 1` → TypeError, `Infinity + -Infinity` → NaN, and the signed zeros compared by `to_bits` — `-0 + 0` → `+0`, `0 + -0` → `+0`, `-0 + -0` → `-0`.

The `impls/add.rs` shrink is subtractive only: the deleted impl's three real arms are subsumed by a path that handles strictly more, and the `todo!()` fallthrough is gone. The `unknownOperation` proof's move to `'/'` was forced and narrows nothing — `'/'` is absent from both `op2Rust` and `op2Js`, is a valid `Op2Id`, and the leaf is still green. Leaf names diffed base-to-head: 3661 → 3677, **zero removals**, 16 added, all `proof["+"].*`.

Four findings.

**1. Two undocumented published breaks, both invisible to a `pub` grep.** `Add for Any<A>` already existed at the base with `type Output = Self`; it is now `Result<Self, Self>`, so external `a + b` on `Any` changes type. And `impl Add for Unpacked<A>` is deleted outright, while `Unpacked` is re-exported from `src/vm/mod.rs:29`. The changelog says only "Add JavaScript-compatible addition for VM values."

Worth flagging the detection gap as much as the breaks: `git diff … | grep -E '^[-+].*\bpub\b'` returns nothing here, because Rust trait impls carry no `pub` keyword. That check does not see either break.

**2. The `-0` asymmetry — the one thing that separates `+` from `-` — is pinned by nothing.** `fjs/nanvm/module.f.mjs:379-395` has no signed-zero case, and `a + b` → `a + b + 0.0` stays green across the whole suite. The machinery exists and is documented for precisely this: `tests/test/harness.rs:35-51`'s `same()` compares numbers by `to_bits` because "`0` does not match `-0`". It is unused for `+`. The behaviour is correct; it is unguarded.

**3. `CANNOT_MIX_NUMBER_AND_BIGINT` (`numeric.rs:5-6`) is still pinned by nothing, and this PR enlarges the unpinned surface.** Replacing the whole message text leaves `cargo test` green. The `main.rs` 9/3 here is mechanical — `.unwrap()` calls in `bigint_add` forced by the new `Result` output — not a pin. `+` makes that const reachable through a third operator (three call sites now). One `assert_eq!` in the shape of `unary_plus_bigint_message` at `main.rs:151-156` closes it.

**4. `nanvm-lib/todo/operator-impl-placement.md:14-15` is now stale, caused by this PR.** It says `impls/add.rs` "mixes it with the unrelated `Add for Unpacked<A>`" — the impl this PR deleted. Its other cites still check out. Separately, the `Sub`/`Mul` asymmetry I raised on #1810 is untouched: `Add for Any` already lived in `any/add.rs`, and `Mul for Any` is still in `impls/mul.rs`.

On the head commit `a0f25c9e`, which adds 13 process lines to `AGENTS.md:23-35` in a PR about addition: the content is grounded — node22 and node24 do run `node --test` — but "run the same Node suite CI runs" is loose in a normative doc. `node --test` is a strict **subset** of `fjs test`: 3615 leaves against 3676, omitting all of `fjs/types/prime_field`, `range_map` and `crypto/sha2`. It sits directly under a line that already says "`fjs test` # or any equivalent runner", which reads as though the two are the same suite.

Smaller: `mulCases` has `functionByOne` and `subCases` has `functionMinusOne`, but `addCases` has no function operand, so `function + 1` → `"function1"` is untested; the README `+` row no longer mentions that it can throw while the `-` row does; `1811.md` has no trailing newline.

Two mutations came back green without being gaps — `b + a` (commutativity) and the ToPrimitive order swap. The latter is green because `value_of` (`primitive_coercion.rs:37`) is a stub returning `None` and every `*_to_string` returns `Some`, so ToPrimitive can neither throw nor have side effects and the `CANNOT_CONVERT_TO_PRIMITIVE_VALUE` branches are dead. That is worth knowing when those stubs are filled in — the order will start to be observable, and nothing currently pins it.

Gates at `dff91a4a` (code) and `a0f25c9e` (`node --test`): `ci-update` round-trips with no diff, 3677/3677, `tsc` 7.0.2 clean, `cargo test -p nanvm-lib` 110+1+5, `cargo clippy` clean, `node --test` 3615/3615. `cargo clippy --all-targets` fails on `src/vm/mod.rs:40` (`clippy::eq_op`), verified identical at the base — not this PR's.

sergey-shandar:
@sasha-gil drift test failed: it means you need to regenerate the code 

```sh
npm run ci-update
```

o2alexanderfedin (commented):
Measured over this PR's own commit `3c6dcc44`; the rest of the delta is the merge of #1810's branch, which I have already reviewed and approved at `7b690b47`.

**`3c6dcc44` edited `fjs/nanvm/module.f.mjs` without regenerating `nanvm-lib/tests/test/generated.rs`, and CI is red on it.**

```
$ npm run ci-update && git status --porcelain
 M nanvm-lib/tests/test/generated.rs
$ git diff
+    check::<A>("negativeZeroPlusZero", (-0f64).to_any() + (0f64).to_any(), (0f64).to_any());
+    check::<A>("zeroPlusNegativeZero", ...);
+    check::<A>("negativeZeroPlusNegativeZero", ...);
```

`node26` reports `completed/failure`; `ci.yml:445-448` runs `ci-update` then `git add -A && git diff --cached --exit-code`.

This is not cosmetic — **the three lines it fails to emit are the whole `-0` fix on the Rust side.** Mutating the `Add` arm in `numeric.rs:42` to `a + b + 0.0` stays **green** at the committed head, and goes **red** on `harness.rs:56` (`negativeZeroPlusNegativeZero: 0.0 is not -0.0`) after `ci-update`. So the finding I raised is fixed on the JS side only, and the missing regeneration is the sole cause. The JS rows themselves are right and discriminate cleanly.

The rest:

- **The two published breaks — fixed.** `- **BREAKING CHANGES:** \`Any + Any\` now returns \`Result\`, and \`Unpacked\` no longer implements \`Add\`.` Both verified real against the merge base. One nit: the continuation line is indented with a literal tab, and `grep -rlP '\t' changelog/` matches this file and nothing else.
- **`operator-impl-placement.md` — fixed, and true rather than merely different.** Every cite checked line by line, and `Status: open` is still right since `Mul for Any` remains in `impls/`. It now omits `any/sub.rs` from its list of `any/` operators — incomplete, not false.
- **README `+` row — fixed**, and accurate against `any/add.rs:18`.
- **`CANNOT_MIX_NUMBER_AND_BIGINT` for `+` — still unaddressed.** The pin that arrived from #1810 asserts `*` and `-` only (`main.rs:165-166`). Giving the `Add` arm its own message literal stays green, so `+`'s message is unpinned even though the shared constant is not.
- Still open from the smaller list: `addCases` has no function operand, so `function + 1` → `"function1"` is untested; and the `AGENTS.md` line added on this branch still says "the same Node suite CI runs" while `node --test` is 3641 leaves against `fjs test`'s 3703 — 62 omitted, across `prime_field`, `range_map`, `bit_vec`, `sha2`, `list`, `bigint`, `emergent_testing`, `nibble_set`, `dev` and `secp`.

Leaves `a0f25c9e` → head: 3677 → 3703, none removed; 3 are this PR's `-0` rows, the other 23 came in with the merge.

o2alexanderfedin (approved):
The regeneration closes it, and not only at the gate. `ci-update` now leaves an empty diff, and mutating the `Add` arm in `numeric.rs:42` to `a + b + 0.0` goes **red** — `harness.rs:56: negativeZeroPlusNegativeZero: 0.0 is not -0.0` — where it was green at the previous head. The Rust-side `-0` coverage genuinely exists now, and the JS side still discriminates on its own. CI is 20/20, including `node26`.

One thing still open, unchanged from last round: `+`'s mixed-kind TypeError message is pinned by nothing. Giving the `Add` arm alone a distinct message literal leaves `cargo test` green — `mixed_numeric_operands_message` asserts `*` and `-`, and `check_throws` only asserts Err-ness. Mutating the shared constant does go red, but on the `*` line, so it says nothing about `+`.

Smaller, also unchanged: `addCases` has no function operand, so `function + 1` → `"function1"` — the concat path unique to `+` — is untested; `operator-impl-placement.md` still omits `any/sub.rs` from its list of `any/` operators now that #1810 has landed; and `1811.md` keeps its literal tab (still the only file in `changelog/` with one) and has no trailing newline.

Worth noting the leaf count hid something: 3703 → 3707 is **5 added and 1 removed**, not +4 — `nixFlakes.threeQuiets`, superseded by `oneQuiet`. All six belong to the merges; this PR added no fjs leaves, which is right for a commit that regenerated the Rust side.

3707/0, `tsc` 7.0.2 clean, `cargo test` 110/1/5, round-trip clean. The two published breaks are correctly and completely described.

PR #1813, opened 2026-08-31
ci: generate flake.lock, and take back two --quiet flags

sergey-shandar:
Every generated flake gets a `flake.lock` beside it, written by the generator and committed. `--quiet` drops from three to one.

```
  nix/flake.lock                  + 49
  nix/node22/flake.lock           + 27
  nix/node24/flake.lock           + 27
  nix/ubuntu-intel32/flake.lock   + 49

- exec nix develop --no-write-lock-file --quiet --quiet --quiet ./nix --command "$@"
+ exec nix develop --no-write-lock-file --quiet ./nix --command "$@"
```

## What the two extra `--quiet`s were paying for

With no lock on disk, every `nix develop` computed one, found it differed from nothing, and — because `--no-write-lock-file` keeps the checkout clean — said so, five lines at a time, on **every step of every Nix job**. Around forty lines per run for `wasm` alone.

Global verbosity is the only lever Nix has against a warning: `--verbose`, `--quiet` and `--debug` are the whole logging category, and verbosity is not a `nix.conf` setting, so `--option` cannot reach it. Getting below `lvlWarn` to hide that one warning hid them all — a failing substituter, a dirty tree, a deprecation notice. The lock removes the cause, and the two flags come off with it.

## Generated from data, so `fjs ci` still runs on Windows

Nothing here runs `nix flake lock`. `todo/65z-ci-nix.md` requires the generator stay Nix-independent and `CONTRIBUTING.md` supports a Windows contributor with no Nix at all, so the two values a lock adds on top of an already-exact revision — `narHash` and `lastModified` — are data in `config/module.f.mjs`, the way `bunSources`' archive hashes already are.

That is not only about Windows. A generated lock stays **inside the drift check**: bump the Nixpkgs pin without its hash and `node26` fails, where a hand-written lock would rot silently the first time a pin moved. The cost lands on whoever moves a pin rather than on everyone who regenerates, and `config/module.f.mjs` records both ways to get the values back — `nix flake metadata` on a machine with Nix, or a CI log, which lists every input's `narHash` in exactly the warning a missing lock produces.

`.gitignore` no longer hides `/nix/*/flake.lock`. That rule existed for a hand-run `nix develop` writing one; these are generated output now, and the drift check has to see them.

## Verification: run against real Nix, not reasoned about

Nix 2.24.9 installed in the build container, which is new — every earlier round in this area read flake text and generated YAML without ever evaluating anything.

**All four flakes resolve entirely from their generated locks.** Zero `not writing modified lock file` warnings, and **no network access at all** — GitHub tarball fetches are blocked in that container, and `nix flake metadata` still succeeded on every flake, which is only possible if nothing needed re-resolving.

**The check is not vacuous.** Remove a lock, or corrupt a single `narHash`, and Nix goes straight back to the network to recompute one:

```
lock present     → 0 re-lock warnings, no fetch
lock removed     → fetches immediately (blocked here, so: error)
narHash corrupt  → fetches immediately (blocked here, so: error)
```

**The verbosity arithmetic is now measured rather than asserted.** The docstrings have claimed this count for several PRs; nobody had run it:

```
quiets=""                         warnings shown: 1
quiets="--quiet"                  warnings shown: 1
quiets="--quiet --quiet"          warnings shown: 1
quiets="--quiet --quiet --quiet"  warnings shown: 0
```

**The nixpkgs values are confirmed three independent ways**, since they were transcribed from a CI log rather than computed: the warning `todo/generated-flake-lock.md` recorded, `git log -1 --format=%ct` for the committer date, and Nix recomputing both from source over `git+https` — a different fetcher, going to the network rather than reading the lock — which returned `sha256-BZWCi9ZR…` and `1787753485`, identical to what is committed. `rust-overlay`'s two are confirmed twice, by the first and second of those.

## The two flags that remain, and what they are actually worth

Both were carried over from before the lock existed, so both were re-measured rather than re-asserted. Both stay, and one of my own claims about them did not survive the measurement.

**`--no-write-lock-file` is a guard, not a fix.** With a correct lock, Nix writes nothing whether or not the flag is passed — the file comes through byte-identical with its mtime untouched, because Nix only writes when the lock it computes differs from the one on disk. What the flag buys is that case: the generator owns this file, and without the flag every Nix step in every job is a writer of a tracked one. Nothing is lost by refusing that, since `npm run ci-update` regenerates the lock from `config/module.f.mjs` and would revert a rewrite anyway — two writers where one is authoritative is churn, not redundancy. A future Nix whose lock schema moves past version 7 is where it would be churn on every step at once.

**`--quiet` costs nothing that reports a problem.** Measured, it removes `this path will be fetched (N MiB download)`, one `copying path '…' from '…'` per store path, `this derivation will be built:` and `building '…'` — all `lvlInfo`, all progress rather than outcome. A warning survives it, and so does a failing build's log, which arrives inside the `lvlError` message as `last N log lines` rather than as the suppressed build output. The one real loss is that a cache miss looks like a cache hit: Nix compiles from source in silence and the job is only slower. That is bounded — the store persists across a job's steps, so substitution happens on the first `./nix/run` and no other.

**A correction, since it was in an earlier version of this description.** I claimed matching Nix's alphabetical key order was "not cosmetic", on the grounds that a differently ordered file would be rewritten and show up as drift. It would not: Nix compares the *parsed* lock, not the text. Reversing every key and re-indenting to four spaces produces no rewrite at all. Matching the order makes these files read like every other `flake.lock` and keeps a diff against one Nix did write about content — that is the whole of it, and the docstring now says so.

## Proofs

`lockText` is pinned in full for a one-input flake, and the `follows` node — `["nixpkgs"]`, the path form Nix uses for an input redirected to another node — for a two-input one. Without it the lock would carry a second Nixpkgs revision the flake never asked for. `threeQuiets` becomes `oneQuiet`, and a `lock` proof asserts the file is written beside every flake.

`todo/generated-flake-lock.md` is deleted; `65z-ci-nix.md` and `66b-dockerfile-nix-integration.md` are updated where they described the old arrangement, including the two task boxes that recorded adding the `.gitignore` rule.

`tsc` clean, 3663/3663 proofs, 100.00 lines/branches/functions, `fjs ci` leaves no diff.

Changelog:

- `ci`: a `flake.lock` is generated beside every `flake.nix` and committed, from
  `narHash` and `lastModified` in `config/module.f.mjs`. Nothing runs
  `nix flake lock`, so `fjs ci` still needs no Nix and still runs on Windows.
- `ci`: the `run` script passes `--quiet` once rather than three times. The
  extra two only hid `not writing modified lock file`, and took every other Nix
  warning with them; the lock removes the cause.
- `ci`: `/nix/*/flake.lock` is no longer in `.gitignore` — those files are
  generated output now, and the `node26` drift check compares them.


o2alexanderfedin (approved):
The change is clean and well proved. The round-trip is byte-clean, the deleted todo's six tasks are all genuinely done, dropping `/nix/*/flake.lock` from `.gitignore` is not just safe but required — `git add -A` honours it, so the drift check would not have seen those files — and the new proofs bite: mutating `owner`, `version`, the `follows` shape, the remaining `--quiet`, or making `writeJob` skip the lock write each goes red naming a leaf. The last of those is caught **only** by the new proof, since the generator does not purge stale files.

Two prose fixes.

**1. `changelog/unreleased/1807.md:22-28` is made false by this PR, in the same unreleased release.** It says the run scripts "pass `--quiet` three times instead of once… three is what it takes to get below warnings", and names `` `fjs/ci/todo/generated-flake-lock.md` `` as owning the work that "would… take two of these flags back off with it". That work is this PR; the file is deleted here; `1813.md` sits in the same directory saying one. The next release notes will carry both, plus a code span pointing at a path that no longer exists.

It is a code span rather than a link, so nothing in CI catches it — and it is the only surviving referrer. The other three were correctly removed.

**2. The commit that calls its description "measured" corrected one of the three files it edited.** `nix/README.md` now lists `this path will be fetched (N MiB download)`, `copying path '…' from '…'` per store path, `this derivation will be built:` and `building '…'`. But this PR's own new text at `fjs/ci/README.md:139-140` and `fjs/ci/todo/65z-ci-nix.md:418-419` still says `copying N paths` — a different Nix message. The repo now says two things about what the one remaining `--quiet` suppresses, and the measured version is in one file of three.

Cosmetic: `nix/README.md:65` still shows `./nix/deno/run …` in a paragraph this PR rewrote; there is no `nix/deno/`.

Two notes rather than findings. A wrong pin value that is *regenerated and committed* passes both gates — I mutated `narHash`, ran `fjs ci`, and the suite stayed green with a clean second round-trip. `config/module.f.mjs:194-196` argues this is loud at Nix runtime, which is plausible and untested here. In practice it is covered from the other side: both `narHash` values and both `lastModified` values check out against real Nix-written locks in the wild, and the node key order matches Nix's own output byte for byte. Separately, the two holes I raised on #1808 are untouched and still open — dropping `'clippy'` from `wasmRust.extensions` or `'deno'` from `devNixJob.packages` each leaves the suite green with a real `nix/flake.nix` diff.

Nix is not installed here, so nothing about runtime behaviour was executed. The evidence that the generated lock is accepted comes from this head's own CI run: `not writing modified lock file` appears zero times in the node26 log, while other Nix warnings still print — which is the point of taking the two flags back.

3663/0, `tsc` clean, `fjs ci` leaves no diff, changelog entries 229 / 205 / 145.

PR #1816, opened 2026-08-31
Upgrade FunctionalScript dependency to 0.48.0

sergey-shandar:
## Summary
Updates the FunctionalScript dependency from version 0.47.0 to 0.48.0 across the CI configuration and module configuration files.

## Changes
- Updated FunctionalScript version in CI workflow configuration (`.github/workflows/ci.yml`) from 0.47.0 to 0.48.0 in two workflow jobs
- Updated FunctionalScript version in module configuration (`fjs/ci/config/module.f.mjs`) from 0.47.0 to 0.48.0

## Details
This is a routine dependency upgrade that ensures the CI pipeline and build configuration use the latest stable release of FunctionalScript. The version is maintained in a centralized configuration file and referenced in the CI workflows to keep them in sync.

https://claude.ai/code/session_01WXvL8CdQhUCbuNs5fiMEMm

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,7 +153,7 @@
           }
         },
         {
-          "run": "npm install -g functionalscript@0.47.0"
+          "run": "npm install -g functionalscript@0.48.0"
         },
         {
           "uses": "actions/checkout@v7.0.1"
@@ -203,7 +203,7 @@
           }
         },
         {
-          "run": "npm install -g functionalscript@0.47.0"
+          "run": "npm install -g functionalscript@0.48.0"
         },
         {
           "uses": "actions/checkout@v7.0.1"
--- a/fjs/ci/config/module.f.mjs
+++ b/fjs/ci/config/module.f.mjs
@@ -25,7 +25,7 @@ export const images = /** @type {const} */({
 // published FunctionalScript release; do not tie it to package.json's current
 // in-repo version.
 // https://www.npmjs.com/package/functionalscript
-export const functionalscript = /** @type {const} */ '0.47.0'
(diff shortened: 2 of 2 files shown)


o2alexanderfedin (approved):
Clean. One constant, two generated lines, round-trip byte-clean — and not hand-edited: reverting `ci.yml` alone to the merge base and regenerating reproduces the committed file exactly. 0.48.0 is published, every version token in the repo agrees, and all 20 checks are green including the two Windows jobs, which are precisely the two the pin reaches.

One line to fix, and it is this PR that stales it: `fjs/ci/todo/built-package-checks.md:27` still says the pin is "a **published release** (0.47.0 today)". That parenthetical was written when the pin *was* 0.47.0, and this is the first move since.

Worth recording rather than acting on: setting the constant to `'0.99.0'` leaves `fjs t` fully green. `ci/proof.f.mjs` interpolates the same constant, so it asserts source-against-generated agreement, never the literal; and the drift check catches editing without regenerating, not a committed wrong-but-consistent value. What actually guards it is `npm install -g` failing in CI — real, but only there. Not a defect of this PR, just where the guarantee lives.

Also not this PR's, but the sweep hit it: `built-package-checks.md:8,13`, `65z-ci-nix.md:26-27` and `fjs/ci/node/module.f.mjs:76,79` all describe this pin's reach as six jobs, or as `ubuntu-intel` plus the two Windows ones. It is two. All four are byte-identical at the merge base. `65z-ci-nix.md`'s is hard-wrapped, so `grep -n "three images"` finds nothing — only a flattened search sees it.

3707/3707 at head and at the merge base with zero leaf-name churn, `node --test` 3645/3645, `tsc` 7.0.2 clean.

PR #1823, opened 2026-09-01
bnf: validate rule mapping types

sergey-shandar:
Add type-safe terminal, sequence, variant, and repetition mapping contracts backed by RTTI.

Implement `checkMap` to infer each touched rule's input type, propagate mapped output types, preserve unmapped rules as the existing AST shape, and reject duplicate, wrong-kind, or exact-type-mismatched declarations. Track generalized metadata composition as follow-up work.

Checks:
- `npm run update`
- `tsc`
- FunctionalScript: 3,749/3,749
- `node --test`
- `cargo clippy`
- `cargo fmt -- --check`

Changelog:
- `bnf`: add RTTI-checked rule mapping contracts and reject mappings whose declared input type does not match the grammar


o2alexanderfedin (commented):
Measured against `c8d71fba` (the head moved from `e7141a7c` mid-review; the branch is now up to date with `main`, so the delta is `git diff adf61aea c8d71fba` — same 10 files). Suite 3749/0 at head against 3743/0 at base, six new leaves, none retired. `tsc` (7.0.2) clean.

**Blocking, and the fix is one line.** `fjs/bnf/lib/json/module.f.mjs:9` has a stray `import { escape } from 'node:querystring'`. The binding is never used — the only other `escape` in the file is the grammar rule key at `:22`. It is the only `node:` import in any `.f.mjs` in the repo.

The cost is not cosmetic. `npm run website` on `main`:

```
browser proof modules: 146 of 146
```

and at this head:

```
skipped fjs/bnf/data/proof.f.mjs: not linkable in a browser (node:querystring)
skipped fjs/bnf/descent/proof.f.mjs: not linkable in a browser (node:querystring)
skipped fjs/bnf/lib/datajs/proof.f.mjs: not linkable in a browser (node:querystring)
skipped fjs/bnf/lib/json/proof.f.mjs: not linkable in a browser (node:querystring)
skipped fjs/bnf/ll1/proof.f.mjs: not linkable in a browser (node:querystring)
skipped fjs/bnf/proof.f.mjs: not linkable in a browser (node:querystring)
browser proof modules: 141 of 147
```

The whole `fjs/bnf` proof tree leaves the browser suite, because each of those transitively imports this module. Both runs exit 0 and the node suite is green, so nothing in CI says a word — `noUnusedLocals` is commented out in `tsconfig.json`, and the skip is a report line, not a failure.

Two non-blocking notes:

- `fjs/bnf/map/rtti/module.f.mjs:45` — weakening `step instanceof Array && step.length === 2 && step[1] === rule` to drop the `step[1] === rule` self-reference test leaves all 3749 green. That check is the whole discriminator between a real `repeat` and any two-branch variant with an empty arm and a two-element sequence, which is the tag `checkMap` then asserts on at `:94`. The fixtures only ever feed a genuine `repeat0Plus`, so the lookalike case is unpinned.
- The four exported `*Info` contracts have no consumers yet: `Base.map` is `unknown`, so `AssertBase` cannot check a map shape, and the proofs type each fixture as `Base` with `map` cast. Only the runtime `ri` equality is enforced today. The changelog claims exactly that and no more, so this is a note rather than a complaint.

The types themselves are sound. I implemented all five of `fjs/bnf/map/types.ts` — `OneMap`, `TerminalMap`, `SequenceMap`, `VariantMap`, `RepeatMap` — under `strict` and `noUncheckedIndexedAccess` with no `any` and no cast, including the readonly-mapped-tuple rest parameter and key-discrimination in `VariantMap`, and confirmed with a wrong-output control that the probe actually reddens. The shadowed-type-parameter defect from the earlier `fjs/bnf/map/types.ts` does not recur; `b14fcbe4` was that same family and is fixed at head by `TupleTs<RI>`.

o2alexanderfedin (approved):
Approved. Measured against `3364f6e0`; the head moved twice during the review, but the later move is the `main` merge that brought in #1821 and the PR's own delta over the merge base is byte-identical file-for-file to `ecdb6a2d`, where I did the work.

Both my findings are fixed, and I checked the blocking one end to end rather than by reading the diff: no `node:` import remains in any `.f.mjs`, and `npm run website` gives `browser proof modules: 147 of 147` with no `skipped` lines — 147 rather than 146 because this PR adds a browser-linkable proof module of its own. The discriminator now carries an `isRepeat` cross-check against the data-level detector, and two independent mutations redden it.

One note, not a request: `repeatItem` calls `ruleToData(rule)` on every invocation, from inside the recursive walk, so the checker is roughly quadratic in the grammar. Fine for a dev-time check, but it is a new cost.

o2alexanderfedin (approved):
Approved. Measured against `5745b3ee`: suite 3766/0, `tsc` (7.0.2) clean, `npm run website` `147 of 147` with no skips. The `main` merge in this branch is inert for the PR's own material — its delta over the merge base is byte-identical file-for-file to `b18b20a6`. Leaf diff inside `fjs/bnf/map/rtti/` since `8799bf3b` is `lazyRepeat` added and nothing removed; the rest of the whole-suite diff is #1818 arriving.

Four commits since my approval at `3364f6e0`, and each holds under its own check.

`36a203e4`'s boundary rule is pinned in both directions rather than merely present: deleting it reddens three `throw.*` leaves, enforcing it only for mapped variants reddens one, only for unmapped variants the other two. Turning `ambiguousNonRepeats` into two throwing leaves is the right inversion — the old proof mapped a top rule and none of its branches, which is exactly the mixed boundary now forbidden.

`b18b20a6` replaces the hand-rolled repeat detector with `bnf/data`'s `repeatItem`. That is a strict simplification — the new code is the old code's second conjunct, so it can only widen, and the widening is the stated one: a lazy alias in the empty or recursive branch now qualifies where an inline `[]` or inline two-array was required. Identity survives the `map[item]` round-trip, and six leaves redden when it is broken.

`5745b3ee` is small and the best-checked thing here. `Assert<Equal<VariantValue<{0: string}>, ['0', string]>>` is a type-level assertion that can actually fail: dropping the `` `${K}` `` normalization makes `Equal` yield `false` and `tsc` rejects it with TS2344. I checked because a type-level check written the wrong way round is green whatever the type does; this one is not.

I re-probed the map types after each reshape rather than assuming the earlier check carried. At this head all six implement under `strict` and `noUncheckedIndexedAccess` with no `any` and no cast — including a numeric-key variant, where the tag now arrives as the string `'0'`, so the normalization is usable and not only declared.

Three notes, none a request:

- The four constructors cast to `any`. A `unique symbol` brand cannot be applied without *a* cast, so that part is unavoidable — but it need not be the widest one. `/** @type {MappedInfo<typeof info>} */ (info)` typechecks clean; I tried it.
- `changelog/unreleased/1823.md` describes only the `bnf/map` work, while `repeatItem` is a new export on `fjs/bnf/data/module.f.mjs`, a pre-existing shipped module. Nothing is owed — an addition is not a break — but the entry is 122 characters against a ~250 budget.
- `NoInfer` on the map positions weakens nothing, but I could not construct a case where it changes the verdict: `RO` is already pinned by `ro`, so both forms reject a wrong map. It is also unexercised here, since every `map` in `proof.f.mjs` is still `any`-cast — the guard is for future callers.

PR #1831, opened 2026-09-01
ci: enable nix-command and flakes from the generated run scripts

sergey-shandar:
`./nix/run` and its siblings failed with `experimental Nix feature
'nix-command' is disabled` on any Nix installation whose `nix.conf` does not
already enable `nix-command` and `flakes`. A stock installer leaves both off,
so the script whose whole purpose is to need no setup needed setup. CI never
saw it: `cachix/install-nix-action` enables both.

Both generated scripts now pass the features themselves, from one named binding
in `fjs/ci/nix/module.f.mjs` — the four `run` scripts and `nix/lock-update.sh`,
since `nix flake` is gated behind the same two features as `nix develop`.
`--extra-experimental-features` adds to the configured set rather than
replacing it, so an installation that already enables them — or more — is
unaffected.

Changelog:
- `ci`: the generated `run` scripts and `nix/lock-update.sh` pass
  `--extra-experimental-features 'nix-command flakes'`, so they work on a stock
  Nix install rather than only where `nix.conf` already enables both.

### DIFF
--- a/changelog/unreleased/1831.md
+++ b/changelog/unreleased/1831.md
@@ -1,3 +1,3 @@
-- `ci`: the generated `run` scripts pass `--extra-experimental-features
-  'nix-command flakes'`, so they work on a stock Nix install rather than only
-  where `nix.conf` already enables both.
+- `ci`: the generated `run` scripts and `nix/lock-update.sh` pass
+  `--extra-experimental-features 'nix-command flakes'`, so they work on a stock
+  Nix install rather than only where `nix.conf` already enables both.
--- a/changelog/unreleased/1831.md
+++ b/changelog/unreleased/1831.md
@@ -1,0 +1,3 @@
+- `ci`: the generated `run` scripts pass `--extra-experimental-features
+  'nix-command flakes'`, so they work on a stock Nix install rather than only
+  where `nix.conf` already enables both.
--- a/fjs/ci/README.md
+++ b/fjs/ci/README.md
@@ -135,11 +135,10 @@ constant, so the two cannot drift.
 
 A generated `run` script sits beside every flake, and a workflow step reads as
 the command it runs — `./nix/run npm run cov` — rather than as a `nix develop`
-invocation repeated fifteen times. The script carries `--no-update-lock-file`
-(error rather than silently resolve a mismatched input — a plain
-`--no-write-lock-file` would leave the checkout untouched but still resolve
-in memory and warn) and one `--quiet`, which drops the `copying N paths`
-substitution chatter and leaves every warning.
+invocation repeated once per step. Its flags live in that one generated place,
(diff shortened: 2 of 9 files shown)


o2alexanderfedin (commented):
Reviewed against `cfa0ba3d`; base is `main` and the merge base is `origin/main` at `79bd1210`.

The generated round trip holds — `node ./fjs/module.mjs ci` reproduces all four `run` scripts byte-identically and they keep mode `755` — and the single-argument form `--extra-experimental-features 'nix-command flakes'` is tidier than passing the flag twice.

**One call site was left behind: `nix/lock-update.sh`.** It runs four `nix flake lock` commands with no `--extra-experimental-features`, while the four `run` scripts and `dev.sh` all carry it at this head. `nix flake` is itself gated behind `nix-command` and `flakes`, so on exactly the machine this PR exists for — one without those features in `nix.conf` — `./nix/run` now works and `npm run lock-update` still does not. Both files come out of the same generator (`runText` and `lockUpdateText` in `fjs/ci/nix/module.f.mjs`), so this is one flag applied to one of the two outputs rather than a separate script someone forgot.

I could not test what the flags do: Nix is not installed here, so everything above is the generated text and the round trip, not runtime behaviour.

o2alexanderfedin (commented):
Sharpening my earlier note, and adding the gates I had not run when I posted it: suite 3804/0, `tsc` clean, `npm run cov` 100/100/100, and the change is pinned — narrowing the flag to `'nix-command'` reddens `nixFlakes.runText`, so the generated text cannot drift silently.

The `lock-update.sh` gap reads worse than "a call site was missed", because this PR writes the sentence it breaks. The new comment at `fjs/ci/nix/module.f.mjs:310` says the flags enable "the two experimental features **every command here** needs", and `lockUpdateText`, in the same file, emits four `nix flake lock` lines without them. `nix flake` is gated behind the same two features as `nix develop`, so the contributor the comment describes — stock Nix, both features off — reads the exact error it quotes, from `npm run lock-update`, right after `./nix/run` starts working for them.

Either interpolate `experimentalFeatures` into `lockUpdateText` as well, or narrow the comment to `nix develop`. One or the other; at the moment the file asserts something its own neighbour disproves.

Also worth correcting something I implied earlier: `dev.sh` already carries these flags at the merge base and is untouched here, so after this PR every `nix` caller in the tree has them except `lock-update.sh`. I had described `dev.sh` as still lacking them, which was stale.

o2alexanderfedin (approved):
Approved. Measured against `abc1a11f`.

The gap is closed: all four `nix flake lock` lines in `nix/lock-update.sh` now carry `--extra-experimental-features 'nix-command flakes'`, so the comment at `fjs/ci/nix/module.f.mjs` saying every command here needs them is true of every command the file generates rather than of three quarters of them. The generated round trip is byte-identical.

PR #1833, opened 2026-09-01
bnf: design the stateful parser and its metadata operations

sergey-shandar:
Documentation only — no code changes, so no changelog entry
([AGENTS.md §5](https://github.com/functionalscript/functionalscript/blob/main/AGENTS.md)).

Stacked on `meta3` (#1831) so the diff is this work alone; GitHub retargets to
`main` once that merges.

`43. Stateful parser` was a six-line stub predating the transformer and metadata
work, sketching an `append` that took whole strings. It is now the design: the
parser is a `StateFold` over one symbol at a time, with `MI` and `MO` as
separate metadata types.

### Decisions, each with its reason rather than only its outcome

- **`StateFold`, not `todo/flow.md`'s `Transducer`.** The latter is design-stage
  — no module under `fjs/` mentions it, against a dozen `StateFold`/`StateScan`
  users — and its `done` step is not worth a second fold shape ahead of the
  graph engine that would consume it.
- **Two metadata types**, because a layer transforms metadata. A tokenizer reads
  a number, emits a token symbol for the next layer and carries the value in its
  output metadata; one `M` makes that value ride uselessly through every later
  rule, or pushes its recovery into a postprocessing pass.
- **The machine has no error channel.** The same decision as `todo/flow.md`'s
  ("the core is total; every stage always completes") and 207 §6's, and for the
  reason `fjs/effects` gives for the *opposite* rule: an effect's `Result` exists
  because a runner may decline a command, and a pure fold dispatches to nobody.
  Refusal belongs to the driver, which is where DESIGN.md §10 lands — with the
  obligation that `O` must be able to say the input was not in the language,
  since otherwise `end` returns something plausible and wrong.
- **The parser stays RTTI-free.** A validatable root output is `checkMap`'s
  alone, keeping 207 §10's separation of the two mapping APIs intact.
- **`Monoid<M>` becomes `translate: (mi: MI) => MO` plus `reduce: Reduce<MO>`**,
  which is *not* required to be associative. So the fold's grouping is part of
  the contract, and `fjs/common/monoid`'s `fold` must not be reused here — it
  combines as a balanced binary tree, licensed by an associativity this
  operation does not promise, and would regroup silently.

### Left open

Where the answer is a judgment rather than a consequence:

- What a match with no children produces, now that the monoid identity is gone —
  an empty `Sequence`, a zero-round `Repeat`, and an EOF terminal were all
  spending it. Either add the identity back, or represent absence as
  `MO | undefined`.
- Whether `MI`/`MO` reaches the map and RTTI types that #1828 already unified.

### Also

`generic-parser-metadata` and `layered-parser` are updated where this
contradicts or answers them, so the four documents agree rather than disagreeing
two sections apart. `checkMap`'s rename gets an issue of its own: it returns a
complete map over every reachable rule, synthesizing identity entries for
undeclared ones, which the name hides.


o2alexanderfedin (commented):
Reviewed against `f5b6fbf9`. **This PR is stacked**: its base is `meta3`, PR #1831's branch, so everything below is measured against `abc1a11f` and not against `main` — diffing against `main` would have shown #1831's Nix material as this PR's. Six files, all `.md`, 325 insertions and no code, so no suite or type gate applies and no changelog entry is owed.

The design writing is unusually well grounded. I spot-checked the claims about shipped code rather than reading past them, and they hold — `Match<M>` and `Remainder<M>` in `ll1/types.ts`, `StateFold` in `types/function/operator`, `sha2`'s `init`/`append`/`end`, `checkMap`'s signature and its `?? ast` default, and the balanced-tree `fold` in `common/monoid` that the re-grouping argument rests on. `Transducer` really does appear in no module under `fjs/`, verified against a positive control.

**One thing to fix.** `layered-parser.md` moves *Meta info propagation* out of Open Questions and states the rule as settled — "a repetition folds its rounds, and an **empty match takes the monoid identity**". But `generic-parser-metadata.md`, edited in this same commit, says to read every "uses the monoid identity" as 43's open question, because a semigroup has no identity and the empty sequence, zero-round repetition and EOF terminal are exactly where it would be spent. And `043-stateful-parser.md`'s first task is "Settle what a match with no children produces", marked as the one thing here that is not mechanical.

So the sub-rule this PR designates as its central unresolved question is published as a settled answer in the file a reader of the layered design would consult, and it is published by removing the bullet from the open list — which is what takes the constraint with it. Pointing at 43's open question instead of asserting the identity is a sentence.

Smaller, same bullet: "A layer's payload is its `M`" sits awkwardly beside this PR's own thesis that a layer is `MI → MO`.

o2alexanderfedin (approved):
Approved. Measured against `e8c1f609`; the base moved from `meta3` to `main` since my last review because #1831 landed, so this is now an ordinary main-based delta.

My point is closed — `layered-parser.md` no longer asserts that an empty match takes the monoid identity, so it stops publishing as settled the one thing 43 names as its open question.

The part I looked hardest at is the deletion of `hostile-proof-values.md`, since a retired todo carrying a live constraint away with it is a failure mode this repo has hit before. It does not happen here: the cross-realm promise item survives in `imports-promises-realms.md` and `share-browser-console-runner.md`, no link anywhere still points at the deleted path, and the `TestResult` JSDoc that used to defer to that file now states the answer itself — a wire report carries `message` and `stack`, a terminal line carries the value's own text, both through the core's `text`. That is the rationale moving into the code it describes rather than evaporating.

`emergent_testing/types.ts` is JSDoc only; no signature moves, so nothing is owed a breaking-change entry.

PR #1836, opened 2026-09-02
emergent_testing: prove the browser runner's own rejection

sergey-shandar:
`runBrowserProofs`' `.catch(runnerFailure)` had no proof. It is the branch that
decides whether a page whose *interpreter* fails shows an `infrastructure-error`
or hangs in `running` forever — the failure mode
[`todo/share-browser-console-runner.md`](https://github.com/functionalscript/functionalscript/blob/main/fjs/emergent_testing/todo/share-browser-console-runner.md)
catalogs as item 11.

No caller could reach it. `runProofs` has no error channel, so every failure a
*proof* can produce arrives as a value it decides about; the only thing that can
reject the run is one of the page's own handlers, and a caller cannot replace
those.

## The seam

`_runBrowserProofsWith` passes the page's operation map through a function
before it is interpreted:

```js
export const _runBrowserProofsWith = operations => (modules, result = () => undefined) => { … }
export const runBrowserProofs = _runBrowserProofsWith(map => map)
```

The published entry point is unchanged — a testing seam, not the API widening
the plan rejected, which would have made every page carry the cost of one proof.

## Two proofs, for two depths

Only one of them reaches a handler, which is why they are two cases and not one:

| proof | what rejects |
|-|-|
| `aThrowingHandlerIsTheRunnersOwnFailure` | a `sandbox` handler throws |
| `anUndispatchableCommandIsTheSameFailure` | no `sandbox` handler at all — `match` panics inside the same awaited loop |

Both assert what the page actually depends on: the report still *resolves*, it
says `infrastructure-error`, and its one row is named after the runner rather
than after a module — the walk is one effect now, so a rejection is not
attributable to the module it happened under.

**Mutation-checked.** Deleting `.catch(runnerFailure)` fails exactly these two
and nothing else; they reject instead of reporting, which is the page-stuck-in-
`running` failure itself.

## Plan status

The runner-failure task is ticked, and the 7b paragraph that described the seam
as delivered now says when it actually arrived. One task is left in the plan:
that both runners answer identically from the same fixtures.

## Verification

`npx tsc` clean · `npm run cov` 100% lines/branches/functions, 0 failures ·
`node --test` 3746 pass / 0 fail · `npm run gen` no diff.

---
_Generated by [Claude Code](https://claude.ai/code/session_016PyLwDNkPQM1uD1Tg5ApBg)_

### DIFF
--- a/changelog/unreleased/1836.md
+++ b/changelog/unreleased/1836.md
@@ -1,0 +1,4 @@
+- `emergent_testing`: a browser run whose own interpreter rejects resolves its
+  published report with `infrastructure-error` — proved, where before the
+  branch existed but nothing exercised it and a regression would have left the
+  page in `running` forever
--- a/fjs/emergent_testing/browser/module.mjs
+++ b/fjs/emergent_testing/browser/module.mjs
@@ -21,7 +21,7 @@
  * } from '../types.ts'
  * @import { Catch, Import, Sandbox, SandboxResult } from '../../effects/common/types.ts'
  * @import { IoChannel } from '../../effects/node/types.ts'
- * @import { Effect, Func } from '../../effects/types.ts'
+ * @import { Effect, Func, ToAsyncOperationMap } from '../../effects/types.ts'
  * @import { Result } from '../../types/result/types.ts'
  * @import { List } from '../../types/list/types.ts'
  */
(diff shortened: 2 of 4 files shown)


o2alexanderfedin (approved):
Approved. Measured against `4e252e0b`; base is `main`. Suite 3806/0 → 3808/0 with two leaves added and none removed, `tsc` clean, coverage 100/100/100, `npm run website` `148 of 148` with no skips.

The proof earns its title. Deleting `.catch(runnerFailure)` reddens exactly the two new leaves, so the branch that existed but nothing exercised is now held. And the seam I care about most is held too: dropping the failed row while keeping the `infrastructure-error` status — reported but not counted — also goes red, against `assertStructurallySame` on `{ tests: 1, passed: 0, failed: 1 }`. A run whose totals disagree with its results is the one shape that tells an automated consumer the suite was smaller rather than broken, and this is the third PR in a row to pin it deliberately.

No published break: `runBrowserProofs` keeps its type and delegates to a new underscore-prefixed export, which is addition rather than movement, so the changelog owes no breaking line.

PR #1837, opened 2026-09-02
emergent_testing: prove the two runners answer the same fixtures

sergey-shandar:
The sharing plan's last property, and the only one the per-runner proofs cannot
state.

`browser/proof.mjs` asserts the page against `fmtImport` and `testResult`;
`catch.proof.mjs` asserts the console traversal against literals. Each is right
about its own side, and neither says that **one suite produces one answer** — a
runner that calls a shared function differently agrees with that function and
still disagrees with the other runner.

## What it does

`equivalence.proof.mjs` runs one fixture through both runners and compares the
transcripts of name, status and totals as a single string:

| fixture entry | what it pins |
|-|-|
| `passes` | a plain leaf |
| `nested` returning `{ child }` | the count is recursive, not one-per-export |
| `fails` | a normalized failure |
| `'a.b'` | a path segment that needs quoting |
| `throw: { threw, didNot }` | throwing is the pass, returning is the failure |

The browser side is `runBrowserProofs`; the console side is `runModuleMap` on a
mock runner whose `sandbox` and `catch` are a real `try`, for the reason
`catch.proof.mjs` gives.

**Comparing the whole transcript is the point.** A disagreement about *which*
leaves ran — a walk that stops early, a returned tree one runner recurses into
and the other does not — fails it too, not only a disagreement about a status.
A second proof pins the transcript itself, because two runners that both walked
nothing would agree.

**Mutation-checked.** Give the page's `collectTests` a root `throws` of `true` —
a divergence in one runner only — and `nameMatchesTheConsoleRunner`,
`expectedThrowStatusMatchesTheSharedOne` and `normalizedResultMatchesTheSharedOne`
all still pass while this fails.

## The plan is finished, and the file is not deleted yet

Every task in
[`todo/share-browser-console-runner.md`](https://github.com/functionalscript/functionalscript/blob/main/fjs/emergent_testing/todo/share-browser-console-runner.md)
is now done, so `todo/README.md` says the issue goes. It is retained here under
that rule's own exception rather than as a tombstone, and the status names the
documents that require it: its pitfall catalog is cited **by item number** from
`browser/module.mjs`, `browser-testing.md`, `65z-tf-test-tree-walker.md`,
`all-argument-limit.md`, `generate-website.md`, `DESIGN.md` and others — 36
references across 20 files — and neither the catalog nor the boundary argument
exists anywhere else yet.

The last unchecked task is that retirement: move the catalog to
`emergent_testing/README.md` and the code-adjacent parts to JSDoc, repoint the
citations, then delete. It is its own change — the move is larger than any step
the file records.

## Verification

`npx tsc` clean · `npm run cov` 100% lines/branches/functions ·
`node --test` 3748 pass / 0 fail · `node ./fjs/module.mjs t` 3810 pass / 0 fail ·
`npm run gen` no diff.

---
_Generated by [Claude Code](https://claude.ai/code/session_016PyLwDNkPQM1uD1Tg5ApBg)_

### DIFF
--- a/changelog/unreleased/1837.md
+++ b/changelog/unreleased/1837.md
@@ -1,0 +1,4 @@
+- `emergent_testing`: the browser runner and `fjs t` are proved to answer
+  identically from the same fixtures — one transcript of names, statuses and
+  totals, compared as a whole, so a disagreement about which leaves ran fails
+  it as loudly as a disagreement about a status
--- a/fjs/emergent_testing/equivalence.proof.mjs
+++ b/fjs/emergent_testing/equivalence.proof.mjs
@@ -1,0 +1,151 @@
+/**
+ * **The two runners answer the same thing about the same fixtures.**
+ *
+ * The last property the sharing plan asks for, and the only one the individual
+ * proofs cannot state: `browser/proof.mjs` asserts the page against `fmtImport`
+ * and `testResult`, and `catch.proof.mjs` asserts the console traversal against
+ * literals, so each is right about its own side. What neither can say is that
+ * *one* suite produces *one* answer — which is the property a reader comparing
+ * a CI log with a browser page depends on.
+ *
+ * So this file runs one fixture through both and compares the transcripts as
+ * text. The comparison is deliberately narrow: name, status, and the run's
+ * totals. `message` and `stack` are the page's own part and have no console
+ * counterpart, and a duration is a measurement rather than an answer.
+ *
+ * `.mjs` rather than `.f.mjs`, for the reason `catch.proof.mjs` gives: the
+ * console side needs a runner that can really `catch`, and the browser side is
+ * a host module.
+ *
+ * @import { RunInstance } from '../effects/mock/types.ts'
+ * @import { Write } from '../effects/node/types.ts'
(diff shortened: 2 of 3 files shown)


o2alexanderfedin (approved):
Approved. Measured against `6234adfc`; base is `main`. Suite 3808/0 → 3810/0, two leaves added and none removed, `tsc` clean, coverage 100, `148 of 148` browser modules.

The thing worth checking in a proof like this is whether it can be vacuous — comparing one runner against a literal, or against itself. It is not: both sides are live, `runBrowserProofs` on one and `runModuleMap` on the other, and a mutation to the browser half alone reddens both equivalence leaves. It compares the leaf names, their statuses and the totals as one joined transcript, so set and order are held rather than a bare count; the fields it leaves out are named in the file's own header with reasons.

Worth stating plainly, since it bounds what the green means: after step 7b the traversal is genuinely shared, so what this proof can detect is divergence in the host wrappers and report assembly, not two independent implementations agreeing. The PR says as much. And the console side drives a purpose-built reporter rather than `fjs t`'s, so it pins the traversal's answers, not the CI log format.

`.proof.mjs` rather than `.f.mjs` is the right placement — it has to call a host module and needs a real `catch`.

PR #1838, opened 2026-09-02
emergent_testing: retire the runner-sharing issue into the README

sergey-shandar:
Every task in `todo/share-browser-console-runner.md` is done, so `todo/README.md`
says the issue goes. What kept it alive through the last two PRs is that things
in it exist nowhere else. They move to `emergent_testing/README.md`, where they
are documentation of the runners rather than the record of a finished plan.

## What moved

- **The pitfall catalog** — thirteen measured ways this was got wrong, cited *by
  item number* from `module.f.mjs`, `types.ts`, `browser/module.f.mjs`,
  `browser/proof.mjs`, `sequential.proof.mjs` and five other issues. The
  numbering is kept, because the citations use it.
- **The boundary argument** — why the browser runner is `.f.mjs` behind a thin
  host, and what asking `instanceof Promise` in business logic cost.
- **Rules the shared core keeps** — no Node built-ins, Node interpreter,
  `node:test` or Playwright in browser modules; the host runner stays native
  JavaScript; terminal formatting and DOM presentation stay out of the shared
  core; the core never asks which host it is on; every remaining difference
  lives in a part with a written reason; a fix lands in the core or in every
  part at once; both runners name a leaf identically.
- **Open questions about the report shape** — whether `path` survives now that
  `name` exists, whether a report declares the root its module keys are relative
  to, and whether a module-level failure belongs in a variant of its own. Three
  questions that want settling together, with the `fjs t`-in-a-subdirectory
  example that makes the second one concrete.

The last two are there because [a review caught that I had
dropped them](https://github.com/functionalscript/functionalscript/pull/1838#pullrequestreview-5084808877):
I moved what was *interesting* and treated the rest of the file as a record of
finished work, when the `### Constraints` section was mostly guidance about work
that has not happened. `types.ts` still said the report-shape question "is open"
while this branch removed its citation, so it pointed nowhere; it points at the
questions again.

## The references

Thirty-six across twenty files, all repointed. Eleven stopped being *true* when
the plan finished, and are rewritten rather than relinked — a link fix that left
these standing would have been the worse outcome:

| was | now |
|-|-|
| "step 4's shared module resolves to…" | the row states what `effects/common` holds; there is no step 4 to resolve |
| "after the sequential traversal from … lands" | it has landed; the task says so and keeps only the constraint |
| "follow up by sharing its pure semantics with `fjs t`" | they are shared |
| `generate-website.md`'s unchecked "move browser-manifest preparation" | checked — functionalscript#1827 landed it and deleted `browser-prepare.mjs` |

## Verification

`npx tsc` clean · `node --test` 3748 pass / 0 fail · `npm run gen` no diff ·
every relative link in every changed markdown file resolves to a file that
exists.

Docs only; the code changes are JSDoc citations.


o2alexanderfedin (commented):
Reviewed against `abf94c5a`; base is `main`. Suite 3810/0 either side with an empty leaf diff both ways, `tsc` clean, and all six touched `.mjs`/`.ts` files are comment-only — every changed line is a ` * ` line — so nothing is owed a breaking-change entry. The retirement itself is done properly in the parts that are easy to get wrong: all 36 inbound references across 20 files are updated, the three anchors resolve to real headings, every measured number survives into the catalog, and the only unchecked task in the deleted file was "retire this file", which this PR does.

**One thing did not survive.** The deleted `### Constraints` section carried an open design question: whether `path` survives now that `name` exists, whether a report declares the root its module keys are relative to, and whether a module-level failure belongs in a variant of its own — stated as three questions that want settling together. At `main` that text lives in exactly one file; at this head it is nowhere. I checked with a positive control, since an absent phrase and a broken grep look identical: 24 files still match "open questions", so the search was live.

What makes it worse than an ordinary omission is that this PR also edits `fjs/emergent_testing/types.ts:80`, dropping "see `todo/share-browser-console-runner.md`" so the sentence now ends "…is open." The type still tells a reader the question is open and no longer tells them where to read it.

Two prohibitions go the same way, with the same control: that browser modules must not import Node built-ins, the Node effect interpreter, `node:test` or Playwright, and that terminal formatting and DOM presentation must not move into the shared semantic core. Both are rules about what future work may not do, which is the kind of thing a README is the right home for — the new section just does not carry them.

`todo/README.md` asks for design decisions to be captured before a file is deleted, and this is the case it has in mind.

sergey-shandar:
You are right on all three, and the shape of the mistake is worth naming: I moved what was *interesting* — the catalog and the boundary argument — and treated the rest of the file as a record of finished work. The `### Constraints` section was not that. Most of it is guidance about work that has not happened yet, which is exactly what survives a retirement.

Fixed in `a1a75dc`, as two new README sections.

**Open questions about the report shape** carries all three, as three numbered questions with the note that they want settling together — including the module-key/root explanation, because the question is not obvious without the `fjs t` in `fjs/types/list` example that makes it concrete.

**Rules the shared core keeps**, opening with "these are constraints on future work, not history": no Node built-ins, Node interpreter, `node:test` or Playwright in browser modules and the host runner stays native JavaScript; terminal formatting and DOM presentation stay out of the shared core; the core never asks which host it is on; every remaining difference lives in a part with a written reason; a fix lands in the core or in every part at once; both runners name a leaf identically, which `equivalence.proof.mjs` holds. I carried the others of that kind rather than only the two you named — the same reasoning applies to each, and stopping at the two you happened to catch would have been the same error one size smaller.

And you are right that `types.ts:80` was the worse half: this branch dropped its citation, so the type told a reader the question was open and no longer where to read it. It points at the new section.

Two constraints I did *not* move, with reasons rather than silently:

- The `then`-named-proof rule is already in this README, under "Convention: only real Promises are awaited", stated more fully than the constraint stated it.
- The website's "filesystem access through `NodeProgram`, no npm script as a second entry point" rule is a `fjs/website` rule, and `website/todo/directory-index-pages.md` — the only file that cited the constraint — now states it inline where it used the citation, rather than pointing at `emergent_testing`.

Thank you for the positive control on the grep. "24 files still match 'open questions', so the search was live" is the step I did not take, and it is the difference between finding this and concluding the text was already gone.

`tsc` clean, `node --test` 3748/0, `gen` no diff.

---

sergey-shandar:
Correction to the comment above: the commit is `0c556642`, not `a1a75dc` — I wrote the sha before pushing.

---

o2alexanderfedin (approved):
Approved. Measured against `0c556642`; the delta over `abf94c5a` is `README.md` and one JSDoc line, no code.

Everything I flagged is back, and back in the right place. The three report-shape questions — `path` beside `name`, whether a report declares the root its module keys are relative to, and whether a module-level failure wants a variant of its own — are in the README, and both prohibitions with them: no Node built-ins, effect interpreter, `node:test` or Playwright in browser modules, and no terminal formatting or DOM presentation in the shared core, with the no-bundling requirement alongside. I checked for the substance rather than the wording, since the phrasing changed.

The `types.ts` pointer is restored too, and now points at the README rather than the deleted file — so the sentence that says the question is open again tells a reader where to read it.

PR #1841, opened 2026-09-02
emergent_testing: the browser runner's purity debt is paid, so the record goes

sergey-shandar:
`todo/browser-runner-functional-script.md` was the migration-debt record for
`browser/module.mjs` holding business logic. It asked for three things; the
module half is done and its record is retired, and the proof-file half is not,
so that keeps a file of its own.

## The measure, and the result

The issue's own measure is **a host touch per definition, not a definition per
file**. Measured across `browser/module.mjs`:

| | |
|-|-|
| definitions | 14 |
| touching a host object | **13** |
| not | `runBrowserProofs` — a partial application of `_runBrowserProofsWith` |

The touches are a clock, `navigator`, `import()`, a DOM node, `setTimeout`, or
the interpreter loop. Nothing left in the file is logic that could be pure: the
pure pieces it still calls — `runProofs`, `reportOf`, `concat`, `toArray` — are
imported from `.f.mjs`. That debt is paid.

**The line count grew while the debt shrank** — 298 to 486 — which is why the
count was never the measure. What grew is glue: the testing seam
(functionalscript#1836), the pending row and the start event
(functionalscript#1840).

## What outlives the retired file

Four things, moved to the README's *Why the browser runner is `.f.mjs` with a
thin host*:

- **The question that moved each definition** — not "does this touch a host
  object" but "*which values* does it need from one". `reportOf` needed two, so
  they are passed in. The loading walk needed a module loader, so that became an
  operation. It appeared to need a fan-out; it needed sequencing, which an
  effect already is.
- **Pure logic that reads a *user* value is not impure, it is effectful** — the
  rule that let `errorDetails` and `text` move as effects over `catch`.
- **The measure itself**, with where it now stands, and what to do with a
  definition that fails it.
- **`browser/proof.mjs`'s size is the standing measurement of how thick the glue
  is.**

Nothing in the repository referenced the deleted file.

## What is still owed keeps a file

The retired record's task 1 had two halves, and only the first is done. The
second — *shrink `browser/proof.mjs` to the DOM adapter's own proofs* — is not,
with that file at 741 lines against the 628 the record named. Reframing it in
the README as a measurement is fair, but it turns an obligation into an
observation, so `todo/browser-proof-file-size.md` carries the obligation,
measured rather than asserted:

| proofs in `browser/proof.mjs` | |
|-|-|
| need the DOM stand-in | 20 |
| only drive `runBrowserProofs` and read its report | **23** |
| total | 43 |

The 23 are the debt. The issue also records what is **not** owed: the module
itself, and the "50–80 lines" target — retired explicitly, because its own
baseline of "298 today" was already 485 before this PR.

## Verification

`npx tsc` clean · `node --test` 3751 pass / 0 fail · `npm run gen` no diff ·
every relative link resolves. Docs only — `browser/*` is byte-identical to the
base, so no changelog entry is required (`AGENTS.md` §5 exempts documentation-
only PRs).


o2alexanderfedin (commented):
Measured against `c3e6ec06` over merge base `cee82a25`. No code changed — `browser/*` is byte-identical to the base — and the gates are green (3813/0, `website` `148 of 148` with nothing skipped, `tsc` 7.0.2 clean). The capture itself is good: I traced the deleted todo's constraints and the substantive ones have live homes, the "13 of 14 definitions touch a host" measurement verifies at HEAD, and nothing in the tree ever pointed at the deleted file (checked against a positive control, not an empty grep).

Two things worth fixing before merge.

**The changelog entry is not required and is over the cap.** `AGENTS.md`: "a PR that doesn't — internal refactors, test-only changes, and PRs that only touch `todo/`, `AGENTS.md`, or other documentation — needs neither". This PR touches only a todo and a README. The entry is also 309 characters in one bullet against the ~250-per-entry cap.

**The todo closes with all three tasks still `[ ]`, and one of them unmet.** Task 1's second half — "shrink `browser/proof.mjs` to the DOM adapter's own proofs" — closes with that file at 741 lines, up from the 628 the todo recorded. The README's answer is to reframe it as "the standing measurement of how thick the glue is", which is a fair measure, but it turns an obligation into an observation without saying so. Same for the "plausible 50–80 lines" target: retiring it is defensible — its own baseline of "298 today" was already 485 before this PR — but the retirement is nowhere on the record, and that is the thing a future reader will look for.

sergey-shandar:
Both right. Fixed in `fbdd3340`.

**The changelog.** `AGENTS.md` exempts "PRs that only touch `todo/`, `AGENTS.md`, or other documentation", which is exactly this one. I added the entry out of habit from the last four PRs without re-reading the rule. Deleted — and the cap point is right too, it was one 309-character bullet.

**The unmet task is the one that matters, and your framing of it is the correction.** Task 1 had two halves. The first — extract the logic — is done. The second — "shrink `browser/proof.mjs` to the DOM adapter's own proofs" — is not, and I reframed it in the README as "the standing measurement of how thick the glue is" without saying that I had swapped an obligation for an observation. The measure is fair; making the swap invisible is the defect.

So the obligation keeps a file: `todo/browser-proof-file-size.md`, rescoped to what is actually left and **measured rather than asserted**. Of the 43 proofs in `browser/proof.mjs`, **20 need the DOM stand-in and 23 only drive `runBrowserProofs` and read its report** — naming, throw inversion, normalized failures, the yields, the runner's rejection routes, the hostile-value family. Those 23 are the debt: a proof that reads a report rather than a document is proving `.f.mjs` logic through a host entry point. The tasks are to decide per proof whether it can run against `runProofs` under `effects/mock`, move the ones that can, and leave the ones that cannot with the host capability that keeps them there — `crossRealmPromiseSilentlyPasses` needs `node:vm` no matter what it proves.

It also records what is **not** owed, since that was the useful half of the retirement: `browser/module.mjs` itself, whose measure passes, and the line-count target — retired explicitly, with the reason that its own "298 today" baseline was already 485 before this PR because the *glue* grew.

`tsc` clean, `node --test` 3751/0, `gen` no diff, links resolve.

---

o2alexanderfedin (approved):
Approved. Delta measured against my previous marker `c3e6ec06`. Both earlier points are answered: the changelog entry is gone (correct — `AGENTS.md` exempts a PR touching only `todo/` and documentation), and the surviving half of the debt is now on the record rather than dropped.

The new todo does the harder half properly. "What is *not* owed" names the 50–80-line target as retired and says why — its own baseline of "298 today" was already 485 — and states the replacement measure instead of leaving the substitution implicit. I checked its numbers rather than taking them: `proof.mjs` is 741 lines and the largest impure proof file (next is 379), and the 20 / 23 / 43 split reproduces against the file's own stated criterion. Gates green: 3813/0, `website` `148 of 148` with nothing skipped, `tsc` 7.0.2 clean.

One number in that file has no arithmetic behind it, unlike every other one: "nearly half of all the impure proof code there is" is 741 of 1779 across the six impure proof files, or 41.7%. Worth either the figure or a looser word — stale counts are this repository's most persistent documentation failure, and an unbacked one is where it starts.

PR #1849, opened 2026-09-02
docs: move DESIGN.md, REVIEW.md and CONTRIBUTORS.md to doc/

sergey-shandar:
Split out of [#1844](https://github.com/functionalscript/functionalscript/pull/1844) at the owner's request, and rebuilt from current `main` rather than cherry-picked — see "Why rebuilt" below.

## What moves

`README.md`, `LICENSE`, `CONTRIBUTING.md` and `AGENTS.md` stay in the root: they are the entry points a reader or a tool looks for at the top level. `DESIGN.md`, `REVIEW.md` and `CONTRIBUTORS.md` are documents those entry points link to, so they move to `doc/`, matching the repository's other terse top-level directory names (`fjs`, `nix`, `spec`, `todo`, `changelog`).

## How the references were repointed

Across 27 files, in three passes rather than one search and replace, because three different things needed three different rules:

- **The moved files' own outgoing links** were written from the root and now sit one level deeper, so they are re-based (`./AGENTS.md` → `../AGENTS.md`). The two links *between* `DESIGN.md` and `REVIEW.md` are unchanged — both moved together.
- **Inbound markdown links** are re-based against each referring file's own directory, not rewritten by pattern. The same relative path means different things from `fjs/bnf/todo/` and from `changelog/`.
- **Bare textual citations**, including five in `.mjs` JSDoc and comments, name the new path (`DESIGN.md §6` → `doc/DESIGN.md §6`). Markdown link *labels* are deliberately left alone, so `[DESIGN.md §10](../../doc/DESIGN.md#...)` still reads correctly.

`CONTRIBUTORS.md` had no inbound references at all. Nothing reads any of the three programmatically — none is in the package's `files` list, and no script or workflow names them.

## Verification

- No reference to the old root paths survives, and no bare citation names one.
- Every relative link in the repository resolves, except four pre-existing pseudo-links in `nanvm-lib/src/vm/README.md` and `spec/todo/2330-property-accessor.md` that this change does not touch.
- `tsc` clean; `node --test` passes, 3754 tests, 0 failures. The five source-file edits are inside comments and JSDoc prose.

## Why rebuilt rather than cherry-picked

In #1844 this concern had spread across three commits: the move itself, a merge from `main`, and a later commit where the `fjs/website/module.f.mjs` repoint rode along with unrelated changelog fixes. That last one existed because [#1842](https://github.com/functionalscript/functionalscript/pull/1842) added a `DESIGN.md` reference *after* that branch was cut, so the fix had to land wherever the branch happened to be.

Building fresh from `main` makes that a non-issue: #1842's reference is simply one of the five source citations the sweep handles, with no follow-up commit. The result is one commit that does one thing.

## Relationship to #1844

#1844 keeps the changelog-at-release-time work and drops this concern. The two touch overlapping lines in a handful of files — anything citing `DESIGN.md` — so whichever merges second will need its references re-based against the other's paths. That is the cost of the split, and it is small: the conflicts are path strings, not logic.

No `Changelog:` section: this changes documentation only, no behavior and no public API — the three moved files were never in the package's `files` list.

https://claude.ai/code/session_014b31HRoyn7A18YENXR18G7

---
_Generated by [Claude Code](https://claude.ai/code/session_014b31HRoyn7A18YENXR18G7)_


o2alexanderfedin (approved):
Approved. Measured against `a7929618` over merge base `485f6fb2`.

The reference that the earlier attempt at this move could not see is handled here: `fjs/website/module.f.mjs:159` now points at `../../doc/DESIGN.md#10-refuse-what-you-cannot-handle`, and `doc/DESIGN.md:312` is `## 10. Refuse what you cannot handle`, so the anchor lands. Splitting the move out onto a base that already contains #1842 is what made that possible. All 33 link targets resolve with their anchors, the 24 outbound links inside the moved files were re-depthed correctly, and the five `.mjs` edits are provably comment-only — stripping whole-line comments from both versions leaves zero diff in each. Nothing in `package.json`, `deno.json`, `Cargo.toml`, `.github/**` or the CI generator names these files, so no workflow follows. Gates: 3816/0, `website` `148 of 148` with nothing skipped, `cov` 100% across all 141 rows, `tsc` clean, generated files round-trip.

Worth considering rather than fixing: `doc/` is the only content directory in the tree without a `README.md` — `changelog/`, `fjs/`, `nanvm-lib/`, `nix/`, `spec/` and `todo/` all have one. `CONTRIBUTORS.md` has no inbound reference anywhere, which it also had none of before the move, so that is not a regression; a short `doc/README.md` would settle both at once.

o2alexanderfedin (approved):
Approved again, at `db6a14b1`. The only change since my approval of `a7929618` is #1848 arriving through `main` — `fjs/effects/node/virtual/proof.f.mjs` gaining the two nested-read fixtures, which I reviewed and approved on that PR. This one's own files are byte-identical, and the move is intact: `doc/` holds the three documents and none remains at the root.

PR #1854, opened 2026-09-02
operator: unify polymorphic addition

sasha-gil:
Centralize addition across number, bigint, and string using a shared same-kind binary-operator contract.

This refactor removes duplicated a => b => a + b implementations while preserving the existing public API. It also preserves numeric inference for higher-order consumers, adds direct proof coverage for all three operand types and the exported bigint.addition, moves the Addition import into the module JSDoc block, and removes the completed polymorphic-addition TODO.

Validation
- `tsgo --noEmit`
- `node module.mjs t` (3819 passed)
- `node --test` (3757 passed)
- `git diff --check`


sergey-shandar:
@codex review

fjs/types/bigint/module.f.mjs
```diff
@@ -22,16 +22,16 @@
...
 
 /**
  * Adds two `bigint` values.
- * TODO: should be combined with `addition` for `number`.
  *
  * @param {bigint} a - The first bigint value.
  * @returns {Unary} A function that takes the second bigint value and returns the sum.
  */
-export const addition = a => b => a + b
+export const addition = /** @type {(a: bigint) => (b: bigint) => bigint} */ (a => additionOp(a))
```
sergey-shandar: `/** @type {T} */(a)` - it's the same as `a as T`. Don't use it. Use type declaration as `/** @type {T} */const a = ...`, which is the same as `const a: T = ...`. 

```suggestion
/** @type {(a: bigint) => (b: bigint) => bigint} */
export const addition = a => additionOp(a)
```

sergey-shandar: Or
```js
/** @type {Reduce<bigint>} */
export const addition = additionOp
```

sergey-shandar:
@codex review

sergey-shandar (approved):

PR #1858, opened 2026-09-02
Add shared operator test cases for `<` (less than)

sasha-gil:
## Summary

- Adds a `lessThanCases` corpus (50 cases) for `<` to the shared operator test data in `fjs/nanvm/module.f.mjs`, following the same pattern established for `%` (see `nanvm-lib/tests/README.md`).
- `nanvm-lib` does not implement `<` yet (see the operator table in `nanvm-lib/README.md`) — `BigInt` has `PartialOrd`, but nothing dispatches it from `Any` — so every case carries a `rust` reason: the generated Rust keeps each as a commented-out TODO, while the JavaScript proof runs all of them against a real JS engine.
- Coverage:
  - `ToPrimitive` coercion of non-numeric operands (never throws, unlike arithmetic).
  - Lexicographic (UTF-16 code unit), not numeric, string-vs-string comparison — e.g. `'10' < '9'` is `true`.
  - `NaN` comparisons are `false` in *both* directions.
  - Mixed `Number`/`BigInt` and `String`/`BigInt` comparisons, which the abstract relational comparison algorithm allows directly — unlike `*`, `-`, `+`, `/`, `%`, `**`, which all throw `TypeError` on mixed number/bigint operands.
- Wires `<` into `fjs/nanvm/proof.f.mjs` (`op2Js`) and `fjs/nanvm/rust/module.f.mjs` (`rustName`, `op2Rust`), and regenerates `nanvm-lib/tests/test/generated.rs`.
- `<` prints as `Any::lt(a, b)` rather than Rust's own `<` operator: `check` expects a `Result<Any<A>, Any<A>>` against an `Any<A>` expectation, which a `PartialOrd`-derived comparison on `Any<A>` would not give back — following the existing `Any::unary_plus` precedent for an operation with no fitting Rust operator to spell.

## Test plan

- [x] `npm test` — 3869/3869 passing, including all 50 new `<` cases evaluated against Node.
- [x] `cargo test -p nanvm-lib` — generated Rust compiles with the new cases commented out.
- [x] `cargo fmt -p nanvm-lib -- --check` — clean.
- [x] `npm run gen` — no drift after regeneration.

https://claude.ai/code/session_01EtpceTFc1653SDSG7Rsggt

---
_Generated by [Claude Code](https://claude.ai/code/session_01EtpceTFc1653SDSG7Rsggt)_


o2alexanderfedin (commented):
The cases themselves check out — I verified the expected values against real JS, and a mutation inside the new group goes red and back green, so they are load-bearing rather than decorative.

One thing to do before merge: **rebase onto `main` and regenerate.** This branch forked at `3461e0ee`, 26 commits back. Since then `eb959f14` changed the Rust printer's `emit` from two lines (`// TODO: reason` then `// statement`) to one (`// TODO: reason: statement`), so the `nanvm-lib/tests/test/generated.rs` block committed here is old-form output and roughly halves when regenerated. GitHub already reports the branch as conflicting, so the rebase is unavoidable anyway — the point is that resolving the conflict by hand is not enough; `node ./fjs/module.mjs r ./fjs/nanvm/update/module.f.mjs` has to run afterwards. CI catches it at `.github/workflows/ci.yml:445-448` (`npm run gen` then `git diff --cached --exit-code`), not in the test suite.

o2alexanderfedin (commented):
One more thing for the same rebase, at the other conflict site. `fjs/nanvm/module.f.mjs:39` documents `data.groups.length // 6`; the tree here actually evaluates to **7** (checked by importing the module, not by counting). The off-by-one is inherited, not introduced — at the fork point `3461e0ee` the comment said `5` against an actual `6`, and this branch faithfully bumped it by one. `main` is currently right at `7`, because `a400f715` happened to land on the correct number. So when you resolve line 39, the answer is neither this branch's `6` nor `main's` `7`: with your group added it should read **8**.

sasha-gil:
Superseded by #1864, which carries this PR's full `lessThanCases` corpus plus the wiring into `proof.f.mjs`, `rust/module.f.mjs`, and `rust/proof.f.mjs` — and additionally implements `<` in `nanvm-lib` itself, so those cases run as real assertions rather than commented-out `TODO`s. Closing this one in favor of that branch.

---

PR #1860, opened 2026-09-02
Add shared operator test cases for `>` (greater than)

sasha-gil:
## Summary

- Adds a `greaterThanCases` corpus (50 cases) for `>` to the shared operator test data in `fjs/nanvm/module.f.mjs`, following the same pattern established for `%` (see `nanvm-lib/tests/README.md`) and mirroring [the `<` PR](https://github.com/functionalscript/functionalscript/pull/1858).
- `nanvm-lib` does not implement `>` yet (see the operator table in `nanvm-lib/README.md`) — `BigInt` has `PartialOrd`, but nothing dispatches it from `Any` — so every case carries a `rust` reason: the generated Rust keeps each as a commented-out TODO, while the JavaScript proof runs all of them against a real JS engine.
- `x > y` is defined as `y < x`, so every case here mirrors a `<` case with its operands swapped, keeping the same boolean: `ToPrimitive` coercion, lexicographic string comparison, `NaN` staying `false` in both directions, and mixed number/bigint and string/bigint comparisons allowed without throwing.
- Wires `>` into `fjs/nanvm/proof.f.mjs` (`op2Js`) and `fjs/nanvm/rust/module.f.mjs` (`rustName`, `op2Rust`), and regenerates `nanvm-lib/tests/test/generated.rs`.
- `>` prints as `Any::gt(a, b)`, following the same `Any::unary_plus` precedent used for `<` and `<=`.

## Test plan

- [x] `npm test` — 3869/3869 passing, including all 50 new `>` cases evaluated against Node.
- [x] `cargo test -p nanvm-lib` — generated Rust compiles with the new cases commented out.
- [x] `cargo fmt -p nanvm-lib -- --check` — clean.
- [x] `npm run gen` — no drift after regeneration.

https://claude.ai/code/session_01EtpceTFc1653SDSG7Rsggt

---
_Generated by [Claude Code](https://claude.ai/code/session_01EtpceTFc1653SDSG7Rsggt)_


o2alexanderfedin (commented):
All 50 expected values match real JS, there is no disagreement with #1858 on the shared pairs, and a mutation inside the group goes red and back green.

**The `greaterThanCases` doc comment is wrong about the family it singles out.** `fjs/nanvm/module.f.mjs:421-428` says every case is a `<` case "with its operands swapped and its same boolean kept — including the coercion family". For that family neither half holds: `nullGreaterThanFive` is `args: [null, 5]`, the same order as `nullLessThanFive`, and its expected value is `false` where the `<` case is `true`. Operands kept, boolean recomputed — the exact opposite of the stated rule. The values are right; the sentence describing them is not. (Worth noting too that the genuinely swapped cases, `5 > null` and `1 > undefined`, are absent — they are the ones the rule as written would have produced.)

**For the rebase.** This branch forked at `3461e0ee`, 26 commits back, and GitHub reports it conflicting. Regenerate `nanvm-lib/tests/test/generated.rs` rather than hand-merging it: `eb959f14` changed the Rust printer's `emit` from two lines to one (`// TODO: reason: statement`), so the committed block is old-form output and halves when regenerated. Run `node ./fjs/module.mjs r ./fjs/nanvm/update/module.f.mjs` after resolving; CI catches a stale file at `.github/workflows/ci.yml:445-448`, not the test suite.

At the other conflict site, `fjs/nanvm/module.f.mjs:39` documents `data.groups.length // 6` while the tree evaluates to **7**. The off-by-one is inherited — the fork point said `5` against an actual `6`. With your group added, line 39 should read **8**, not this branch's `6` nor `main`'s `7`.

sasha-gil:
Superseded by #1864, which carries this PR's full `greaterThanCases` corpus plus the wiring into `proof.f.mjs`, `rust/module.f.mjs`, and `rust/proof.f.mjs` — and additionally implements `>` in `nanvm-lib` itself, so those cases run as real assertions rather than commented-out `TODO`s. Closing this one in favor of that branch.

---

PR #1868, opened 2026-09-03
nanvm-lib: implement typeof

sasha-gil:
## Summary

Stage 2 of the remaining `nanvm-lib` operators (see the "JS Operator Implementation Status" table in `nanvm-lib/README.md`): `typeof`.

**Based on #1867** (Stage 1: `!`, `&&`, `||`, `??`, `?:`) rather than `main`, since `typeof` reuses the `NonEdagGroup` extension `?:` introduced there (the EDAG has no `typeof` node either, so this adds a second variant alongside `'unaryPlus'` and `'ternary'`). This PR's diff is Stage 2 only — the base will need retargeting to `main` once #1867 merges.

- **`typeof`**: `Any::type_of()` ([`any/type_of.rs`](nanvm-lib/src/vm/any/type_of.rs)) — a plain method, since Rust has no unary operator to spell it with and `typeof` is itself a reserved word. A small `Dispatch` impl (`TypeOf`) maps each `Unpacked` variant to its tag: `"undefined"`, `"object"` (both `null` and the two container types), `"boolean"`, `"number"`, `"bigint"`, `"string"`, `"function"`.

Unlike Stage 1's value-selecting operators, `typeof` always returns a fresh string rather than an operand unchanged, so there's no operand/`expected` identity concern to work around in the corpus this time.

Corpus cases in `fjs/nanvm/module.f.mjs`, wired into `fjs/nanvm/proof.f.mjs` (`op1Js`) and `fjs/nanvm/rust/module.f.mjs` (`op1Rust`, `rustName` — `'type_of'` rather than `'typeof'` since the latter is reserved in Rust). `nanvm-lib/tests/test/generated.rs` regenerated via `npm run gen`, no drift beyond the new cases.

README operator table updated.

### Not in this PR

Stage 3 (bitwise `&`/`|`/`^`/`~` and shift `<<`/`>>`/`>>>`) is separate, tracked in the README table.

## Test plan

- [x] `node ./fjs/module.mjs t ./fjs/nanvm/module.f.mjs` (full `fjs test` suite, 4242 pass / 0 fail)
- [x] `node --test` (full Node suite, 4180 pass / 0 fail)
- [x] `npm run gen` — `nanvm-lib/tests/test/generated.rs` regenerated, diff limited to the new cases
- [x] `cargo test` (in `nanvm-lib`)
- [x] `cargo clippy --lib -- -D warnings`
- [x] `cargo fmt -- --check`
- [x] `tsc`

https://claude.ai/code/session_016r91tf9sadPY8kgPXwfRu9

---
_Generated by [Claude Code](https://claude.ai/code/session_016r91tf9sadPY8kgPXwfRu9)_

### DIFF
--- a/nanvm-lib/src/vm/any/type_of.rs
+++ b/nanvm-lib/src/vm/any/type_of.rs
@@ -1,0 +1,57 @@
+use crate::vm::{
+    Any, Array, BigInt, Function, IVm, Object, String, dispatch::Dispatch, nullish::Nullish,
+};
+
+/// The tag `Any::type_of` returns, one per `Unpacked` variant.
+struct TypeOf;
+
+impl<A: IVm> Dispatch<A> for TypeOf {
+    type Result = &'static str;
+
+    fn nullish(self, v: Nullish) -> Self::Result {
+        match v {
+            // `null` is famously an object in JavaScript's `typeof`.
+            Nullish::Null => "object",
+            Nullish::Undefined => "undefined",
+        }
+    }
+
+    fn bool(self, _: bool) -> Self::Result {
+        "boolean"
+    }
+
+    fn number(self, _: f64) -> Self::Result {
+        "number"
+    }
+
+    fn string(self, _: String<A>) -> Self::Result {
+        "string"
+    }
+
+    fn bigint(self, _: BigInt<A>) -> Self::Result {
+        "bigint"
+    }
+
+    fn object(self, _: Object<A>) -> Self::Result {
+        "object"
+    }
+
+    fn array(self, _: Array<A>) -> Self::Result {
+        "object"
+    }
+
+    fn function(self, _: Function<A>) -> Self::Result {
+        "function"
+    }
+}
+
+impl<A: IVm> Any<A> {
+    /// `typeof`. Not a `core::ops` trait — Rust has no unary operator to
+    /// spell it with (and `typeof` itself is a reserved word) — so this is a
+    /// plain method, the same as `unary_plus`. Never throws, but stays a
+    /// `Result` to match every other operator's shape.
+    /// <https://tc39.es/ecma262/#sec-typeof-operator>
+    pub fn type_of(self) -> Result<Any<A>, Any<A>> {
+        Ok(self.dispatch(TypeOf).into())
+    }
+}
--- a/fjs/edag/todo/typeof-operator.md
+++ b/fjs/edag/todo/typeof-operator.md
@@ -1,0 +1,81 @@
+## `typeof` has no EDAG node — and isn't in the operators spec at all yet
+
+**Priority:** P2
+**Status:** open
+
+### Problem
+
+`nanvm-lib`'s `typeof` (`Any::typeof_`, added in
+functionalscript/functionalscript#1868) has no EDAG counterpart, for a more
+basic reason than `?:`'s
+([`ternary-conditional-node.md`](./ternary-conditional-node.md)): unlike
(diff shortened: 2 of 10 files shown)


o2alexanderfedin (approved):
Stacked on #1867 — delta measured against that branch's head `4523fccb66dc5b07cb7c6ef20b4f8224f635ce8f`, not `main`. The only commit of this PR's own is `f453b389`.

Approved. `typeof null` is `"object"` in both the implementation (`type_of.rs:14`, with the wart named in a comment) and the assertions (`module.f.mjs:1062`, `generated.rs:692`), and flipping it to `"null"` fails as an assertion rather than a compile error — so it is genuinely tested, not merely written down.

One word: `fjs/nanvm/module.f.mjs:1053` calls `typeof` "the corpus's **other** group with no canonical EDAG id". `NonEdagGroup` now has three arms — `unaryPlus`, `ternary`, `typeof`.

sasha-gil:
Thanks! Fixed in 13a4c82 — "another of the corpus's groups" (three now: `unaryPlus`, `ternary`, `typeof`).

---

nanvm-lib/src/vm/any/type_of.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+impl<A: IVm> Any<A> {
+    /// `typeof`. Not a `core::ops` trait — Rust has no unary operator to
+    /// spell it with (and `typeof` itself is a reserved word) — so this is a
+    /// plain method, the same as `unary_plus`. Never throws, but stays a
+    /// `Result` to match every other operator's shape.
+    /// <https://tc39.es/ecma262/#sec-typeof-operator>
+    pub fn type_of(self) -> Result<Any<A>, Any<A>> {
```
sergey-shandar: To be consistent with FJS and JS, I would call it `typeof_`. 

`_` at the end is a consistent suffix to resolve name collisions.

sergey-shandar:
> **Based on #1867** (Stage 1: `!`, `&&`, `||`, `??`, `?:`) rather than `main`, since `typeof` reuses the `NonEdagGroup` extension `?:` introduced there (the EDAG has no `typeof` node either, so this adds a second variant alongside `'unaryPlus'` and `'ternary'`). This PR's diff is Stage 2 only — the base will need retargeting to `main` once #1867 merges.

Add a TODO to implement `typeof` in EDAG: `fjs/edag/todo/`. Priority P2.

sergey-shandar (changes requested):

nanvm-lib/src/vm/any/type_of.rs
sasha-gil: Done in f9e2882 — renamed the method and the file (`any/type_of.rs` → `any/typeof_.rs`), and regenerated `generated.rs`.

---

sasha-gil:
Filed in 3f70b00: [`fjs/edag/todo/typeof-operator.md`](https://github.com/functionalscript/functionalscript/blob/claude/nanvm-lib-typeof-op/fjs/edag/todo/typeof-operator.md) (P2). One thing worth flagging while filing this: unlike `?:` (which already has a decided EDAG design, just not yet implemented — see `fjs/edag/todo/ternary-conditional-node.md`), `typeof` doesn't appear in `spec/todo/2340-operators.md`'s operator table at all — not even as `not allowed` the way `==`/`!=` are. So the todo's first task is actually deciding whether `typeof` is admitted as FJS syntax, not just wiring the node.

---

sergey-shandar (approved):

PR #1870, opened 2026-09-03
Add ebnf-migration plan as alternative to grammar-bucket

sergey-shandar:
## Summary

This PR introduces a new migration plan document (`fjs/todo/ebnf-migration.md`) that proposes an alternative approach to the existing `grammar-bucket` plan for building the EBNF module alongside the classical BNF module.

## Key Changes

- **New document**: `fjs/todo/ebnf-migration.md` (268 lines)
  - Proposes building `fjs/ebnf/` as a new module with one-way dependency from `bnf` to `ebnf`
  - Introduces a triage-based approach (move/rewrite/retire) rather than the staged refactoring in grammar-bucket
  - Defines 7 implementation stages with clear deliverables and task checklist
  - Includes detailed module layout, triage tables for code organization, and consumer port strategy
  - Addresses the djs tokenizer and parser grammar conflicts that require LL(1) resolution

- **Updated document**: `fjs/todo/grammar-bucket.md`
  - Added "Related" section linking to the new ebnf-migration plan
  - Notes that only one of the two plans will be followed

## Notable Details

- The plan emphasizes **triage over blind copying**: every piece of `bnf/` is assigned to move (relocate), rewrite (new implementation), or retire (delete with `bnf/`)
- Establishes a mechanical boundary check to prevent `ebnf/` from importing `bnf/` at any level
- Proposes a comparison proof strategy using `descentEquivalence` to validate both front ends against the same grammar and AST
- Identifies specific LL(1) conflicts in djs grammars that must be resolved during the port
- Provides a clear 7-stage implementation roadmap with dependencies and testing requirements

https://claude.ai/code/session_01NLd1QuFocrAaEb8d8R9feG


o2alexanderfedin (commented):
Reviewing the draft. The good news first: dropping the arrow table takes the three arrow errors with it, and the coupling inventory at `:48-63` is the post-correction text — every claim in it matches the imports. All 22 bnf issues are accounted for in the triage table, and no dangling link to the deleted `grammar-bucket.md` survives; the remaining mentions are prose calling it retired.

**The reader-list error recurs in a new place.** `ebnf-migration.md:231-232` gives `fjs/djs/tokenizer` as depending on `terminal/`, `unicode/`, `data/`, `ll1/`, and the parser as "the above plus `token_symbol/`". Both omit `matcher/`, which all four djs files import: `tokenizer/module.f.mjs:6`, `tokenizer/private.ts:7`, `parser/module.f.mjs:11-12`, `parser/private.ts:7`. This is the same shape of mistake the old plan's reader sentence had, in a list that drives the porting order.

**One constraint did not survive the replacement, and the new stage order breaks on it.** `grammar-bucket` put `unicode/` at stage 2 *because* "`token_symbol` uses `unicodeRange`, so until this exists it cannot be pointed away from the front end". Here `token_symbol/` moves at stage 3 (`:295`) while `ebnf/unicode/` is built at stage 4 (`:299`) — and `:210` states the dependency outright, "`token_symbol/` … imports `unicode/`". As numbered, stage 3 leaves `ebnf/ → bnf/`, which principle 2 forbids and the stage-0 CI check is meant to catch. The rewritten `unicode-rules.md` writes "(stages 3 and 4)" without resolving it.

Two smaller ones:

- `ebnf-migration.md:241` — the tokenizer conflict row does not follow from the plan's own method. Building the dispatch map of each rule's closure, the smallest failing rules are `multilineContent`/`85` (12-rule closure), not the punctuator variant `90` (81). The parser row is right, and the 91/167 counts check out.
- `fjs/bnf/todo/unicode-rules.md` — the proposal now says the adapter serves EBNF only and `bnf` keeps its helpers, but the task list still strips `fjs/bnf/module.f.mjs` and `bnf/data` and ports `bnf/lib` and `djs` "here as well".

On the framing added in `48367c0b`: garbage collection with consumers as roots fits the post-fork drift, but stages 1–4 move the machinery ahead of any consumer and `:218` deletes each `bnf/` path in the same PR, so moved code never faces the walk. The roots there are the triage table, not the consumers.

sergey-shandar:
Addressed in c599d97, by removing restrictions rather than adding steps.

**The common cause behind the Codex "stage 1 breaks djs" thread, the descent-adapter thread, and the GC-roots remark:** the plan still said each move deletes the `fjs/bnf/…` path in the same PR. That rule came from grammar-bucket's one-hop constraint and has no reason to exist here, so it is gone. Nothing under `fjs/bnf/` is deleted or moved away before stage 7: a move copies into `ebnf/`, `bnf/` keeps its copy, exports and consumers (`fjs/djs` included) untouched until each consumer's own port, and repointing a `bnf/` module at `ebnf/` is allowed under the direction rule but never required. Stage 7 is the only breaking change. The GC paragraph now says plainly that what `ebnf/` starts from is the triage's choice and the walk governs everything after.

**Alexander's points:**
- `matcher/` added to the djs dependency list; all four djs files import it.
- Stage order: `unicode/` now lands in stage 3 with `matcher/`, and `token_symbol/` in stage 4 after it. Stages 1–4 are also stated to be numbered for reference only, ordered by the dependencies each module names, so nothing else is forced into sequence.
- Tokenizer conflicts: you are right. `multilineContent` and rule 85 are mutually recursive, so my "all children clean" filter skipped both. Re-checked: the punctuator variant's closure does not contain `multilineContent` and still fails with that pair neutralized, so both are independent conflicts. The table now has both rows with closure sizes (12 and 81), and the `*/`-vs-`*` two-symbol-lookahead nature of the first.
- `unicode-rules.md` is now EBNF-only end to end: no task strips `fjs/bnf/module.f.mjs` or `bnf/data`, and the lib and djs grammars adopt the adapter at stages 5 and 6, not there.

**Other Codex threads:** `todo/parser-serializer-restructure.md` no longer points at `fjs/grammar/unicode/`; `proof-recognizer-and-fixtures.md` ships `ll1Recognizer` in the `ebnf/` harness and keeps the descent comparison as a `bnf`-local adapter over the shared `Recognizer` type and `assertRecognizes`, which the `bnf → ebnf` direction permits.

---

sergey-shandar:
Addressed in 80c8f8e (on top of the merge from main), again by removing prescriptions rather than adding them.

All three new findings asked the plan to fix an order or a method. The plan now says the opposite explicitly: **the stages are numbered for reference, not for order, and prescribe no method.** The only constraints are the module dependencies each piece names (`token_symbol/` needs `unicode/`, `ll1/` needs `data/` and `matcher/`, a ported grammar needs what it imports, deleting `bnf/` needs no consumer left on it) and the `ebnf` never imports `bnf` rule. Any order, overlap, split or merge of stages that respects those is the developer's call, and so is whether a "move" is a copy, a re-export or a rewrite.

- **"Settle the EBNF representation before stage 1":** not made a prerequisite. `ebnf-front-end`'s open problems are answered by whoever needs the answer, when they need it, recorded in `ebnf/README.md` or the issue, and revisable when the code shows an answer wrong. The issue's task line now gives the dependency map among the problems (8 decides the AST tables; 4 and 7 follow it; 1, 3 and 6 shape the lowering) as a map, not an order. Since #1865 has merged, `BoundedArray` is dropped as a blocker.
- **"Preserve the public data type":** `terminal-range-shared-type` now requires nothing of `bnf/`. Its two declarations stay, or either becomes an alias of `ebnf/terminal/`'s when someone prefers one declaration to two. Both keep every public path exporting what it exports today, so nothing breaks before `bnf/` is deleted.
- **"Wait for stage 4 before the LL(1) harness":** the issue is unblocked once `fjs/ebnf/ll1/` exists, whichever stage that turns out to be, and it may be written earlier inside `bnf/` against `bnf/ll1` and moved later if that is more convenient.

---

o2alexanderfedin (commented):
All three of my findings are addressed. The djs dependency list now names `matcher/` — imports re-checked at `tokenizer/module.f.mjs:6`, `tokenizer/private.ts:7`, `parser/module.f.mjs:11-12`, `parser/private.ts:7`. `unicode/` moved to stage 3 and `token_symbol/` to stage 4, so the `ebnf/ → bnf/` edge is gone whatever the ordering. And `unicode-rules.md`'s tasks are additive now, so it no longer contradicts its own proposal.

**The renumbering did not reach the citations.** `ebnf/unicode/` is created at stage 3 (`ebnf-migration.md:326-328`), and the triage table agrees — `:280` says "until stage 3 implements it". Five places still say 4, or count from the old order:

- `ebnf-migration.md:384` — "unicode-rules — stage 4", contradicting `:280` in the same file.
- `ebnf-migration.md:383` — the reference backend grows "at stages 3 and 6", but `ll1/` is now stage 4.
- `fjs/bnf/todo/unicode-rules.md` contradicts itself: `:10` "This issue is that plan's stage 3" and `:56` "stage 3", against `:179` "stage 4".
- `fjs/bnf/todo/ebnf-front-end.md:14` — "stage 4 there" for the text terminals.
- `fjs/bnf/todo/bnf-grammar-single-owner.md:71` — "that plan's stage 4 creates `fjs/ebnf/unicode/module.f.mjs`"; also `:97` "fixed from stage 4", and `:81` "At stage 2 … `unicode/` already exists at its final path", which is now stage 3.

This is the kind of thing that reads as settled until someone follows a number, so worth a sweep rather than fixing the two I happened to open.

The new numeric claims all hold — I re-ran `toData` and `dispatchMap`: 167 tokenizer rules, 91 parser rules, `multilineContent` closure 12, punctuator closure 81, and the two closures are disjoint.

sergey-shandar:
Addressed in 20fb47d (on top of the second merge from main).

**Alexander's sweep, and the Codex "label as stage 3" thread — fixed at the root rather than one number at a time.** Every issue outside the epic now cites the piece of work by name (`ebnf/terminal/`, "the `unicode/` piece", "the consumer port") and never by stage number, and the epic says that is the rule, so a renumbering there cannot strand a citation again. Inside the epic the two stale numbers are corrected (`unicode/` is stage 3, `ll1/` is stage 4). Thanks for re-running the closure counts.

**Boundary check:** you are right that a text scan is exactly what AGENTS.md §6 forbids, and no parser-backed tool for this exists in the repo. So the plan no longer asks for one. The direction rule is written down in `fjs/AGENTS.md` and held by review, and its mechanical check is stage 7 itself: deleting `bnf/` fails `tsc` and the suite on any `ebnf → bnf` import, JSDoc `@import` included. A parser-backed lint may be proposed by anyone who wants the check earlier; it is an option, not a task.

**Open front-end problems:** the absorb row now carries every problem still open when the issue is absorbed, 4 and 5 included, rather than an enumerated subset.

**rule-visitor:** rebased on the EBNF data layer. The visitor lives in `fjs/ebnf/data/`, gains a `repeat` case for the bounded repeat, is blocked only on that module existing, and has no alphabet-split blocker since the EBNF lowering expands strings before the data form exists. `bnf/` is untouched and may adopt the visitor if someone wants one discriminator there.

---

sergey-shandar:
Addressed in fa2fc75.

**Byte adapter:** `ebnf/byte/` is now in the layout beside `unicode/`, with the same owner as the rest of unicode-rules. It lands whenever its first consumer wants it (the recognizer backend, per that issue); nothing in the plan needs it earlier and nothing forbids it earlier. The stage-3 task says which half of unicode-rules it settles and which half is owed to that consumer.

**DJS descent exports:** checked the importers. `jsMatcher` and `descentParserCpOnly` are used only by the tokenizer's own proof, so the port replaces them with LL(1) equivalents and updates the proof in the same PR. The plan now says that is a change to `fjs/djs/tokenizer`'s own public API, declared by the port, and scopes the "one breaking change" claim to `fjs/bnf/` paths. `proof-recognizer-and-fixtures` now leaves that export to the djs port rather than saying to leave it alone.

---

sergey-shandar:
Addressed in a45e5dd.

**Rule-to-name bridge:** correct, the transformer protocol keys on rule identity through `Entry.rule` and needs a way to reach the generated `RuleSet` name, which is what `toDataWithRules` supplies today. The triage now retires only the *classical* `GrammarData` and `RuleNameMap`, and adds a rewrite row: the EBNF lowering returns its own identity-to-name map beside the rule set and entry, in whatever shape the `data/` rewrite chooses. Stage 2 says the same, citing ebnf-front-end's "rule identity must survive" requirement.

---

sergey-shandar:
@codex review

sergey-shandar:
Addressed in 3d8196f.

**fsc rename:** the consumer inventory now says the djs consumers are named by today's paths, that `parser-serializer-restructure` renames them to `fjs/fsc/` with the BNF dependency intact, and that if that lands first the port simply follows them there. No ordering between the two plans is introduced; the port targets whatever path the consumer has when it happens.

**detectRepeat:** one contract in both plans now. It retires with the classical `data/`: an EBNF set, hand-written or deserialized, spells the primitive, and an opt-in normalizer of the right-recursive shape may be added to `ebnf/data/` by whoever wants one, but nothing plans it. `ebnf-front-end.md` says the same instead of keeping it.

---

sergey-shandar:
Addressed in 53d21ae, by relaxing rather than prescribing.

**Recognizer pipeline:** the diagram and the "author it, `toData` it, compile" line are now front-end-agnostic. A grammar is authored in a front end and lowered by that front end's lowering — the EBNF one once `fjs/ebnf/data/` exists, classical `toData` for as long as `bnf/` lives and someone finds it convenient — and the builders never see which; both produce a valid rule set.

**Classical `string` case:** the paragraph that directed removing `string` from `fjs/bnf/types.ts` and its conversion from `bnf/data` is rewritten for the EBNF representation: a string is one terminal per code point there, applying that is the one text interpretation the EBNF core carries, and `fjs/bnf/` is not touched. The combinator paragraph is relaxed the same way: the EBNF front end may give `commaJoin0Plus` a symbol-taking core form with a Unicode wrapper, or keep the string spelling, since a string is a legal EBNF rule.

**Inbound links at deletion:** stage 7 and its task now say that each moved issue takes its inbound links with it, and that the deletion PR repoints every other inbound reference from outside `fjs/bnf/` — about two dozen files today, the two you named among them — at `fjs/ebnf/` or at whatever replaced the target, by search, since neither `tsc` nor the suite reads Markdown.

---

sergey-shandar:
Addressed in c1e23ec, by leaving the decisions to whoever makes them.

**Bounded repeats in the recognizer builders:** the issue no longer pins the input union to `Variant | Sequence | TerminalRange`. It now says the IR carries alternation, concatenation, terminal ranges and a bounded repetition, that the exact union is whatever `fjs/ebnf/data/` defines when the builders are written, and that how a builder compiles a bounded `Repeat` (unrolled to a bound, a counter in the state, or something else) is that builder's design decision, taken then, not here.

**Classical grammars and the adapter:** `bnf-grammar-single-owner` no longer says the split removes the classical helpers and `string` case or breaks the grammars. Nothing is removed from classical BNF before `bnf/` is deleted, so nothing breaks in place; the grammar adopts `fjs/ebnf/unicode` when it is ported to `fjs/ebnf/lib`, which is the migration's consumer port and not this issue's. The "before the port" import block that showed a classical grammar importing the EBNF adapter is gone, with a sentence saying why that is not an option (the adapter returns EBNF forms) and why it is not needed (the classical helpers stay).

---

PR #1878, opened 2026-09-04
ebnf: rule constructors over the generic range set

sergey-shandar:
On `main`, now that #1874 has merged. Five new files, nothing else touched.

## What this adds

A first `fjs/ebnf` layer: the rule vocabulary an EBNF grammar is written in, with terminals carried as `range_set` values.

- `fjs/ebnf/types.ts` — the rule algebra. A `DataRule` is a number, a string, a `Tuple` (sequence) or a `Variant` (named alternatives); a `Rule` is a `DataRule` or a `Thunk`. A thunk is a tagged tuple behind a function, so a grammar can name a rule before it is defined: `['const', …]`, `['set', …boundaries]`, `['repeat', min, max, rule]`.
- `fjs/ebnf/module.f.mjs` — the constructors, split the way [ebnf-range-set](https://github.com/functionalscript/functionalscript/blob/main/fjs/bnf/todo/ebnf-range-set.md) asks for.
- `fjs/ebnf/proof.f.mjs` — proofs for every constructor, its rejections, and the `const` type parameters.
- `fjs/ebnf/lib/json/module.f.mjs` and `fjs/ebnf/lib/json/proof.f.mjs` — JSON's lexical rules as the first real user: string with its escape variants, the `uint`/float-suffix number shape, and whitespace, each read down to its boundaries by the proof. The composite `value`/`array`/`object` rules stay commented out until the sequence combinators they need land.

## A set is a value; `oneOf` makes it a rule

The range-set algebra is `types/range_set`'s, over values, and there is no second copy of it here. `range`, `rangeEncode` and `set` return a `RangeSet`; a grammar composes them with `union` and `difference`; `oneOf` is the single injection from a value to a rule, and so the one place a set is checked for being a terminal's. That is what the JSON string body reads like at the call site:

```js
c: oneOf(difference(range(` ${unicodeMax}`))(set('"\\')))
```

A raw set is never a rule: as an element of a sequence it would read as its boundaries, two symbols in a row.

## What is refused

Every constructor that builds a value out of its arguments checks what it is about to hand back, rather than only the shape of the input:

- **Ordinary symbols are the non-negative safe integers.** `rangeEncode` refuses a negative endpoint rather than building a set the lowering's intersection with the domain would silently clip to a plausible but different one; `-1` is EOF's, a rule of its own and no member of any set. Above the top the successor is the number itself, so `Infinity`, `Number.MAX_VALUE` and a fraction are out too.
- **The top symbol is an open tail.** `2 ** 53` is not a safe integer, so no set can carry it as a boundary: `rangeEncode(b, b)` at `Number.MAX_SAFE_INTEGER` is `[b]`, its only spelling, not `[b, 2 ** 53]`.
- **`oneOf` refuses a set no terminal can be made of** — the empty one (a rule that can never match is a grammar error, though the empty *value* is fine, being what `union` folds from), one reaching below zero, one carrying an unsafe boundary, and a list that is not a canonical set at all.
- **`repeat` requires a count below, a count or `Infinity` above, and no decrease between them.** A negative, fractional or reversed cardinality is no repetition a parser can carry out. `option` is `repeat(0, 1)` rather than a second spelling of the same tuple.

## Notes

- Symbols are counted as code points, not UTF-16 units — `range` reads its bounds off `stringToCodePointList`, which is what makes an astral bound like `0x10FFFF` work and what makes a lone astral symbol a one-symbol (rejected) argument.
- Terminal sets are half-open above (`'0'..'9'` is `[48, 58]`). The `+ 1` lives here, at the layer that knows a symbol is an integer, and never in `types/range_set`, where boundaries are only ever compared.
- The `Assert<Equal<…>>` typedefs in the proof each pass a *variant literal*: a lone string literal is inferred the same with or without `const`, so an assertion written that way would be inert. Dropping the `const` on `repeat`, `times` or `option` fails `tsc` and nothing else in the suite — checked by actually removing each one.

## Testing

- `npx tsc` — clean.
- `node ./fjs/module.mjs t` — 4499 passing, 0 failing, including 45 new `fjs/ebnf` proofs.
- `npm run cov` — 100% lines/branches/functions on both new modules, and the tree-wide gate passes.
- `npm run gen` — no drift.

https://claude.ai/code/session_01WpcwphKdV63F2AxoPKGbmv


o2alexanderfedin (commented):
Stacked on #1874 — delta measured against that branch's head `4c0fd03a47d8b51f52332a22e90487538955e3e4`. All five files in the delta are this PR's own; nothing from #1874 is in it.

**The two front ends define the same lexical language.** Since no runner exists for the EBNF algebra yet, we built a throwaway recognizer over `range_set`'s own `contains` and compared every `fjs/ebnf/lib/json` export against its `fjs/bnf/lib/json` counterpart under `descentParser` — about 18,800 inputs, exhaustive to length 3 per rule alphabet plus the RFC edges (`\u00AG`, `\uD800`, `\u041`, `\u{41}`, raw control characters, `01`, `1e+`, unterminated, astral). Zero disagreements, with controls verified in both directions. Both new `module.f.mjs` files reach 100/100/100, both mutations go red as assertion failures, and `browser proof modules: 149 of 149`.

Three things.

**`union` and `remove` are rule combinators, not value operations.** `fjs/ebnf/module.f.mjs:89-91,97-101` — both take `SetInfo` and return `SetInfo`, unwrapping with `getSet` and rewrapping in a thunk. Both design documents specify these as value operations over `RangeSet`, with exactly one injection from a set to a rule; that is the answer recorded against `ebnf-front-end`'s Problem 5. As written, the set algebra lives at the rule level again, which is the shape the design set out to leave behind. Relatedly `remove` builds `intersection(a)(complement(b))` where `fjs/types/range_set` already exports `difference` — same output, one fewer round trip.

**`repeat1Plus` is in the design's constructor list and is not shipped.** `fjs/ebnf/lib/json/module.f.mjs:44` spells `digits = [digit, digits0]` — precisely the `[T, Repeat0Plus<T>]` shape the design's table says `repeat1Plus` exists to flatten. The exports are `repeat`, `repeat0Plus`, `times` and `option`.

**A style break, easy to fix while it is one line.** `fjs/ebnf/lib/json/module.f.mjs:14` is the only semicolon-terminated import in any `.f.mjs` in the tree, and it uses double quotes where everything else uses single. Its named-import list is unsorted too.

o2alexanderfedin (commented):
Now targets `main` directly, since #1874 merged. Nothing under `fjs/ebnf/` changed since my last review — the delta is the merge plus nine `.md` files — so the language-equivalence result from before still stands on identical bytes. Gates green, both new `module.f.mjs` files at 100/100/100, both mutations red.

**Two new ones, and both are the function disagreeing with its own header comment.** `rangeEncode` asserts only that `a` and `b` are safe integers with `a <= b`:

```
rangeEncode(-5, -1)                    ->  ['set', -5, 0]
rangeEncode(0, Number.MAX_SAFE_INTEGER) ->  ['set', 0, 9007199254740992]
```

The first is what `ebnf-range-set.md:235-239` says must be rejected — "an endpoint below `0` is a mistake at the call site, and it must fail there rather than reach the lowering, whose intersection with the domain would clip `[-1, 2]` to the plausible but different `[0, 2]` without a word". That is exactly the silent-wrong-answer the design set out to prevent, and it is reachable today.

The second emits `9007199254740992` as a boundary, which is not a safe integer. `:232-234` says `2 ** 53` "is not safe and is rejected" and the open tail `[a]` is the spelling for `b === MAX_SAFE_INTEGER`. The header comment at `:46-50` claims this case is refused; the assert checks `a` and `b`, never `b + 1`.

**The three from last round are still open.** `union`/`remove` remain `SetInfo → SetInfo` (`:88-102`) while `ebnf-range-set.md:150-160` still mandates `oneOf` as the single injection over set values and `:256` still records that as the answer to Problem 5 — the spec did not move under the code. `remove` still rebuilds `intersection ∘ complement` where `range_set` exports `difference`. `repeat1Plus` is still absent while `ebnf-front-end.md:167,228` list it and `lib/json/module.f.mjs:44` still spells `digits = [digit, digits0]`.

On the style point I raised: the semicolon at `lib/json/module.f.mjs:14` is genuinely unique — 1 of 1299 `import` lines across 284 `.f.mjs` files — and the named list is still unsorted. **I withdraw the double-quote half of that claim**: 28 double-quoted import lines exist tree-wide, including this PR's own `fjs/ebnf/module.f.mjs:19-23`. That part was wrong.

PR #1882, opened 2026-09-04
ebnf: an EBNF front end over a range-set algebra

sergey-shandar:
`fjs/ebnf` is a new front end for writing grammars: a small set of rule
constructors over the existing `types/range_set` algebra, plus the JSON
grammar written with them as the first real consumer.

A rule is data — a number, a string, a tuple, or a variant — or a thunk that
returns one. The thunk is what lets a grammar name itself, so a JSON value can
contain JSON values without the definition chasing its own tail, and it is why
`types.ts` spells the recursive shapes as `Info<...>` tuples rather than as
mutually recursive object types.

A terminal is a range set rather than a single symbol. `range('09')` takes the
two bounds as one two-symbol string, so the pair cannot drift apart at a call
site, and it reads the bounds off the code point list rather than off
`.length` — an astral symbol is one symbol here, not the two UTF-16 units that
spell it. `set`, `union`, and `remove` are then the algebra's operations lifted
to that encoding, which is what makes a JSON string's body expressible as one
terminal: everything from a space up, minus the quote and the backslash.

`repeat` carries its bounds and its rule verbatim; `option`, `times`,
`repeatFrom` and `join` are that one constructor with the bounds fixed.

`fjs/ebnf/lib/json` is the grammar itself. There is no recognizer for these
rules yet, so both proofs are structural: they force a rule into plain data by
calling every thunk it meets and compare that against hand-written data.
`json` names itself, so forcing it would not terminate — its proof expands the
one thunk by hand, which is also the clearest statement of why the thunk is
there.

Checks run on this branch: `tsc`, `npm test` (4479 pass, 0 fail), `npm run cov`
at 100% lines, branches, and functions, and `npm run gen` with no drift. No
Rust changed.

Changelog:
- `ebnf`: a new front end for writing grammars — rule constructors over the
  `types/range_set` algebra, with the JSON grammar in `ebnf/lib/json`

https://claude.ai/code/session_01CiVMiWos8Lrf96Qtuwar7i

---
_Generated by [Claude Code](https://claude.ai/code/session_01CiVMiWos8Lrf96Qtuwar7i)_


sergey-shandar:
The five open review findings are addressed as of `e848f988`. Four are filed as tracked issues rather than fixed in this PR, at the author's direction, since further work is stacking on top of it. All four reproduce — the commands are in the issues.

| Finding | Where it went |
|---|---|
| Refuse out-of-domain repetition bounds | [`fjs/ebnf/todo/repeat-bounds.md`](../blob/claude/ebnf-branch-pr-nrroty/fjs/ebnf/todo/repeat-bounds.md) |
| Refuse malformed UTF-16 in `set` | [`fjs/ebnf/todo/malformed-utf16-symbols.md`](../blob/claude/ebnf-branch-pr-nrroty/fjs/ebnf/todo/malformed-utf16-symbols.md) |
| Preserve the maximum safe-integer symbol | [`fjs/ebnf/todo/top-symbol-open-tail.md`](../blob/claude/ebnf-branch-pr-nrroty/fjs/ebnf/todo/top-symbol-open-tail.md) |
| Pin the literal grammar constants | [`fjs/ebnf/lib/json/todo/pin-literal-constants.md`](../blob/claude/ebnf-branch-pr-nrroty/fjs/ebnf/lib/json/todo/pin-literal-constants.md) |

Two of those — the unchecked `repeat` bounds and the `errorMask` leak in `set` — answer a mistaken input with a plausible value, which AGENTS.md §1 says may not be deferred. Each issue records that, so neither reads as ordinary backlog.

On **Keep the JSON and DataJS ports atomic**: the contract is not broken here. It exists because `bnf/lib/datajs` imports eight rule values from `bnf/lib/json`, so a port that moves json alone hands classical combinators ebnf thunks. This PR adds a spelling beside an untouched `bnf/lib/json`, which datajs goes on importing, so that hazard does not arise. Both `ebnf-migration.md` and `ebnf-front-end.md` now say which change the requirement binds.

One correction to an earlier round, for the record: the fix for **Reject invalid symbol endpoints** overshot. Requiring `b + 1` to stay safe refuses `rangeEncode(a, MAX_SAFE_INTEGER)`, which `ebnf-range-set.md` spells as the open tail `[a]`. That is the third issue above.

---

sergey-shandar:
Round 5, as of `5d6d81ba`.

**Fixed** — three sites the earlier amendments didn't reach, all of them mine to finish: the port checklist and boundary statement in `bnf-grammar-single-owner.md`, the module triage in `ebnf-migration.md`, and `unicode-rules.md`'s own note (`b30db884`); then the range-set adapter task and the stale `types/array` paragraph (`5d6d81ba`). Following any of them would have rebuilt the boundary the amendment rejects, which is the failure the amendment existed to prevent.

**Not a defect: "Prefix private module constants with underscores".** The `_` rule is about *exported* names that are not public API — "even when module linkage requires an export: exportability is linkage, not API status". It does not reach ordinary module-local constants, and the tree is consistent on that. The classical counterpart of this very grammar, `fjs/bnf/lib/json/module.f.mjs`, declares `onenine`, `digits0`, `digits`, `number` and `value` unprefixed; `fjs/types/range_set/module.f.mjs` has `inUniverse`, `isBoundary`, `validated`, `mergeSorted`; `fjs/text/utf16/module.f.mjs` has `codePointToUtf16`, `isInU16Range`, `u16`; `fjs/bnf/module.f.mjs` has `offset`, `mask`, `terminalSize`, `maxSymbol`. None of the names cited on this PR is exported. Renaming them would make these two modules the only ones in the tree following that reading.

**Open, with the author** — `digit` is the one lexical piece from `bnf-grammar-single-owner.md`'s reuse list that this module does not export. That is a code change, and code on this PR is paused pending the author's approval while further work stacks on top of it.

---

PR #1887, opened 2026-09-04
ebnf/ll1: the LL(1) backend, building the typed AST

sergey-shandar:
The `ebnf/ll1/` piece of [ebnf-migration](https://github.com/functionalscript/functionalscript/blob/ebnf-ll1/fjs/todo/ebnf-migration.md), designed in [`fjs/ebnf/ll1/README.md`](https://github.com/functionalscript/functionalscript/blob/ebnf-ll1/fjs/ebnf/ll1/README.md). It was stacked on [#1886](https://github.com/functionalscript/functionalscript/pull/1886) and stands alone against `main` now that it has merged. It is the reference backend: a predictive parser over the data `RuleSet` whose tree is the typed AST of `ebnf/ast`, so `rewrite` in `ebnf/map` takes what it builds as it is.

**The module.** `parser(rule)` returns a `Parser<Ast<typeof rule>>` and `parserRuleSet(ruleSet, entry)` the same over a hand-written set, validated on entry; a match is `['ok', [tree, end]]` or `['error', index]`, both indices physical, with the end of input synthesized once after the last symbol. The input holds ordinary symbols only — `-1` in it would read as that end of input — and the alphabet is otherwise the caller's. `firstMap(ruleSet)` is the analysis a whole set gets: every rule's first set as a range set, EOF's `-1` among the symbols.

**The tree is `Ast<R>`.** Of the two things `ebnf/map` left the backend — build `Ast<R>` values, or fold the map into the parse — it builds the values: a symbol for a set, `[]` for EOF, an array for a sequence, `[tag, node]` for a variant, one flat array for a repetition whatever its bounds. That answers what remained of the front-end design's Problem 8, and the proof pins it twice: a JSON document's tree node by node, and a larger one through `rewrite` with nothing mapped, which refuses a tree that is not the rule's. The bounded repeat is matched as the data README promised — a round forced below `min`, optional until `max`, an optional one starting exactly when the lookahead is in the item's first set — so `times(3)('')` matches empty three times and `option('')` zero.

**What is refused, before any input.** A grammar one symbol of lookahead cannot decide, naming the rule: **left recursion**, direct or through a nullable prefix, a variant's branch, or a bounded repeat of itself; a **first/first conflict**, naming the branch and the shared symbols; and a **first/follow conflict**, from the follow sets the parser computes over the rules its entry reaches. That last one covers both shapes of decision — a rule that can match empty, and a repetition with a round to spare, `max > min`, which is a decision even where the rule must match once. So `[option('x'), 'x']` and `[repeat(1, 2)('x'), 'x']` are refused rather than accepted and failed on the input they should match, and `[times(2)('x'), 'x']` parses, since a forced round decides nothing. A rule the entry does not reach is dead, not wrong, and is left alone.

**The machine** is a loop over one immutable state, the suspended frames on a heap stack and what to do next, since nesting depth grows with the input: 5000 nested brackets and a 10000-round repetition are in the proof. LL(1) never backtracks, so the first failure is the match's.

**Not here, and recorded.** No metadata channel: the input is symbols and the tree carries no positions — `fjs/ebnf/ll1/todo/metadata.md`, owed to the djs port. `ebnf/matcher/` and `showAst` are retired from the migration plan, since the backend shipped without them and nothing else shares them; the map and data READMEs and Problem 8 of the classical design issue record the `Ast<R>` choice.

**Checks.** `tsc`, `node --test` (4713 passing), `npm run cov` at 100% lines, branches and functions across the repository; `npm run gen` changes nothing.

https://claude.ai/code/session_0161RGMVTmu4yCS3uwz6Q3Ey


o2alexanderfedin (approved):
Stacked on #1886 — delta measured against the merge base `d3f9126ce898fb38d5c707f339bb2901f70c4433`, which is an older commit of that branch, not its current head. Own files: `fjs/ebnf/ll1/*`; the rest of `fjs/ebnf/` arrives via the base and is reviewed there.

Approved. Two things make this convincing rather than merely green.

**The language matches the classical backend.** `bnf/ll1`+`bnf/lib/json` against `ebnf/ll1`+`ebnf/lib/json`, full-consumption on both sides: 86 pinned inputs plus 80,000 fuzz strings, zero disagreements, with 2,261 fuzz accepts so the probe has power. Every edge that has caught real bugs here agrees, and agrees with `JSON.parse`. On top of that, 10,000 random rule sets against an exhaustive backtracking reference — 628,000 checks, no over- or under-acceptance — and breaking the lookahead drives that to 71 failures, which is what shows the comparison can fail.

**The conflict refusal is real and pinned.** Hand-written ambiguous sets produce `['first/first conflict','v','q',97,98]` — right rule, right offending tags, still right two rules deep — plus left recursion and first/follow cases; 16 refusals are pinned in `proof.f.mjs`, and defeating the `clash.length === 0` assert turns them red.

Two doc nits:

`fjs/ebnf/ll1/module.f.mjs:226-227` opens the JSDoc block twice — `/**` then `/**` — so it renders as `/** /**`.

`fjs/ebnf/ll1/README.md:108` says of `[option('x'), 'x']` that "the classical backend accepted it and failed on that input". `bnf/ll1` throws while building that grammar, so it never accepts it; the sentence is true of `bnf/descent`. Line 90 uses the same phrase where `bnf/ll1` is the subject.

o2alexanderfedin (commented):
Rebased onto `main`. Every file under `fjs/ebnf/ll1/` is byte-identical to the head I approved, so the LL(1) evidence stands unchanged — I did not re-derive it. The PR's own delta is three documents, and the gates pass after the rebase.

`90e7130c` checks out: `fjs/ebnf/data/README.md` really did say `{ tag, sequence }` when it landed and says `Ast<R>` now, with only two contrastive mentions left under `fjs/ebnf/`.

**One thing in the new text.** `fjs/ebnf/map/README.md:155-157` says the walk "refuses a *rule* the lowering refuses, where it reads one — an info that is no tuple or carries a field past its arity, a `const` over a thunk, a set or a repetition whose bounds are outside the domain". That reads as the complete list, and it is short by two:

```
rewrite([])(0.5)(0.5)          ->  ['not a symbol', 0.5]
rewrite([])('\uD800')('\uD800')  ->  ['malformed UTF-16', '\uD800']
```

`rule-restrictions.md`, which the paragraph cites, lists both and says `rewrite` "has grown a refusal for each of those" — so the README is under-inclusive against the code and against the document it points at.

The two nits from my last review are still open, since those bytes did not move: `fjs/ebnf/ll1/module.f.mjs:226-227` opens the JSDoc block twice, and `fjs/ebnf/ll1/README.md:108` still says the classical backend "accepted it and failed on that input" for `[option('x'), 'x']`, which `bnf/ll1` refuses at build.

Issue #1893, opened 2026-09-05
`evo(cas)(key)` throws a raw TypeError when the memory slot was never allocated


o2alexanderfedin:
Handing `evo` a cache key whose `memory` slot was never allocated does not produce a channel error. On `0.48.0` it throws:

```
TypeError: Cannot read properties of undefined (reading 'bySubject')
```

Minimal reproduction, no filesystem:

```js
import { virtual, emptyState } from 'functionalscript/fjs/effects/node/virtual/module.f.mjs'
import { evo } from 'functionalscript/fjs/cas/evo/module.f.mjs'
import { fileCas } from 'functionalscript/fjs/cas/module.f.mjs'
import { sha256 } from 'functionalscript/fjs/crypto/sha2/module.f.mjs'

const cas = fileCas(sha256)('/home')
virtual({ ...emptyState, memory: [] })(evo(cas)(0).list('false'))
// TypeError: Cannot read properties of undefined (reading 'bySubject')
```

`0` is a well-formed key shape; what is missing is the `initEvo` that would have allocated the slot behind it.

## Not a live fault for us

Our only production construction path calls `initEvo` first and threads the key, so we cannot reach this and have no workaround in place. Filing it because the message is a diagnostics gap, not because anything is broken downstream.

## The ask, as a choice

Whether a never-allocated slot is *programmer error* or a *runtime condition* is your call and both are defensible. What is hard to defend either way is the message: `Cannot read properties of undefined (reading 'bySubject')` is V8's sentence rather than anyone's design — it names an internal field of the cache record and mentions neither `evo`, nor the key, nor `initEvo`.

Two shapes:

1. **Return `evoError`** — e.g. `evoError('evo cache slot <key> was never allocated; call initEvo first')`. Costs a branch per operation, and widens `list`/`head` from `NotImplemented` to the full `EvoChannel`. That widening is worth weighing: the `NotImplemented`-only typing of `list`/`head` is load-bearing for consumers — it is exactly what tells a caller that only a dispatch failure can come out of `list`.
2. **Panic by name** — keep it programmer error, but throw your own sentence instead of V8's. No type changes, no per-operation cost, actionable message.

Shape 2 looks like the better trade from outside, precisely because it preserves that typing distinction — but that is a judgement about your intent, so it is a preference rather than the request.

Happy to send a PR for either shape.

o2alexanderfedin:
[...]

PR #1899, opened 2026-09-07
effects/node/virtual: refuse a never-created memory slot

o2alexanderfedin:
Fixes the crash reported in #1893, one layer below where it was reported.

## What was wrong

`virtual`'s `memRead` read the slot table by index and wrapped whatever came
back in `ok`:

```js
memRead: key => state =>
    [state, ok(state.memoryValues[asBase(key)])],
```

For a key `memCreate` never handed out that is `ok(undefined)` — a plausible
wrong value, which `doc/DESIGN.md` §10 ("Refuse what you cannot handle")
names as the one outcome that is never acceptable. Nothing failed at the read;
the program failed wherever the `undefined` was first used. In the reported
case that was `evo`'s cache:

```js
import { emptyState, virtual } from './fjs/effects/node/virtual/module.f.mjs'
import { evo } from './fjs/cas/evo/module.f.mjs'
import { fileCas } from './fjs/cas/module.f.mjs'
import { sha256 } from './fjs/crypto/sha2/module.f.mjs'
import { asNominal } from './fjs/effects/memory/module.f.mjs'

const cas = fileCas(sha256)('.')
const key = asNominal('mem0')          // a slot `initEvo` never allocated
virtual(emptyState)(evo(cas)(key).list())
virtual(emptyState)(evo(cas)(key).head('doc'))
```

Before — both lines:

```
TypeError: Cannot read properties of undefined (reading 'bySubject')
```

That message is V8's. It names an internal cache field and mentions neither the
key, the slot, nor `initEvo`. After:

```
memory key not found: mem0
```

(The issue's snippet spells the fixture as `{ ...emptyState, memory: [] }`; the
real `State` fields are `memoryNext`/`memoryValues`, so the version above is
what actually reproduces on `main`. `evo` is the reporter's path to it, not the
only one — `fjs/mcp` allocates a slot through the same effect.)

`memWrite` had the mirror of the same hole: it spread the key into
`memoryValues` unconditionally, so writing to a key that was never created
silently allocated it, and every later read of that key then succeeded too. Both
are fixed here; the read is the reported half.

## Why a panic, and why here

**The condition is already a panic in this repository, for this exact
operation.** `fjs/effects/node/memory/module.mjs` — the real interpreter this
virtual one stands in for — has refused it since it was written:

```js
const missingKey = id => new Error(`memory key not found: ${id}`)
// …
memRead: async key => {
    const id = asBase(key)
    if (!store.has(id)) { throw missingKey(id) }
    return ok(store.get(id))
},
```

`fjs/effects/memory/proof.f.mjs`'s mock does the same for `memWrite`, with
`assert(id in state.values, id)` and a `throw:` proof beside it. The virtual
runner was the only one of the three that refused nothing. So this is not a new
policy — it is closing a divergence, and the change takes the existing sentence
verbatim rather than inventing one. That matters beyond tidiness: a proof run
against the virtual runner is only evidence about the Node one where the two
answer alike.

Named panics are also the ordinary idiom in production `module.f.mjs` generally
— `types/prime_field` throws `'1/0'`, `types/btree/remove` `'invalid node'`,
`nanvm` `['unknown shared value', name]`, `bnf` `` `Invalid range ${ab}.` `` —
and `assert(cond, msg)` is exactly the form
`doc/DESIGN.md` §10 ("Refuse what you cannot handle") prescribes for
an input that violates something the caller was supposed to guarantee. A memory
key is such an input: the only way to obtain one is `memCreate`, so a key that
was never created is a caller bug and there is nothing sensible to hand back.

**The alternative offered on the issue was to return `evoError` instead.** That
was rejected. It would widen `Evo.list` and `Evo.head` in `fjs/cas/evo/types.ts`
from `NotImplemented` to the full `EvoChannel` — a breaking public API change —
to describe a condition that is not a runtime failure of `evo` at all. It would
also be the wrong layer twice over: the guard would be dead against the real
Node runner, which already throws before `evo` sees anything, and every other
slot consumer would need its own copy of it. Nothing under `fjs/cas/evo/` is
touched by this pull request.

**Presence is the test, not the value.** `memCreate(undefined)` allocates a real
slot, so the guard is `hasOwn`, never `=== undefined`; a proof pins that.

## Checks

| Check | Result |
| --- | --- |
| `node --test` on `main` (baseline) | 5495 pass / 0 fail, exit 0 |
| `node --test` with this change | **5499 pass / 0 fail, exit 0** (+4 new proofs) |
| `npm run cov` | **100.00 lines / 100.00 branches / 100.00 funcs**, all files, exit 0 |
| `npm run gen`, then `node --test` again | no generated-file drift (`git status` clean but for the two edited files); 5499 pass / 0 fail, exit 0 |

No pre-existing test changed its result — including under the `memWrite` guard,
which nothing in the suite trips.

The four new proofs are in the co-located `fjs/effects/node/virtual/proof.f.mjs`:
a create/write/read round trip, a slot holding `undefined` (which must still
read cleanly), and the two refusals under a `throw:` key per
`fjs/AGENTS.md` §1.5.
I checked the two refusal proofs are load-bearing by deleting the guards and
re-running them: both then return cleanly, which the runner scores as failures.

## `tsc` — run, with the pinned compiler

**Corrected.** An earlier revision of this description reported `tsc` as unrun and owed to CI.
That was wrong, and the reason was that I stopped looking too early: npm does not carry
typescript-go at `7.0.2` (only `7.0.0-dev.*` prereleases), and I treated that as the answer.
It is not — the release exists upstream under the tag `typescript/v7.0.2`, with a prebuilt
`typescript-darwin-arm64.tgz` among its assets.

So the check is run, with **the exact compiler `fjs/ci/config/module.f.mjs` pins**:

```
$ tsc --version
Version 7.0.2
$ tsc          # on this branch
$ echo $?
0
```

Clean, no output. The same compiler on pristine `main` is also clean, so the zero is this
branch's and not a baseline that happened to be quiet.

The change adds no types and no new exports. `assert` and `Object.hasOwn` were
already in scope in the module (lines 15 and 59); the proof's two unallocated
keys are annotated `/** @type {Key<number>} */` the way
`fjs/effects/memory/proof.f.mjs` annotates its own. That is a reason to expect
it to be clean, not evidence that it is.

Changelog:
- `effects/node/virtual`: reading or writing a memory slot that `memCreate`
  never allocated now panics with `memory key not found: <key>`, the sentence
  the Node interpreter already uses. The read previously returned `undefined`
  as a success, which surfaced as an unrelated `TypeError` in whichever caller
  used the value first.


sergey-shandar:
@codex review

sergey-shandar (approved):

PR #1908, opened 2026-09-09
ebnf/lib/json: fold the grammar to a JSON value or an error

sergey-shandar:
Stacked on #1907: a JSON reader built as a rewrite set over the EBNF JSON grammar, folded by the LL(1) backend as it parses.

## What it does

`fjs/ebnf/lib/json/parser` maps the grammar in `fjs/ebnf/lib/json` to a JSON value, one mapping per rule the grammar holds:

- `string` → the text it spells, every escape decoded;
- `number` → its value, or the error it is;
- `value` → what its branch made of it: a container folds its items, a keyword is its constant;
- `json` → its value's symbol, so `parser(json, mappings)` yields one symbol carrying the value.

`parse(text)` runs the set over `[json, eof]` and returns `Result<Unknown, Error>`.

**The mapped output is a JSON value or an error, never a value that is no JSON.** A mapping reports nothing, so the one thing in a JSON text that is no JSON value — a number the finite `number` range cannot hold, `1e400` — is an `['error', ['range', lexeme]]` standing where the number would; a container holding one is that error too, and the first in document order is the document's. A parse that fails builds no tree and reports `['error', ['syntax', index]]`.

The input is the UTF-16 alphabet of the new `fjs/ebnf/utf16` — its `Utf16` metadata, and `units(text)`, a text as symbols — code units rather than code points, so a string is the sequence of units it spells: a lone surrogate is one unit in and one unit out, raw or escaped, as `JSON.parse` reads it, where a decoder to code points would refuse it before the grammar saw it. Nesting depth stays on the machine's heap stack: 5000 levels of brackets and 10000-item arrays parse in the proof.

## What the grammar needed

For each mapping's function to be typed from its rule, the grammar's helpers now build rules as precise as ones written out — the first three rows of `widened-rule-signatures`, plus its `value` row for JSON:

- `cj`, `array`, `object` and `createValue` take `const` type parameters and return `Container<Item>` and `Value<P, V>` from a new `fjs/ebnf/lib/json/types.ts`;
- `string`, `number` and `hex` are pinned (closing `pin-literal-constants`); `number` and `value` are exported;
- `value` is annotated `JsonValue`, a recursive alias, `() => readonly ['const', Value<typeof string, JsonValue>]`, which TypeScript admits because the reference sits inside a function type.

`Ast` of a rule that names itself is infinite, and it stays finite to the checker by one change in `fjs/ebnf/ast/types.ts`: the variant row builds each branch through an alias, `_Branch`, whose tuple TypeScript defers, where the tuple written directly into the mapped type met itself (TS2615). The grammar's proof gains a `types` entry pinning the seven tags of `Children<typeof value>`, the document's shape, and the refusal of a shape the grammar cannot produce — at depth as at the top, which the old `Const<Variant>` typing could not say. Declaration emit was checked: no `any` or elided types in the JSON or `ast` declarations.

The narrowing is a source-incompatible change to `fjs/ebnf/lib/json`'s declared types, declared below: `createValue(p, v)` is typed by its seven tags, so reading a key it has not — `createValue(p, v).missing`, which the abstract `Variant` admitted as a `Rule` — is now a type error, and `Ast<typeof json, I>` refuses a tree the old `Const<Variant>` typing admitted, a foreign tag at depth say. Migration: read through the abstraction where that is meant — annotate the binding `Variant`, as the grammar's own `contract` proof now does — and a tree pinned as `Ast<typeof json, I>` has to be one the grammar builds.

## Issues

- `fjs/ebnf/lib/json/todo/pin-literal-constants.md` — closed, deleted.
- `fjs/ebnf/lib/todo/widened-rule-signatures.md` — records what landed and what the annotation cost; the DataJS `value` row stays open.
- `fjs/bnf/todo/ebnf-front-end.md` — Problem 7 answered from it.
- `fjs/media/json/todo/self-contained-tokenizer.md` — notes that the value route now exists beside the token-stream stage it describes, and which of its four contracts the reader keeps (depth, code units) and does not (a numeric policy per codec, a public token stream). It replaces nothing in `fjs/media/json`.
- `fjs/todo/ebnf-migration.md` — `utf16/` added to the layout.

## Checks

`tsc` clean; `node --test` 5469 passed, 0 failed, exit 0; `npm run cov` 100% lines, branches and functions across all files; `npm run gen` is a no-op.

Changelog:
- **BREAKING CHANGES:** `ebnf/lib/json`: `cj`, `array`, `object` and `createValue` return the precise shapes they build (`Container<Item>`, `Value<P, V>`) rather than `Tuple` and `Variant`, `string` is a pinned tuple rather than `Rule`, and `value` is typed `JsonValue`; a key `createValue`'s result has not is a type error rather than a `Rule`, and `Ast<typeof json, I>` admits only trees the grammar builds
- `ebnf/lib/json`: `parser/` parses JSON through the grammar's rewrite set, returning a JSON value or an error; `number` and `value` are exported
- `ebnf/utf16`: the code-unit alphabet — `Utf16`, `utf16`, `units`
- `ebnf/ast`: a rule that names itself through a variant has a finite `Ast`

https://claude.ai/code/session_01Y9osdSCmhpZ4NcSDWd3McY


fjs/ebnf/lib/json/parser/types.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import type { Unknown } from '../../../../media/json/types.ts'
+import type { Result } from '../../../../types/result/types.ts'
+
+/**
+ * The input alphabet: UTF-16 code units, one symbol each, with nothing the
+ * grammar ignored about them. A caller that knows more — a position, say —
+ * hands over a record that carries this one's `id` beside it.
+ */
+export type Utf16 = { readonly id: 'utf16' }
```
sergey-shandar: This input alphabet should have its own home.

sergey-shandar: Moved to `fjs/ebnf/utf16/` in 1208f0c: `Utf16` in its `types.ts`, and `utf16` (the shared record) with `units(text)` in its `module.f.mjs`, with a proof of its own. The reader imports it, and `ebnf-migration`'s layout lists `utf16/` beside `unicode/`. Picked a directory named for the alphabet rather than putting it under the planned `unicode/`, since that one is for rules over code points and this is the code-unit alphabet.

---

o2alexanderfedin (approved):
Looks good. Delta measured against its own base, PR #1907's head `edc7d8a5bb41abb6f375551ac3f01b76c06ca7b7`; `tsc -p .` exits 0 and `node ./fjs/module.mjs t` exits 0 (pass: 5531, fail: 0, i.e. +13 over 1907's 5518). Differential-tested the fold against `JSON.parse` over several hundred thousand inputs — escapes, lone and paired surrogates, `-0`, underflow, precision, duplicate keys, deep nesting — with no divergence beyond the documented out-of-range number case.

o2alexanderfedin (approved):
Reviewed only the delta since my previous approval (`1208f0c9..e34ca784`): a clean merge of `origin/claude/sleepy-galileo-kvuvni`, carrying only 1907's two repointed line citations in `fjs/ebnf/todo/symbol-domain-owner.md`. Verified against this head's tree that `fjs/ebnf/ll1/module.f.mjs:42` is `eofSymbol` and `:52` is `isSymbol`, and that the merge introduced nothing else. Still approved.

o2alexanderfedin (approved):
Approving the delta only: the branch was rewritten onto `main` (base is now `main` at `e6324f811ef97b9f4cc09c8796a488d42d7abcc9`, which is #1907 landing), and the resulting tree at `e9641066` is byte-identical to the tree at the previously approved head — this PR's own content measured against its new base is an exact match for the diff already reviewed, every merge in the range has an empty combined diff, and the only new content in the range (`todo/git-name-resolution.md`) arrived from `main` via #1904 rather than from this PR. tsc/suite/differential runs skipped: the delta touches neither the fold nor the reader.

PR #1910, opened 2026-09-09
media/json: read JSON through the EBNF grammar, one policy per codec

sergey-shandar:
Stacked on #1908: `fjs/media/json/parser` is now the EBNF reader that PR built, moved out of `fjs/ebnf/lib/json/parser` and generalized over the codec's numeric policy, and both codecs' `parse` run it. The container state machine over a token stream is deleted.

## What it does

`fjs/media/json/parser` exports two things:

- `mappings(policy)` — the rewrite set over the grammar in `fjs/ebnf/lib/json`, one mapping per rule it holds: `string` to the text it spells with every escape decoded, `number` to what `policy` makes of its **exact lexeme**, `value` to what its branch made of it, `json` to its value's symbol.
- `parse(policy)` — the parser over `[json, eof]` with that set folded in, returning `(text: string) => Result<ParseUnknown<P>, string>`.

`NumberPolicy<P>` is now `(lexeme: string) => Result<P, string>`. The standard codec keeps reading every number as JavaScript does (`1e400` is `Infinity`), and the extended codec keeps reading bare integers as `bigint` straight from the lexeme and refusing what the finite range cannot hold — the numeric-policy contract `self-contained-tokenizer` held a value mapping could not keep. It is kept by building the set per policy: the policy's error stands in the number's place, a container holding one is that error too, and the first in document order is the parse's.

A parse that fails reports where — `unexpected end`, or `unexpected symbol at N` with `N` the code unit the LL(1) backend stopped at — rather than `unexpected token`. The error type stays `string`, so every consumer of `json.parse` and `extended.parse` (`djs/transpiler`, `protocol/mcp/stdio`, `media/revision`, `media/lock`, `media/note`) compiles and behaves as before; one DJS proof that pinned the old message is updated.

`fjs/ebnf` no longer depends on `fjs/media/json`: the reader was the one importer of `Unknown` from there.

## What stays

- `fjs/media/json/tokenizer` is public and unchanged; it is no longer on `parse`'s path. Its own proof covers it in full.
- Depth: the proof parses 5000 nested brackets, 6000 sibling containers and 12000 items, as the old machine's proof required.
- Strings are code-unit sequences: a lone surrogate is one unit in and one out, raw or escaped.

## Issues

- `fjs/media/json/todo/parser-container-stack-bookkeeping.md` — described the deleted machine; deleted.
- `fjs/media/json/todo/self-contained-tokenizer.md` — status rewritten: the value route landed as the reader, three of its four contracts are kept through the mapping and the fourth by leaving `tokenize` where it is; what remains is what becomes of `tokenize`.
- `fjs/media/json/README.md`, `standard-parse-serialize.md`, `fjs/djs/todo/parse-error-location-format.md`, `fjs/ebnf/ll1/README.md`, `widened-rule-signatures.md` — pipeline and links updated.

## Checks

`tsc` clean; `node --test` 5422 passed, 0 failed, exit 0; `npm run cov` 100% lines, branches and functions across all files; `npm run gen` is a no-op; declaration emit of the reader has no `any`.

Changelog:
- **BREAKING CHANGES:** `media/json/parser`: `parse(policy)` takes the text rather than a token list; `NumberPolicy<P>` is `(lexeme: string) => Result<P, string>` rather than a function of a `NumberToken`; the parser's state types (`_JsonState`, `_StateParse`, …) are gone; a syntax error's message names the index the parse failed at
- `media/json`: `parse` in both codecs reads through the JSON grammar of `ebnf/lib/json`, folded by `ebnf/ll1`; `ebnf/lib/json/parser` moved to `media/json/parser`

https://claude.ai/code/session_01Aw8spamVSoymum2CuFQ3jS

---
_Generated by [Claude Code](https://claude.ai/code/session_01Aw8spamVSoymum2CuFQ3jS)_


o2alexanderfedin (commented):
Reviewed as a stacked PR: delta measured against its own base, PR #1908's head `1208f0c992404235c4547da9f487928ee9c20113`, not `main`. `tsc -p .` exits 0; `node ./fjs/module.mjs t` exits 0 (pass: 5484, fail: 0).

One thing to fix — a doc link broken by a deletion in this PR:

This PR deletes `fjs/media/json/todo/parser-container-stack-bookkeeping.md`, but `fjs/djs/todo/serializer-children-helper.md` still links to it twice and is not touched here:

- `fjs/djs/todo/serializer-children-helper.md:41` — "…the serializer counterpart of [the JSON parser's container-stack bookkeeping](../../media/json/todo/parser-container-stack-bookkeeping.md)"
- `fjs/djs/todo/serializer-children-helper.md:78` — the same link under `### Related`

Both are prose links, not code samples, and the target resolves at `main` and at #1908's head but not at this one. The `:78` entry also still describes it as "the same container-kind merge, on the JSON parser", which now names a file that no longer exists.

Since the deletion looks intentional (the bookkeeping it described is what the fold replaces), the fix is probably to update the two referring passages rather than restore the file.

sergey-shandar:
Addressed the review in `facc1f5`: the deletion of `parser-container-stack-bookkeeping.md` was intentional, so both passages in `fjs/djs/todo/serializer-children-helper.md` now say what became of the counterpart — the JSON parser is a rewrite set folded into the grammar's parse, with no container-kind branches to merge — and link `fjs/media/json/parser` in its place. A repository-wide sweep finds no other reference to the deleted file.

`tsc` clean; `node --test` 5422 passed, 0 failed, exit 0.

---

o2alexanderfedin (approved):
Delta since my last review is clean: the dangling link to the deleted `parser-container-stack-bookkeeping.md` is resolved — both citations in `fjs/djs/todo/serializer-children-helper.md` now point at `fjs/media/json/parser/module.f.mjs`, the name appears nowhere in the tree, and all three links in that file resolve. Measured against base `e34ca78423f372ca4cad54cc3d452e3161d2cb6c` (`claude/ebnf-json-parser`).

o2alexanderfedin (approved):
Restack only — no content change from my approval at `facc1f5d`, so my approval stands.

Measured against base branch `claude/ebnf-json-parser` @ `e96410664b361a3aadd79e10fe0e142906949de6`: the contribution `e9641066...c0291c22` is byte-identical to the approved `e34ca784...facc1f5d` (diff-of-diffs empty, positive-controlled against a wrong base, which does report a difference), and the new range is linear — zero merge commits, against 1 in the old range. The only tree difference versus the approved head is `todo/git-name-resolution.md`, blob-identical to `origin/main`'s copy and citing nothing in this stack.

`tsc` and the suite skipped deliberately: the content delta relative to the approved state is zero code.

PR #1913, opened 2026-09-09
todo: namespace Git attestations and DISOT metadata

sergey-shandar:
Rename the experimental Git attestation headers to the FunctionalScript vendor namespace:

- `didsig` -> `vnd.fjs.didsig`
- `tstsig` -> `vnd.fjs.ttssig`

`vnd.fjs.didsig` is intentionally DID-native rather than GPG/OpenPGP-specific. Its signature algorithm and key representation come from the DID verification method, so DID methods using `secp256k1`, P-256, Ed25519, or future algorithms are not required to map to an OpenPGP key/signature format. The existing Git `gpgsig` remains the conventional Git-signature layer.

`tts` means Trusted Time-Stamp; the doubled `t` is intentional. The vendor prefix makes the extension origin explicit while also avoiding Git's special `gpgsig*` payload handling.

Also make `dialect: "vnd.fjs.disot"` a required part of the logical `.disot` metadata value, shared by `.disot.json`, `.disot.data.js`, and any future lossless encoding.

Updates the trusted-timestamp and Git name-resolution TODOs only; no runtime behavior changes.

### DIFF
--- a/todo/git-name-resolution.md
+++ b/todo/git-name-resolution.md
@@ -55,6 +55,19 @@ They are two serializations of **one logical DISOT metadata value**:
 - `.disot.json` — JSON;
 - `.disot.data.js` — DataJS.
 
+That logical value MUST identify its format explicitly:
+
+```json
+{
+  "dialect": "vnd.fjs.disot"
+}
+```
+
+`vnd.fjs.disot` follows the same vendor-style `vnd.fjs.*` dialect convention as
+other FunctionalScript formats. The dialect is part of the metadata value, not
+inferred from the filename, so every supported `.disot.*` encoding carries the
+same required `dialect` field.
+
 ### Metadata lookup boundary
 
 DISOT metadata is recognized **only at the root of the Git commit tree**.
@@ -84,6 +97,7 @@ Preferred form:
 
 ```json
 {
+  "dialect": "vnd.fjs.disot",
   "name": "/secp256k1:23a/coolJsModule"
 }
 ```
@@ -102,6 +116,7 @@ A DISOT metadata file MAY use a relative self-name:
 
 ```json
 {
+  "dialect": "vnd.fjs.disot",
   "name": "./coolJsModule"
 }
 ```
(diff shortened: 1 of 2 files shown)


o2alexanderfedin (commented):
The `vnd.fjs.*` namespacing and the required `dialect: "vnd.fjs.disot"` both look right — the field name and the vendor-tree convention match `fjs/media/lock` and `fjs/media/revision` exactly, every `.disot` example in `todo/git-name-resolution.md` now carries the field, and no stale `didsig`/`tstsig` reference survives outside the deliberate rationale sentence.

One question on the stem, though — `vnd.fjs.gpgsig`.

The header is defined as "a DID-addressed author/attestor signature", its section is "DID signatures", it is produced by `DIDSign(did, key, Hash(A))`, and each value binds a DID plus a verification-method identifier. Nothing about it is a GPG signature. Both goals stated in the description — the explicit FJS vendor namespace, and staying clear of Git's `gpgsig*` payload handling — are met just as well by `vnd.fjs.didsig`, which keeps the accurate stem the old name had.

Two knock-on effects of the swap:

- the surviving rationale now reads "`vnd.fjs.gpgsig` deliberately does **not** begin with `gpgsig`", which lands oddly for a name that ends with it — the point being made is about the prefix, but the sentence has to fight its own subject;
- the verifier projections make the reader hold `gpgsig` and `vnd.fjs.gpgsig` apart across sections 2 and 4 ("remove all standard Git signature headers / remove all `vnd.fjs.gpgsig` headers"), which is exactly the ambiguity `didsig` never had.

If the intent is instead a generic FJS signature slot that happens to carry DID signatures today, that would explain the stem — but the doc currently defines the header as DID-only, so it would be worth saying so.

`vnd.fjs.ttssig` reads fine and its rationale is in the doc; the `dialect` half of the PR I have no concerns about.

o2alexanderfedin (approved):
Rename to `vnd.fjs.didsig` is complete across both docs and the DID-native rationale is now stated explicitly; no stale references remain.

PR #1924, opened 2026-09-10
ebnf: name the rules under JSON's string and read combinator scaffolding once

sergey-shandar:
The JSON reader mapped four rules — `string`, `number`, `value`, `json` — and walked everything under them by hand, because the rules under `string` were object literals written inline and the pairs `cj` and `join` build are returned to nobody, so nothing could key a mapping by them. The DataJS reader copied the container walk and the two AST tests verbatim.

`fjs/ebnf/lib/json` now exports `hex`, `escape` and `character`, and the reader maps `escape` to the character it spells and `character` to its text, so `string` is the join of its characters' symbols; the reader's `private.ts` is gone with the rules it recovered through `ReturnType`. The three mappings are one export, `stringMappings`, since the backend refuses a mapping of a rule the grammar does not hold and a grammar holding `string` holds all three — a proof pins that refusal.

The scaffolding a combinator builds is read once, beside the combinator: `joined` in `fjs/ebnf` reads the list `join` matched, `items` in `fjs/ebnf/lib/json` the container `cj` matched, and `unmapped` and `symbolAt` in the new `fjs/ebnf/ast/module.f.mjs` are the two tests a mapping makes. Both readers import them.

A reader generic over the alphabets cannot be typed by `Children`: `Meta<O>` with `O` a type parameter is a deferred conditional, from which TypeScript infers nothing. So `Unmapped` and `JoinNode` in `fjs/ebnf/ast/types.ts` and `ContainerNode` in `fjs/ebnf/lib/json/types.ts` type the two readers by the shape of their node, with the item's node as their one type parameter, and `Join` in `fjs/ebnf/types.ts` names what `join` builds. Type assertions in the proofs pin that a real `Children` node fits those shapes; `fjs/ebnf/ast/README.md` records the reason.

`AGENTS.md` gains two habits: propose and ask when something would improve or simplify the task, and export what another module needs rather than copying it.

Checks: `tsc`, `fjs test` (5531 pass), `npm run cov` (100% lines, branches, functions on all files), `npm run gen` (no diff).

Changelog:
- **BREAKING CHANGES:** `media/json/parser`: `stringMapping` is `stringMappings`, the mappings of `string`, `character` and `escape` as one list
- `ebnf/lib/json`: exports `hex`, `escape`, `character` and `items`; `ebnf`: exports `joined` and the `Join` type; `ebnf/ast`: new `module.f.mjs` with `unmapped` and `symbolAt`, and the `Unmapped` and `JoinNode` types

https://claude.ai/code/session_012RKwVzkr9az9ba6hSqF67Z


o2alexanderfedin (approved):
Verified behaviour-preserving against the merge-base over ~2.8k adversarial inputs and every repo .json across all three readers; gates green; breaking rename declared with importers updated.

o2alexanderfedin (approved):
Delta over my approval at 8337f025, measured against base 2a365b8d: a merge of the moved main (its `joined` parameter rename to `list` avoids main's new module-private `node`) plus a type-only `@import` of `hex`, whose sole use is `typeof hex` — tsc clean, 5547/0.

PR #1928, opened 2026-09-10
nix: a setup script for a container without Nix

sergey-shandar:
A Claude Code on the web container has no Nix, no `tsc`, and a Node that is not the one CI pins. `nix/setup.sh` turns such a container — a Claude Code environment, or any other that starts from an image without Nix — into one where `./nix/run` and `./dev.sh` work at once. Nothing is specific to Claude Code and nothing lives in `.claude/`.

The script runs as root and needs nothing of the repository on disk: Nix fetches the flake from `main` by URL, and the script reads the lock from that same fetch. It installs Nix with the official single-user installer (the install runs as root, so `/etc/nix/nix.conf` unsets `build-users-group` first and enables the `nix-command flakes` features `./nix/run` also asks for), fetches every locked `github:` input of `nix/flake.lock` over git, and enters the shell once, so the shell's whole closure is on disk. It leaves nothing behind but the Nix store and Nix's own caches, and an environment that caches its filesystem after the script keeps all of it.

The git fetch is a workaround: a Claude Code environment's GitHub proxy scopes tarball downloads to the session's repository and answers 403 for `oxalica/rust-overlay`, while git access to public repositories is not scoped. A shallow git fetch of the locked revision produces the store path the lock names (same content hash), so Nix finds the input in the store and never asks GitHub for the tarball. The input list is read from the lock with `nix eval`, so no JSON tool is needed and the list follows the lock.

A Claude Code environment runs the script once, before any session, and caches the result; its **Setup script** field holds one line, which fetches the script from `main`:

```bash
curl -fsSL https://raw.githubusercontent.com/functionalscript/functionalscript/main/nix/setup.sh | bash
```

The session's shell already has `/root/.nix-profile/bin` on its `PATH`, so a session then runs every command as `./nix/run <command>`, with `./nix/run npm ci` first; `AGENTS.md` says so. `CONTRIBUTING.md` documents the script and the one-liner, and `nix/README.md` lists it as the second hand-written script under `nix/`.

Measured in a web container, `x86_64-linux`, 4 cores, from a container without Nix and from outside any checkout: 2 min 34 s, within the five-minute limit the environment cache requires. After it, `./nix/run node --version` in a checkout takes two seconds and reports `v26.8.1`, and `tsc` (7.0.2), `node --test` (5540 pass, 0 fail), `cargo clippy -- -D warnings` and `cargo fmt -- --check` pass in the shell.

https://claude.ai/code/session_014WovW2cRSKc1eaLmVLujvg


o2alexanderfedin (commented):
Verified at `45954286`: `node ./fjs/module.mjs ci` leaves the tree byte-clean (positive-controlled by flipping the pinned `bun` version — `.github/workflows/ci.yml` and `nix/flake.nix` regenerate, no `SyntaxError`), `tsc -p .` exits 0, and `node ./fjs/module.mjs t` reports 5547 pass / 0 fail. The `.gitignore` re-include behaves as intended: a new `.claude/hooks/*` file is tracked while `.claude/settings.local.json` stays ignored. `CLAUDE_CODE_REMOTE`, `CLAUDE_ENV_FILE` and `CLAUDE_PROJECT_DIR` are all documented hook variables and `timeout` is in seconds, so 1800 is valid.

Two things.

**The two scripts carry the same code twice.** The `nix eval ... | while read ... nix flake prefetch` block is byte-identical in `.claude/setup.sh` and `.claude/hooks/session-start.sh` — 11 lines — and the `/etc/nix/nix.conf` write plus the installer `curl` differ only in the profile path (`/root/.nix-profile` against `$HOME/.nix-profile`) and in how `USER` is set. `doc/DESIGN.md` §4 calls DRY "a core principle of FunctionalScript, not just a stylistic preference", and its "only extract once the second real consumer exists" condition is met here; `AGENTS.md` §1 adds that "a copy is a second implementation that will drift". The two copies already disagree on whether the environment is root, which is that drift starting. `setup.sh` clones the repository anyway, so cloning before installing Nix would let both scripts call one shared step out of the clone.

**`CONTRIBUTING.md` leaves out `cargo fetch`.** The new paragraph says the hook "runs `npm ci`"; the hook runs `./nix/run npm ci` and `./nix/run cargo fetch`. The rest of the document pairs the two commands wherever it names them.

PR #1954, opened 2026-09-10
edag: spell unary + and - as Op12, one tag at two arities

sergey-shandar:
Implements the issue filed in [#1953](https://github.com/functionalscript/functionalscript/pull/1953), on top of that branch, and deletes the issue file.

**EDAG.** A third operation vocabulary, `Op12Id = '+' | '-'`, whose tags are legal at both arities: `['-', a]` is negation and `['-', a, b]` subtraction, `['+', a]` is JS unary plus and `['+', a, b]` addition. The schema spells it as `op12`, an `or` of the two closed tuples, so the node's length decides. `neg` leaves `Op1Id` and `+`/`-` leave `Op2Id`, keeping the three vocabularies pairwise disjoint, and the proof pins that disjointness. Amnesia dispatches the two tags on length; its proof pins both arities and that `+0n` throws, the one input that separates unary `+` from `Number`.

The word tag `neg` existed because rtti tuples once had open trailing positions, so `['-', a]` also matched `['-', a, b]`. Tuples are closed and validated by length now, and the chain steps already end by the same arity rule. `todo/edag-stage1-discussion.md` records the reversal and states why unary `+` is an EDAG operation without being FunctionalScript syntax; `spec/todo/2340-operators.md` is unchanged.

**Corpus.** `Group12` carries an explicit `arity`, the one place the id fixes nothing, and `arityOf` reads it first. `groupKey` names the two arities of one tag distinctly (`'-/1'`, `'-/2'`) with one owner for both consumers: the proof's test keys and the printer's `rustName` table use it. The `unaryPlus` group moves off `NonEdagGroup` onto `{ op: '+', arity: 1 }`, retiring 26 escape-path cases; `ternary` and `typeof` remain there. `neg` and `unary_plus` keep their Rust function names, so `npm run gen` leaves `nanvm-lib/tests/test/generated.rs` byte-identical, which was the acceptance check.

**Issues.** `fjs/edag/todo/op12-groups.md` is deleted; its rationale now lives in the `op12Id` JSDoc and the EDAG README. `nanvm-lib/todo/replace-unary-plus-with-number.md` and `fjs/edag/todo/ternary-conditional-node.md` no longer link to it.

**Checks.** `tsc` clean. `npm run cov`: 5378 tests, 100% lines, branches and functions. `node --test` on the final tree: 5379 passed, 0 failed. `npm run gen`: no diff. Rust is untouched, so `cargo test` was not run; the unchanged generated file is the evidence.

Changelog:
[...]

PR #1957, opened 2026-09-10
edag: add typeof as a unary operation

sergey-shandar:
Stacked on [#1954](https://github.com/functionalscript/functionalscript/pull/1954). Closes `fjs/edag/todo/typeof-operator.md`.

**EDAG.** `typeof` joins `Op1Id` and the schema, with a one-line amnesia handler. Its proof pins one tag per kind of value the evaluator can produce, including `null` as `'object'` and a closure as `'function'`. The `op1Id` JSDoc records that `typeof` is the type tag of its operand, always a fresh string, so it is the one unary operation with no identity concern.

**Why the blocker is gone.** The typeof issue held the node back until the language decided whether `typeof` is FunctionalScript syntax, so the schema would not answer that question implicitly. The stage-1 discussion, as of #1954, separates the two: the EDAG admits every pure operation, and the language spells a subset of them, decided in `spec/todo/2340-operators.md`. That spec gains a `typeof` row marked EDAG-only and a paragraph saying the syntax question is open, so nothing is decided by this PR that the spec does not say.

**Corpus.** The `typeof` group moves off `NonEdagGroup` onto a canonical `Group1`, so its 15 cases lower to `['typeof', x]` and validate against the schema like every other unary group, and `crossCheck` now covers them against host JavaScript (14 new tests). `ternary` is the one `NonEdagGroup` arm left, so `arityOf` no longer needs to ask which arm it holds. The printer's `rustName` and `op1Rust` already keyed `typeof` by its own name, so `npm run gen` leaves `nanvm-lib/tests/test/generated.rs` byte-identical. The printer proof's sample corpus swaps its escaped group from `typeof` to `ternary`, so the escape path stays pinned.

Escaped corpus cases go from 82 before #1954 to 43 after this PR: 26 `functionValue` operands, 15 `ternary`, and the two `function` cases of the unary `+` and `typeof` groups.

**Checks.** `tsc` clean. `node --test`: 5419 passed, 0 failed. `npm run gen`: no diff. Rust untouched.

Changelog:
- **BREAKING CHANGES:** `edag`: `Op1Id` gains `typeof`, so a consumer that dispatches exhaustively over the unary vocabulary (as amnesia's `Map` does) needs a `typeof` case.
- `nanvm`: `NonEdagGroup` loses its `typeof` arm; the `typeof` group is a `Group1` with `op: 'typeof'`.

---


o2alexanderfedin (approved):
Verified against base `claude/happy-tesla-72lm4z-op12` @ `c142527602e321b68bffa9442c6d5baf823ed116` (not yet in `main`, so that is the yardstick, not `merge-base`): `tsc -p .` clean, `node ./fjs/module.mjs t` 5481/0, `npm run gen` byte-clean and positive-controlled, `data.groups.length` re-derived as 27 matching its doc comment, escaped cases re-derived as 43, and mutating `typeof: o1(a => typeof a)` reddens exactly the three `typeof`/`null` assertions while dropping `'typeof'` from `op1Id` reddens 20 — the new operation's proof goes red.

PR #1967, opened 2026-09-11
ebnf, djs/tokenizer: two review leftovers of #1952

sergey-shandar:
Two Codex findings arrived on #1952 after it merged; this takes them, documentation only, against `main`.

- `fjs/ebnf/terminal/todo/ebnf-range-set.md`: the `str` task said `str('true')` is an ordered `Sequence`, the data layer's type, while `unicode-rules.md` says a `Tuple`, the front end's; an implementer following both got two contracts. It now says `Tuple`, with sequence semantics, and points at the other issue saying the same.
- `fjs/djs/tokenizer/todo/serialized-proof-expectations.md`: the Related entry said `remove-native-json.md` counts the same sites, but that issue counts direct `JSON.stringify` calls, so it counts the eight and not the eleven through the `stringifyAsTree` alias. The entry says which.

Checks: markdown only; both relative links resolve.

Changelog:
- two issues' cross-references match the tree: `str` returns a `Tuple` in both issues that describe it, and the serialized-expectations issue says which of its sites the native-JSON issue counts.

https://claude.ai/code/session_014CZRKbBvWAYpYtgVpUa3UH

---
_Generated by [Claude Code](https://claude.ai/code/session_014CZRKbBvWAYpYtgVpUa3UH)_


o2alexanderfedin (approved):
Approve. Docs-only delta, measured against merge-base `b5e62449`; both changes check out by content.

**`fjs/ebnf/terminal/todo/ebnf-range-set.md`** — `Sequence` → `Tuple` is the right correction: `Tuple = readonly Rule[]` is a real member of the front-end `Rule` union in `fjs/ebnf/types.ts`, and no `Sequence` type exists anywhere in `fjs/ebnf/` outside prose. The new parenthetical is true too: `unicode/todo/unicode-rules.md` says "`str` — an ordered `Tuple` of one-symbol terminals, one per code point, exactly as a bare `string` lowers", so the two documents now agree, and the link resolves.

**`fjs/djs/tokenizer/todo/serialized-proof-expectations.md`** — the counts hold. `fjs/djs/tokenizer/proof.f.mjs` has exactly eight `JSON.stringify` assertion sites (a ninth occurrence is the comment above the `stringify` alias) and eleven `assertEq(stringify(...))` sites; 8 + 11 = the "Nineteen assertions" the Problem section opens with. `media/json/todo/remove-native-json.md` lists `fjs/djs/tokenizer/proof.f.mjs (8)` in its expected-output row and headlines "115 call sites call `JSON.stringify` **directly**", so "that issue counts direct calls" is a fair description of it.

`node ./fjs/module.mjs t` is green at this head: 5582 pass, 0 fail.

Non-blocking, and pre-existing rather than introduced here: the same file still writes `Sequence` at two earlier spots in the `oneOf` paragraph. That frame is arguably the classical vocabulary, so it may be deliberate — worth a glance next time the file is open.

PR #1968, opened 2026-09-11
nanvm: one JavaScript reference table, and one owner for a group's tree

sergey-shandar:
Stacked on [#1963](https://github.com/functionalscript/functionalscript/pull/1963). Cleanup only: no corpus case changes, no schema changes, and `nanvm-lib/tests/test/generated.rs` is untouched.

**Three reference tables become one.** `op1Js`, `op2Js`, and `op3Js` were keyed by operation id, so `crossCheck` had to pick between them by arity and then call through a three-way chain. A group key already distinguishes `+` at one operand from `+` at two, which is why the printer's `rustName` is keyed that way; the same key works here. One table, one `reference`, and `crossCheck` applies it with a spread over the case's arguments, so the arity dispatch is gone.

**One owner for a group's proof object.** `group` and `crossCheck` each ended with the same six lines: split the cases on whether they throw, put the throwing ones under a `throw` key, assemble. That is `tree` now.

They stay two trees on purpose, and the comment now says why rather than leaving it to be rediscovered: a leaf that asserts a throw can only assert one call, and there is no `try`/`catch` available, so amnesia and the JavaScript reference cannot share a throwing leaf. Everything around that they can share, and now do.

**A silent skip becomes a list of one.** Both halves treat a missing table entry as "do not cross-check this group", which is correct for `own` — its `Object.getOwnPropertyDescriptor` read deliberately disagrees with amnesia's stricter key check — and would be wrong for anything else. Nothing said so except a comment, and a group added without a reference, or one whose key was respelled, would simply stop being checked with no failure anywhere. `referenceCoverage` asserts it in both directions. Positive-controlled three ways before committing:

| Mutation | Result |
|---|---|
| Drop the `typeof` entry | `["no JavaScript reference for","typeof"]` |
| Give `own` an entry | `["own is excluded deliberately; see the js table"]` |
| Respell `-/1` as `-/one` | `["no JavaScript reference for","-/1"]` |

**Checks.** `tsc` clean. `npm run cov`: 5542 passed, 0 failed, 100% lines, branches, and functions. The count is the previous 5541 plus `referenceCoverage`, so no existing leaf was lost or renamed. `npm run gen`: no diff.

No changelog entry: nothing exported changes, and the tables were module-private.

---


o2alexanderfedin (commented):
Delta measured against **`43856f6ff6800e69e60a51aed5d74a30bf8f5909`** (#1963's head, stacked — asserted *not* an ancestor of `origin/main` `b0176b4`). One file, `fjs/nanvm/proof.f.mjs`.

The consolidation is exact, and I checked it entry by entry rather than by size. Evaluating the old `op1Js`/`op2Js`/`op3Js` literals and the new `js` literal and mapping the two `Op12Id`s onto their `groupKey`s (`-`→`-/1`,`-/2`; `+` likewise): 27 old entries → 27 new, nothing missing, nothing extra, no key collided, and every function body is character-identical after whitespace normalisation. Importing `data` and mapping `groupKey` gives 27 groups, no duplicate key, exactly one group key absent from `js` (`own`), exactly one `js` key that is not a group key (`===`) — so both sentences the new JSDoc adds are true. The `group`/`crossCheck` merge into `tree` is leaf-for-leaf identical: diffing the suite's `fjs/nanvm/proof.f.mjs` test names at base and head gives 1843 → 1844 with the single addition `proof.referenceCoverage()` (positive-controlled with an injected fake name).

`node ./fjs/module.mjs t` 5604/0; `tsc -p .` clean; `ci` + `nanvm/update` then `git add -A && git diff --cached --exit-code` green, with a canary appended to `generated.rs` proving the updater really rewrites it. `referenceCoverage` is red in both directions: deleting `'~'` fails with `no JavaScript reference for ~` (and silently drops 24 crossCheck leaves — exactly what it now pins), adding an `own` entry fails with `own is excluded deliberately`. No `SyntaxError` in either log.

One stale doc this delta creates. `nanvm-lib/todo/replace-unary-plus-with-number.md` still names the table this PR deletes — "``fjs/nanvm/proof.f.mjs``'s `op1Js` gains `Number: a => Number(a)`" and the task "add `Number` to `op1Js`" — and the reason it gives, "a group whose id has no entry is silently **not cross-checked**", is what `referenceCoverage` just made loud. Both want rewording to `js` and to the new assertion.

o2alexanderfedin (commented):
Delta review since `0836c4b`. The old base `claude/happy-tesla-72lm4z-ternary` (#1963) has merged, so I measured this against `merge-base(origin/main, head)` = `d7068d4`.

**This is a pure rebase.** `0836c4b` was a merge commit over `43856f6`; `d914a6a` is a single commit over `d7068d4`. Normalizing index lines and hunk offsets, `diff(43856f6..0836c4b)` and `diff(d7068d4..d914a6a)` are byte-identical — 306 lines each, zero difference (positive control: appending one fake `+` line makes the comparison exit 1). The delta touches exactly one file, `fjs/nanvm/proof.f.mjs`. So the evidence from the previous pass stands unchanged and I did not re-derive it.

Re-verified at this head: `node ./fjs/module.mjs t` → 5613 pass, 0 fail; `tsc -p .` clean; `node ./fjs/module.mjs ci` plus `fjs r ./fjs/nanvm/update/module.f.mjs` leave the tracked tree clean (controlled: appending two lines to `nanvm-lib/tests/test/generated.rs` makes the same gate exit 1).

**The one open finding is still standing.** `nanvm-lib/todo/replace-unary-plus-with-number.md` is untouched by this branch and still names `op1Js` twice — "`fjs/nanvm/proof.f.mjs`'s `op1Js` gains `Number: a => Number(a)`" in the Proposal, and "add `Number` to `op1Js`" in Tasks. After this PR that table does not exist: a repo-wide grep at `d914a6a` finds `op1Js`/`op2Js`/`op3Js` in that todo file and nowhere else. The same paragraph's premise is also now false — it says "a group whose id has no entry is silently not cross-checked", which is precisely what this PR's `referenceCoverage` stops being true. A one-line edit repointing both to `js` (keyed by `groupKey`) and dropping the "silently" clause would close it.

sergey-shandar:
Finding closed in `5f88022`, and the branch is rebuilt on current `main` (`b5e0436`).

`nanvm-lib/todo/replace-unary-plus-with-number.md` named `op1Js` in two places and argued from a premise this branch falsified. Both are repointed: the Proposal asks for an entry in the `js` table "keyed by `groupKey` like every other entry", and the Tasks line says the same. The "silently not cross-checked" clause is gone — the paragraph now says a group with no entry is not cross-checked at all, that `referenceCoverage` fails on one rather than letting it go unannounced, and so the entry is required and not merely advisable. `jsOnly.throw.unusedOperation` still repoints to `own`, which is still the one key that deliberately has none.

A repo-wide grep for `op1Js`, `op2Js` and `op3Js` at this head finds nothing.

Also on this branch: the two `fjs/nanvm` overview headers listed `Op1Id`, `Op2Id` and `Op12Id` as where a group's operation comes from, which stopped being complete when `?:` landed in #1963. Both now name `Op3Id` too. That commit (`af9b2c4`) was written against the ternary branch before it merged, so it is re-homed here rather than left on merged history.

Gates at `5f88022`: `tsc -p .` clean on both 6.0.2 and CI's 7.0.2, `node ./fjs/module.mjs t` 5617 pass / 0 fail, `npm run gen` leaves the tree clean.

---

o2alexanderfedin (approved):
Delta review over my COMMENTED review at `d914a6a0`, measured against `merge-base(origin/main, head)` = `b5e04368`. `M..H` is one merge of `origin/main` (datajs only — pure base movement) plus two doc-only commits; `fjs/nanvm/proof.f.mjs` is byte-identical to `M`, so the table evidence from the previous pass stands unchanged and I did not re-derive it.

**My open finding is fixed.** `nanvm-lib/todo/replace-unary-plus-with-number.md` now asks for an entry in the `js` table "keyed by `groupKey` like every other entry", and both the prose and the checklist item name `js` rather than the deleted `op1Js`. A repository-wide grep for `op1Js|op2Js|op3Js` at this head returns nothing (exit 1, with a positive control on `referenceCoverage` returning exit 0), so no reference to the removed tables survives anywhere. The "silently not cross-checked" premise is gone too, replaced by "`referenceCoverage` fails on such a group rather than letting it go unchecked unannounced" — which matches `referenceCoverage`, and the repoint target `own` matches the sole key its second assertion excludes.

The other commit checks out as well: `Op3Id` is a real export in `fjs/edag/types.ts` and `types.ts` already has a `readonly op: Op3Id` group variant, both new `{@link}`s resolve against the existing type imports, and `Op12Id = '+' | '-'` is exactly the "two tags" the new sentence claims.

Ran from a clean worktree at this SHA: `node ./fjs/module.mjs t` (5617 pass, 0 fail), `tsc -p .` (exit 0), and the regeneration gate `node ./fjs/module.mjs ci` + `r ./fjs/nanvm/update/module.f.mjs` with a canary line appended to `generated.rs` — the canary was rewritten away and `git diff --cached --exit-code` was clean, so the gate ran and the generated file is current.

PR #1971, opened 2026-09-11
edag: amnesia takes the nodes a caller established, and nanvm drops its walker

sergey-shandar:
Stacked on [#1968](https://github.com/functionalscript/functionalscript/pull/1968), which is stacked on [#1963](https://github.com/functionalscript/functionalscript/pull/1963). `nanvm-lib/tests/test/generated.rs` is untouched and no Rust changes.

**The duplication.** The corpus's equality cases ask an identity question — `arrayByItself` is `true` where `arrayByEqualArray` is `false` — and amnesia forgets by design, so the proof carried a second evaluator: 24 lines re-implementing amnesia's `undefined`, `=>`, `[]`, `{}` and `===` handlers around a memo it could consult. Two evaluators for one node kind is exactly what this corpus exists to avoid.

**The EDAG side.** `Context` gains an optional `memo`: nodes whose values the caller has already established, consulted by identity before dispatch. It is `execution-models.md`'s §2.2, "memoize only shared nodes", in the form where the caller did the analysis. It deliberately does not make amnesia the model: nothing is remembered that the caller did not supply, and nothing crosses a call boundary, since a call is a new invocation and per-invocation is the memo's scope.

Immutability is why it takes that shape. A growing memo would need `Map#set`, which FunctionalScript does not have, so a general memoizing executor means threading state through every handler. A caller that already knows which of its nodes are shared can establish them in order instead, which is what `sharedMemo` was doing anyway.

**What goes.** The walker, and with it two things that only existed to serve it:

- The reference table's `'==='` entry. The equality cases reached the JavaScript `===` directly through the walker; they now run amnesia's `'==='` handler like every other case, so that entry has no caller.
- `jsOnly.throw.objectSpread`. It pinned that the walker refused a spread read as a property — a defect amnesia does not have, since it implements spread.

**Positive controls.** Three mutations of the new behaviour, each restored after:

| Mutation | Result |
|---|---|
| Ignore the memo | amnesia 1 failure, corpus 5 |
| Let the memo cross a call boundary | amnesia 9 failures |
| Match by structure instead of identity | corpus 7 failures |

**Checks.** `tsc` clean. `npm run cov`: 5542 passed, 0 failed, 100% lines, branches and functions. `npm run gen`: no diff. The count is unchanged from #1968 because the new `established` section replaces the deleted spread pin one for one.

**Not in this PR.** The `eq` section is still its own authoring shape: `Eq`, `EqCase`, `LoweredEq`, `lowerEq` and the printer's `eqFn` remain, and folding them into an ordinary `'==='` group is `fjs/nanvm/todo/unify-eq-into-a-group.md`. That fold changes `generated.rs` and the hand-written `harness.rs`, so it belongs on its own. This PR removes the duplicated *evaluator*, which was the part that needed something from EDAG.

Changelog:
- `edag`: `amnesia`'s `Context` gains an optional `memo` of already-established nodes, consulted by identity. Existing callers are unaffected — absent `memo` is the previous behaviour exactly.

---

### DIFF
--- a/fjs/edag/amnesia/README.md
+++ b/fjs/edag/amnesia/README.md
@@ -49,15 +49,6 @@ vm(context)(['===', shared, shared])   // false — the model says true
 ```
 
 One node reached from two operand positions is one value; here it is two
-— unless the caller says otherwise. `Context`'s `memo` carries nodes whose
-values are already established, consulted by identity before anything is
-computed, which is [execution-models](../execution-models.md)'s §2.2 with the
-analysis done by the caller. It does not make this the model: nothing is
-remembered that the caller did not supply, and nothing crosses a call
-boundary. It is enough for a caller that knows which of its nodes are shared,
-which is how `fjs/nanvm`'s equality cases ask an identity question of an
-evaluator that otherwise forgets.
-
 arrays, so identity comparisons silently give the wrong answer. The cost
 compounds in the same way. A chain of 22 shared additions,
 `['+', x, x]` over the same `x`, is 23 distinct nodes:
@@ -72,6 +63,24 @@ Exponential in depth, and the leaf alone is visited 4,194,304 times.
 same reason, which is where the word for it comes from. The models that do
 preserve identity, and what each is for, are in
 [execution-models.md](../execution-models.md).
+
+### ... unless the caller says otherwise
+
+`Context`'s `memo` carries nodes whose values are already established,
+consulted by identity before anything is computed — which is
+[execution-models](../execution-models.md)'s §2.2 with the analysis done by
+the caller rather than by a pass here. It does not make this the model:
+nothing is remembered that the caller did not supply, nothing is added while
+walking, and nothing crosses a call boundary, since a call is a new
+invocation. It is enough for a caller that knows which of its own nodes are
+shared, which is how `fjs/nanvm`'s `'==='` cases ask an identity question of
+an evaluator that otherwise forgets.
+
+```js
+const shared = ['[]', [1, 2]]
+const memo = [[shared, vm(context)(shared)]]
+vm({ ...context, memo })(['===', shared, shared])   // true
+```
 
 ### It hands out the host
 
--- a/fjs/edag/amnesia/README.md
+++ b/fjs/edag/amnesia/README.md
@@ -49,6 +49,15 @@ vm(context)(['===', shared, shared])   // false — the model says true
 ```
 
 One node reached from two operand positions is one value; here it is two
+— unless the caller says otherwise. `Context`'s `memo` carries nodes whose
+values are already established, consulted by identity before anything is
+computed, which is [execution-models](../execution-models.md)'s §2.2 with the
+analysis done by the caller. It does not make this the model: nothing is
+remembered that the caller did not supply, and nothing crosses a call
+boundary. It is enough for a caller that knows which of its nodes are shared,
+which is how `fjs/nanvm`'s equality cases ask an identity question of an
+evaluator that otherwise forgets.
+
 arrays, so identity comparisons silently give the wrong answer. The cost
 compounds in the same way. A chain of 22 shared additions,
 `['+', x, x]` over the same `x`, is 23 distinct nodes:
--- a/fjs/edag/amnesia/module.f.mjs
+++ b/fjs/edag/amnesia/module.f.mjs
@@ -236,6 +236,10 @@ const map = {
     '===': o2((a, b) => a === b),
     '=>': (x, [, frameExp, body]) => {
         const frame = vm(x)(frameExp)
+        // The body is a new invocation, so it starts with nothing
+        // established: the enclosing `memo` does not cross the boundary, the
+        // same way the model's per-invocation memo does not. The captured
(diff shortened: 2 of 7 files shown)


o2alexanderfedin (commented):
Delta measured against **`d914a6a081e9093ea42bdac7bb62a8d6cf5f0531`** (#1968's head), asserted not an ancestor of `origin/main`.

The mechanism holds. `node ./fjs/module.mjs t` 5613/0, `tsc -p .` clean, `ci` + the `nanvm/update` runner leave `git diff --cached --exit-code` empty (canary: a line appended to `generated.rs` is rewritten away by the runner, so the second step really ran). Three mutations, each restored and re-greened: memo ignored → 6 red (`amnesia.established` + `eq.arrayByItself`/`stringArrayByItself`/`objectByItself`, `lambda`, `nestedSharing`); matched by structure instead of identity → 7 red, exactly the cases that must stay `false`; memo carried across the `=>` boundary → `established` red alone. No `SyntaxError` in any log. A `^export` sweep base→head shows no name appearing or disappearing, so nothing needs a `**BREAKING CHANGES:**` line. Wrong-memo probing: a stranger node is simply not found; a garbage or `undefined` value comes back verbatim (the `find`-result check is the right one); out-of-order is silently wrong — but `lowerEq` refuses a forward reference (`['unknown shared value','base']`), so the one caller cannot spell it. The printer still pins `objectSpread`.

Three prose items:

1. **`fjs/edag/amnesia/README.md`** — the new paragraph was inserted *mid-sentence*. "…here it is two — unless the caller says otherwise." then, after the inserted block, an orphan paragraph opening "arrays, so identity comparisons silently give the wrong answer." The original read "here it is two arrays, so…".
2. **`fjs/edag/amnesia/module.f.mjs`** header still says "every incoming edge evaluates its target again" without qualification — the one claim `memo` falsifies, and the one the README and `types.ts` both took care to qualify.
3. **`fjs/nanvm/todo/corpus-as-conformance-vectors.md`** names "the proof's inline evaluator" three times (body, task list, Related). It existed at the base; it does not at head.

sergey-shandar:
All three prose items fixed in `df32c96`; the branch is rebuilt on #1968's new head so it is no longer stale under its base.

**The split sentence.** You are right that the paragraph went in mid-sentence. "here it is two arrays, so identity comparisons silently give the wrong answer" is whole again, and the `memo` material has its own section, `### ... unless the caller says otherwise`, placed after the passage it qualifies rather than inside it. It carries a runnable example:

```js
const shared = ['[]', [1, 2]]
const memo = [[shared, vm(context)(shared)]]
vm({ ...context, memo })(['===', shared, shared])   // true
```

Checked at this head: `true` with the memo, `false` without.

**The unqualified header claim.** `module.f.mjs` now reads "It remembers no node values **of its own** — every incoming edge evaluates its target again", and says a caller may hand over established nodes as `Context`'s `memo`, "the one thing this walk does not recompute".

**The dead walker.** `fjs/nanvm/todo/corpus-as-conformance-vectors.md` named "the proof's inline evaluator" in the body, the task list and the Related section. All three now name `amnesia`, which is what `interpret-edag` will actually replace. The body adds that the inline evaluator is already gone, so what is left to migrate is the evaluator itself and not a second one beside it.

Gates at `2e04370`: `tsc -p .` clean on 6.0.2 and on CI's 7.0.2, `node ./fjs/module.mjs t` 5617 pass / 0 fail, `npm run gen` leaves the tree clean.

---

o2alexanderfedin (approved):
Delta review of `e7269bcb..2e043701`. Stacked: the delta is measured against base SHA `5f8802210b21246a18c0ce08c9a93f470fcd143d` (PR #1968's head), which I asserted is **not** an ancestor of `origin/main` (`merge-base --is-ancestor` exits 1). Since my last pass the PR added exactly one commit of its own, `df32c966`, plus a merge of base movement — `diff(df32c966, HEAD)` touches only base-side files, so the merge is clean in both directions.

All three open findings are fixed.

1. **README sentence spliced mid-sentence** — repaired. Flattened, the README now reads `here it is two arrays, so identity comparisons silently give the wrong answer.` as one sentence, and the `memo` paragraph has moved into its own `### ... unless the caller says otherwise` section after it. Its new example runs verbatim: with `memo = [[shared, vm(context)(shared)]]`, `vm({ ...context, memo })(['===', shared, shared])` is `true`, and the unmemoized block above it is `false` — both comments match.

2. **Module header's unqualified claim** — fixed. It now reads "remembers no node values **of its own**" and names `memo` as the one thing the walk does not recompute, which matches `vm`'s `memo?.find(([n]) => n === e)` before dispatch.

3. **`corpus-as-conformance-vectors.md` naming a removed evaluator** — fixed. Flattened, "inline evaluator" survives exactly once, in the sentence saying it is already gone; the prose, the task, and the Related entry now name `amnesia`, which is what `fjs/nanvm/proof.f.mjs` actually uses.

Ran at the head: `node ./fjs/module.mjs t` (5617 pass, 0 fail) and `tsc -p .` (clean). The evidence established at `e7269bcb` carries forward — `df32c966` changes prose only, no code.

PR #1976, opened 2026-09-11
nanvm: four relational groups from one table of rows

sergey-shandar:
Stacked on [#1974](https://github.com/functionalscript/functionalscript/pull/1974). Closes `fjs/nanvm/todo/comparison-cases-table.md`.

**The duplication.** `<`, `<=`, `>` and `>=` spelled the same 50 argument pairs four times. I checked rather than assumed: all four lists are 50 cases long, their `args` sequences are byte-identical, and every one of the 200 names splits cleanly into a left half, the relation's word, and a right half. Nothing irregular anywhere.

The four are pinned to each other by invariants no code stated — `>` is the reversed `<`, `>=` the reversed `<=`, each loose relation differs from its strict one only at equality — so drift between the lists was invisible. And because the printer walks the same cases, a pair added to one list and missed in another quietly narrowed the Rust suite too.

**The shape.** One table of rows, each carrying its two name halves, its arguments, and its four answers side by side, plus four one-line derivations that pick a column and splice in the relation word. This is the move `numberCoercionCases` already makes for `+n` and `-n`, with its rationale written in the file, and deliberately no cleverer than it: the four answers stay four independent booleans rather than one derived ordering, because a relation that does *not* follow the rule is exactly what these cases exist to catch.

The table makes an invariant legible that nothing checked before. 14 rows are `false` in all four columns — precisely the ones `NaN` reaches after coercion — and on the other 36, `le` is `!gt` and `ge` is `!lt`, with zero violations.

The comments the four lists carried survive, restated once against the columns they now sit beside: equal operands and equal infinities are loose-only, `NaN` is `false` everywhere, and the relations compare a `Number` against a `BigInt` directly instead of throwing.

**How the rows got there.** Transcribed, not retyped. A script parsed the four lists, asserted they agreed on arguments and name stems, and emitted the rows with each argument pair's source text kept verbatim, so `-0`, `5n` and `functionValue` could not be mis-spelled on the way across.

**Nothing changed.** The corpus serializes identically before and after — 28 groups, 819 cases, with `-0`, `NaN` and bigints spelled out so none of them compares equal by accident — and `generated.rs` is byte-identical.

Positive controls, each restored afterwards:

- flipping one boolean in one column turns the suite red (1 failure) and moves exactly one line of `generated.rs`;
- respelling one relation's word moves 50 lines of `generated.rs`.

**Checks.** `tsc -p .` clean on 7.0.2 and 6.0.2. `node ./fjs/module.mjs t`: 5704 passed, 0 failed. `npm run cov`: 100% lines, branches and functions. `npm run gen`: stable. `cargo test`: 185 + 1 + 5 passed, 0 failed. `cargo fmt --check`: clean.

`cargo clippy -- -D warnings` fails identically on the base branch: four pre-existing errors, untouched here.

Changelog: no public API change — `comparisonCases` and `relationCases` are module-private, and every exported name and every case keeps its spelling.

---


sergey-shandar:
Closing as a duplicate. [#1977](https://github.com/functionalscript/functionalscript/pull/1977) was opened on this same head branch and supersedes this one; further commits here update that PR.

One difference worth knowing when reviewing: this PR was based on `claude/happy-tesla-72lm4z-unify`, so its delta was the three files of the comparison-table change alone. [#1977](https://github.com/functionalscript/functionalscript/pull/1977) is based on `main`, so it carries all ten commits of the stack — [#1968](https://github.com/functionalscript/functionalscript/pull/1968), [#1971](https://github.com/functionalscript/functionalscript/pull/1971) and [#1974](https://github.com/functionalscript/functionalscript/pull/1974) as well as this change, across 16 files.

---

PR #1977, opened 2026-09-11
nanvm: four relational groups from one table of rows

sergey-shandar:
Stacked on [#1974](https://github.com/functionalscript/functionalscript/pull/1974). Closes `fjs/nanvm/todo/comparison-cases-table.md`.

This PR was opened against `main` and so carried the whole stack; it is now based on `claude/happy-tesla-72lm4z-unify`, so its delta is three files and one commit. The title and body below describe that delta. The `eq` unification, the shared-node lowering, the `memo`, and the Rust `strict_eq` helper are [#1971](https://github.com/functionalscript/functionalscript/pull/1971) and [#1974](https://github.com/functionalscript/functionalscript/pull/1974), both approved, and are not re-proposed here.

**The duplication.** `<`, `<=`, `>` and `>=` spelled the same 50 argument pairs four times. I checked rather than assumed: all four lists are 50 cases long, their `args` sequences are byte-identical, and every one of the 200 names splits cleanly into a left half, the relation's word, and a right half. Nothing irregular anywhere.

The four are pinned to each other by invariants no code stated — `>` is the reversed `<`, `>=` the reversed `<=`, each loose relation differs from its strict one only at equality — so drift between the lists was invisible. And because the printer walks the same cases, a pair added to one list and missed in another quietly narrowed the Rust suite too.

**The shape.** One table of rows, each carrying its two name halves, its arguments, and its four answers side by side, plus four one-line derivations that pick a column and splice in the relation's word. This is the move `numberCoercionCases` already makes for `+n` and `-n`, with its rationale written in the file, and deliberately no cleverer than it: the four answers stay four independent booleans rather than one derived ordering, because a relation that does *not* follow the rule is exactly what these cases exist to catch.

The table makes an invariant legible that nothing checked before. 14 rows are `false` in all four columns, precisely the ones `NaN` reaches after coercion, and on the other 36 `le` is `!gt` and `ge` is `!lt`, with zero violations.

The comments the four lists carried survive, restated once against the columns they now sit beside: equal operands and equal infinities are loose-only, `NaN` is `false` everywhere, and the relations compare a `Number` against a `BigInt` directly instead of throwing.

**How the rows got there.** Transcribed, not retyped. A script parsed the four lists, asserted they agreed on arguments and name stems, and emitted the rows with each argument pair's source text kept verbatim, so `-0`, `5n` and `functionValue` could not be mis-spelled on the way across.

**Nothing changed.** The corpus serializes identically before and after — 28 groups, 819 cases, with `-0`, `NaN` and bigints spelled out so none of them compares equal by accident — and `generated.rs` is byte-identical.

Positive controls, each restored afterwards:

- flipping one boolean in one column turns the suite red (1 failure) and moves exactly one line of `generated.rs`;
- respelling one relation's word moves 50 lines of `generated.rs`.

**Checks.** `tsc -p .` clean on the 7.0.2 CI pins and on 6.0.2. `node ./fjs/module.mjs t`: 5709 passed, 0 failed. `npm run cov`: 100% lines, branches and functions. `npm run gen`: stable. `cargo test`: 185 + 1 + 5 passed, 0 failed. `cargo fmt --check`: clean.

`cargo clippy -- -D warnings` fails identically on the base branch: four pre-existing errors, untouched here.

Changelog:
- No public API change. `comparisonCases` and `relationCases` are module-private, and a sweep of `^export` across `fjs/nanvm/module.f.mjs`, `types.ts`, `proof.f.mjs` and `rust/module.f.mjs` shows base and head identical. Every case keeps its name, arguments and expectation.


sergey-shandar:
Both Codex findings addressed. Replying here rather than on the threads: the GitHub API that returns inline review threads has been refusing every call for the last half hour with a user-level rate-limit error, so I could not get the comment ids to reply in place.

Both findings were raised when this PR was based on `main` and so carried all ten commits of the stack. It is now based on `claude/happy-tesla-72lm4z-unify`, which is what settles them.

**P1, declare the breaking corpus API change.** `caseExp`'s new leading `shared` argument and the removal of `lowerEq` and the `Eq` types are [#1974](https://github.com/functionalscript/functionalscript/pull/1974)'s change, and that PR's description declares them under `**BREAKING CHANGES:**` with a migration line. They are no longer in this PR's delta.

This PR's own delta changes no public name. I swept `^export` across `fjs/nanvm/module.f.mjs`, `types.ts`, `proof.f.mjs` and `rust/module.f.mjs` between base and head: identical in all four. `comparisonCases` and `relationCases` are module-private. The body now carries a `Changelog:` line saying exactly that, so release tooling has a declaration to read rather than a missing one.

**P2, split out the comparison-table refactor.** Agreed, and it is split: the comparison table is this PR alone, one commit over three files, and the equality and shared-node work is [#1971](https://github.com/functionalscript/functionalscript/pull/1971) and [#1974](https://github.com/functionalscript/functionalscript/pull/1974), both approved. The two were only ever combined because this PR was opened against `main` on a branch that already had the stack beneath it.

The body was also still describing the whole stack, which was wrong for the retargeted base. It now describes this delta, and says where the rest lives.

---

o2alexanderfedin (approved):
Approving. Measured against `merge-base(origin/main, head)` = `91834450` — the stack has merged, so that is `origin/main` and the delta is this one commit's three files.

**The folding is a bijection.** I imported `data` at base and at head and serialized all 28 groups case by case — name, args (with `-0`, `NaN`, bigints and `ref` nodes spelled distinctly) and `expected`. The two dumps are identical, 819 cases on both sides, and `generated.rs` is the same blob (`eb83dbc`) at base and head. Positive control: flipping one boolean in `comparisonCases` makes the dumps differ on exactly one line.

**The JSDoc's numbers hold.** Reconstructing the table from the four groups: 50 rows each, `args` byte-identical across all four, and every one of the 200 names splitting into the same `left`/`right` halves around the relation's word. 14 rows are `false` in all four columns; on the other 36, `le === !gt` and `ge === !lt` with zero violations. `data.groups.length // 28` in the module `@example` is still 28, and all six comment themes the four lists carried survive against the columns.

**`referenceCoverage` still bites both ways.** Deleting the `'<'` entry from `js` → red with `no JavaScript reference for,<`, and the run is exactly 50 leaves short, so the loss is reported rather than silent. Adding an `own` entry → red with `own is excluded deliberately`. Both restored; neither log contains `SyntaxError`.

**Gates.** `node ./fjs/module.mjs t`: 5709 passed, 0 failed, exit 0. `tsc -p .` (7.0.2): exit 0. Regeneration with a canary appended to `generated.rs`: `fjs ci` plus `fjs r ./fjs/nanvm/update/module.f.mjs` overwrote the canary and left `git add -A && git diff --cached --exit-code` clean.

**No published break.** `^export` sweep over every `.mjs`/`.ts` at both SHAs: 2090 lines, identical (positive-controlled). `comparisonCases` and `relationCases` are module-private.

No surviving reference to the deleted `comparison-cases-table.md` anywhere at head.

PR #1990, opened 2026-09-11
edag: 28 schema pins moved to where they are checked

sergey-shandar:
Against `main`, now that [#1986](https://github.com/functionalscript/functionalscript/pull/1986) and the rest of the stack have merged. Two commits, four files. The `fjs/edag/proof.f.mjs` task of `todo/inert-type-level-proofs.md`.

**What was wrong.** `proof.f.mjs`'s `consistency` entry was a body of nothing but typedefs, so none of them bound to a statement and all 28 `Assert<Check<…>>` pins were green whatever they claimed. These are the pins `types.ts`'s own header says keep it and the rtti schema from drifting, and they would not have noticed if it had.

**The move.** All 28 go verbatim to module scope in `types.ts`, where a type alias resolves on sight.

**The thing I checked before moving anything.** Every pin is `Check<SomeType, typeof someSchemaValue>`, pairing a type in `types.ts` with a value in `module.f.mjs`, so `types.ts` has to reach back into the module it describes. That is a cycle, and `types.ts` is the published type API that must hold no runtime implementation.

An `import type` settles both: the cycle is type-only and no runtime import is emitted, so the file stays type source as §3.2 requires. I probed it with one true pin and one deliberately false one before touching the real thing, and only the false one reported TS2344.

**Proved bitten, not merely moved.** All 28 falsified at once, each claim replaced by `Assert<Equal<1, 2>>` with the name kept, then one `tsc` run: 28 flagged, none missed. The repository sweep agrees independently — single-line inert falls from 62 to 34, with `fjs/edag/proof.f.mjs` off the list entirely.

| | before | after |
|---|--:|--:|
| inert, single-line | 62 | 34 |
| inert, total | 70 | 42 |
| files with inert pins | 10 | 9 |

**What goes with them.** The `consistency` entry, since it would now assert nothing, and the 28 `@import` names it alone used. The suite drops one test relative to its base: that entry ran and counted as passing while checking nothing, which is the defect in miniature.

**The containment comment is finished here too.** The three chain-state pins cover the three arms the wider state adds and the two containments; the shared segment is unpinned in either direction, since a change there lands on both sides of the `Exclude` and cancels. With `_OptionLambda` now in this file, that is the pin covering it, and I measured both directions at this head: an arm added to `OptionLambda` gives 2 errors and an arm removed gives 2.

**The second commit** repoints `fjs/AGENTS.md` §1.4, which described `consistency` in the present tense as its worked example of the inert form. `fjs/effects/proof.f.mjs`'s `signatures` carries it now; `consistency` stays as the past extreme with a pointer to where its pins went.

**Checks.** `tsc -p .` clean on the 7.0.2 CI pins and on 6.0.2. `node ./fjs/module.mjs t`: 5751 passed, 0 failed. `npm run gen`: stable.

Changelog: no public API change. The pins are `_`-prefixed private aliases, no exported name changes, and the schema is untouched.

`fjs/edag/amnesia/proof.f.mjs`'s 8 remain and finish the directory. I left them for their own change: some pin local inference rather than a published type, so whether each has a `types.ts` home is a judgment worth reviewing on its own.

### DIFF
--- a/fjs/AGENTS.md
+++ b/fjs/AGENTS.md
@@ -111,10 +111,12 @@ the last thing in a function body compiles. Put any statement after it and the
 same line is TS2344.
 
 So a proof entry that *ends* with its typedefs checks nothing from there on.
-That is what `fjs/edag/proof.f.mjs`'s `consistency` is — a body of nothing but
-typedefs, all 28 of its `Assert<Check<…>>` pins green whatever they claim —
-and `fjs/effects/proof.f.mjs`'s `signatures`, whose last eight go the same
-way. Where a typedef is followed by an `assert` call, as in
+`fjs/effects/proof.f.mjs`'s `signatures` is the standing example: its last
+eight `Assert<Equal<…>>` have no statement after them and are green whatever
+they claim. `fjs/edag/proof.f.mjs`'s `consistency` was the extreme of it — a
+body of nothing but typedefs, so all 28 of its `Assert<Check<…>>` pins were
+inert — which is why those now sit at module scope in `fjs/edag/types.ts`.
+Where a typedef is followed by an `assert` call, as in
 `fjs/ebnf/ll1/proof.f.mjs`'s `constParameter`, it is checked and does its job.
 
 **Prefer module scope in a `.ts` file**, where a type alias is resolved
--- a/fjs/edag/proof.f.mjs
+++ b/fjs/edag/proof.f.mjs
@@ -10,14 +10,10 @@
  * which `comma` is now the sole route to; it pins the operand array's
  * element schema, and claims nothing about what a `,` means.
  *
- * @import { Assert } from '../asserts/types.ts'
  * @import { ValidationError } from '../rtti/common/types.ts'
- * @import { Check, Check3, Unknown } from '../rtti/ts/types.ts'
+ * @import { Unknown } from '../rtti/ts/types.ts'
  * @import { StringMap } from '../types/object/types.ts'
  * @import {
- *  _exp,
- *  _optionLambda,
- *  _optionPropertyLambda,
  *  array,
  *  call,
  *  comma,
@@ -40,33 +36,8 @@
  * } from './module.f.mjs'
  * @import {
  *  Array,
- *  Call,
- *  Comma,
- *  Dot,
  *  Exp,
- *  Exps,
- *  Items,
- *  NumberCast,
  *  Object,
- *  Op0,
- *  Op0Id,
- *  Op1,
- *  Op1Id,
- *  Op12,
- *  Op12Id,
- *  Op2,
- *  Op2Id,
- *  Op3,
- *  Op3Id,
- *  OptionCall,
- *  OptionDot,
- *  OptionLambda,
- *  OptionPropertyLambda,
- *  Primitive,
- *  Properties,
- *  Property,
- *  PropertyLambda,
- *  Spread,
  * } from './types.ts'
  */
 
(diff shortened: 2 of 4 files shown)


o2alexanderfedin (commented):
Stacked: delta measured against `180925622e1ed2863e227ab24b10a7387bc4f2f6` (#1986's head, read live), asserted **not** an ancestor of `origin/main`.

**The claim holds, measured both ways.** Falsifying all 28 pins at once (first type argument → `never`, names kept):

| | `tsc -p .` |
|---|---|
| base, 28 falsified in `proof.f.mjs` | **exit 0, zero errors** |
| base, same + one `void 0` after them | exit 1, 28 errors — positive control, tsc *was* looking |
| head, 28 falsified in `types.ts` | **exit 1, 28 errors**, none outside `types.ts` |

28 distinct alias names in the head log — `_ExpAssert … _Op3`, one error each, no pin missed and none double-counted. That is the count derived by measurement, and it matches the 28 claimed.

**#1985's open finding.** Dropping `readonly['|.', Index]` from `OptionLambda` (schema untouched): base → **no error anywhere in `fjs/edag/types.ts`**, the `Exclude` gate silent as reported. Head → `TS2344` at `_OptionLambda` **and** `_OptionPropertyLambda`. The gate itself is still one-sided, and the new comment now says so outright and names `_OptionLambda` as the pin that covers it — accurate at this head.

**Other checks.** `tsc -p .` clean base and head. Suite 5729 → 5728, exactly the one empty `consistency` test, 0 failures. `npm run gen` + `git add -A && git diff --cached --exit-code` clean, canary-confirmed live (appending to `generated.rs` is overwritten by regeneration). No `^export` line changes in the delta, so no `**BREAKING CHANGES:**` owed. `Check3` for "the three schemas that are thunks behind a `Phantom`" is exactly `_exp`, `_optionLambda`, `_optionPropertyLambda`.

**One thing to fix.** `fjs/AGENTS.md` §1.4 still reads, present tense:

> That is what `fjs/edag/proof.f.mjs`'s `consistency` is — a body of nothing but typedefs, all 28 of its `Assert<Check<…>>` pins green whatever they claim

This PR deletes that entry. §1.4's worked example of the inert form now names something that does not exist. `types.ts` was moved to past tense in the same delta; §1.4 wants the same treatment (`fjs/effects/proof.f.mjs`'s `signatures` still stands and can carry the example alone).

sergey-shandar:
Fixed in `c5d6172`. You are right, and it is the kind of staleness this whole branch exists to remove: §1.4's worked example of the inert form named an entry this branch deletes.

`fjs/effects/proof.f.mjs`'s `signatures` is the standing example now — its last eight assertions still end their block and are still green whatever they claim. `consistency` stays as the past extreme, a body of nothing but typedefs, with a pointer to where its 28 pins went, since that is what the rest of the branch is about.

I swept the rest rather than fixing only the line you cited. `fjs/edag/types.ts` was already past tense in both places, and the three mentions in `todo/inert-type-level-proofs.md` describe what §3.2 used to say or record the removal, so this was the only present-tense claim about something that no longer exists.

Thanks for re-deriving the 28 rather than taking the count. Your base-versus-head table is the check I most wanted an independent run of, and the `void 0` control — 28 errors appearing the moment a statement follows the falsified block — is a cleaner demonstration of the binding rule than the scratch-file matrix I put in [#1986](https://github.com/functionalscript/functionalscript/pull/1986).

Re-verified at this head: `tsc -p .` clean on 7.0.2 and 6.0.2, 5728 tests passing, `npm run gen` stable, no new lines over 80 columns.

---

o2alexanderfedin (approved):
Approving. Delta measured against `6005ed30cfcc19f7d5575c91c147a47e87079232`, PR #1986's head (asserted not an ancestor of `origin/main`). The `dirty` mergeable state is the base having moved under it, not a defect.

**My open finding is fixed.** `fjs/AGENTS.md` §1.4 no longer names `consistency` in the present tense: it now reads "`fjs/edag/proof.f.mjs`'s `consistency` **was** the extreme of it … which is why those now sit at module scope in `fjs/edag/types.ts`", with `effects`'s `signatures` promoted to the standing example. Correct for the tree this PR leaves.

**The 28, re-derived by falsification at the new base.** All 28 first type arguments → `never`, names kept: at the base `tsc -p .` exits **0, zero errors**; the middle control (same falsified file plus one `void 0` after the pins) exits **1 with 28 TS2344**, all in `proof.f.mjs` — so the base's zero is inertness, not a skipped file. At the head the same falsification exits **1 with 28 errors, all in `types.ts`**, and the 28 error lines carry 28 distinct alias names.

**The gain direction, now measured.** Adding a fresh arm `readonly['|~', Exp]` to `OptionLambda` at this head gives **TS2344 at `_OptionLambda` and `_OptionPropertyLambda`** (plus a collateral TS7030 in `amnesia/module.f.mjs`). The removal of `['|.', Index]` gives TS2344 at the same two. So the pins close **both** directions, and `types.ts`'s "which either change makes disagree" is measured true. The three chain pins stayed green under both mutations, as their comment says.

Also run: `tsc -p .` exit 0 and `node ./fjs/module.mjs t` exit 0 (5746 pass) at the head — exactly one leaf fewer than the base's 5747, the inert `consistency` entry. `^export` sweep base→head: no difference. No generated path is touched, so no regeneration gate applies. `todo/inert-type-level-proofs.md`'s table sums to 42 across nine files, matching 70 − 28.

PR #2004, opened 2026-09-11
AGENTS: an assertion owes a mutation, not only a falsification

sergey-shandar:
Documentation only. Two edits, both consequences of #1993 and #1995.

## The gap

§1.4 already asks that a new `Assert` be falsified once and seen to fail. That check answers one question: does the compiler evaluate the line at all. #1995 showed it is not the whole question.

`fjs/rtti/ts/types.ts`'s `_UnionKeepsBranchCorrelation` passed it. It resolved, it held a true claim, and falsifying the claim reported it correctly. It still went green when the mechanism its comment credited it with was removed from `TupleTs`, because its witness was `readonly [1, true]`, a value neither the correct answer nor the broken one admits.

Two different questions, and an assertion can pass the first while failing the second:

| check | asks |
| --- | --- |
| falsify the assertion | is the line evaluated |
| break the mechanism it names | would the line notice |

## What §1.4 gains

Where an assertion's comment credits it with a mechanism, break that mechanism once and read which line reports. The named one, or nothing, which means the assertion is about something other than its prose says. Either the assertion or the comment is then wrong, and a row claiming more than it holds is worse than one claiming nothing, since a reader stops looking. The three mutations behind `_RestTuple`'s header are the worked example.

The scope is written into the rule on purpose: this is for an assertion you write or touch, not a standing audit. There is no enumeration of "mechanisms" to sweep, so a repository-wide pass would be judgement per assertion with no definition of done, which is the machinery root `AGENTS.md` §6 says to leave unbuilt. Saying so in the section makes the absence read as a decision rather than an oversight.

## What `REVIEWING.md` gains

Its type-level section told reviewers to ask for the assertion "in the proof". That predates #1993 and now reads as if a proof were the default home, when it is only right for the local inference §1.4 carves out. It now asks for module scope, names the exception and its statement-after requirement, and points at the two checks a reviewer can ask the author to have run.

## Checks

No source changes. `tsc -p .` exits 0, `fjs test` passes 5780 of 5780, and `npm run gen` produces no diff.

https://claude.ai/code/session_01NXCZLTT4sRcrSc95AJviTk

---
_Generated by [Claude Code](https://claude.ai/code/session_01NXCZLTT4sRcrSc95AJviTk)_

### DIFF
--- a/doc/REVIEWING.md
+++ b/doc/REVIEWING.md
@@ -99,9 +99,15 @@ observable — the runtime already refuses the invalid value, and the valid one
 was valid before — so do not ask for one; the simplest type that holds is the
 right one whether or not a walk is near the limit
 ([DESIGN.md §1](./DESIGN.md#1-simplicity-first)). Ask for the simplest type that states the
-contract, an `Assert<Equal<…>>` in the proof where the inference matters
-([fjs/AGENTS.md §1.4](../fjs/AGENTS.md#14-assert-type-level-facts-with-assertequal)),
-and a `todo/` for anything tighter.
+contract, an `Assert<Equal<…>>` where the claim matters, and a `todo/` for
+anything tighter. That assertion belongs at module scope in a `types.ts`,
+where an alias resolves whether or not anything follows it; ask for one in a
+proof only for a claim about a local inference, which is the case
+[fjs/AGENTS.md §1.4](../fjs/AGENTS.md#14-assert-type-level-facts-with-assertequal)
+carves out, and then only with a statement after it. That section also says
+what to ask of the assertion itself: falsified once to show it is evaluated,
+and, where its comment credits it with a mechanism, that mechanism broken once
+to show it reports.
 
 ## Bots
 
--- a/fjs/AGENTS.md
+++ b/fjs/AGENTS.md
@@ -140,6 +140,32 @@ write** — falsify it once, see it fail, restore it. A form copied from
 somewhere that works is not evidence, since what decides it is what follows
 the line, not the line.
 
+**That check is the first of two, and it is the weaker one.** Falsifying the
+claim asks whether the compiler evaluates the line at all. It says nothing
+about whether the line would notice the thing it is about breaking, and an
+assertion can pass the first test and fail the second: its claim resolves, its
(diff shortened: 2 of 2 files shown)


o2alexanderfedin (approved):
Approved. Delta measured against `merge-base(origin/main, head)` = `af1e560c`.

I tested the rule rather than read it. All in `fjs/rtti/ts/types.ts`, `tsc -p .` exit codes asserted, each mutation confirmed applied by a non-empty `git diff --stat`, logs free of `SyntaxError`:

- Mutation 3 (`T extends Tuple` → `[T] extends [Tuple]` in `TupleTs`): exit 1, TS2344 at `_UnionKeepsBranchCorrelation` and `_UnionGuardsEachBranch`, nothing else.
- Same mutation, witness restored to the historical `readonly [1, true]` (5cd63146): exit 1, only `_UnionGuardsEachBranch` — the old row stays green.
- Clean implementation, that old row's claim flipped: exit 1, TS2344 at its own line.

So the pair the text describes is real: a row that passes §1.4's falsification and that no implementation mutation reddens, beside one that reddens. Mutations 1 and 2 hold too — removing the length guard reports `_VariadicPrefixRejectsMixedPrefix` alone, the peel's outermost fallback → `readonly []` reports `_OptionalMember` alone; none of the three moves `_RestTuple` or `_NonFixedLength`.

Two notes, neither blocking:

- "break that mechanism once" under-determines the mutation. In `fjs/edag/types.ts`, adding an arm to `OptionLambda` leaves `_OptionPropertyAdds` and both containment rows green — only `_OptionLambda`/`_OptionPropertyLambda` report — while dropping one of the three named arms would report. Same mechanism, opposite verdicts from the choice.
- The trigger is an assertion's *comment* and the instrument is `tsc`, so a runtime proof credited by its *name* — `hmac`'s `throw.oddBlock` — sits outside the wording, though the diagnostic transposes: removing the whole-bytes guard reports `oddBlock`, alone.

No contradiction with §1.4's module-scope rule; the `REVIEWING.md` paragraph restates its carve-out. `#14-…` and `#6-external-tools` resolve; link sweep at base and head identical (10 pre-existing each, positive-controlled). Suite 5780/0, `tsc` 0.

PR #2006, opened 2026-09-11
git/walk: the width check reaches tryEntries and tryEntry, so they declare it

sergey-shandar:
Picks up the one review comment from #1989 that its merge did not carry, and
declares the break that went in with it.

#1989 merged at `70c7bca6`, one commit before the fix for
[this P1 on `peel`'s width check](https://github.com/functionalscript/functionalscript/pull/1989#discussion_r3992617211).
Two things are therefore in `main` unfinished.

**The `@throws` was on `peel` only.** `peel` asserts the id it is given is
`oidBytes` wide. `tryEntries` and `tryEntry` hand the caller's id straight to
it and inherit the throw, and neither said so. Both now carry an `@throws`.
That is the whole diff: seven lines of JSDoc, no code change.

Which calls actually throw, measured against an 8-bit id at `oidBytes` 20
rather than reasoned about:

```
tryEntries THREW: not an id of the width,255
tryEntry one component THREW: not an id of the width,255
tryEntry empty path: no throw
```

The last line is why `tryEntry`'s `@throws` carries a sentence the other two
do not. A path of no components returns `null` before `rootOf(id)` is
reached, so the id is never read and the assertion never runs.

**The break shipped undeclared.** `peel`'s assertion is itself in `main`, at
`653f1bba`. It throws where the old code called the supplied `Read` and
answered its effect, and `Oid` carries no width, so nothing catches the mix
at compile time. I added a `**BREAKING CHANGES:**` item to #1989's
description in answer to the review, but the merge commit `37e14461` carries
the earlier text, whose `Changelog:` section names only `git/repo`. Nothing
derives a break from a diff, and `0.49.0` is already published, so the
declaration has to land on a pull request in the open window or the next
release picks a patch number for a release that breaks. This is that pull
request, and the section below is the declaration.

Changelog:
- **BREAKING CHANGES:** `git/walk`: `peel`, and `tryEntries` and `tryEntry`
  through it, throw on an id that is not `oidBytes` wide, where they used to
  call the supplied `Read` and answer its effect. `Oid` carries no width, so
  no type catches the mix, and a SHA-256 id was getting a plausible answer
  out of a SHA-1 walk. Pass the id at the width the repository uses, which
  is `git/store`'s `oidBytes`. A path of no components still answers `null`
  before the id is read, so `tryEntry` does not throw there. The check
  itself shipped in #1989; this declares it.
- `git/tag`, `git/commit`: a tag whose `type` value holds an LF, and a commit
  whose `parent` names the same id as its own `tree`, are refused, as Git's
  own parse refuses them. Also shipped in #1989, and not breaking — they
  narrow what two readers accept to what Git accepts — but worth a release
  note

https://claude.ai/code/session_01EdU8wW4oihFS1gj4dUAcPS

---
_Generated by [Claude Code](https://claude.ai/code/session_01EdU8wW4oihFS1gj4dUAcPS)_

### DIFF
--- a/fjs/git/walk/module.f.mjs
+++ b/fjs/git/walk/module.f.mjs
@@ -183,6 +183,9 @@ export const peel = (read, oidBytes) => {
  * it refuses one without, so this reads that tree and does not judge the
  * commit again.
  *
+ * @throws On an id that is not `oidBytes` wide, since the walk starts at
+ * {@link peel} and that is its check.
+ *
  * @template {Operation} O
  * @param {Read<O>} read
  * @param {OidBytes} oidBytes
@@ -260,6 +263,10 @@ const componentStep = at => ({ name, last }) => ({ entries }) => {
  * The entry's own object is not read, so the caller reads the blob with
  * the same `Read` and a submodule entry comes back rather than a missing
  * object.
+ *
+ * @throws On an id that is not `oidBytes` wide, since the walk starts at
+ * {@link peel} and that is its check. A path of no components is the one
+ * that does not: it answers `null` before the id is read.
  *
  * @template {Operation} O
  * @param {Read<O>} read


o2alexanderfedin (commented):
Measured against merge-base(origin/main, head) = 29f637bae226e12028f3aa68c318198a083b2726 — the delta is the two `@throws` blocks in `fjs/git/walk/module.f.mjs` and nothing else.

**The declarations are true.** I checked both by running, not reading. `peel(read, 20)` and `tryEntries(read, 20)` and `tryEntry(read, 20)(id, [name])` each throw on a 32-byte id, all three with the same payload `not an id of the width` — the same width, the same refusal, the same message as `peel`, because `peeled(id)` is evaluated eagerly and there is exactly one `assert` in the file. And the exception `tryEntry` carves out holds exactly: `run({})([])(tryEntry(loud, 20)(wide, []))` answers `["ok", null]` with a `Read` that throws if called, so a path of no components really does answer before the id is read. Nothing else in the file is falsified — "Nothing here panics on what it read" still stands, since this panics on an argument.

**But the assert is still not pinned, which was the point.** Deleting `assert(length(id) === bits, ...)` from `peel` leaves the suite fully green: `node ./fjs/module.mjs t` → pass 5856, fail 0, exit 0, `git diff --stat` one deletion, no `SyntaxError` in the log. The harness is working — mutating `only`'s `found.length === 1` to `>= 1` reddens `proof.entryRefused` at 5855/1. So there are now three `@throws` in this module and zero `throw:` entries in `walk/proof.f.mjs`, while the identical assert in `fjs/git/store` is pinned by `proof.throw.width`. `fjs/AGENTS.md` §1.5 is explicit that a throwing contract is tested under a `throw` key. Declaring the contract is worth having; a `throw: { width: () => peeled(wideId) }` — and one for `tryEntries`/`tryEntry` — is what would keep it.

Clean tree, `tsc -p .` exit 0.

sergey-shandar:
@o2alexanderfedin You were right, and this merged before I could act on it, so the fix is #2008.

Your deletion test is the one that mattered: three `@throws` and zero `throw:` entries, while the identical check in `fjs/git/store` is pinned by `proof.throw.width`. A declaration nothing enforces rots on the next edit.

Worth recording, because it nearly cost me the fix. Writing the cases the obvious way — over the proof's existing `Read`, which is the store's — pins nothing. That reader asserts the width itself at `fjs/git/store/module.f.mjs:126`, so the walk throws whichever of the two checks survives. I wrote that version first and it passed with `peel`'s assert deleted:

```
assertion deleted → pass: 5859, fail: 0, exit 0
```

A test that looks like it pins something and does not. The cases in #2008 run over `liar` instead, which answers any id without looking at its width, so the only check left in the call is the one under test:

```
assert present  → pass: 5859, fail: 0, exit 0
assert deleted  → proof.throw.width, entriesWidth, entryWidth all error
                  pass: 5856, fail: 3, exit 1
```

Exactly three, and nothing else moves.

`tryEntry`'s case passes a path of one component deliberately, since a path of none answers before the id is read and is the one call of the three that does not throw.

---

PR #2008, opened 2026-09-11
git/walk: the width check is pinned by a proof, not only declared

sergey-shandar:
Takes the review point on #2006 that its merge did not carry.

The review was right and the reasoning was exact: #2006 added three
`@throws` to `fjs/git/walk` and zero `throw:` entries to its proof, while the
identical check in `fjs/git/store` is pinned by `proof.throw.width`.
`fjs/AGENTS.md` §1.5 puts a throwing contract under a `throw` key. Declaring
a contract nothing enforces is a declaration that rots on the next edit.

**The trap, since it caught me.** The obvious proof is to run `peel` over the
proof's existing `Read`, which is the store's. That reader asserts the width
itself, at `fjs/git/store/module.f.mjs:126`, so a walk built on it throws
whichever of the two checks survives. I wrote that version first and it
passed with `peel`'s assert deleted — a test that pins nothing while looking
like it pins something. The cases here run over `liar` instead, the proof's
reader that answers any id without looking at its width, so the only check
left in the call is the one under test.

**Measured both ways rather than argued.** With the assert in place:

```
Number of tests: pass: 5859, fail: 0, total: 5859     exit 0
```

With `assert(length(id) === bits, ...)` deleted from `peel`:

```
proof.throw.width(): error
proof.throw.entriesWidth(): error
proof.throw.entryWidth(): error
Number of tests: pass: 5856, fail: 3, total: 5859     exit 1
```

Exactly the three, and nothing else moves. That is the property the review
asked for and the first version did not have.

`tryEntry`'s case passes a path of one component on purpose. A path of none
answers `null` before the id is read, so it is the one call of the three that
does not throw, which is what its `@throws` already says.

`idOf` now goes through a new `idAt`, which takes the hex rather than a
number, so the 32-byte id and the 20-byte ones are built one way instead of
two.

Coverage stays at 100% on lines, branches and functions.

No changelog entry: no observable behavior changes here. The break this pins
was declared on #2006.

https://claude.ai/code/session_01EdU8wW4oihFS1gj4dUAcPS

---
_Generated by [Claude Code](https://claude.ai/code/session_01EdU8wW4oihFS1gj4dUAcPS)_

### DIFF
--- a/fjs/git/walk/proof.f.mjs
+++ b/fjs/git/walk/proof.f.mjs
@@ -271,12 +271,22 @@ const name = latin1
 /** @type {(e: TreeEntry) => readonly [number, string, string]} */
 const seen = e => [mode(e), codePointListToString(e.name), hex(e.oid)]
 
-/** An id the hex digits of a number spell, at 20 bytes. */
-const idOf = /** @type {(k: number) => Oid} */ (k => {
-    const v = tryFromHex(latin1(hexOf(k)))
+/** The id a string of hex digits spells, at whatever width it is long. */
+const idAt = /** @type {(hex: string) => Oid} */ (hex => {
+    const v = tryFromHex(latin1(hex))
     assert(v !== null)
     return v
 })
+
+/** An id the hex digits of a number spell, at 20 bytes. */
+const idOf = /** @type {(k: number) => Oid} */ (k => idAt(hexOf(k)))
+
+/**
+ * An id of 32 bytes: the width a SHA-256 repository uses, and not this
+ * walk's. Which digits it spells says nothing, since the width is refused
+ * before any object is asked for.
+ */
+const wideId = idAt('8031c3b5f0c291f374148e59909ea8a8f83538e9a412bac9b1f8072e6e6be27f')
 
 /** The 40 hex digits a number spells, as an id is written. */
 const hexOf = /** @type {(k: number) => string} */ (k => k.toString(16).padStart(40, '0'))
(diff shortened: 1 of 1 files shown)


o2alexanderfedin (approved):
Approve. Measured against `merge-base(origin/main, head)` = 527b1b74b91c2c402ebdd5189b94a2fafeebd4ba; the delta is 26 lines in `fjs/git/walk/proof.f.mjs` alone.

This closes my #2006 finding, and I tested the fix rather than read it.

**The pin holds.** Deleting the one `assert` in `peel` (`assert(length(id) === bits, ...)`) — `git diff --stat` non-empty, one deletion — turns the suite red: `5856 pass / 3 fail`, exit 1, no `SyntaxError` in the log. The three are exactly the new entries:

```
proof.throw.width(): error
proof.throw.entriesWidth(): error
proof.throw.entryWidth(): error
```

Restored, the suite is `5859 / 0` (exit 0) and `tsc -p .` exits 0. The +3 over the old 5856 is these three, and the run log marks each `# EXPECTED TO THROW`, so they are registered under the `throw` key as `fjs/AGENTS.md` §1.5 requires.

**They throw for the right reason.** Each of the three throws the width assert's own payload, `['not an id of the width', <id>]` — not some unrelated failure on the way. That module has exactly one `assert`, so with it pinned there is no surviving unpinned one.

**The comments check out.** `liar` answers any id without reading its width; `fjs/git/store/module.f.mjs` carries the byte-identical assert; and `tryEntry(liar, 20)(wideId, [])` really does answer `['ok', null]` without throwing, which is what the `entryWidth` comment claims justifies giving it a component. `wideId`'s hex is 64 digits, i.e. the 32 bytes its doc says.

No prose in the tree is falsified by the delta.

PR #2043, opened 2026-09-13
datajs: the public module.f.mjs, and the four entry points are try*

sergey-shandar:
The last piece of stage 4's reader: the codec's own `fjs/media/datajs/module.f.mjs`, and with it the naming decision the reader's byte path ([#2040](https://github.com/functionalscript/functionalscript/pull/2040)) deferred to it.

## The surface

The codec had no module of its own — the reader was reached at `parser/module.f.mjs`, the writer at `serializer/module.f.mjs` — and the reader's two names had landed bare (`parse`, `parseBytes`) beside the writer's `trySerialize` and `tryStringify`. The design said all four are `try*`, and said the rename waits for the module that carries all four, so half the names are never renamed ahead of the other half.

```js
export { tryParse, tryParseBytes } from './parser/module.f.mjs'
export { trySerialize, tryStringify } from './serializer/module.f.mjs'
```

**Every entry point is fallible, and the names say so.** The reader's two are renamed to `tryParse` and `tryParseBytes`; every importer outside the directory — `fsc`'s two proofs, the corpus's three writer-side proofs, the matrix generator — now imports the public module rather than a sub-module. `fsc/module.f.mjs` keeps its `serializer/` import because it also takes `_numberSerialize`, which is linkage, not surface.

No `**BREAKING CHANGES:**` entry: nothing under the old names ever shipped in a release (no changelog file mentions `fjs/media/datajs`).

## The module's proof

Pins the four signatures the design fixed, as type-level assertions, and proves the four compose as one codec — the deeper proofs stay where each entry point lives:

- `roundTrip` — `tryStringify ∘ tryParse` is the identity on a normalized document carrying sharing, `1n`, `undefined`, `NaN`, `-0` and a four-byte scalar;
- `bytes` — `tryParseBytes` over that document's UTF-8 bytes reads the same graph, by `difference`;
- `chunks` — `concat ∘ trySerialize` is `tryStringify`;
- `refused` — a trailing comma, a lone BOM, a function: each side refuses.

## The reader's issue retires

With the surface landed the reader owes nothing, and `todo/parser-serializer.md`'s own last task says to delete it in the PR that finishes the reader, "but not before the shared material … the public API of §Layout and the value domain of §1, has moved into [the writer's issue] or into the module's README". Both moved into the README, as two sections: **Every entry point is fallible, and the names say so** (the layout, the four signatures, why the byte path is a conformance obligation and refuses rather than strips a BOM) and **One type-level trap in the value domain** (`{a: undefined}` and `{}` share a type and are different documents, and what that obliges each side to prove). Its four citers — the README, the writer's issue, the plan, the YAML issue — point there now.

Net −137 lines: the retired file was the bulk of it.

## What stage 4 still has

The writer's issue, `serializer.md`, with its remaining items: the explicit-stack walk (2,600 nested arrays make `tryStringify` throw where it owes an `error`), the quadratic-steps measurement, and a readable layout if one is wanted. The plan's stage-4 item says so, and says stage 4 is done when that file goes.

## Checks

`tsc` clean; `node ./fjs/module.mjs t` 6837 pass / 0 fail (up 7, the new proof leaves); `npm run cov` 100% lines/branches/functions; `npm run gen` leaves no diff; every relative link and cross-document anchor in the changed files resolves, and no `parser-serializer.md` citation remains. The branch carries `main` at `e8cdf0f3`, which is #2040's merge.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TzayQtZmCzP9rtExwHXtF2)_


o2alexanderfedin (approved):
Approved. Delta measured against `origin/main` at `e8cdf0f3`. Suite 6837/6837 (+7 over main, and the seven are exactly the new proof's leaves), `tsc` clean, `cov` 100%, `gen` leaves no diff.

The omitted `**BREAKING CHANGES:**` is right, and it holds against the artefacts rather than the changelog. The last publish is 0.49.0 (2026-09-02); at its release commit `ff4d1a71`, `fjs/media/datajs/` held only `todo/parser-serializer.md` and no `.mjs` at all, and the 0.47/0.48/0.49 tarballs carry no `fjs/media/datajs` path. `parse` first landed `9c1de927` (2026-09-09) and `parseBytes` in #2040 — both after that publish, so neither old name ever shipped. The `files` globs do cover the path, so the `try*` names ship on the next release.

The deleted issue's material survives: both anchored citers resolve to real README headings, no citation to the file remains anywhere (base had ten), and link and anchor sweeps show zero new dangling entries with injected controls caught.

The four leaves are pinned, not just green — breaking `tryParseBytes`' re-encode reddens `proof.bytes`, a spurious chunk in `trySerialize` reddens `roundTrip` and `chunks`, and swallowing the writer's refusal reddens `refused.value`.

PR #2045, opened 2026-09-14
datajs: the writer keeps the reader's depth contract, over an explicit stack and the graph's post-order

sergey-shandar:
The one real defect left in stage 4's writer: both passes recursed on the call stack, where the reader walks an explicit one and keeps a 5,000-level depth contract — so the writer could not write back every document the reader accepts. Measured before this change: 1,000 nested arrays wrote and 1,100 threw, and the reader's own 5,000-level document made `tryStringify` throw `RangeError: Maximum call stack size exceeded` where `§Layout and API` promises an `error` at worst. `serializer.md` §1 had it booked as a deferred crash with the input named.

## Pass 1: the read, over an explicit stack

The reader's shape exactly — a frame per container being read, its members read in order (`private.ts` gains `_Frame`, `_Stack`, `_Todo`, `_Next`, `_State`, mirroring the reader's). The recursive trio `read → readNode → readMembers → read` becomes four functions over a state and one loop:

- `open` — a container met for the first time, as a frame: kind settled before descriptors, exactly as before (the `Date`/`Map`/`Set`/`Number` argument stands), holding the properties the read will follow in observable order;
- `round` — the next member of the frame on top, its descriptor validated **as its turn comes** — so a refusal is still the first in observable order, as `readMembers` had it — or, when none is left, the frame closed into its node and a reference handed down;
- `enter` — one value of the caller's: leaf, known container, new container, or refusal;
- `read` — the loop.

Sharing, cycle termination and the refusal order are unchanged: `started` still holds a container from before its members are read, `_link` still refuses the cycle over the finished graph. The member refusal from `_memberValue` goes through `mapOk`, not a local branch, since only a host-built value reaches it — the first draft had the branch and `cov` reported it at 99.99%.

## Pass 2: the write, over the graph's post-order

No stack needed, and that's a measurement rather than a hope. `_link` leaves the nodes in post-order — a node comes after every node it refers to — so each node's chunks can be built **in index order from the chunks of nodes already built**. A reference to an inline node is a thunk over that node's chunks (`() => chunks[i]`), forced only as the document is read out, and the list's `next` trampolines thunks in a loop rather than recursing: a 20,000-deep lazy chunk structure reads out in 125 ms. The recursive `value ⇄ inline` pair stays as it was, minus the recursion; `arrayWrap`/`objectWrap` are still JSON's.

## Proof

`proof.depth.writesBack` — the reader's 5,000-level document written back byte for byte, which is the contract stated as an equation. `proof.depth.below` — under the 2,600 arrays that used to throw, a function is refused with `a function is not a DataJS value`, and a shared deep chain is hoisted as any other.

**Cost, and where it comes from.** The two leaves take about 1.9 s together, and that is the quadratic step the issue already names, now measurable because depth no longer throws first: `new Set([...started, value])` per container. Measured: 2,600 deep 0.3 s, 5,000 deep 1.1 s, 20,000 deep 23 s — and 5,000 persistent `Set` adds alone are 0.98 s of the 1.1 s. That measurement is now in `serializer.md` §1, under the task that decides whether the step is worth changing; this PR does not change it, because the decision is the owner's and the sanctioned immutable idiom is exactly that copy.

## Documents

`serializer.md` §1's "Depth is a defect rather than a cost" becomes "Depth is the input's, not the call stack's", the task is ticked, and the quadratic task carries the measurement. The module README's writer section and the module header say how each pass keeps depth. The plan's stage 4 now names the quadratic decision and the optional readable layout as all that is left before `serializer.md` goes.

## Checks

`tsc` clean; `node ./fjs/module.mjs t` 6839 pass / 0 fail (up 2); `npm run cov` 100% lines/branches/functions; `npm run gen` leaves no diff; every relative link and anchor in the changed documents resolves. Based on `main` at `c7c32cfa` (#2043's merge).

---


o2alexanderfedin (commented):
Delta measured against `origin/main` at `c7c32cfa`. Suite 6839/6839 (+2, exactly `proof.depth.writesBack` and `proof.depth.below`), `tsc` clean, `cov` 100%, `gen` leaves no diff, 0 dangling links or anchors in the three changed documents.

The contract holds, measured at both ends. At `c7c32cfa` the throw starts between 1,000 and 1,100 — your "1,000 wrote, 2,000 threw" is right, but the real boundary is lower than 2,000. At this head nothing throws at any depth I could reach: 5,000 and 5,001 write, and so do 10,000, 20,000 and 100,000 (385 s, no throw). The reader's 5,000-level document round-trips byte-exact, and so do 6,000 and 8,000; neither side has a numeric cap. Making the write pass eager reddens `proof.depth.writesBack` with the old `RangeError`.

The post-order claim is stronger than the body says: `_link` **verifies** `nodeRefs(node).every(ref => ref < i)` rather than assuming it, and `chunks` is fully built before any thunk is forced, so an unbuilt entry is unreachable. Cycles still answer `error('a cycle')`, and a 3,000-case random-DAG fuzz preserved sharing with no failures. Your timings reproduce here within the usual load spread (5,000 deep 1.08–1.40 s; 20,000 deep 13.4–17.3 s against your 23 s).

Two small things:

1. `fjs/media/datajs/todo/serializer.md:497` says the `started` copy is "85% of a 5,000-deep write's 1.1 s", but §1 of the same file measures it as 0.98 s of 1.1 s, which is 89%.
2. `fjs/media/datajs/serializer/proof.f.mjs:290` cites "§Layout and API" with no link, and that section is in `todo/serializer.md`, not the neighbouring `README.md`.

Worth noting for later: the 5,000-level leaf pins the contract but not much above it — a regression costing one call frame per level still passes `writesBack` and only breaks around 9,000.

sergey-shandar:
Both fixed in aca3029f. The task now quotes the measurement as §1 states it — the `started` copy is 0.98 s of a 5,000-deep write's 1.1 s — and the proof's comment says `todo/serializer.md` §Layout and API, since the README has no such section. The description's "2,000 threw" is corrected to the boundary you measured, between 1,000 and 1,100.

The note about the pin's reach landed in the same task rather than in the thread: the 5,000-level leaf pins the contract and little above it — a regression of one call frame per level passes it and breaks only near 9,000 — and pinning higher costs the quadratic step's time until that step is decided, so the two wait on the same decision.

`tsc` clean, suite 6839/0 on the head.

---

o2alexanderfedin (approved):
Approved. Delta measured against the merge-base `c7c32cfa`. Suite 6839/6839, `tsc` clean.

Both fixed: `serializer.md:497` and §1 now both say 0.98 s of 1.1 s, so there is no percentage left to disagree with, and the proof's citation names `todo/serializer.md` where `#### Layout and API` does exist. I re-confirmed the substance independently — 5,000 nested arrays return `ok`, and the reader's 5,000-level document writes back byte-exact at 10,016 characters.

One nit: from `fjs/media/datajs/serializer/` the bare `todo/serializer.md` resolves nowhere; siblings such as `fjs/crypto/sha1/module.f.mjs` write `../todo/…`.

PR #2047, opened 2026-09-14
sul: document compression research and reproducible experiments

sergey-shandar:
SUL exposes repeated subtrees, but that alone does not establish how compactly a document can be stored or how it compares with existing approaches. Preserve the complete investigation beside the SUL module so the design, evidence, and limitations can be reviewed and reproduced.

The research covers SUL's construction and backing-storage requirements; comparisons with ordinary deduplication, Prolly Trees, and content-defined Merkle trees; and a lossless grammar codec using local dictionary references and adaptive arithmetic coding. It includes the cost model, traversal and pruning rules, format, related literature, applications, and proposed next experiments.

Include the complete eight-input measurements for all 160 grammar variants, gzip/Brotli/Zstandard baselines, input hashes, and one-shot timings. Preserve the six experimental JavaScript files as a reviewable source listing and a runnable archive with the original SUL dependency snapshot and MIT license. The archive is explicitly a research artifact; it introduces no supported codec API or fjs command. Generated encoded outputs are reproducible from the included benchmark and are omitted from the archive.

Measurements are pinned to FunctionalScript commit d16a9ebf39b30a42e1b795bc329774d867196841 and Node v24.19.0. Repeated structures compress strongly, but high-quality Brotli and Zstandard generally produce smaller outputs. Arithmetic-only and raw-grammar controls separate subtree reuse from statistical coding gains. The report does not claim an optimal grammar or a measured advantage over Prolly/CDMT.

The follow-up tests the proposed first-occurrence dictionary order explicitly: deduplicate a frontier, recursively compress the retained dictionary contents together at lower levels, and skip a level when its complete recipe-plus-dictionary representation loses. It compares actual SUL word generations and expanded Patricia-height frontiers, each with and without recursive dictionary deduplication. All streams use adaptive arithmetic models or a smaller raw alternative.

The ten-input follow-up includes 40 final encodings and 9,976 candidate cost comparisons. With word generations, recursion improves eight inputs over the one-level control and ties on two. Repeated sentences shrink from 1,344 to 511 bytes relative to that control. Against the earlier complete codec, JSON improves from 15,498 to 14,726 bytes and unique sentences from 7,718 to 5,904 bytes; the earlier codec remains smaller on several other inputs. Preserve the original archive unchanged and add the follow-up report, source listing, compact measurements, and a second runnable archive containing all candidate costs. [Follow-up report](https://github.com/functionalscript/functionalscript/blob/06c01ab6c2d9cfbbbd2931ef00d1cdc834e76cb2/fjs/sul/research/ordered-dictionaries.md).

The byte-alphabet follow-up runs the same ordered codec with raw byte leaves and a matched raw-bit control, using the existing SUL word/hash machinery from the first generation. This explicitly bypasses the native bit-specific literal pipeline and changes root IDs. Adaptive binary arithmetic coding, dictionary ordering, recipes, and fallbacks are unchanged. Across the same ten inputs, bytes improve five, tie two, and lose three against the native-bit recursive hierarchy; against the matched raw-bit control they improve four, tie two, and lose four. Source code shrinks from 44,254 to 35,024 bytes, JSON from 14,726 to 11,536, and repeated sentences from 511 to 430. All 40 new encodings and 3,972 candidate cost comparisons are preserved in a third archive with its report and source listing. The report includes padding costs and prototype bounds, and both previous archives remain unchanged. [Byte-alphabet report](https://github.com/functionalscript/functionalscript/blob/a03029b808bcadab67220a959f13fbe5a62e4628/fjs/sul/research/byte-alphabet.md).

The Brotli dictionary follow-up tests 13,504 shared base entries and all 121 standard transforms as level-zero tokens, with literal-byte coverage for unmatched content. It preserves the adaptive binary coder and recursive first-occurrence dictionary procedure, compares direct/one-level/recursive coding, and includes every token/transform reference plus 28 bytes of framing in dictionary modes. The static dictionary is a pinned decoder resource with an explicit distribution cost. Against byte leaves, the full profile improves source from 35,024 to 31,447 bytes, JSON from 11,536 to 10,229, and the repeated phrase from 188 to 139. It wins five, ties two, and loses three on the original ten inputs; two new short examples also improve. Base entries alone beat full transforms on unique sentences (3,889 versus 4,760 bytes). The report preserves these tradeoffs and the established Brotli/Zstandard comparisons. A fourth archive contains both pinned MIT dependency snapshots, six new source files, the shared builder change, 72 final measurements, and 2,866 internal cost comparisons. [Brotli dictionary report](https://github.com/functionalscript/functionalscript/blob/b8b7f86428f52471ed4fa7cd34eab3d9ea985f60/fjs/sul/research/brotli-dictionary.md).

The byte-order follow-up searches permutations of all 256 level-zero symbols while retaining original byte payloads, higher-level SUL ordering, recursive first-occurrence dictionaries, and adaptive arithmetic coding. Exact input group count selects the order. Across ten fixtures, groups decrease on nine and tie on one, but complete compression improves on four, ties on three, and loses on three. JSON shrinks from 11,536 to 10,432 bytes and unique sentences from 5,025 to 4,250; source grows from 35,024 to 35,570 despite 17.01% fewer groups. The report includes all ten permutations, two held-out grouping checks, and five additional successor-chain/DAG constructions per fixture, including frequent-first and rare-first choices. Neither starting-frequency rule consistently wins. Fixed-frequency tables remain a separate model-tuning hypothesis. A fifth archive preserves eight source files, the pinned dependency, measurements, and previous byte controls. [Byte-order and successor-graph report](https://github.com/functionalscript/functionalscript/blob/53d792ba4df9c7530994ff37f77ce3f19fb127ed/fjs/sul/research/byte-ordering.md).

The shared-frequency follow-up adds a local registry prototype with immutable generic, text, ASCII and JSON model objects. Their SHA-256 identifiers are stored as compression metadata; the SUL alphabet, hierarchy and first-occurrence dictionary policy remain unchanged. Profiles model the existing binary coding contexts, rather than assuming document-local dictionary IDs have shared meanings. Training uses 22 real whole documents (61,466 bytes), with 12 different test documents (26,537 bytes) and complete corpus provenance/licenses. Four fixed profiles and the same four profiles as small adaptive priors are compared with the original adaptive model. All 108 SUL candidates decode correctly, but every test selects the original adaptive model: 14,392 bytes total, versus Brotli's 7,566 and Zstandard's 9,090. The 37-byte profile-reference cost and uncached profile distribution sizes are explicit; removing identifier overhead still yields no profile win. Preserve this negative result, all four profiles, the registry, source listing, and a sixth runnable archive. [Shared-frequency report](https://github.com/functionalscript/functionalscript/blob/9c92d67b372ac0d1a278fa8263c238650a2172b9/fjs/sul/research/frequency-profiles.md).

Validation:

- Type-checking passed with the repository-pinned TypeScript 7.0.2.
- Full `node --test` passed: 6,794 tests, zero failures.
- Extracted prototype tests passed: 20 cases across the original and all three follow-up suites, covering short bitstrings, every byte value, decreasing byte words, nested dictionaries, first-occurrence order, skipped levels, raw recipes, unequal-height frontiers, non-byte-aligned bit inputs, and byte-input validation.
- Rerunning the archived benchmark reproduced every input hash, all 160 candidate sizes/statistics, and every baseline size exactly; all outputs decoded correctly. Timings were excluded from equality comparisons.
- All 40 follow-up encodings decoded exactly; all eight reused fixture hashes matched the original experiment. The CLI file round-trip was byte-identical.
- The extracted byte benchmark reproduced all ten input hashes, all 40 compressed files byte for byte, all 3,972 candidate costs, selected layers, statistics, and baseline sizes. Timings were excluded; the archived CLI restored its input exactly.
- The Brotli transform implementation matched the pinned C oracle for all 1,633,984 word/transform combinations. All 72 new final encodings, all 72 dictionary candidates, and 12 byte controls decoded exactly. The extracted benchmark reproduced all 72 files byte for byte and all 2,866 candidate costs; the CLI restored its input exactly. Timing differences were excluded.
- All 20 byte-order encodings decoded exactly, and all ten natural-order controls reproduced the prior byte files byte for byte. Six extracted prototype tests passed. The extracted source benchmark reproduced both files and every non-timing measurement; the extracted graph run reproduced all 50 additional permutations and grouping measurements. Accepted graph edges, a directed-cycle example, a descending chain, empty input, complete alphabets, and input hashes were checked.
- Five profile tests and all 11 existing ordered/byte cases passed. All 22 training encodings, 108 SUL test candidates, and 24 standard-codec outputs decoded exactly. The extracted rerun reproduced all four profile objects, the registry, all 108 files byte for byte, and all non-timing measurements. The extracted CLI restored a hash-selected profiled document exactly. The adaptive factory also reproduced the earlier repeated-phrase byte encoding byte for byte.
- Archive checksums, source-listing equality, local documentation link targets, ZIP integrity, and `git diff --check` passed. A scoped Git attribute preserves the binary archive under the repository's text normalization default.
- `cargo clippy` and `cargo fmt -- --check` could not run because Cargo is unavailable in this environment. No Rust files changed. This PR remains a draft pending the repository-required Cargo checks.

Further design questions remain documented in the report: cost-based rule selection, stronger probability models, shared dictionaries across real revisions, and independent compressed blocks for synchronization. The existing serialization TODOs and CAS Strategy 3 plan remain open.


o2alexanderfedin (commented):
Delta measured against `origin/main` at `88faac36`. Suite 6856/6856, `tsc` clean.

The Brotli commit holds on every axis I could check it against. Every figure in the prose recomputes from `brotli-results.json` — the percentages, the win/tie/loss tallies, the candidate costs with their fallbacks, the coverage numbers, the 16→336→70 and 19→399→78 token arithmetic and the worked 396-byte example — with flipped digits as controls. The CSV is a field-by-field projection and byte-identical to the archive's own regenerated copy, all 19 manifest digests verify, the seven quoted listings are byte-verbatim against `unzip -p`, and the 30 vendored fjs files match the pinned commit `d16a9ebf`.

The part worth saying out loud: the Brotli claims check against **RFC 7932 itself**, not just against the prototype's output. Appendix A's 122,784 bytes and CRC-32 `0x5136cb04` both match the shipped `dictionary.bin`, which is byte-identical to `dictionary_inc.h` at the pinned google/brotli commit; NDBITS matches the vendored table and sums to 13,504 words over lengths 4–24; 121 transforms confirmed in both the RFC and `transform.c`; the 14- and 21-bit token widths are exactly minimal for the stated ID formulas; and the eight-zero-byte base entry is real, at index 1002 of the length-8 bucket.

`brotli-prototype.zip` is covered by `-text` before it landed — `check-attr` reports `text: unset`, blob bytes equal worktree bytes equal the manifest digest at 535,849, and `unzip -t` passes on the blob. Links: 10 dangling at base and the same 10 here, none new.

One thing: the body still says "Full `node --test` passed: 6,794 tests, zero failures." The repository's own gate reports 6,856, and nothing I measured produces 6,794.

o2alexanderfedin (commented):
Delta measured against the merge-base `88faac36` — main has moved past it. Suite 6856/6856, `tsc` clean.

The new commit holds on every axis. 480 derived checks across `order-results.json`, `order-heldout.json`, `successor-results.json` and `byte-orders.json` reconcile — both tables, the 7×10 successor table, the means as N/M, the `groupReduction`/`compressionReduction` formulas, every win-tie-loss tally, the 546 and 775 byte deltas, the 3847→4598 and 153→532 unique counts, 2,772 recipe comparisons summed from trials, 17.3 s summed from `searchMs`, the seed `0x53554c31`, 105 active symbols and `(1+1/256)^256` = 2.713 — with nine flipped-digit controls each going red.

Two things lift this above a data check. The held-out table was **reproduced end to end** from a rebuilt source stream with an independent group counter — 24,614 / 23,441 / 24,216 / 24,142 plus every histogram, and the `source_code` fixture's sha256 with natural 24,821 and optimized 20,599. And all five successor-graph candidates were rebuilt **from the prose description alone**: the 256-value orders, `candidateEdges`, `acceptedEdges`, `rejectedCycles` and `descendingTransitionCount` all match, with a red control.

The external claims check out too: the topological sort is Kahn's with correct transitive-closure cycle rejection and `rank(a) > rank(b)` re-derived on every accepted edge; `transitionScore` is exactly the LOP objective and the cited LOLIB URL is live; xorshift32 uses the standard 13/17/5 shifts; `HEADER = 14` backs the 14-byte frame and the golden files begin `SULO\x01`. The 48 self-loops and 538 mutual 2-cycles confirm the "not automatically a DAG" claim empirically rather than by assertion.

Manifest 8/8, CSV an exact projection, all eight in-prose listings byte-identical to `unzip -p`, the five committed data files byte-identical to their in-archive copies, archive clean across 78 entries, `.gitattributes` `-text` landing in the same commit as the zip with blob bytes equal to worktree bytes — all positive-controlled. Links: 33 dangling at base, the same 33 here, none new.

One thing: the body still says "Full `node --test` passed: 6,794 tests"; the runner reports 6,856. Everything else in the body reconciles, including "50 additional permutations" as 10 fixtures × 5 candidates.

o2alexanderfedin (commented):
Delta measured against the merge-base `88faac36`, not current main. Suite 6856/6856, `tsc` clean, `ci` gate clean.

The new commit holds: 585 figure assertions against the committed data pass with none failing, and four mutation controls each reddened. Manifest digests, `.gitattributes`, byte-verbatim listings against `unzip -p`, archive hygiene and the link sweep (19 dangling at base, the same 19 here) all check out, each positive-controlled, and the RFC 7932 sections still reconcile.

Two things:

1. `fjs/sul/research/frequency-profiles.md:71-75` omits two of `select-corpus.py`'s filters. Applied literally, the rule selects a different source set — a pool of 309 against the actual 210 — though the committed outputs are unaffected.
2. The body still quotes 6,794 tests; the runner reports 6,856.

PR #2048, opened 2026-09-14
datajs: the writer's cost is a sort, not a square — a persistent set for the containers entered, and sorted occurrences for sharing

sergey-shandar:
The two quadratic steps `serializer.md` §1 named, measured in [#2045](https://github.com/functionalscript/functionalscript/pull/2045) once depth stopped throwing first, and decided here by changing them. `fjs compile` writes every module through this writer, so "a document with enough containers" was the compiler's own output.

## What was quadratic

- **`started`**, the set of containers entered, grew by one per container through `new Set([...prev, value])` — the immutable idiom, and it copies every container so far each time. At 5,000 nested arrays that copy alone was 0.98 s of a 1.1 s write; 20,000 took 23 s.
- **`shared`** counted reference occurrences with `refs.indexOf(ref) !== i`, an `indexOf` per occurrence.

## `fjs/types/set`: a persistent set with a logarithmic add

New module, `module.f.mjs` + `types.ts` + `proof.f.mjs`. It keeps `Set`'s equality (SameValueZero, so objects by identity) and lays the elements out as **the binary representation of the set's size**: the entry at index `k` holds exactly `2^k` elements or is `null`. An add carries like a binary counter — a one-element `Set` into the first empty entry, every full entry on the way merged into what is carried — so each element is copied at most once per bit of the size, and `has` asks at most `log2(n) + 1` `Set`s. Every operation is a construction; nothing is mutated. Adding what is already present returns the same set (`===`), so the layout invariant holds without duplicates.

```js
const set = add(a)(add(1)(empty))
has(a)(set)          // true
add(a)(set) === set  // true
```

Nothing in `fjs/types` did identity-keyed persistence — `sorted_set`, `btree`, `patricia_trie` all need comparable keys, and objects aren't — which is why it's a new module rather than a use.

**Proof:** the carry sequence `1`, `10`, `11`, `100`; the layout invariant (entry `k` is `null` or size `2^k`) and every element present at every size 0..70 and at 20,000 (15 entries); persistence (the old set is unchanged by an add); adding a present value is the same set; `Set`'s equality — objects by identity, `NaN` itself, `-0` as `0`.

## The writer

`_Walk.started` is that set (`has(value)(walk.started)`, `add(value)(walk.started)`), and `shared` sorts the occurrences (`toSorted`, ES2023, already used in three modules) and takes a repeat to be a neighbour equal to the one before it. Nothing else changes; `_link` and the write are as #2045 left them.

## Measured

| | before | after |
|---|---|---|
| 5,000 nested arrays | 1.1 s | 87 ms |
| 20,000 nested arrays | 23 s | 0.3 s |
| 100,000 nested arrays | — | 1.3 s |
| 100,000 empty objects in one array | — | 1.1 s |
| 100,000 references to one object | — | 0.5 s |

With the cost gone, `proof.depth.writesBack` pins **20,000 levels** — four times the reader's 5,000-level contract, so a regression costing one call frame per level, which 5,000 would survive (the note from #2045's review), is caught. That leaf costs ~0.7 s, most of it the reader parsing 20,000 brackets; `below` dropped from 0.55 s to 30 ms.

## Documents

`serializer.md` §1's "One cost is now measured rather than assumed" becomes "The cost is a sort, not a square", with the before and after numbers; its task is ticked, and its Status says what remains is the readable-layout decision alone. The module README's writer section and the module header say why neither pass squares the container count; `private.ts` says what `started` is; the plan's stage 4 names only the readable layout as what is left before `serializer.md` goes.

## Checks

`tsc` clean; `node ./fjs/module.mjs t` 6862 pass / 0 fail (up 6 against `main`, the set's leaves; `main` itself is up 17 since #2045 with #2046); `npm run cov` 100% lines/branches/functions; `npm run gen` leaves no diff; every link in the changed documents resolves. Based on `main` at `88faac36`.

---


o2alexanderfedin (approved):
Approved. Delta measured against `origin/main` at `88faac36`. Suite 6862/6862 (+6, exactly the six named `fjs/types/set` leaves), `tsc` clean, `cov` 100%, `gen` leaves no diff.

The complexity claim holds on measurement, not on reading. Medians of five, same machine, main versus this head: 2,600 deep 403 ms → 59 ms, 5,000 1,504 → 113, 10,000 3,594 → 339, 20,000 20,575 → 539, 40,000 83,949 → 938. Main's 1,000→40,000 is 1,077× for 40× depth, against 1,600× predicted by n² and about 62× by n log n; this head's is 18×, and 5,000→40,000 is 8.3× for 8× depth. Your own figures reproduce to order, including all three 100k shapes.

Output is unchanged where it must be: 945 inputs across the conformance corpora and hand-built cases give 0 disagreements between the two revisions, 936 written and 9 refused identically, with a control corrupting every 37th reporting 26. Sharing decided the same on siblings, across depths, last-member-only, thrice in one array, root-shared, equal-but-distinct and a 50-deep chain — and `[p,p]` with `p=[c]` still leaves `c` inline, which is the occurrences-not-paths rule. Mutations redden `fjs/types/set/proof.f.mjs proof.carries` and `serializer/proof.f.mjs proof.sharing.occurrencesNotPaths`. The old §1 figures were reframed rather than left standing.

One nit: `fjs/media/datajs/todo/serializer.md:198` (not yours) still says "proved by the reader's own 5,000-level document written back byte for byte", while `proof.depth.writesBack` now writes 20,000.

PR #2049, opened 2026-09-14
fsc/edag: a program linked into one EDAG, its imports resolved and bound

sergey-shandar:
The second half of Stage 1 in [`compile-modules-to-edag`](https://github.com/functionalscript/functionalscript/blob/main/fjs/fsc/todo/compile-modules-to-edag.md), on top of #2046's `unresolved`: `resolve` links a program from its root path into one EDAG.

## What it does

`resolve(path)` is an `Effect<ReadFile, Exp, ParseError>`, the shape `transpile` has with a graph in place of a value. It reads and parses the module, resolves each of its imports the same way, recursively, and binds every import in its parameter's place. The result carries no path and no parameter: the `Unresolved` layer is the compiler's and is gone once the link is done.

- **Binding happens where a reference is lowered.** The importing module is lowered over the imported modules' EDAGs, so the graph is built once with the imported node where the parameter would be. A rewrite of a finished `Unresolved` was the alternative, and it needs a memo keyed by node identity to keep a shared node one, which is either a mutable map or a quadratic immutable one; lowering from the AST needs neither. The issue's Stage 1 text now says so, and the cache issue notes it owes that rewrite for a cached `Unresolved`, which has no AST.
- **One link, one node per module.** Resolved modules are memoized by path within the link, so a diamond of imports joins at one node — pinned by identity in the proof — and a module met again while it is still being entered is a `circular dependency`, as in `transpile`.
- **A `.json` import** is the tree its document denotes, as `transpile` reads it.
- **The record of resolved modules boxes each EDAG**, since a module's EDAG may be `null`, which is also what an absent record says; pinned by a module importing `export default null;` twice.
- **Failures are `transpile`'s**: a parse error at its token, a missing file or a cycle with no position, and an imported module's own refusal from `unresolved` is the link's.

The transpiler's module reader, import-path resolution and JSON reader are exported with the `_` prefix for the linker, and its own `.json` path now reads through the exported one; no behaviour of `transpile` changes.

## Verification

`tsc` clean; all 6863 proofs pass (9 new: alone, bound, diamond, paths, null module, JSON, and the refusals); `npm run cov` at 100% on every file; `npm run gen` unchanged; the broken-link count stays at its baseline of 9.

Four more Stage 1 tasks are ticked with citations. What Stage 1 still owes is the parser's `a.b` and `a[b]`, and the compiler path that writes the EDAG.

### Changelog

- `fsc/edag`: `resolve` links the program at a path into one EDAG, its imports resolved recursively and bound in place, memoized by path within the link.
- `fsc/transpiler`: `_parseModule`, `_importPath` and `_parseJson` exported for the linker.

---


o2alexanderfedin (approved):
Approved. Delta measured against `origin/main` at `88faac36`. Suite 6863/6863, `tsc` clean.

One node per module holds on every shape I could think to break it with: a diamond, the same module under two names, and two spellings that normalise to one file (`./b.f.js` versus `./sub/../b.f.js`, and `../dir/b.f.js`) all give one node, `===`-asserted, because `_importPath` normalises before the memo key. Cycles of length 1, 2 and 3 give `circular dependency` and a missing import gives `file not found`, both with `null` metadata. Deleting the memo lookup reddens `proof.resolve.diamond`; deleting the cycle check overflows the stack, so that guard carries weight. The shadowed-member rule still holds through your `transpiler` edit — `{a:s,a:s}` stays unshared and compiles, `{a:s,b:s}` still hoists.

One nit: the body says "9 new" proofs; the delta is 7, and seven named leaves were added.

PR #2050, opened 2026-09-14
datajs: the writer's issue retires, its parameter narrowed to the data model and the readable layout decided against for now

sergey-shandar:
The last of stage 4. `fjs/media/datajs/todo/serializer.md` held two decisions open and nothing else the stage still owed; both are taken here, recorded in the module README, and the file goes — "in the PR that finishes it", as its own last task said. Net −484 lines.

## Decision 1: the writer takes `Unknown`, not `unknown`

```ts
export const trySerialize: (value: Unknown) => Result<List<string>, string>
export const tryStringify: (value: Unknown) => Result<string, string>
```

The first draft took `unknown`, on the ground that refusing what is outside the model is the writer's job and a signature taking `Unknown` would assert what the writer has to check. That left a gap nothing could close: a corpus set is a DataJS data module, so no vector can carry a function, an accessor or a cycle, and `fjs/AGENTS.md` §1.6 keeps host-built values out of a `.f.mjs` proof — so the refusals only a host can reach had no proof but the one over the data such a value would carry (`_memberValue`, `_elementNames`, `_link`). The corpus README recorded that from its side as "the writer's decision to take".

Taken: the type is the contract at the call — a FunctionalScript caller cannot hand the writer a value outside the model, and `tsc` says so — and **the refusals stay**, because the specification refuses each of a hole, a symbol key, an accessor, a non-enumerable property, an extra own property on an array, a non-plain object and a cycle, and a host that casts is owed a refusal rather than a wrong document. No code path changes: `read`/`enter` keep `unknown` internally, since they see whatever a host hands in. The writer's proof hands its refusal cases cast (`asHanded`), as a host boundary would; the corpus's `normalize` proof and the matrix proof type what they hand as the values of the model they are.

The compiler already hands `Unknown` (`fsc`'s `Denotation`), so nothing there moves. No `**BREAKING CHANGES:**` entry: `fjs/media/datajs` has never shipped in a release.

## Decision 2: no readable layout, for now

Normalized form is what the compiler and hashing need, and nothing asks for another layout. The README's normalized-form section now says so: when something does ask, it's a new issue, and `tryNormalize` then names this writer.

## The file goes

Everything else it recorded is already carried — the two passes and why cycle refusal lives in `_link`, the descriptor-first read and the kind-before-descriptors argument, the refusals, no `sort` seam, normalized form, hoisting by occurrence, the depth contract, the cost — by the module header, the writer's JSDoc, the README's writer sections and `types.ts`. Its nine citers point where each thing now lives: the module README, the writer's proof comment, the corpus README (its "one question is open" paragraph is now the answer), `shared-empty-array.md`, `new-array-out-of-subset.md` (whose serializer task was gated on this decision), the YAML issue (to the spec's §What may be serialized), the plan, and two `not-applicable` reasons that named "this writer's recursion limit" as its bug — they say what is true now, and `matrix.md` is regenerated from them.

The plan ticks stage 4 and its Status says every stage has landed; retiring the plan itself is the next PR.

## Checks

`tsc` clean; `node ./fjs/module.mjs t` 6869 pass / 0 fail; `npm run cov` 100% lines/branches/functions; `npm run gen` leaves no diff beyond the regenerated matrix rows; every link and anchor in the changed documents resolves, and no `serializer.md` citation remains. Based on `main` at `112e2bce` (#2048's merge).

---
_Generated by [Claude Code](https://claude.ai/code/session_01TzayQtZmCzP9rtExwHXtF2)_


o2alexanderfedin (approved):
Approved. Delta measured against `origin/main` at `112e2bce`. Suite 6869/6869 at both revisions — delta 0, and no leaf added or removed. `tsc` clean, regeneration leaves no diff.

The deletion is clean on both counts that matter. Citations: 13 hits across 9 files at the base, 0 here, and a separate flattened sweep for the heading text `Layout and API` goes 5 → 0 — with a hard-wrapped `[…](./todo/\nserializer.md)` injected as control and caught by both sweeps. The two new anchors resolve by heading text, to `spec/datajs/README.md` §"What may be serialized" and `fjs/media/datajs/README.md` §"Every entry point is fallible, and the names say so".

The narrowing is real and type-level only: `unknown` → `Unknown` makes a `tsc` probe passing a plain `unknown` red here and green at the base, while runtime output is bit-identical across 17 cases — depths 5,000, 20,000 and 40,000 all `ok`, `{a:s,b:s}` still hoists where a single reference does not, and a cycle still refuses. No exported name is removed or renamed, and I checked the "never shipped" claim independently: `serializer/module.f.mjs` did not exist at the 0.49.0 release commit and was added on 2026-09-11, so nothing is owed a declaration.

One small thing: `todo/parser-serializer-restructure.md:3-4` still reads `**Priority:** P1 — stage 4 is urgent` directly above the `**Status:** done — every stage, 4 included, has landed` line this PR writes, and the "Item 2 is what to start" text below is stale the same way. You say that file is next, so it is short-lived, but the two lines disagree as of this commit.

PR #2065, opened 2026-09-16
docs: update DISOT overview

sergey-shandar:
## Summary

Update `todo/sergey/disot.md` as a concise overview of the DISOT model.

The document now builds the concepts in this order:

`documents + hashes -> metadocuments/history -> signatures + timestamps -> directories -> global names -> relative names -> snapshots -> Git`

The Git section treats Git primarily as a **decentralized, local-first document format and history model**, not as a protocol. It uses concrete mappings (`blob = document`, `tree = directory`, `commit = revision metadocument`) and states what Git already provides versus what DISOT adds.

Git already provides content-addressed documents/directories, revision and merge history, local repositories, transport/hosting ecosystem, and digital signatures. DISOT adds trusted timestamps, decentralized names independent of branches/hosts, and generalized lock/snapshot resolution. Git submodules are noted as a limited existing example of exact dependency pinning.

The main Git compatibility disadvantage called out is SHA-1: most existing Git services still use SHA-1 object IDs, so DISOT needs compatibility plus stronger hash identities/migration.

IPFS, Nostr, and AT Protocol are compared briefly by example and presented as complementary systems rather than replacements.

## Scope

Documentation only. The PR adds `todo/sergey/disot.md` and its linked `todo/sergey/doc-def.md`; the latter remains unchanged from the requested source branch.

Detailed wire formats and implementation decisions remain in the linked naming, timestamp-signature, CLI, and SHA-1 TODOs.

## Validation

The latest revision is intentionally much shorter and example-driven. Markdown structure and fenced examples were checked while editing. Full repository suites and Mermaid rendering were not run, so the PR remains draft.


o2alexanderfedin (approved):
Reviewed at `1956ebabde2902d9a165ca9fba9b6ed8a2aa04a6` (note: head moved during review — this is two commits, `86794328` + `1956ebab`, +474/−0). Measured in a detached worktree at that SHA against merge-base `30d98cce`.

`git diff --name-status` is exactly two added files, `todo/sergey/disot.md` (+435) and `todo/sergey/doc-def.md` (+39) — docs-only confirmed, nothing modified or deleted.

Nearly all of `disot.md` is plan (what DISOT will be), which is the point of a `todo/` file. I checked the parts that are testable against the tree:

- **Links.** All five internal targets resolve at this head: `./doc-def.md`, `../git-name-resolution.md`, `../git-trusted-timestamp-signatures.md`, `../../fjs/todo/disot-cli-epic.md`, `../git-sha1-collisions.md`. The `../../` depth from `todo/sergey/` is right. Positive-controlled with an injected missing path. The four external URLs return 200.
- **Companion-spec claims.** The embedded-evidence chain `A → B(+vnd.fjs.didsig) → C(+vnd.fjs.ttssig) → D(+gpgsig)` matches `git-trusted-timestamp-signatures.md` exactly, including the "`tts` means Trusted Time-Stamp" expansion and "didsig is DID-native, not OpenPGP-only". "Multiple recognized root encodings are an ambiguity to reject", root-only lookup, singular `name`, relative self-name needing one authenticated author DID, ancestry-scoped relinquishment, and archive-retention being open all match `git-name-resolution.md`. Scoped/recursive locks match its lock-map section.
- **Internal consistency.** The DAG, revision (`abcd…`/`bcd2…`/`d0c0…`/`r002…`), directory and snapshot examples agree with their diagrams and with each other; the 8-leaf Merkle proof (`h001`, `h01`, `h1`) is correct.
- **Cross-file.** `doc-def.md`’s document/file definitions and hash-identifier examples match how `disot.md` restates them; no term defined twice.

Gates from the worktree: `node ./fjs/module.mjs t` exit 0, 7052/7052 pass; `tsc -p .` exit 0; `node ./fjs/module.mjs ci` exit 0 with a clean staged diff. Check-runs at this head: 16 success, 2 still in progress (`node26`, `macos-intel`), 0 failures.

No factual errors found. Approving.

o2alexanderfedin (approved):
Approve. Delta measured against `1956ebab` (the head I approved); `git diff --name-status` over that range is exactly one modified file, `todo/sergey/disot.md` (−325/+248).

This is a condensing rewrite, so I re-verified every claim I had checked before **by content at the new head**, not by assuming it survived:

- **Kept and still correct:** the evidence chain `A → B (vnd.fjs.didsig) → C (vnd.fjs.ttssig) → D (optional ordinary Git signature)` matches the projections and the chronology block in `todo/git-trusted-timestamp-signatures.md`; `name` stays a single string in the example, matching "the initial P3 schema keeps `name` singular"; the relative-name expansion (`/did:example:alice/` + `./json` → `/did:example:alice/json`) matches the `Relative references` rule in `todo/git-name-resolution.md`; the revision, directory, name-revision and snapshot examples all agree with their own diagrams.
- **Dropped, not distorted:** the `tts` expansion, root-only `.disot.*` lookup, the multiple-encodings rejection, the relative-self-name rule, ancestry-scoped relinquishment, the open archive-retention question, and the 8-leaf Merkle proof. None was replaced by a weaker approximation that now reads false — the companion TODOs still own that detail.
- **"The name history is immutable"** does not conflict with ancestry-scoped relinquishment: `git-name-resolution.md` states that archiving does not delete history and that relinquishment is itself causal and recorded.

Git facts tested against `/usr/bin/git` 2.50.1, exit codes asserted: a submodule gitlink does pin an exact commit (`160000 commit <sha>`); a branch is a mutable ref and the displaced commit object survives; history round-trips through a bundle and a plain directory copy with no server (`fsck` clean); `gpgsig` is a real commit header. NIP-34 fetched and is indeed "git stuff". IPFS/IPNS, AT Protocol, and "most Git services still use SHA-1" are plausible but I did not verify them.

All 5 internal links resolve (sweep positive-controlled with an injected dangling link). Suite 7052/7052, `tsc -p .` clean, `ci` gate leaves no diff. All 19 CI checks green.

PR #2066, opened 2026-09-16
nanvm-lib: Array indexing and `.length` for `.`/`[]` (Stage 1)

sasha-gil:
## Summary

Stage 1 of `nanvm-lib/todo/member-access-operator.md`, the staged plan for the last unimplemented basic operator in the MVP roadmap (`.` / `[]`, the EDAG's `['.', receiver, index]`).

- `Array::member_access` (`vm/array/member_access.rs`, new, `pub(crate)`): an in-bounds integer key — given as a `Number` or its canonical decimal string — reads the element; the string key `"length"` reads `Array::length()`; any other key reads `None`. Never panics on an out-of-range index (`index < len` is checked before ever indexing; `Array`'s own `Index<u32>` is intentionally left panicking out of bounds, matching ordinary Rust/`Vec` convention).
- `Any::member_access` (`vm/any/member_access.rs`, new): the dispatcher — a nullish receiver throws the same `TypeError` `own_property` does, an `Array` receiver delegates to `Array::member_access` (mapping `None` to `undefined`), every other receiver is an explicit `todo!()` for Stage 2 (`String`) / Stage 3 (`Object` generalization).
- File layout mirrors `own_property`'s existing split between `Any::own_property` (dispatcher) and `Object::own_property` (mechanics), per the convention in `nanvm-lib/todo/operator-impl-placement.md`.
- `nanvm-lib/README.md`'s operator table and the plan doc updated to match.

Explicitly out of scope (see the plan doc's Non-goals): prototype-chain walking and built-in methods (`.map`, `.push`, `.slice`, getters), and the EDAG's chain-step nodes (`|.`, `?.`, etc.).

## Test plan

- [x] 12 new unit tests: numeric/canonical-string indexing, `.length`, non-canonical strings (`"01"`, `"+0"`, `"1.0"`, `" 0"`, `"-0"`, `""`), `-0` key parity with real JS, out-of-range, nullish-throws, and the Stage 2/3 `todo!()` gap
- [x] `cargo test -p nanvm-lib` — 197 passed (full suite, incl. doctests)
- [x] `cargo fmt -p nanvm-lib -- --check` — clean
- [x] `cargo clippy -p nanvm-lib --lib` — clean (the two `--all-targets` failures are pre-existing/unrelated, confirmed via `git stash`)

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Xjs4wQV6tLvf9fec4F4pom

---
[...]

PR #2067, opened 2026-09-16
nanvm-lib: String indexing and `.length` for `.`/`[]` (Stage 2)

sasha-gil:
## Summary

Stage 2 of `nanvm-lib/todo/member-access-operator.md`. Stacks on #2066 (Stage 1) — this diff is the incremental change on top of it.

- `String::member_access` (`vm/string/member_access.rs`, new, `pub(crate)`): mirrors Stage 1's `Array` arm — an in-bounds integer key (`Number` or its canonical decimal string) reads the single UTF-16 code unit as a one-character `String<A>`, matching JS `str[i]` (not `.charAt`, still out of scope); `"length"` reads the UTF-16 length; any other key reads `None`.
- A lone unpaired surrogate is read back verbatim, since `String<A>` stores and indexes raw UTF-16 code units without validating — same as real JS. Covered by its own test.
- `canonical_index`/`string_to_index` moved out of `vm/array/member_access.rs` into a new shared `vm/member_access.rs` (the same flat cross-type pattern `dispatch.rs`/`nullish.rs`/`primitive.rs` already use directly under `vm/`), so `Array` and `String` share one key-classification implementation instead of two.
- `string_to_index` rewritten to scan the key's UTF-16 code units directly via `String<A>`'s own `Index`/`SizedIndex`, rather than allocating a `std::string::String` and lossily decoding via `char::decode_utf16` just to check "is this a run of ASCII digits" — found in review, since this now runs on every string-keyed access to *either* receiver type, not just `Array`'s.
- `Any::member_access` now dispatches `Array` and `String` receivers; only `Object` (Stage 3) and the `Number`/`Boolean`/`BigInt`/`Function` fallback remain `todo!()`.
- `nanvm-lib/README.md` and the plan doc updated to mark Stages 1–2 done.

## Test plan

- [x] 11 new unit tests for `String::member_access` (code-unit indexing, `.length`, non-canonical strings, `-0` parity, out-of-range, the unpaired-surrogate case) plus 2 new dispatcher-level tests in `Any::member_access`
- [x] `cargo test -p nanvm-lib` — 208 passed (full suite, incl. doctests), up from 197 in Stage 1
- [x] `cargo fmt -p nanvm-lib -- --check` — clean
- [x] `cargo clippy -p nanvm-lib --lib` — clean
- [x] `string_to_index`'s rewrite manually verified behaviorally identical to the original across every branch (leading zeros, non-digit lead, overflow, empty string, surrogate case) before and after the fix, plus full suite re-run

---

https://claude.ai/code/session_01Xjs4wQV6tLvf9fec4F4pom

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xjs4wQV6tLvf9fec4F4pom)_


o2alexanderfedin (approved):
Approving.

Delta measured against `a2cbc120` — #2066's head and this PR's base branch, re-read live from the API and confirmed **not** yet merged (`merge-base --is-ancestor a2cbc120 origin/main` → rc=1), so `a2cbc120..6cb21d9d` (+265/−64, 8 files) is what I reviewed.

**Gates.** From `nanvm-lib/` at the head worktree: `cargo test` rc=0, `cargo test --release` rc=0, `cargo clippy -- -D warnings` rc=0, `cargo clippy --release -- -D warnings` rc=0 (cargo 1.95.0). This machine has a broken global `build.rustc = "rust-fv-driver"` in `~/.cargo/config.toml` that SIGABRTs; I bypassed it with `CARGO_BUILD_RUSTC=rustc`, so those codes are real. GitHub check-runs at this SHA: all 15 completed/success; `mergeable=true`, `mergeable_state=clean`. I did **not** run the i686 cross jobs — no toolchain here.

**String semantics.** `String<A>` is already UTF-16 code units, and `member_access` indexes them directly, which is exactly what JS `str[i]` does. Probed with added tests: `"héllo".length === 5` and `[1]` is `é`; `"\u{1F600}b"` gives length 3 with `[0]`/`[1]` the lone surrogates `D83D`/`DE00` and `[2]` `"b"` — agrees with JS and with the doc's stated contract; `"abc"[3]`, `[-1]`, `[0.5]`, `[NaN]`, `[Infinity]`, `[4294967296]` all `undefined`; `"0"` works, `"00"`/`"01"`/`"0.0"`/`" 0"`/`"+0"`/`"-0"`/`""` do not; `"length"` agrees with `.length`, and `""` gives 0. No panic path: `i < len` filters before `Index<u32>`, and `string_to_index` uses `checked_mul`/`checked_add`, so a huge digit key returns `None` rather than wrapping. Same in release.

**Refactor.** Ran an identical 27-key probe against `Array::member_access` at `a2cbc120` and at the head: byte-identical output, with a control mutation proving the comparison can report a difference. All ten array tests survive the move. Plan doc and README: every named path resolves, and the `dispatch.rs`/`nullish.rs`/`primitive.rs` precedent claim is true.

o2alexanderfedin (approved):
Approving the delta since my last review at `6cb21d9d`. That delta is exactly the merge commit `64eb554b`, which brings in `bb09db11` from the base branch `claude/compassionate-brahmagupta-p2fxwh` — the delta was measured against base SHA `bb09db118218aa34ab9189b5a07162f3e8eb59b5` (also the merge base), not main. Combined tree change: `nanvm-lib/todo/member-access-operator.md` only, +6 −4.

**The merge is automatic, not hand-authored.** Reproducing the 3-way merge of the two parents at their merge base (`a2cbc120`) with Git 2.50.1 `merge-tree --write-tree` yields tree `95bab013…`, byte-identical to the head's tree, exit 0 with no conflict. Controls confirm the harness discriminates: a deliberately wrong merge base yields a different tree, and a probe commit conflicting on the same file exits 1 with the three conflict stages listed.

**The merged doc still describes the tree.** The Stage-1 fix lands correctly on top of Stage 2: the doc now quotes the README's `.`/`[]` row as its *original* wording and says so explicitly, and that sentence is indeed absent from `nanvm-lib/README.md` at this head (positive control on a string that is present). Every named file and API in the doc exists at this head (22/22, bogus-path control negative): `vm/member_access.rs` (`canonical_index`/`string_to_index`), `vm/{array,string,any}/member_access.rs`, the two `Index` `TODO`s, `op2Id` containing `'own'`, `Index = number | NumberCast | string`, the `Chains` heading, and the reused `CANNOT_CONVERT_NULLISH_TO_OBJECT` constant.

**Gates green at the head** (worktree at `64eb554b`, `CARGO_BUILD_RUSTC=rustc`, exit codes asserted): `cargo test` 0 (208+1+5 pass), `cargo test --release` 0, `cargo clippy -- -D warnings` 0, `cargo clippy --release -- -D warnings` 0. The i686 cross jobs need a toolchain this machine lacks, so I did not run them. GitHub CI: all 14 checks pass.

sergey-shandar (approved):

PR #2070, opened 2026-09-16
edag: analysis, the table of a program for a writer and a VM

sergey-shandar:
The first task of [`fjs/edag/todo/analysis.md`](https://github.com/functionalscript/functionalscript/blob/claude/affectionate-carson-rwkrzh/fjs/edag/todo/analysis.md): one pure module, [`fjs/edag/analysis`](https://github.com/functionalscript/functionalscript/blob/claude/affectionate-carson-rwkrzh/fjs/edag/analysis/module.f.mjs), that reads an EDAG once and returns what a writer and a JavaScript-compatible executor both need, in a form that needs no identity-keyed structure downstream.

- **`analysis(exp)`** returns `{ root, nodes, scope, shared }`: every operation node once, in walk order — depth first from the root, operands as written, a node after its operands and on the first edge that reaches it — each naming its operands by index, `['#', i]`, with a primitive in place; the scope of each entry, `-1` at the module level or the index of the `=>` whose body holds it; and the shared entries in index order. `export default 1;` is an empty table with the root `1`.
- **Merged before counted, within one scope.** A node whose result identity is decided by its inputs — a plain read (`.`, `?.` without a continuation, `own`), an operator, the comma, `args`/`frame`/`undefined` — is the same entry as another in the same scope spelled the same over the same inputs, an entry by index and a primitive by `Object.is`, so `[cfg.a, cfg.a]` is one entry reached twice, `1 / 0` and `1 / -0` stay two, and two operators over `NaN` are one. A constructor, `[]`, `{}`, `=>`, and a call in any spelling, `()`, `?.()`, or a chain that continues, never merge.
- **Shared is written more than once.** An identity-minting entry is written once, hoisted where more than one place reaches it; a merged entry is written at every place that reaches it. So an entry's count is the sum, over the edges into it, of the places its parent is written, and `shared` is every entry counted more than once. This is what makes `[r, r]` over one access node of `a` and `[a.x, a.x]` over two the same table, with `a` hoisted in both, which the writer's round trip needs. An edge from a lazy position is an edge; the table records sharing, the executor decides when.
- **Refused.** A node reached from two scopes, which the compiler never emits and the EDAG's scope rule forbids, throws where it is met.
- **Types.** `Node` in [`types.ts`](https://github.com/functionalscript/functionalscript/blob/claude/affectionate-carson-rwkrzh/fjs/edag/analysis/types.ts) is derived from `Exp` position by position, so a node kind added to the EDAG is a node kind here with no second declaration; the three chain-step types collapse to one `Step`, since which step may follow which is the schema's question and neither consumer asks it. Four pins, each falsified once and seen to fail.
- **Proof** at 100% lines, branches and functions: walk order on the design's own example, constructors, reads, `Object.is`, operators and lazy edges, calls, scopes (frame in the enclosing scope, sibling bodies apart, a body in a body), every node kind and step, and the three ways to share across a boundary.
- The EDAG README and `execution-models.md` §2.2 point at the module; the todo's first task is ticked and its `shared` wording follows the count.

The accumulators are immutable and copied per node, so a build is quadratic in the node count — the sizes compiled today, with the DataJS serializer's finished list and one `Map` at the end as the shape to move to if it ever matters.

`tsc`, `fjs test`, `node --test` (7032 pass, 0 fail) and coverage of the new module at 100% all green; `npm run gen` changes nothing.

---


o2alexanderfedin (approved):
Approving — the table is correct as far as I could push it, and the proofs pin it.

**Mutation-tested the load-bearing branches** (one at a time in a worktree at `831b04f2`, each re-run with `node ./fjs/module.mjs t`, exit code asserted, no `SyntaxError` in any log, tree restored clean between runs). All six went **red**:

| mutation in `module.f.mjs` | result |
|---|---|
| `places`: `weight = mergeable(node) ? counts[i] : 1` → `weight = 1` | fail 4 |
| `analysis`: `n > 1 ? [i] : []` → `n > 0` | fail 6 |
| `twin`: drop `entry.scope === scope` | fail 2 |
| `mergeable`: `.`/`?.` `node.length === 3` → `=== 4` | fail 4 |
| `same`: `Object.is(a, b)` → `a === b` | fail 1 |
| `node`: delete the cross-scope `assert` | fail 3 |

So the count model, the shared threshold, the scope partition of merging, the chain-continues rule, `Object.is`, and the refusal are each pinned by a case that fails without them. The over-approximation in `places` (a hoisted merged entry still hands its operands its own count) is conservative in the safe direction — it can only hoist more, never less — and is exactly what makes `[r, r]` and `[a.x, a.x]` the same table, which `access` pins.

**Verified:** suite `pass: 7094, fail: 0`, exit 0; `tsc -p .` exit 0, positive-controlled by breaking `_Op1` in `types.ts` (TS2344) and `_Scope` in `private.ts` (TS2345) and seeing both go red. No `try`/`catch`, no regex literal or `RegExp`, no host globals in either `.f.mjs` (each grep positive-controlled on a fixture). `analysis` is the only export and both doc links resolve; the `[x]` in `todo/analysis.md` names nine proof cases and all nine are in `proof.f.mjs`. CI all green.

PR #2075, opened 2026-09-16
edag: the is operator, Object.is as an op2 node

sergey-shandar:
The node half of [`fjs/edag/todo/is-operator.md`](https://github.com/functionalscript/functionalscript/blob/claude/affectionate-carson-rwkrzh/fjs/edag/todo/is-operator.md): `['is', a, b]`, `Object.is` as an EDAG operation, the equality the language's guarantees are stated in — the analysis merges inputs by it, DataJS guarantees its normalized form by it — and one a program had no way to say, since `Object.is` is a function of a namespace the language cannot name as a value.

- **Schema.** `is` joins the `op2` ids in `fjs/edag`'s schema, `types.ts` and the README's node table, beside `===` and `!==`, with the `op2Id` doc saying what it is; the schema proof's literal id list pins it as it pins the other twenty-one.
- **Operations.** One row in the shared table, `Object.is`, so amnesia and memo get it together. Amnesia's proof pins the four cases beside `===`'s answers: `NaN` with `NaN` is `true` where `===` says `false`, `0` with `-0` is `false` where `===` says `true`, an object with itself `true` under both, two equal objects `false` under both. The operations and analysis proofs each add a line; the analysis's handler table gained the tag, which `tsc` demanded, as the `Over` derivation promised.
- **Native.** `nanvm-lib` has no SameValue operation yet, so the corpus in `fjs/nanvm` does not pin `is`: a case is pinned only where both executors can run it, and `is` joins the corpus with the Rust operation, as `===` did with `strict_eq`. The todo says so and keeps that task open.
- **What waits.** Lowering `Object.is(a, b)` to the node and the writer spelling it back both wait on calls in the language, so those tasks stay open and the file stays.

`tsc`, `fjs test`, `node --test` (7049 pass, 0 fail), coverage of the edag modules at 100%, `npm run gen` changes nothing.

---


o2alexanderfedin (approved):
Approving — the operator is genuinely pinned, not merely present.

**Mutation-tested, each restored and re-greened:**

- \`operations/module.f.mjs\`: \`is: o2(Object.is)\` → \`o2((a, b) => a === b)\` → **RED**, 2 failures, exactly the two leaves that should care: \`edag/amnesia/proof.f.mjs proof.op2\` and \`edag/operations/proof.f.mjs proof.operators\`. Both the \`NaN\` and the \`-0\` halves are load-bearing.
- \`analysis/module.f.mjs\`: deleting \`is: o2,\` from \`handlers\` → **RED** (`analysis/proof.f.mjs proof.operators`).
- \`analysis/module.f.mjs\`: adding \`case 'is'\` to \`mergeable\`'s no-merge list → **RED**, same leaf. So the new line pins the classification, not just the presence of a handler — and `is` merging is right: its result is a function of its inputs.
- \`module.f.mjs\`: dropping \`'is'\` from \`op2Id\` → **RED** (\`edag/proof.f.mjs proof.op2.ok\`).
- \`types.ts\`: renaming the member → \`tsc\` reports 10 errors including \`types.ts(356,26): Type 'false' does not satisfy the constraint 'true'\`, so the type and the runtime schema are tied.

**Checks:** `node ./fjs/module.mjs t` exit 0, pass 7111 / fail 0. `tsc -p .` exit 0. `node ./fjs/module.mjs ci` + `git add -A && git diff --cached` clean. CI all green.

**Docs verified:** the four id lists (`op2Id`, `Op2Id`, `op2Ids`, README row) agree member-for-member, 22 each, so `twenty-one` → `twenty-two` is right by measurement. `memo` and `amnesia` both import `operation`, so the `[x]` on the operations-table item holds. `nanvm-lib` has no SameValue op and `fjs/nanvm`'s corpus has no `is` — as the Native bullet says; `strict_eq` exists in `tests/test/harness.rs`. Every link in the touched docs resolves.

No findings.

PR #2079, opened 2026-09-17
effects/node: one readChunks for both fjs/cas loops

o2alexanderfedin:
Replaces two duplicated file-reading loops in `fjs/cas` with one reusable
`readChunks`, then adds what a response with a declared length needs: a byte
bound, and an advance by the length actually read. It does not implement HTTP
streaming; it supplies one building block.

The advance is the part worth a second look. Both old loops stepped by
`chunkBytes` whatever came back — sound while an empty read ends the fold,
since a short read is then the last one. Under a bound it is not, and the
recipient gets the promised number of bytes with a hole in the middle that
nothing downstream can detect.

One Codex finding is declined rather than fixed: a purely computational source
with a large bound overflows the stack. The recursion is unchanged from the
loops it replaces — `nonEmpty` takes its tail as an ordinary argument
(`effects/list/module.f.mjs:31`), so a chain of pure cells is built before the
first is pulled. What this change adds is the source parameter that lets a
caller reach it; every source with a host behind it performs a command and
suspends, verified at a 13 GB bound.

Seven proof leaves, each watched to fail. Advancing by `chunkBytes` reddens
`boundedAdvancesByWhatItGotNotByChunkBytes` alone; ending the stream instead of
failing it reddens `aBoundedStreamThatEndsShortFailsTheCell` alone. That second
one passed for the wrong reason first — it read only the first cell and the
refusal lives in the second — so the helper now drains. `tsc` 0 on the pinned
7.0.2, 7175 proofs pass.


sergey-shandar:
@codex review

sergey-shandar (approved):

sergey-shandar:
This PR replaces two duplicated file-reading loops with one reusable readChunks function, then adds the behavior needed for future HTTP responses with a declared length. It does not implement HTTP streaming yet; it supplies one building block.

sergey-shandar (changes requested):
Update title and description. More signal less noise. Keep it simple.

sergey-shandar:
@codex review

o2alexanderfedin:
Rewrote both. Title drops the bound clause; description opens with the three sentences you gave and keeps only the defect, the one finding I declined, and what the tests caught.

Two claims from the commit messages I dropped as false, having checked: only `types.ts` was reworded for the `readBytes` fill loop, not `module.f.mjs`; and "the pre-existing implementation overflows identically" is true of the recursion shape but not of the old code, which hardwired `readBytes` and had no bound, so nothing could drive it purely. The description says what is actually true instead.

sergey-shandar (approved):

PR #2093, opened 2026-09-17
nanvm-harness: wire the Rust code generator into the harness

sasha-gil:
## Summary

#2 (harness + walking skeleton, PR #2071) and #3 (Rust code generator, PR #2073) were built independently and never connected — `nanvm-harness`'s own doc comment still said the generator "does not exist yet," so the harness kept running hand-written `synthetic/` stand-ins instead of real `fjs compile` output. This closes that gap, the literal remaining piece of `nanvm-lib/todo/mvp-roadmap.md`'s MVP definition.

- `nanvm-harness/fixtures/{number,boolean,string}.mjs` are now real inputs to `fjs compile`, not "documentation only" — each compiles to a sibling `.rs` file (generated, committed, marked `linguist-generated=true`).
- `npm run gen` grew three `fjs compile` invocations, so the existing whole-tree CI drift check covers them with zero new CI wiring.
- `nanvm-harness/src/lib.rs`/`main.rs` pull in `number`/`boolean`/`string` via `#[path = "../fixtures/*.rs"]` instead of the hand-written `synthetic/` module, which is deleted outright.
- Stale "doesn't exist yet" / "documentation only" comments are gone, including a cross-file one in `nanvm-lib/src/vm/any/to_json.rs` that a self-review caught referencing the old "synthetic fixtures" framing.
- Ticked the now-complete roadmap items in `nanvm-lib/todo/mvp-roadmap.md` and `todo/fjs-nanvm-integration.md`, and fixed the pre-existing stale "Complete all basic FunctionalScript operators" checkbox (already `[x]` per `nanvm-lib/README.md`'s own operator table).

## Test plan

- [x] `cargo test --workspace` — all suites green (harness's 3 constant-export assertions now produced by the real generator+harness pipeline, `nanvm-lib`'s 235 unit tests, the generated operator corpus, doctests)
- [x] `cargo run -p nanvm-harness` — prints `42`
- [x] `cargo fmt --check` / `cargo clippy --workspace -- -D warnings` — clean
- [x] `npm run gen` re-run — zero drift (`git diff --cached --exit-code` would pass)
- [x] `node ./fjs/module.mjs t` — full FJS proof suite, 7486/7486 passing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CDgWHWHmnhTAmd9df7MTYi

---
_Generated by [Claude Code](https://claude.ai/code/session_01CDgWHWHmnhTAmd9df7MTYi)_

PR #2098, opened 2026-09-18
Release 0.50.0

sergey-shandar:
Bumps `package.json` to `0.50.0` and adds `changelog/0.50.0.md`. Merging to `main` triggers the `npm publish` workflow. No unrelated work: three files, and `changelog/unreleased/` does not exist to delete.

**The title is `Release 0.50.0`, not "Release version 0.50.0".** [`RELEASE.md` step 1](https://github.com/functionalscript/functionalscript/blob/main/changelog/RELEASE.md#1-find-the-release-boundary) has the next release author find this boundary with `git log --first-parent --format='%H %s' | grep 'Release 0.50.0'`, and the original title does not contain that substring, so the search would return nothing — the same empty result a wrong boundary gives. Renamed before anything depends on it.

## Window

198 pull requests, between `46f62f17` (`Release 0.49.0`) and `8446f9ab`.

The boundary was **confirmed by its `package.json` diff**, not its title — `git show --first-parent 46f62f17 -- package.json` shows `0.48.0` → `0.49.0`. **No line in the window lacks a `(#NNN)`**, so step 2's authorless-commit rule had nothing to do.

The count was cross-checked against GitHub, and the two disagree in a way worth writing down rather than waving at. `is:merged base:main merged:<window>` returns **192**, which is 191 in-window plus `#1851`, the boundary release itself. The six the search misses each have a reason:

- **Five had a base that is not `main`** — #1936, #2013, #2022, #2038 and #2095 were stacked on another branch and merged into it, so `base:main` never matched them. Their merge commits are still on `main`'s first-parent walk, which is why the `git log` listing is the authority and the search is only the cross-check.
- **One merged after the range ended** — #2093's `merged_at` is `05:06:13Z`, past the tip commit's `05:00:15Z`.

191 + 5 + 1 = 197, plus #2092 below = 198.

## Minor, not patch

Sixty merge bodies mention `**BREAKING CHANGES:**`. Four are **not** declarations, and each says so in as many words: #1852 is documentation-only, #2043 and #2050 state that nothing under the old names ever shipped in a release, and #2080 is a spec document assigning the declaration to the pull request that will implement the rule. That leaves **57 declarations**, and step 4's check — that they were not undone inside the window — leaves plenty standing: `fjs/bnf` and `fjs/djs` are gone from the tree, `Tuple` is `FixedArray`, `types/range_set` is a toggle list, `NodeOp` carries `Inflate` and `ReadWhole`, and `fjs/media/json/tokenizer` is absent.

Pre-1.0 the minor is the only upgrade boundary a resolver enforces, so one surviving break moves it: `0.49.0` → `0.50.0`.

## What the window undid, and therefore is not in the notes

This is the step that per-pull-request entries could not do, and this window leans on it hard. **Four modules are new here**, so a declaration against a shape they held earlier in the window is invisible to a consumer of `0.49.0`:

- `fjs/ebnf` — `ebnf/map` arrived (#1886) and was retired (#1907); `ebnf/meta` arrived (#1905) and was renamed `ebnf/ast` (#1907). Neither shipped. What did is `fjs/bnf`'s deletion.
- `fjs/git` — `packDir` (#2041) became `packsIn` (#2062) before either shipped.
- `fjs/media/datajs` (only a `todo/` file at `0.49.0`) — the `vectors` type churn of #1973, #1992, #1997 and #1999 is internal, and #2043 and #2050 say so themselves.
- `fjs/website/page` and `fjs/website/demo` — so `Dir.demo` (#2002) and `DemoEvent`'s `click` arm (#2017) are churn. `website`'s `main` did exist, and its break is an entry.

Within `fsc` the same applies to `.edag.f.js` → `.edag.data.js` (#2051, #2084), `hasAccess` added and removed (#2053, #2054), and `unreached` → `anchors` (#2046, #2057).

One entry is deliberately cited against a change that was itself superseded: #1894 declared that `media/json`'s `tokenize` reports a malformed string literal as errors alone, and #2035 then removed `fjs/media/json/tokenizer` outright. The net effect a reader needs is the removal, so that is the entry, and #1894 is cited under it rather than dropped.

## #2092 landed while this branch sat, and that is the whole point of step 7

`main` moved between the first scan and the merge-time re-scan: [#2092](https://github.com/functionalscript/functionalscript/pull/2092) (unary `-` as a grammar-level prefix) merged at `08:00Z`. It ships with this release whether or not this branch ever merged it, so the re-list is not a formality — it added three breaking entries (the tokenizer never signs a numeric literal, a negative number is an operation in the AST, `a[-1.5]` no longer parses) and the unary `-` note on the language entry. The branch was rebased onto it and every gate re-run.

It also retired two refusals this window had just introduced — `access on a numeric literal` (#2060) and `call on a numeric literal` (#2087) — which is step 4 again, in a single window.

## Grouping

54 entries for 198 pull requests. Roughly half the window produces none: designs, `todo/` files, documentation, proofs and internal refactors are invisible to a user of the package.

Where several pull requests moved one thing, the entry describes where it landed. `fjs compile`'s output is the clearest case — #2031 made the module output normalized DataJS, #2051 added an EDAG output under `.edag.f.js`, #2073 added `.rs`, and #2084 routed everything by extension and renamed the EDAG output. One entry says what the compiler does today:

> FunctionalScript for `.js`, DataJS for `.data.js`, the EDAG for `.edag.data.js`, Rust for `.rs` and JSON for `.json`, with `.mjs` spellings too; anything else is refused.

**No migration was invented.** Every breaking entry carries the one its declaring pull request gave — "Replace `Operand` with `Value`", "Build the shared nodes with `sharedExp(data.shared)`", "test `kind` against `fjs/js/keywords`" — and where none was given the entry says what changed and stops. The `ci` entries owe none at all, and the leading one says so.

Entry length was held to the documented budget: all 54 fit in 262 characters, against `changelog/README.md`'s ~250 and `0.49.0`'s longest of 260, and the file wraps at 80 columns as `0.49.0` does.

## Step 7 checklist

- [x] the version in `package.json` matches the changelog file name
- [x] every pull request in the window was read, and the count was cross-checked — with the six-way discrepancy resolved above rather than rounded off
- [x] every `**BREAKING CHANGES:**` declaration is either in an entry or accounted for as undone
- [x] **immediately before merging, fetch and re-run step 2 against `origin/main`** — run twice here. The first re-list caught #2092 and changed the notes; the second, after rebasing, found the window unchanged at 198 and `origin/main` still `8446f9ab`. The reviewer-facing half stands: a pull request landing on `main` while this sits open ships with this release whether or not this branch merged it, so **re-fetch and re-list immediately before pressing merge** — `origin/main` moves only on fetch, so re-reading it without one is a test that cannot fail
- [x] the tip scanned is `8446f9ab`
- [x] the final scan compared the new `ls-tree` against step 3's record — both empty, which is the comparison at full stretch and not an exemption from it: nothing recorded and nothing now means no late arrival and no deletion
- [x] `changelog/unreleased/` is deleted in this same pull request, if it exists — it does not exist on `origin/main`, so there was nothing to read into the entries and nothing to delete

## Verification

`tsc` clean. `npm run gen` produces no drift.

Both suites `AGENTS.md` requires pass, on the rebased tree, and both numbers are quoted because they discover different sets: the repository's own runner, `node ./fjs/module.mjs t`, reports **7501 tests, 0 failures**; `node --test` reports **7439 tests, 0 failures**.

No `Changelog:` section: a release pull request ships no code of its own.

https://claude.ai/code/session_0136Wq1FxvNVyLhDcrnNMDsr


sergey-shandar:
@codex review

o2alexanderfedin (commented):
Already merged — this is a follow-up note, not a merge blocker. One entry promises a refusal the released compiler does not perform.

**`-NaN` is not refused.** The `NaN`/`Infinity` entry ends "`-NaN` is refused", cited to #2013 and #2015. That refusal was retired later in the same window by #2092, which this file cites three bullets earlier: its table reads `-NaN` | before `unexpected token` | now `['-', NaN]`. At the release head `db53b0ae`:

```
export default -NaN;   ->  .data.js  rc 0  export default NaN;
                       ->  .js       rc 0  export default NaN;
                       ->  .edag.data.js rc 0  export default NaN;
                       ->  .rs       rc 0  (f64::NAN).to_any()
                       ->  .json     rc 1  no JSON spelling for NaN
```

The only refusal left is the JSON one, and plain `NaN` gets it too — it is the entry's own first clause, not a sign check. A control source (`export default @;`) exits 1 with `unexpected token`, so the harness can report a refusal.

This is the case RELEASE.md step 4 names: "a break introduced and reverted inside the window is **not a break** — not in the entries and not in the version number." The clause reads as a live restriction to anyone migrating. Suggested fix on `main`: drop the four words, leaving "…written as those words where a non-finite number was written `null`. They are reserved words, so binding one is refused." (`const NaN = 1;` → `reserved word`, confirmed.)

Everything else checked clean. Breaking sweep over the 849 merges in `ff4d1a71..8446f9ab`: 63 carry `**BREAKING CHANGES:**`; every declaration not in the file is on a module that did not exist at the `0.49.0` boundary (`fjs/ebnf`, `fjs/git`, `fjs/media/datajs` and its vectors, `website/demo` — verified by `ls-tree` at `ff4d1a71`) or on `fjs/djs`, whose removal the file declares. The #2092 bullets carry the corrected wording, and I confirmed it: `export default -1;` folds to the leaf, `(-1f64).to_any()` at rc 0.

Also verified by running: the eight accepted output extensions and `.txt` refused; all six `.json` refusals at exit 1 with no file; member order preserved; `;` required; `a[-1.5]` refused and `a["-1.5"]` accepted; self-naming `const` → `const not found`; every named export move in `fsc/*`, `types/*`, `ci/rust`, `js/prototype`, `path`, `effects/node`, `emergent_testing/browser`, `nanvm` and `edag/*`. Version bump agrees in all three places. Suite 7501 pass / 0 fail, `tsc -p .` rc 0 (red on an injected error), `fjs ci` no drift.

PR #2102, opened 2026-09-18
nanvm-lib: implement ECMAScript integer-keys-first object enumeration order

sasha-gil:
## Summary

Follow-up to functionalscript/functionalscript#2095, which recursed `Any::to_json` into arrays/objects but refused any object holding an array-index-like key (`JsonError::IntegerLikeObjectKey`) rather than risk printing member order that disagrees with `JSON.stringify`. That reviewer's own approval on #2095 noted the refusal cost 72 of 587 differential test cases that would have serialized *correctly* anyway (their insertion order happened to already match ECMA order) — this closes that gap by implementing the real ordering instead of refusing it.

- `nanvm-lib/src/vm/any/to_json.rs`: `object()` now partitions its deduped keys (last-write-wins, first-insertion position — unchanged) into array-index keys and the rest, sorts the array-index group ascending by numeric value, and prints that group first — matching ECMAScript's `[[OwnPropertyKeys]]` order, which `JSON.stringify` inherits.
- `is_array_index_key` (bool) becomes `array_index_value` (`Option<u32>`), reusing the parsed value for the sort instead of re-parsing.
- `JsonError::IntegerLikeObjectKey` is removed — no object shape refuses serialization anymore, only individual unrepresentable scalars still do (`undefined`, `BigInt`, non-finite numbers, functions). **Breaking change to a public, re-exported enum** — see the `Changelog:` section below.
- Tests: replaced the three refusal tests with positive ordering tests (ascending-ahead-of-other-keys, numeric-not-lexicographic sort, zero-sorts-first, duplicate-integer-key dedup-then-sort interaction, nested-in-array now serializes); kept the three "not an array index" boundary tests (`"01"`, `"4294967295"`, overflow) since that classification logic is unchanged.

## Test plan

- [x] `cargo test --workspace` — all suites green (`nanvm-lib`'s 250 unit tests, `nanvm-harness`'s 7 fixture tests unaffected since none use integer-like keys, doctests)
- [x] `cargo run -p nanvm-harness` — prints `42`
- [x] `cargo fmt --check` / `cargo clippy --workspace -- -D warnings` — clean
- [x] `npm run gen` re-run — zero drift
- [x] `node ./fjs/module.mjs t` — full FJS proof suite, 7501/7501 passing

Changelog:
- **BREAKING CHANGES:** `nanvm-lib`: `JsonError` drops the `IntegerLikeObjectKey` variant — `Any::to_json` now serializes objects with array-index-like keys in ECMAScript's integer-keys-first order instead of refusing them; an exhaustive match on `JsonError` needs updating

https://claude.ai/code/session_018WGrCcsz8rxTbGHvoWWNrk

---
_Generated by [Claude Code](https://claude.ai/code/session_018WGrCcsz8rxTbGHvoWWNrk)_


sergey-shandar:
@codex review

sergey-shandar (approved):

PR #2108, opened 2026-09-19
Add Stage 1 AOT callable functions

sasha-gil:
## Summary

Add the Stage 1 foundation for callable-function support in the FunctionalScript Rust compiler.

The change introduces the staged design in `nanvm-lib/todo/callable-function-objects.md` and implements the first AOT slice: a capture-free function export is lowered to a generated Rust `f0` with the ABI

```rust
fn f0<A: IVm>(args: &Array<A>) -> Result<Any<A>, Any<A>>
```

The generated module invokes that function through the same fallible `Result` ABI.

## What changed

- Add the detailed staged callable-function specification and Stage 1 boundaries.
- Extend the shared EDAG-to-Rust printer with explicit contexts for:
  - argument references via `args.clone()`;
  - fallible lowering and `?` propagation;
  - the legacy function-value placeholder used only by the operator corpus.
- Generate capture-free root functions and rest-argument functions in `fjs compile` Rust output.
- Change generated Rust module output to `Result<Any<A>, Any<A>>` so errors from function bodies and property access are preserved.
- Refuse unsupported function values and unsupported receiver/index shapes instead of emitting misleading or panicking Rust.
- Add function and rest-function fixtures, harness coverage, exact generator proofs, and update generated expectations and documentation.

## Scope and follow-up

This is Stage 1 only. It does not add runtime `Function<A>` construction or dispatch, capturing closures, dynamic calls, recursion, or general static-call lowering through arbitrary module graphs. Those remain staged follow-up work described by the specification.

The implementation currently covers capture-free root function exports, including fallible arithmetic, property access, and nested fallible values in generated bodies.

## Validation

- `tsc --noEmit`
- `npm run gen`
- `npm test` (7,512 tests)
- `git diff --check`

Rust validation (`cargo test -p nanvm-harness`, `cargo clippy`, and `cargo fmt -- --check`) could not be run in the development environment because `cargo` was unavailable on `PATH`; CI should provide the authoritative Rust validation.


sasha-gil:
Compared this against the detailed Stage 1 implementation spec I wrote in #2107 ([`compile-noncapturing-functions-to-rust.md`](https://github.com/functionalscript/functionalscript/blob/claude/zealous-pasteur-t59fvh/fjs/fsc/todo/compile-noncapturing-functions-to-rust.md)). Sharing the comparison here since it's concrete and testable against this branch; not asking for any specific change.

**Relationship**: this branch's `callable-function-objects.md` is a snapshot of #2107 mid-flight (before the citation fixes in `86c2103`), so the shared design doc lines up. The two efforts then went different directions on what "Stage 1" means in practice, and neither currently covers what the other does.

**The main thing I'd push back on**: `functionRoot` treats a bare exported function (`export default () => 42;`) as something the *code generator* should auto-invoke with zero arguments, baked into `pub fn module` itself. In real JS, that module's default export **is the function**, not `42` — a JS engine importing it gets the function object. `console-program.md`'s "run it if it's a function" is described as the *harness's* convenience when driving execution, not a property of module compilation, and this repo's own principle ("every FunctionalScript module... denotes in FunctionalScript the value that a JavaScript engine gives it") argues for refusing this case at Stage 1 (no `Function<A>` exists yet to represent it faithfully) rather than silently computing a different value. My spec refuses a bare exported function and requires the source to spell an explicit call (`const f = (...a) => …; export default f(41);`) instead — which is also why, conversely, **this branch has no call-node (`['()', …]`) handling at all**: the two PRs implement non-overlapping halves of Stage 1 (this one: bare-function-as-root; #2107's spec: explicit static calls).

**Confirmed gaps against my Task list** (independent of Copilot's findings below):
- Task 1 (make the pipeline fallible): only the new function-body path is properly fallible. The pre-existing `Any::own_property(…).unwrap()` for ordinary module bodies is unchanged — just wrapped in `Ok(…)` — so a property-access failure there still panics.
- Task 3 (`args[i]`): not implemented — no change to `indexExpr`'s numeric-index refusal, so only whole-`args` forwarding works, not indexed reads.
- Task 4 (scope-aware `let`-sharing inside a function body): `functionLines` hardcodes `shared: []`. The language's own canonical example (`spec/README.md`'s `const pair = [first, first]`) isn't attempted or tested, and would currently double-construct rather than share.

**Your own Copilot review already caught the sharpest bugs** (`args.clone()` missing `.to_any()`, a nested `.` receiver not routed through `nested()` in fallible mode, and `functionRoot` only matching when the arrow is the *literal* root) — I independently hit that same `functionRoot` issue by reading `resolve` in `fjs/fsc/edag/module.f.mjs`: it wraps an export in a `,` node ahead of the real value whenever an earlier `const`/import goes unreached, so a shape check on the root refuses an otherwise-trivial module with one unused import beside its function export. I've folded that + the fallible-composition issue into the spec as explicit design constraints (`c969669`), since a general reachability walk sidesteps both for free instead of pattern-matching the root shape.

CI is currently red on 8 jobs, consistent with the `args.clone()` type error.

---

o2alexanderfedin (commented):
**1. A property chain inside a function body compiles to Rust that does not build.**

`export default (...args) => [({a:{b:1}}).a.b];` emits `Any::own_property(Any::own_property(…, string_any("a")), string_any("b"))`. The inner call is `Result<Any<A>, Any<A>>`; the outer wants `Any<A>`. rustc gives `error[E0308]` and suggests the missing `?` (reproduced by dropping the output into `nanvm-harness` as a module). In `fjs/edag/rust/module.f.mjs`'s `'.'` branch the base is printed with `f(a)`, not `nested(a)`, so the `?` that `nested` exists to append is never added in a fallible body. #2107 (`80b535a1`) names this exact case — "Fallibility composes, and the printer must thread it through every operand position", Task 1, `{ a: { b: 1 } }.a.b`. Changing that one call to `options.fallible ? nested(a) : f(a)` produces correct Rust **and leaves the suite at 7512/0** — nothing covers a `.` chain in a generated body. (Mutation control: `${options.args}.clone().to_any()` → `Array::default().to_any()` reddens `fjs/fsc/rust/proof.f.mjs` `generateFunction` and `generateFallibleNestedValues`, 7510/2.)

**2. Both red checks are this PR's own.** `wasm`: `cargo fmt -- --check` exits 1 on `nanvm-harness/src/lib.rs` — the `run` signature and the `use crate::{…}` line exceed rustfmt's width (reproduced locally). `node26`: `npm run cov` demands 100% branches and reports 99.93, from exactly the two files changed here — `fjs/edag/rust/module.f.mjs` 97.21%, `fjs/fsc/rust/module.f.mjs` 97.73%.

**3. No `**BREAKING CHANGES:**` block.** Every emitted module's signature changed from `pub fn module<A: IVm>() -> Any<A>` to `Result<Any<A>, Any<A>>`, and `nanvm-harness::run`'s parameter and error type changed with it.

**4. Two citations in `callable-function-objects.md` do not hold.** `a => b => a + b` is called a worked example "in function-frame and edag-stage1-discussion"; it is in neither (it lives in `todo/demo/README.md`). 3111's worked example is `const a = i => b(i + 3)` / `const b = i => i % 5 === 0 ? i : a(i)`. And subject 10's `a`/`b` are `const a = 0` / `const b = 3`, plain numbers — its mutual-recursion example is `const x = { a: () => x.b(), b: () => x.a() }`.

Green here: `cargo test`, `cargo test --release`, `cargo clippy -- -D warnings` and its release twin all exit 0; `node ./fjs/module.mjs t` 7512/0; `tsc -p .` 0 with a control failure; regeneration leaves no diff. The i686 cross jobs were not run locally.

o2alexanderfedin (commented):
Status only — my findings from `7d9a2650` stand unchanged; this is about something outside the diff.

The branch no longer merges. GitHub reports `mergeable=false, dirty`, and reproducing it with `merge-tree --write-tree` against `main` at `b444eb33` gives content conflicts in `fjs/edag/rust/module.f.mjs`, `fjs/fsc/module.f.mjs`, `fjs/fsc/proof.f.mjs`, `fjs/fsc/rust/proof.f.mjs` and four regenerated `nanvm-harness/fixtures/*.rs`. The branch is a long way behind — the operator work, the parser's block-and-return change and the Rust string escaping have all landed since it forked, and they touch the same lowering this PR rewrites.

The two red checks are the same ones I reported: `wasm` is `cargo fmt --check` on `nanvm-harness/src/lib.rs`, and `node26` is the branch-coverage gate at 99.93%.

sergey-shandar:
@sasha-gil the PR has a lot of conflicts and fails on CI. Should we redo it on top of main?

sergey-shandar:
@codex review

sergey-shandar (approved):

PR #2109, opened 2026-09-19
website: pad a section's list links for a finger, not a mouse

Trinidadec:
Implements the touch-targets todo: pads a section's list links to 27px under `@media (any-pointer: coarse)`, leaving desktop lists dense.

Chose `any-pointer: coarse` over `pointer: coarse` per the todo's own flagged gap: a touch-screen laptop's primary pointer is its trackpad, so `pointer: coarse` would leave its touchscreen at 19px.

Verified with Playwright against the generated site: 27px at a 390px touch viewport, 19px unchanged at 1280px desktop (screenshots attached below).

Retires `touch-targets.md`; the reasoning moves to the website README per AGENTS.md's documentation table.

`node --test`: 7509/7509 passing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2110, opened 2026-09-19
website: a favicon, "fs" drawn as path strokes

Trinidadec:
Implements the favicon todo: a script "fs" ligature — the project's own initials — drawn as round-capped path strokes (geometry, not a character in a font), taking --link's two colours via the SVG's own prefers-color-scheme query.

`favicon.ico` at the root, `favicon.svg` in `fjs/website/`, both committed static files linked from every page head, per the todo's own reasoning for why both must be declared.

Also fixes a real bug found along the way: `.gitattributes`' blanket `text eol=lf` rule was silently stripping the CR out of the embedded PNG's magic-byte CRLF, corrupting the binary file in the index. Added `*.ico binary` and verified the committed blob is byte-identical to what was generated.

Retires `favicon.md`; the reasoning (why geometry not a font glyph, why both files) moves to the website README per AGENTS.md's documentation table.

`node --test`: 7509/7509 passing, coverage 100/100/100 on the touched files.


Trinidadec:
Redrew the mark: it's now a script "fs" ligature (the project's own initials) instead of a lambda, per feedback after the first push. Same construction — round-capped path strokes, no font dependency — verified legible at a true unscaled 16px render before committing.

o2alexanderfedin (commented):
The favicon lands, and the page emits what the retired design document asked for. Reviewed at `5cdb19f8` in a worktree; delta measured against the merge base `89e6502f`.

**The drawing is real.** `fjs/website/favicon.svg` rasterises (rsvg-convert) to a green script "fs" — clear at 32px, thin but still readable at 16. Every path coordinate plus the 1.8 stroke sits inside `0 0 24 24`; `stroke` comes from the embedded `<style>`, `fill="none"` is on each path, and nothing reaches for an external resource. `favicon.ico` is a valid two-entry ICO (16 and 32, PNG-embedded, every chunk CRC intact — so `*.ico binary` did its job), and its dominant pixel is `(19,115,51)` = `#137333`, the light `--link` the README claims. The two files draw the same mark.

**The emitted head names files that exist.** Importing `page` gives `<link rel="icon" href="/favicon.ico" sizes="32x32">` and `<link rel="icon" type="image/svg+xml" href="/fjs/website/favicon.svg">`; both paths resolve in the tree.

**Retiring `todo/favicon.md` left nothing dangling.** No reference to it survives anywhere (exit code asserted; a `touch-targets.md` control hits). A tree-wide markdown link sweep reports the same 9 pre-existing breakages at this head, at the merge base, and at `origin/main` — an injected fake link takes it to 10, so the sweep can count.

**Gates**: suite 7535/0; `ci` regenerates nothing; `tsc -p .` (7.0.2) rc 0, and rc 1 when I put a `42` into `faviconLinks`. All checks green. `faviconLinks` is added and nothing is removed, so no published break.

## One gap: the root page's copy is unasserted

`page.head` reddens for a directory page — I pointed the SVG link at a path that does not exist and it failed. But deleting `...faviconLinks` from `rootPage` in `fjs/website/module.f.mjs` leaves the suite at 7535/0. The site's front door is the one page that could lose its icon without a test noticing.

Trinidadec:
Merged main back in — GitHub auto-merged it into this branch again and got a delete/modify conflict wrong the first time (resurrected fjs/website/todo/touch-targets.md, which #2109 already deleted, and dropped 30 lines from the README). Redid the merge properly: touch-targets.md stays deleted, both README sections (favicon + touch-targets) present, no duplication. 7562/7562 passing, favicon.ico verified byte-identical to the generated original.

Trinidadec:
Fixed in 7c3101198: added the same assertion the directory-page head test has, to the root-page end-to-end test in fjs/website/proof.f.mjs. Verified it actually catches the regression first — removed faviconLinks from rootPage, confirmed proof.run() fails on exactly the two new lines, then restored it. 7562/7562 passing, coverage still 100/100/100 on the touched files.

sergey-shandar:
We need to find the original icon that we are using for https://github.com/functionalscript

Or, use a new icon everywhere:
- GitHub org,
- BlueSky account.

o2alexanderfedin (approved):
Delta measured over `5cdb19f8`; the branch's own new commit is just `7c310119`.

**My open finding is closed, and for the right reason.** Deleting `...faviconLinks` from `rootPage` now reddens `import("./fjs/website/proof.f.mjs").proof.run()` — 7561 pass / 1 fail, against 7562/0 clean. The new assertions read the root page's own emitted `index.html` (`pageAt(generated.root, [])`), not some other page's.

I also checked the guard is not a one-mutation guard. Replacing the spread with an inline copy that is wrong in one way each — `href` pointing at `/favicon-nope.ico`, `rel` changed to `shortcut icon`, the SVG MIME type changed to `image/svg`, the SVG link dropped entirely — reddened the same leaf every time (7561/1, no `SyntaxError` in any log). Restored and re-greened at 7562/0 after each.

The favicon assets and the page wiring are untouched since `5cdb19f8`: `favicon.ico`, `fjs/website/favicon.svg` and `fjs/website/module.f.mjs` are the same blob hashes, and `fjs/website/style/module.f.mjs` differs only by main's `any-pointer: coarse` tap-target hunk absorbed through the merge — the `faviconLinks` region is byte-identical.

Gates from the head SHA: suite 7562/0; `node ./fjs/module.mjs ci` then `git add -A && git diff --cached --exit-code` rc 0, nothing left to regenerate; `tsc -p .` (7.1.0-dev) rc 0, and rc 2 with a deliberate type error in the touched file so the 0 means something; `gh pr checks 2110` all pass.

sergey-shandar (approved):

PR #2112, opened 2026-09-19
fsc: refuse unsupported bare import specifiers

sergey-shandar:
Stacked on #2111 (current parent `aa9e7146`, merged into this branch by `4079e6ad`). Merge the parent first, then retarget this PR to `main`.

## Summary

Take the next small P1 module-resolution step from #2103: stop interpreting bare imports as sibling filesystem paths.

```js
import value from "pkg";   // unsupported host/package resolution: compilation error
import value from "./pkg"; // explicit relative path: existing resolution
```

Add one classification check to the parent's shared `_importSources` boundary, before percent decoding, path normalization or dependency reads. Both value transpilation and EDAG linking already use it. Unsupported names/subpaths/scoped packages and other non-path forms return the existing `Result<..., ParseError>` channel, not a panic or a fabricated value.

The implementation change relative to #2111 remains eight added lines, including comments. The other two changed files contain proofs and the existing TODO update. No new module, exported API, configuration or generated file.

## Scope

- Original specifiers must start with `./`, `../` or `/` to proceed through the existing path resolver. Bare names, `#` aliases and scheme URLs need a host resolver and are refused until supported. This is a category boundary, not a blacklist of package names.
- Parsing and unresolved compilation still represent the original JavaScript import. Refusal here belongs to dependency resolution; CLI input filenames remain unchanged.
- Preserve the parent's raw URL dot-segment processing, query/fragment refusal, segment validation, scalar conversion and error propagation. Existing path behavior is not advertised as complete URL semantics.
- Package resolution, import maps, query/fragment module identity and cache redesign remain separate work.

## Regressions

Actual FJS proofs exercise both compiler paths against misleading existing local files: `pkg`, `pkg/submodule`, scoped packages, `.hidden`, `..hidden`, a normalized bare path and an escaped leading dot. Each corresponding explicit `./` control succeeds. Additional cases cover unused imports, relative-then-bare imports, JSON attributes, empty/non-path specifiers and normal CLI error/no-output behavior across all five output routes.

Existing local-file fixtures use `./b`, `./c`, etc., preserving their sharing/cycle/missing-file purposes rather than relying on the broken bare-name fallback.

## Latest stack update — `4079e6ad`

Merged #2111 at `aa9e7146` into the existing #2112 head `1f52d7ca` using a two-parent merge commit; no rebase or force-push. Code and proofs merged cleanly. Resolved the shared TODO's conflicting task insertions by retaining both the query/fragment restriction and bare-import refusal. The parent's encoded-filename controls and all prior fixes remain intact.

The diff against the updated parent remains three files, 104 additions and 11 deletions. No additional feature or semantic decision was introduced.

## Verification of this merged revision

Direct GitHub cloning still fails on DNS resolution. Verification uses the `package-tarball` artifact from `03e2300a` with the exact merged sources applied. Common-ancestor and earlier-child artifacts were used to reconstruct the three-way inputs; artifact checksums and the actual GitHub blob IDs were verified. This is a runnable package, not a complete repository checkout.

- **Package-root `node --test`: 7,467 passed, zero failures, exit 0**, Node 22.16.0. The complete run finished, including browser proofs; no tests were skipped.
- **Compiler subtree: 1,765 passed**, with **100% line, branch and function coverage** for both `fsc/transpiler/module.f.mjs` and `fsc/edag/module.f.mjs`; all three coverage thresholds were enforced.
- Focused **TypeScript 5.8.3** checking passed for the merged implementation/proof and linker against artifact dependency declarations and the original private types. Stale generated declarations for the checked implementations were removed in an isolated typecheck copy so their actual source was checked.
- `node --check`, `git diff --check`, and the remote three-file comparison passed. All three uploaded blobs match the locally tested files.

**Full repository pinned checks still require this revision's CI.** No full checkout, repository generation, pinned repository-wide type-check or Rust gate pass is claimed. The successful package tests and focused checks do not replace those gates.

Changelog:
- **BREAKING CHANGES:** Non-path import specifiers no longer fall back to sibling files. Use explicit `./` or `../` for local imports. Host/package/scheme resolution remains unsupported; CLI entry filenames are unchanged.


sergey-shandar:
Updated the branch in merge commit `1f52d7ca`, incorporating #2111's latest `03e2300a` without rebasing or force-pushing. Resolved the shared TODO conflict by retaining both completed tasks; the implementation and proofs merged cleanly. URL dot-segment handling, bare-import refusal, compiler diagnostics, and the parent's moved hex-decoder TODO are preserved. GitHub now reports the PR as conflict-free; the diff against its parent remains the same three-file feature.

Verified on the parent's CI package artifact with exact merged sources: package-root `node --test` completed with **7,466 passes, zero failures, exit 0**. Compiler-subtree coverage passed **1,764 tests** with **100% lines/branches/functions** for transpiler and linker. Focused TypeScript 5.8.3, syntax checks, blob hashes and whitespace checks passed. This is not a complete checkout or a full pinned repository type-check/generation pass; the new CI run is in progress.

PR #2120, opened 2026-09-19
fsc/parser: preserve blocks and explicit returns

sergey-shandar:
The source-tree mapper erased return-only blocks and stored other blocks as declarations plus an implicit result. That loses statement structure needed by the planned instruction-pattern matcher.

Preserve every function block as an ordered list of tagged const declarations and an explicit return. The existing fold consumes these statements and produces the same executable AST and EDAG. Grammar, accepted syntax, diagnostics, and compilation results stay unchanged. Bare returns, ASI, and general statement support remain separate work.

An internal syntax reader lets the FunctionalScript proofs inspect the tree before lowering. It does not perform binding checks or early-error validation; compilation still goes through parseFromTokens and its existing fold. Proofs cover return-only and nested blocks, declaration order and source positions, existing syntax refusals, and equivalent EDAGs.

This is the first representation slice of the existing statement-aware-intrinsics TODO, which remains open. It targets main, which now includes the merged #2119.

Validation on the published tree:
- Full `node --test` with CI's Node 26.8.2: **7,562 passed**, zero failures, cancellations or skips.
- Parser-subtree Node coverage run: **170 passed**, **100% lines, branches and functions** for `fjs/fsc/parser/module.f.mjs`.
- Pinned TypeScript 7.0.2, declaration emission, `npm run gen`, and whitespace checks passed. Generation left tracked output unchanged.
- The published Git tree matches the checked local tree exactly.
- Draft pending CI. Cargo is unavailable locally, so Rust lint and format checks were not run; no Rust files changed.

Changelog:
- **BREAKING CHANGES:** `fsc/parser`: the public source `Node` block variant is now `['block', statements]`, where `statements` contains tagged `['const', declaration]` entries followed by `['return', expression]`. Return-only source blocks are retained. Consumers of source nodes/rewrite mappings must handle the new shape; `parseFromTokens` keeps its existing executable-AST result.


o2alexanderfedin (approved):
Approving. Compiled 16 arrow-body fixtures through `fjs compile` to `.js`, `.rs` and `.data.js` — 48 runs — at this head and at `45fbcea1`, comparing exit code, diagnostic text and emitted output: **byte-identical**, with `export default 1;` green as a control and six of the 48 succeeding, so the comparison could have shown a difference. `(...a) => { return 1; }` and `(...a) => { const b = 1; return b; }` both emit `export default (...$a)=>1;`, the same text as `(...a) => 1` — the claim that lowering erases the new shape holds where it matters.

Refusals still name the offending token, checked by column: the `}` of `{ return 1 }` (37) and of `{}` (27), the `;` of `{ return; }` (34), the `{` of a nested block (28), the second `return` of `{ return 1; return 2; }` (38), and `return` at top level (29) — the last agreeing with Node's own `Illegal return statement`.

Three mutations redden named leaves, each restored and re-greened: collapsing a return-only block reddens `parser/proof.sourceBlocks.explicitReturn` and `.nestedBlocks`; reversing declaration order reddens `.orderedDeclarations`, `edag/proof.bodyConst` and `serializer/proof.law`; ignoring the `return` tag reddens `edag/proof.explicitReturns` and 12 more. No `SyntaxError` in any log.

`types.ts` adds `Block` and removes no exported name; the `Node` shape change is declared under **BREAKING CHANGES:** in the description. Suite 7562/0, `tsc -p .` clean under TypeScript 7.0.2 — proved real by retyping `['return', Node]` as `['return', string]`, which fails at `module.f.mjs(419,54)`. `fjs ci` leaves the tree unchanged; all 19 GitHub checks pass.

Coordination, not a defect here: #2106 (`0bdd7107c7d80ab148479335b92112b84f0ea6f5`) keeps the old `['block', readonly Const[], Node]` variant while adding `'~'` and `BinaryTag`, so `merge-tree` conflicts with `dab4e790` in `fjs/fsc/parser/types.ts`; whichever lands second must carry both. #2108 (`7d9a2650432cc537e7ee2ceb91d3b1e6a2ac538a`) merges clean and touches no parser source.

PR #2125, opened 2026-09-19
website: an empty page is probably a stale directory, not a bug

Trinidadec:
Resolves the stale-generated-pages todo, which framed itself as a decision among three options rather than a task list.

Chose the cheapest: no behaviour change. The symptom (an empty page for a directory that was deleted or renamed on a branch switch, since the generator only writes and git won't remove a directory still holding an untracked generated file) is a working-tree-only condition — a real deploy always builds from a fresh checkout, so it can never ship. And git clean -Xd already clears it.

The other two options both make the generator riskier to fix something that isn't actually broken: skipping directories with only generated output can't distinguish a stale leftover from a directory whose real files were legitimately deleted in the same commit; deleting what the build no longer writes is the only option that converges, but adds a destructive rm-like step to a build that has none today.

So: one paragraph in fjs/website/module.f.mjs's @module doc — where the todo itself said to put it, since that's where someone debugging an unexplained empty page would look.

node --test: 7562/7562 passing.


sergey-shandar:
@codex review

o2alexanderfedin (approved):
Approving. The change is prose only — nine lines of `@module` documentation and the retirement of the design document that asked for them — and every claim in the note holds.

I checked the note against the tree rather than taking it: nothing under `fjs/website` unlinks or removes (the generator only writes); `index.html` is in `.gitignore`, and a leftover directory holding only one is in fact removed by `git clean -Xd` (built a throwaway repo and ran it); `generatedName` filters `index.html` out of a listing, so such a directory gets a page with nothing in it.

I also ran the generator over a virtual tree both ways, since the title hedges. A directory holding only a stale `index.html` and a genuinely empty directory produce byte-identical pages — the two really are indistinguishable from inside the build. That is the argument for putting the explanation where someone debugging will read it instead of emitting a claim the code cannot support, and "probably" is the honest word.

Deleting `stale-generated-pages.md` leaves nothing dangling: a tree-wide sweep for the filename and for its heading returns no hits (exit 1, with `generate-website` as a positive control returning hits), and the umbrella list is the referrer and was updated. A full relative-link sweep finds the same 9 pre-existing false hits at this head and at the merge base — 3028 links checked, unchanged by this PR; an injected fake link raised it to 10.

Gates: suite 7562/7562, `ci` clean with no missing regeneration, `tsc -p .` clean (a deliberate type error gave exactly 1), all CI checks green.

Coordination, not a defect: #2110 (`7c310119`) also touches `fjs/website/todo/generate-website.md`, on the checklist line directly above yours. `/usr/bin/git merge-tree --write-tree` against `1bd2c2e0` conflicts there; `module.f.mjs` auto-merges.

sergey-shandar (approved):

PR #2126, opened 2026-09-19
media/json: a demo — JSON as you type, parsed then written back

Trinidadec:
A demo for fjs/media/json: type JSON, see it parsed by this module's own reader and written back in normalized form (sorted keys, one line) by its writer — or the parser's own error message if it doesn't parse. A round trip, not an echo: key order and whitespace aren't part of what a JSON document denotes.

Pure, no operations needed (parse/stringify are pure functions of the text), so update declares never and returns through pureOk — same shape as the sha2 demo.

**Also fixes a small stylesheet gap this surfaced.** textarea is the first multi-line field a demo has used:
- Its own baseline sits at the bottom edge, so a label placed before it (as every other demo does) floated to the bottom of the box instead of the top. Added `textarea { vertical-align: top }`.
- It wasn't in the `font: inherit` list the first demo (sha2's input) already needed, so its font was the generic `monospace` keyword rather than this site's own stack. Added it to that rule — exactly the "every control, not only the ones the site has today" gap the README already names.

Checked in a real browser (Playwright against a locally built site, with a small local static server since Python's http.server serves .mjs as text/plain and browsers refuse to execute that as a module): sorted/compact output for valid input, a clear error for malformed input typed mid-edit, label above the field in both.

`node --test`: 7565/7565 passing. `npm run cov`: 100/100/100 repo-wide.


o2alexanderfedin (approved):
Approving.

**The round trip agrees with node.** I ran `roundTrip` against `JSON.stringify(JSON.parse(x))` over 3555 inputs — every escape, `\ud800` alone, non-BMP, empty/duplicate/integer-like keys, `0`/`-0`/`1e400`/`1e-400`/`5e-324`/`1.7976931348623157e308`, 3000 random finite f64 bit patterns (the ties-to-even printing rule), and whitespace between every token pair. Two disagreements, both exactly the documented normalization: `{"b":…,"a":…}` and `{"-0":1,"0":2}`, where node hoists integer-like keys and this sorts. Nothing else differed — and the harness proved it can report a difference, since it reported those two.

**Malformed input is refused, with the position.** 32 texts node rejects — trailing comma, unquoted key, single quotes, `NaN`/`Infinity`/`undefined`, `+1`, `01`, unterminated string and object, a lone `,`, the empty string, `"\x41"`, a raw tab in a string — all `Error: … at N` naming the offending character. Zero mismatches, with a valid-document control that stayed accepted.

**The new proofs bite.** Four mutations, each reddening the leaf that names it and no other: `sort`→`identity` and swallowing the error → `proof.demo.roundTrip`; dropping `name="json"` → `.view`; ignoring `event.kind` → `.update`. No `SyntaxError` in any log. Restored: 7565 pass, 0 fail.

Gates: `node ./fjs/module.mjs t` green; `ci` leaves the tree clean; `tsc -p .` (7.1.0-dev) rc 0, and a seeded `(text: number)` made it rc 2 on these very files. The sheet gains only `textarea` in the `font: inherit` list and `textarea { vertical-align: top }` — and this demo is the tree's only `textarea`, so the comment's "first of its kind" is true. #2127 carries that identical hunk; `/usr/bin/git merge-tree --write-tree` merges the two heads clean (tree `d0491399`). No exported name removed or renamed.

sergey-shandar (approved):

o2alexanderfedin (approved):
Approve stands at `cefd8cf4`. The head moved only by absorbing `main`: this branch added no commit of its own, and `fjs/media/json` and `fjs/website/style` are byte-identical to `eee6cb26`, which is what I verified. Nothing to re-measure.

PR #2127, opened 2026-09-19
media/datajs: a demo — a document as the graph it denotes

Trinidadec:
Type a DataJS document, see its nodes and edges: a container per distinct reference, a leaf per occurrence, sharing drawn as one node with two incoming edges rather than two look-alike nodes. Plain JSON has no way to write that reference twice, so there's nothing there to draw as a graph — a DataJS `const` is the one part of this format whose entire job is spelling sharing, and this demo draws exactly what the spec says every document denotes: a DAG.

Reads reference identity (`Object.is`) off the value `tryParse` already built, reuses `leafSerialize`/`keySerialize` for label spelling. **A node's rank is the longest path from the root**, computed in a separate pass after the walk (Bellman-Ford relaxation, not a topological sort) — not the depth the walk happened to first reach it at. That started as the simpler rule and had a real bug: a shared node reached again later from higher up the document could draw an edge pointing backward up the page once a shorter path had already claimed a shallower rank. Longest-path ranking guarantees every edge points down by at least one rank, always.

**Two rendering bugs found by actually looking at the output, not by reasoning about the code:**
- Edges fanning out from one shared point had midpoints closer together than their children — two short labels could overlap. Moved labels 2/3 of the way to the child.
- A shared node reached again from above an intervening rank drew exactly on top of the one-rank edge that first reached it (same two endpoints, same straight line) — the second label was invisible. Every edge is now a quadratic bezier (straight is just the degenerate case), and a skip-level edge bows sideways so it can't land on another edge's line.

Verified in a real browser via Playwright against a locally built site at every step — sharing, parallel-edge merging, the bow fix, and the longest-path fix all screenshotted, not just asserted in the test suite. Several tests assert exact hand-computed SVG path/text coordinates, since the layout is fully deterministic; each one was checked by making the bug it guards against, watching it fail, then reverting.

**Review round (o2alexanderfedin), five findings, all fixed:**
- Moved six `_`-private JSDoc typedefs out of `demo.f.mjs` into a new `private.ts` — AGENTS.md forbids a file-scope `@typedef` in an authored `.mjs` anywhere in the repo.
- Replaced four regex-based occurrence counts in `proof.f.mjs` with the `occurrences = (text, needle) => text.split(needle).length - 1` idiom already used in `fjs/ci/rust/proof.f.mjs` — the repo bans regular expressions outside two exceptions at the Node host boundary.
- Fixed two stale comments left over from the first-discovery ranking this PR replaced.
- `ranked` was quadratic in the wrong thing (all edges re-filtered per node, every round, unconditionally for `nodes.length` rounds). Now groups incoming edges by target once and stops the moment a round changes nothing. Reviewer's own benchmark: a 1500-element document 6322ms → 357ms, an 800-node/400-shared-reference document 1520ms → 237ms, with rendered output confirmed byte-identical across 1500 random documents before and after.
- Added two tests for behaviour nothing was pinning: `edgeLabelPosition` (exact coordinate) and `equalLeavesDoNotShare` (a leaf never dedupes by value the way a container dedupes by reference).

Also touches `fjs/website/style/module.f.mjs`: adds `textarea` to the `font: inherit` rule and its own `vertical-align: top`, needed for this demo's own textarea field. Originated in #2126 before that PR merged; now that #2126 is merged into `main`, the two copies are byte-identical and this branch carries no duplicate.

`node --test`: 7599/7599 passing. `npm run cov`: 100/100/100.


Trinidadec:
Reworked the ranking after finding a real gap: the original "rank by first discovery" heuristic only bowed forward skip-level edges, but a node reached again from *higher up* the document (a shorter first path, then a longer one later) could still draw a straight edge pointing backward up the page. Confirmed with `const $0=[1]; export default {"a":$0,"b":{"c":{"d":$0}}};` — the "d" edge drew from y=168 up to y=76.

Switched to ranking by longest path from the root (Bellman-Ford relaxation over the discovered graph, not a topological sort — simpler for a graph this size and gives the same guarantee). This makes every edge point down by at least one rank, always — no more backward edges, and the existing bow logic still handles the remaining forward-skip case. Added a test with hand-computed exact SVG path coordinates for the case that broke it, same rigor as the earlier `skipLevelBow` test.

Also stress-tested whether crossing-minimization (reordering nodes within a rank) is needed: threw a deliberately adversarial case at it (three siblings cross-referencing shared arrays in scrambled creation order) and the result stayed legible with the current simple left-to-right-by-creation ordering. Did not add it — nothing found that actually needs it at this demo's scale.

`node --test`: 7571/7571 passing.

Trinidadec:
Changed the default document so the first thing a reader sees also demonstrates longest-path ranking, not just sharing: `const $0=[1,2]; export default {"a":$0,"b":{"c":$0}};` — "a" and "c" reach the same array by routes of different length (direct vs. through "b"), so the array ranks by the longer one and "a"'s edge is the one that visibly bows past a rank rather than the one that decided where the array sits. The old default had both sharing edges from the same parent at equal length, so it never actually showed the layering this demo now does.

o2alexanderfedin (commented):
Measured at `7fbe1f76` and re-checked at `4bd5e04a`; the last commit was not reviewed, so treat the geometry results as measured one commit back.

The graph does denote the document, and that was worth establishing properly: 500 random documents matched an independent reference walk on node and merged-edge counts with no mismatches, a second 500 matched the multisets of node and edge labels including `-0`/`0`, `NaN`, `1n` and non-canonical keys, and a tree-walking expectation gives different numbers, so the harness can disagree. The longest-path change holds too — 1700 shared-reference documents compared against an independent memoised DFS, 0 mismatches, while scoring shortest-path gives 118 and first-edge 318. No edge points backwards or within a rank: 11,318 edges, none violating, against 84 violations at `59dd8841`.

Five things:

- `demo.f.mjs` has six file-scope JSDoc `@typedef`s. `fjs/AGENTS.md:411`: "No authored `.mjs` may contain a **file-scope** JSDoc `@typedef` — anywhere in …", and points such a claim at module scope in a `.ts` file. They are `_`-private, so a sibling `private.ts` is the home.
- `proof.f.mjs` uses four regex literals. `fjs/AGENTS.md` §"No regular expressions". Two regexes exist in all of `fjs/**/*.f.mjs` today, both at the Node host boundary.
- Line 224, `graphSvg`'s own doc, still reads "ranked by depth of first discovery" — the rule this PR replaced. Line 183's "a rank fills before the walk descends" is stale the same way: ranks are assigned after the walk now. (Line 23's mention is a historical note and reads correctly.)
- `ranked` runs `nodes.length` relaxation rounds unconditionally — `Array.from({ length: nodes.length }, () => null).reduce(relax, initial)` — with no fixpoint check, and each round filters all edges per node. Same document, `59dd8841` → `7fbe1f76`: a 1500-element array went 138 ms → 3.9 s, a 400-shared-reference document 160 ms → 3.0 s. That is per keystroke, and the 1500-element case settles in one round.
- Two branches nothing holds: reverting `t = 0.65` to `0.5` keeps the suite green, so no edge-label position is asserted; and merging equal leaves into one node — against the module's own "a leaf gets one node per occurrence" — is green too.

Suite 7571/0, `tsc` rc 0 against a control, `ci` leaves nothing to regenerate. The `textarea` rule is byte-identical to #2126's and merges to one copy.

o2alexanderfedin (commented):
Marker refresh only. The head moved to `7409d167` by absorbing `main`; this branch added no commit of its own, and `fjs/media/datajs` and `fjs/website/style` are byte-identical to `4bd5e04a`. Every finding in my review there stands unchanged, and I am not restating them.

Trinidadec:
All five findings fixed:

1. **Typedefs**: moved the six `_`-private types out of `demo.f.mjs` into a new `private.ts`, per AGENTS.md's ban on file-scope `@typedef` in an authored `.mjs`.
2. **Regex**: replaced all four `.match(/.../g)` occurrence-counts in `proof.f.mjs` with the `occurrences = (text, needle) => text.split(needle).length - 1` idiom already used in `fjs/ci/rust/proof.f.mjs`.
3. **Stale docs**: fixed both comments still describing first-discovery ranking (`byRank`'s and `graphSvg`'s).
4. **Performance**: `ranked` now groups incoming edges by target once (was re-filtering all edges per node per round) and stops as soon as a round changes nothing (was always running `nodes.length` rounds). Rebenchmarked your exact two cases: 1500-element document back to 108ms (was 3.9s), 400-shared-reference document back to 71ms (was 3.0s).
5. **Coverage gaps**: added `edgeLabelPosition` (asserts the exact `y="62"` coordinate at `t=0.65` vs. what `t=0.5` would produce, `y="56"`) and `equalLeavesDoNotShare` (asserts `[1,1]` produces two leaf nodes, not one shared like a container). Verified both catch their respective regressions by making each mutation, confirming the new test fails, then reverting.

`node --test`: 7599/7599 passing. `npm run cov`: 100/100/100.

o2alexanderfedin (commented):
Delta measured over `7409d167`. All five earlier findings are fixed; one thing is left, and it is in the PR description rather than the code.

**The description still teaches the rule this PR replaced.** "Layout ranks a node by depth of first discovery, left to right within a rank" — that is the first version, corrected inside the module but not here, and this text becomes the merge commit. Same paragraph: "`node --test`: 7570/7570". The suite is `node ./fjs/module.mjs t`, and it reports **7599/0** at this head (7597 at `7409d167`, +2 for the two new proof entries).

What I checked, and how:

- **File-scope `@typedef`s**: 0 in `demo.f.mjs` (6 at `7409d167`). The six moved verbatim to `private.ts`, all `_`-prefixed, none re-declared in both — `types.ts` is byte-identical to its base blob, so nothing public moved.
- **Regex literals in `proof.f.mjs`**: 0 (4 before). `occurrences` counts by `split`, which is what its doc says.
- **Perf**, same process, best of four runs: a 1500-element flat array **6322 ms → 357 ms**; an 800-node document with 400 shared references **1520 ms → 237 ms**. Rendered HTML is byte-identical at both heads over 1500 random shared-reference documents (0 mismatches; the same harness reports 1500 when I change the relaxation step to `rank + 2`), so the fixpoint changed the cost and not the ranks.
- **Edge direction survives the rewrite**: 12,513 edges over that corpus, 0 pointing up or sideways; the pre-fix first-discovery commit gives **560** on the identical corpus.
- **Both mutations now redden.** `t = 0.65` → `0.5` fails `demo.graph.edgeLabelPosition`; merging equal leaves fails `demo.graph.equalLeavesDoNotShare`. One failure each, no `SyntaxError`, green again after restoring.
- `tsc -p .` clean (rc 2 when I retype `_Node.rank` as `string`, so `private.ts` is really checked); `node ./fjs/module.mjs ci` leaves nothing to regenerate; `gh pr checks` all pass.

sergey-shandar (approved):

PR #2135, opened 2026-09-19
docs: separate DISOT principles from the Git projection

sergey-shandar:
The DISOT overview mixes format-independent principles with their Git projection, making it harder to distinguish the model from one concrete representation.

Split `todo/sergey/disot.md` into:

- `todo/sergey/disot/README.md`: documents, immutable history, signatures and trusted time, global and relative names, snapshots, and transport independence.
- `todo/sergey/disot/git.md`: Git rationale, object mappings, signature metadata, submodules, branch and SHA-1 limitations, and comparisons with other decentralized systems.

Move `todo/sergey/doc-def.md` into the same directory without changing its content. Preserve the existing examples and diagrams, add navigation between the pages, and correct relative links, including the existing broken DISOT CLI epic link.

Validation: `node --test` completed with 7,529 passing tests and no failures. Checked all relative links and the snapshot anchor, confirmed there are no references to the removed paths, and ran `git diff --check`. The published Git tree matches the verified local tree. TypeScript, Cargo, and Nix are unavailable in this environment, so the remaining repository checks were not run; this PR remains draft pending those checks. Only Markdown documentation changes.


o2alexanderfedin (commented):
The split is clean: I accounted for all 358 deleted lines paragraph-by-paragraph (whitespace-normalised, fuzzy-matched, positive-controlled by planting a sentence that is present and one that is not). Nothing was lost. The only paragraph gone without a successor is the `immutable documents + ... = DISOT` banner, which `007a0948` removes on purpose; the two ASCII tables became Markdown tables with identical cells, and "The important property is not Git hosting…" is re-stated transport-neutrally. `doc-def.md` is byte-identical across the move (blob `aaa0a1c` both sides).

One thing to fix before this leaves draft.

**`todo/sergey/disot-origin.md` will have a dangling link once this merges.** That file landed on `main` in `776a01dd` — after this branch's merge base — and line 5 links `[document](./doc-def.md)`, i.e. `todo/sergey/doc-def.md`. This PR moves that file to `todo/sergey/disot/doc-def.md`. I reproduced the merge (`git merge-tree --write-tree`, git 2.50.1, clean) and resolved every relative Markdown link and heading anchor in the resulting tree: `main` alone and this head alone each have 11 broken links; the merged tree has 12, the extra one being exactly that link. The sweep was positive-controlled by injecting a fake path and a fake anchor (both reported) alongside a real path and a real anchor (neither reported). Either rebase and update the link to `./disot/doc-def.md`, or leave a stub.

Everything else checks out: no reference to `sergey/disot.md` exists anywhere in the tree, before or after; `node ./fjs/module.mjs t` passes 7591/7591; `node ./fjs/module.mjs ci` leaves the tree clean; all CI checks are green. The Medium citation is well-formed and makes no claim the text depends on — I did not verify its content (Medium returns 403 here).

o2alexanderfedin (commented):
Delta measured over `4b49d6ca`. The branch added no commit of its own — the five it owns have the same patch-ids as before — but it merged `main`, and that merge is what I want to flag, because it turned my earlier finding from a prediction into a fact on the branch.

At `4b49d6ca` I noted that `todo/sergey/disot-origin.md` links `./doc-def.md`, that this file had arrived on `main` after the merge base, and that the link would dangle **on merge** because this PR moves `doc-def.md` into `disot/`. The merge has now happened on the branch itself: at `eca0deb1`, `disot-origin.md:5` still reads `[document](./doc-def.md)`, `git cat-file -e` on `todo/sergey/doc-def.md` exits 128, and the file is at `todo/sergey/disot/doc-def.md`. The merge commit brought the referrer over without adjusting the reference. One-line fix: `./disot/doc-def.md`.

Everything else in my review stands — the paragraph-set comparison showing nothing was lost in the split, the deliberate removals accounted for, the rename byte-identical.

The PR also left draft since my last look.

o2alexanderfedin (commented):
The new head only takes in `main`. It adds no work of its own, and the link I flagged before is still broken.

I proved the absorption in two ways.

1. I reproduced the merge of the two parents with `git merge-tree --write-tree`. The result is tree `d898b605768e4c77c8c47d129a60c27b0f720f76`. That is exactly the merge commit's own tree. There was no conflict, so nothing was resolved by hand.
2. I compared the PR's diff over its merge base, before and after the merge. The two diffs are the same byte for byte. Both have patch-id `6ff0b57f1fe8a6db2c96974ebfe7c2d7347316ea`.

So the documents did not change since my review at `eca0deb1`, and my finding from that review still stands. At `7aac3565`, `todo/sergey/disot-origin.md` line 5 still reads `[document](./doc-def.md)`. `git cat-file -e` on `todo/sergey/doc-def.md` exits 128. The file is at `todo/sergey/disot/doc-def.md`. The one-line fix is to change the link to `./disot/doc-def.md`.

What I did not check. I did not read the documents again, because their content is unchanged. The PR's own diff is four Markdown files under `todo/sergey/`, and that diff did not move. The merge also brought in the rest of `main`. I did not run the test suite or the regeneration check over it, because `main` passed its own checks.

sasha-gil (approved):

PR #2138, opened 2026-09-19
media/rust: spell control and bidirectional characters as \u{...}

sergey-shandar:
Stacked on #2137. Step 1 of `fjs/media/rust/todo/strings-rust-cannot-spell.md`.

`stringLiteral` threw on a control character, U+0000 to U+001F and U+007F, and the throw reached the CLI uncaught; it passed a bidirectional control, U+202A to U+202E and U+2066 to U+2069, through to a literal `rustc` refuses under `text_direction_codepoint_in_literal`. Rust's `\u{…}` escape holds every one of them, so the writer now spells them and the refusal has no reason left. The signature is unchanged, so no importer moves.

**Proof.** `fjs/media/rust/proof.f.mjs` pins both ends of each range, a character between the two bidirectional ranges and one above them, and the lone surrogate still passing through. A new `nanvm-harness` fixture, `escapes.mjs`, round-trips five of the code points through a generated module on the naive VM; `npm run gen` compiles it and produced no drift.

**Left for step 2.** A lone surrogate still passes through as it stands, and `i64Literal` still throws for a bigint outside `i64`; the todo's remaining task turns both into the `Result` the printers carry.

Checks run: `tsc`, `fjs test`, `node --test`, `npm run cov` at 100%, `cargo test -p nanvm-harness`, `cargo clippy`, `cargo fmt -- --check`, `npm run gen`.

https://claude.ai/code/session_01A2jsFNE5PsuS6D7YFbhiuf

---
_Generated by [Claude Code](https://claude.ai/code/session_01A2jsFNE5PsuS6D7YFbhiuf)_


o2alexanderfedin (approved):
Approving. Delta measured over `3ab7fede`, the head of the base branch `claude/inspiring-archimedes-qz3s3s`; that base has since merged into `main`, and it is `merge-base(origin/main, 4990908c)`.

The escaped set is exactly right, measured rather than sampled. I compiled every non-surrogate code point below U+10000 as the printer now writes it — 65,024 literals in eight `rustc` crates, all rc 0, no diagnostics — and separately compiled the same code points written **raw**, which rejected exactly nine: U+202A–U+202E and U+2066–U+2069. That non-zero is what makes the zero believable. The printer escapes those nine plus the C0 controls and U+007F and nothing else; U+2029, U+202F, U+2065, U+206A, U+FEFF, U+200B, U+200E, U+200F and all 1,048,576 astral code points pass through untouched.

The escapes still name the right characters: a generated Rust program over 71,143 literals asserts each holds one `char` whose scalar value equals the intended code point — rc 0, and rc 101 after I flipped one escape by hand. `cargo test` at the workspace root is rc 0; `escaped_characters` reddens with a left/right diff when I change one code point in `escapes.rs`, so it pins the value, not the compile.

Gates: `fjs test` 7589/7589; `tsc -p .` rc 0; `npm run gen` leaves no drift, and re-writes `escapes.rs` after I corrupt it; `cargo clippy -- -D warnings` and `cargo fmt --check` rc 0; all 14 PR checks pass. Narrowing either bidi range by one code point at either end reddens `fjs/media/rust/proof.f.mjs proof.stringLiteral()`.

No `**BREAKING CHANGES:**` block is owed. Over all 1,112,064 non-surrogate code points the new printer differs from the old only where the old one threw (30) or emitted a literal `rustc` refuses (the nine); nothing that compiled before changes shape, and the export set is unchanged.

One non-blocking nit: widening either range outward by one — U+202F, U+2065 — leaves the suite green. The nearest guard above is `‰` at U+2030 and there is none below U+202A, so an over-escape of the neighbours would go unnoticed. Harmless to the value; worth an `assertEq` if you want the range pinned on both sides.

PR #2140, opened 2026-09-19
media/rust: refuse a lone surrogate and an out-of-range bigint as a Result

sergey-shandar:
Step 2, the last task of `fjs/media/rust/todo/strings-rust-cannot-spell.md`, which this pull request deletes. Step 1 was #2138.

`stringLiteral` wrote a lone surrogate through to a file `rustc` could not read, and `i64Literal` threw for a bigint outside `i64`, the throw reaching the CLI uncaught. Both now answer `Result<string, unknown>`, the refusal carrying the value itself: the literal layer commits to nothing about a reason's shape, and the printer above it names one.

- **`fjs/edag/rust`** wraps each refusal as its `[reason, detail]` pair where a string or bigint is printed, as a primitive, an object key or an index, through two small helpers, `stringExpr` and `bigintExpr`. `primitiveExpr` returns the `Result` the rest of `expExpr` already carries. The surrogate detail is written by `JSON.stringify`, which spells it as its escape rather than the unpaired code unit a diagnostic cannot show.
- **`fjs/fsc/rust`** is unchanged: `reasonText` renders the pair as before, so `fjs compile` reports `no Rust spelling for this module: no Rust string literal for a lone surrogate in: "\ud800"` and `… no Rust i64 for: 9223372036854775808`, exit 1, where it exited 0 with an unreadable file or died with an uncaught throw.
- **`fjs/nanvm/rust`** unwraps the `Result` for the corpus's own case names, a refusal there being a defect in the corpus, as it already unwraps the printer.

**Proofs.** `fjs/media/rust` pins both surrogate halves and one past either end of `i64` as refusals carrying the value; `fjs/edag/rust` pins the reason at each of the three print sites and the two throw-through-`unwrap` cases; `fjs/fsc/rust` pins both CLI-level messages. No `throw` case remains in `fjs/media/rust`.

The compatibility epic's link to the todo now points at the module, saying the defect is fixed.

Checks run: `tsc`, `fjs test` (7600), `node --test` (7538), `npm run cov` at 100%, `npm run gen` with no drift, `cargo test`, `cargo clippy`, `cargo fmt -- --check`. `cargo clippy --all-targets` reports a pre-existing `clippy::eq_op` in nanvm-lib's own tests, identical on `main`, which this change does not touch.

Changelog:
- **BREAKING CHANGES:** `media/rust`: `stringLiteral` and `i64Literal` return `Result<string, unknown>`, refusing a string with a lone surrogate and a bigint outside `i64` where they wrote invalid UTF-8 or threw.

https://claude.ai/code/session_01A2jsFNE5PsuS6D7YFbhiuf

---
_Generated by [Claude Code](https://claude.ai/code/session_01A2jsFNE5PsuS6D7YFbhiuf)_


o2alexanderfedin (approved):
Approved.

The refusal fires exactly where the value has no spelling and nowhere else. Verified by compiling 21 modules through `node ./fjs/module.mjs c`, exit codes asserted, with `export default 1;` as a control at rc 0 in the same run: a lone high surrogate, a lone low surrogate, one adjacent to text, two in one string and a reversed pair are all refused at rc 1 with the message this PR's description quotes, and no `.rs` file is written; `"😀"` (both as a pair escape and as a literal), U+10FFFF, `"퟿"` and `""` compile. Sweeping every code unit U+D800–U+DFFF alone gives 2048 refusals and 0 acceptances, while 13818 sampled well-formed pairs give 0 refusals — the iteration-by-code-point check does not swallow a pair, which is the failure a per-code-unit check would have.

`i64Literal` accepts exactly [-2^63, 2^63-1], matching `bigint_any(v: i64)`, the helper the printer emits; 2^63, -2^63-1, 2^64-1 and 2^127-1 are refused as diagnostics, not crashes. No `throw` remains in `fjs/media/rust`; `fjs/nanvm/rust`'s `unwrap` is the documented corpus-contract failure.

Round trip: 9531 accepted strings (every code point to U+2100, the bidi and specials set, an astral sample) emitted through `stringLiteral` into one `rustc`-compiled program — builds, and every string's code points match JavaScript's exactly. The harness reports 1 mismatch when one expectation is corrupted.

Mutations, each reddening named leaves: accepting lone surrogates again → `media/rust.stringLiteral`, `edag/rust.literalRefusals`, `edag/rust.throw.loneSurrogate`, `fsc/rust.toRust.refusedLoneSurrogate`; `i64Max` +1 → 3 leaves; `i64Min` −1 → 2; a per-code-unit surrogate check → `media/rust.stringLiteral`.

Gates: suite 7600/0 (main 7597 — net +3 leaves), `tsc -p .` rc 0 (control failure TS2322), `ci` + `nanvm/update` + the harness fixtures leave no drift (control: a printer change moves 967 lines of `generated.rs`), `cargo test` rc 0, all 19 GitHub checks pass with none pending. The deleted todo's two open tasks are both what this PR does, and nothing in the tree references it.

